PR #2, opened 2016-08-19
edit the formated double " to the right ``

ESP32DE:


Spritetm:
Thanks for contributing! Fyi, this is fixed in our internal Git, and will be uploaded the next time we push that to Github. We don't have automatic pushing in place just yet, so it'll take until after the weekeind before you see this change.

Spritetm:
Fixed in latest master.

PR #4, opened 2016-09-01
Replace outdated use of c_types.h with stdint

ESP32DE:
Replace outdated use of c_types.h with stdint
uint32 -> uint32_t


projectgus:
Thanks Rudi! I've added this PR and #5 into our internal change queue.

For future reference, if you make changes on the same branch then you can add commits to an existing PR. You don't need to create a new PR each time (but github will do this if you create a new branch each time.)

ESP32DE:
ah - thanks - have not seen "Add more commits by pushing to the" - ok now i know.
you guys are hard working team! where we can - we support you :+1:

PR #21, opened 2016-09-22
Backport the static allocation feature from FreeRTOS V9.0.0

danicampora:
This feature allows to use static buffers (or from a pool of memory which is not
controlled by FreeRTOS). 
In order to reduce the impact of the changes, the static feature has only been added 
to the queus (and in consequence to the semaphores and the mutexes) and the tasks. 
The Timer task is always dynamically allocated and also the idle task(s), which in the
case of the ESP-IDF is ok, since we always need to have dynamic allocation enabled.

### DIFF
--- a/components/freertos/include/freertos/FreeRTOS.h
+++ b/components/freertos/include/freertos/FreeRTOS.h
@@ -74,6 +74,7 @@
  * Include the generic headers required for the FreeRTOS port being used.
  */
 #include <stddef.h>
+#include "sys/reent.h"
 
 /*
  * If stdint.h cannot be located then:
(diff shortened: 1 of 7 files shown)


danicampora:
This change is the last requirement needed in order to enable MicroPython threading on the ESP32. MicroPython threads (and locks) are created using native FreeRTOS objects that must live in the pool of memory owned by MicroPython's garbage collector.

danicampora:
I fixed the bug of not saving the CoreID during task initialization and pushed again.

negativekelvin:
Just wondering what are the considerations for backporting these features vs moving completely to freertos 9?  

> FreeRTOS V9.x.x is a drop-in compatible replacement for FreeRTOS V8.x.x

igrr:
Latest FreeRTOS version for which an Xtensa port is available is 8.2.0. It has been extensively modified to support dual core setup. So while version 9.0.0 may be a drop in replacement from the application perspective, it is not drop-in from the port perspective.

danicampora:
@igrr @Spritetm is there any update on this? Have you had a chance to look at it? Thanks :-)

projectgus:
Hi @danicompara,

Sorry for the slow update. This commit is in our internal review/test queue for merging. This week will be quiet because all of China is on holidays (Golden Week), but we should get back to you shortly after.

Angus

danicampora:
@projectgus thanks for the feedback!

igrr:
Turned out that test failures we saw were unrelated to FreeRTOS changes. This is now merged in master. Daniel, sorry that it took so long.

danicampora:
No probs @igrr , I'm happy that it's merged now :-)

PR #70, opened 2016-11-03
examples: add a "wifi scan" example

noduino:


bvernoux:
It will be really better to add example with fully open source license (not GPL/LGPL) for any reuse like using Apache 2.0 or BSD as it is the purpose of example ...
Could you change that ?

igrr:
Even better, do the same as we do for our examples:

```
    This example code is in the Public Domain (or CC0 licensed, at your option.)
```

icamgo:
have changed the license to Apache license 2.0

igrr:
Thanks for the PR.
I am not sure about your addition of `u8glib` and `U8glib` libraries.
First, please check https://github.com/espressif/esp-idf/blob/master/CONTRIBUTING.rst#before-contributing, in particular the first point.
Second, these libraries do not seem to be related to the topic of your PR ('add a wifi scan example).
Third, adding two folders which differ only in their case is not the best idea. On Windows and macOS filesystems are case-insensitive by default. So guess what happens when someone checks out the repository on such a system? :)

Next, for the code which you add, please observe ESP-IDF style guide.
It's not yet merged into master, I have put a temporary copy of it here: https://gist.github.com/igrr/008c58b228d885d191e8dcafaab8eb16

Yet another note: the example file says something like "Licensed to the Apache Software Foundation (ASF)". This should not be the case... Please review [contributor license agreement](https://github.com/espressif/esp-idf/blob/master/docs/contributor-agreement.rst).

icamgo:
please forget the commits not related to "wifi scan" example

projectgus:
@icamgo Thanks for sending this in, very much appreciate the contribution.

We would really prefer all our examples to be Public Domain/CC0 licensed. This way a user can combine a few of them into one program without picking up extra license Copyright notices along the way.

How would you feel about using the license header from one of the other examples, rather than Apache? ie this header: https://github.com/espressif/esp-idf/blob/master/examples/02_blink/main/blink.c#L3

Thanks again,

Angus

examples/06_wifi_scan/main/app_main.c
```diff
@@ -0,0 +1,76 @@
...
+	esp_wifi_set_mode(WIFI_MODE_STA);
+	esp_wifi_start();
+}
+
+void app_main()
+{
+	nvs_flash_init();
+	system_init();
+	wifi_init();
+	printf("Welcome to Noduino Quantum\r\n");
```
projectgus: Could you change this to a generic message "Welcome to esp-idf wifi_scan demo" or something, please?

examples/06_wifi_scan/main/component.mk
```diff
@@ -0,0 +1,10 @@
+#
+# Main Makefile. This is basically the same as a component makefile.
+#
+# This Makefile should, at the very least, just include $(SDK_PATH)/make/component_common.mk. By default,
```
projectgus: The newly updated esp-idf build system doesn't require including component_common.mk, so you can delete everything from this line down.

examples/06_wifi_scan/main/app_main.c
```diff
@@ -0,0 +1,76 @@
+/*
+ * MaiKe Labs (2016 - 2026)
```
projectgus: I don't understand 2026...? I'm 100% OK with you crediting yourself and MaiKe Labs here, but could you please combine these into one line, ie
```
WiFi scan example

Written by Jack Tan of MaiKe Labs <jiankemeng@gmail.com>

This example code is in the Public Domain ...
```

projectgus:
Hi @icamgo ,

Sorry for so long between updates. We're still happy to look at merging this, if you're able to please sign the CLA as prompted above.

Angus

projectgus:
Closing due to inactivity.

PR #79, opened 2016-11-06
Non-Volatile Storage (NVS) Example

krzychb:
* Demonstrates how to read and write a value using NVS. The value tracks number of ESP32 module restarts.
* Shows how to use basic diagnostics to check if read / write operation was successful.


examples/07_nvs_read_write/main/nvs_read_write.c
```diff
@@ -0,0 +1,62 @@
...
+    system_init();
+
+    nvs_handle handle_to_settings;
+    esp_err_t err;
+    int32_t restart_counter = 0;
+
+    // Open the NVS
+    printf("Opening Non-Volatile Storage (NVS) ... ");
+    err = nvs_open("settings", NVS_READWRITE, &handle_to_settings);
+    printf((err != ESP_OK) ? "Failed!\n" : "OK\n");
```
igrr: This will print "Failed" in case of an error, but will continue to `nvs_get_i32` with an uninitialized `handle_to_settings`, likely causing UB.

examples/07_nvs_read_write/main/nvs_read_write.c
```diff
@@ -0,0 +1,62 @@
...
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "esp_system.h"
+#include "nvs_flash.h"
+#include "nvs.h"
+
+void app_main()
+{
+    nvs_flash_init();
+    system_init();
```
igrr: `system_init` not needed here since WiFi stack is not used. Plus, this function will be removed shortly.

examples/07_nvs_read_write/main/nvs_read_write.c
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    // Write to the NVS
+    printf("Updating restart counter in NVS ... ");
+    restart_counter++;
+    err = nvs_set_i32(handle_to_settings, "restart_conter", restart_counter);
+    printf((err != ESP_OK) ? "Failed!\n" : "OK\n");
+
+    // Close the NVS
+    nvs_close(handle_to_settings);
```
igrr: You need to call `nvs_commit` before calling `nvs_close`, otherwise the changes done using `nvs_set` are not guaranteed to be written to flash.

examples/07_nvs_read_write/README.md
```diff
@@ -0,0 +1,7 @@
+# Non-Volatile Storage (NVS) Read and Write Example
+
+Demonstrates how to read and write a value using NVS. The value tracks number of ESP32 module restarts.
+
```
igrr: In addition to using a fixed size int value, it may be worth demonstrating how to read and write variable-length data (str and blob) using NVS. In particular, demonstrate how to get the length of value, allocate the buffer of the sufficient size, read from nvs into a buffer, and release the buffer. I can imagine that using nvs API for variable length data storage may be less obvious, therefore example would be handy. There is some example code in the header file, it may be worth moving it here.

examples/07_nvs_read_write/main/nvs_read_write.c
krzychb: I did it on purpose to make the example simpler by not adding an extra `if` condition. My assumption was that issue with uninitialized `handle_to_settings` will be captured and displayed as an error in code below. Granted, I did not test this scenario.

examples/07_nvs_read_write/main/nvs_read_write.c
krzychb: Noted.

examples/07_nvs_read_write/main/nvs_read_write.c
igrr: It should be, but that is currently being done in another branch, along with `system_init` deprecation.

examples/07_nvs_read_write/main/nvs_read_write.c
igrr: If you want to test it, the easiest way is to change namespace name to the one which doesn't exist yes, and open it with `NVS_READ` option.

examples/08_nvs_rw_blob/main/nvs_rw_blob.c
```diff
@@ -0,0 +1,183 @@
...
+
+    // Read the size of memory space required for blob
+    size_t required_size = 0;  // value will default to 0, if not set yet in NVS
+    err = nvs_get_blob(my_handle, "run_time", NULL, &required_size);
+    if (err != ESP_OK && err != ESP_ERR_NVS_NOT_FOUND) return err;
+
+    // Read previously saved blob if available
+    uint32_t* run_time;
+    if (required_size > 0) {
+        run_time = malloc(required_size);
```
igrr: How about allocating `required_size + sizeof(uint32_t)` here in advance, to avoid doing `realloc` later?

igrr:
Except for the `malloc`/`realloc` thing, looks good to me. I will put this up for internal review.

examples/08_nvs_rw_blob/main/nvs_rw_blob.c
krzychb: Yeah - this part was bugging me.
Great tip and nice chunk of code stripped - thanks a lot!

PR #98, opened 2016-11-14
StaticQueue_t needs to have the same size as xQUEUE.

danicampora:
Without this change, building FreeRTOS with static allocation enabled succeeds, but trying to create a queue from a static buffer causes a runtime assert because the size of static and dynamic queues differ.


danicampora:
@igrr @Spritetm have you been able to check this one? It's a no-brainer ;-)

igrr:
Yes! It is merged and should be in Github master soon.

igrr:
Merged in 3830305, thanks.

Issue #110, opened 2016-11-19
Pullup is not configured by gpio_config 


ggsubs:
The following config is not enabling pullup:

```
	 gpio_config_t io_conf;
	 io_conf.intr_type = GPIO_INTR_NEGEDGE; 
	 io_conf.mode = GPIO_MODE_INPUT; 
	 io_conf.pin_bit_mask = GPIO_SEL_21;
	 io_conf.pull_down_en = 0; 
	 io_conf.pull_up_en = 1; 
	 gpio_config(&io_conf); 
```
Implicit gpio_set_pull_mode is needed after gpio_config to enable pull-up:

```gpio_set_pull_mode(21, GPIO_PULLUP_ONLY);```

igrr:
Thanks @ggsubs, your fix has been merged into master.

PR #111, opened 2016-11-19
Pullup is not configured by gpio_config #110

ggsubs:


jrobeson:
the last 3 commits don't match the title of the PR

projectgus:
Thanks for submitting these, @ggsubs, and also for opening issues #110 & #112. We have separate fixes for these issues in the pipeline (assuming the issue #112 you're seeing is the same issue we found on OS X embedding text formatted files - we're going to keep the example certs in PEM format for now as they're easier for people to work with.)

I'm going to close this PR, but please let us know if either of the fixes we merge don't resolve these issues satisfactorily.

Issue #137, opened 2016-12-03
Build system uses a lot of undefined make variables


rojer:
I am integrating ESP32 IDF build into our own build system, which runs with `--warn-undefined-variables` in `MAKEFLAGS`. I find it very useful and would rather keep it enabled. I also like esp-idf build system, it looks promising, but currently is extremely "--w-u-v dirty" and produces a lot of noise.

Is there a specific reason why you `--warn-undefined-variables` is not used?

rojer:
FWIW, #138 is my attempt at doing it. seems fairly non-invasive, and i think `--w-u-v` help keeps one's sanity while writing make rules.

projectgus:
Closing in favour of ongoing discussion/PR in #138.

PR #138, opened 2016-12-03
Add --warn-undefined-variables to MAKEFLAGS

rojer:
Fix warnings where undefined vars are used.
Most are trivial, except the $$(FLAGS_$$(subst .o,,$$(@))) but this one
produces "weird" variable names anyway and i verfified that they are
always empty.

A few are left in the bootloader.

Fixes espressif/esp-idf#137


projectgus:
Thanks for raising the issue, to be honest I did not know make had an option to warn on undefined variables.

The `FLAGS_basename` feature is documented, although you're entirely right that that it doesn't appear to be used at the moment: https://github.com/espressif/esp-idf/blob/master/docs/build_system.rst#optional-component-specific-variables

(There are use cases for this, when upstream code has some warnings but the component contains both upstream and non-upstream code, and you don't want to disable those warnings in your code or edit the upstream code. I think I've used it in some upstream code that isn't in the repo yet, but the make feature somehow got in first.)

In a general sense... maybe you can explain to me the benefit of enabling this option? The Make default behaviour is "a variable which is undefined expands to nothing", and most of the changes add `?= \n` which means "if the variable is undefined, set it to nothing".

On the one hand, I see that this way any typos or other mis-namings of variables will be found fairly quickly because they produce a warning.

On the other hand, the `?=` for some of these variables seems to create an implication that the user might want to or be able to override them (we can blame Make here for having basically no definition of encapsulation or a "public interface"). The variables that were explicitly set to empty beforehand were onces for which overriding was supported (except for where we needed to set the variable type before `+=` was invokved). Or at least that was the general idea, there might be some inconsistencies to this that have crept in.

Also, setting CONFIG_xxx variables to empty values seems, in my mind, to not quite match with the way Kconfig works.

I'm not saying we won't merge this, but I'd like to understand the rationale behind the kinds of problems it solves.

rojer:
> On the one hand, I see that this way any typos or other mis-namings of variables will be found fairly quickly because they produce a warning.

this, and pretty much only this. it helps immensely when refactoring makefiles, assuming previous state was "clean". i found this invaluable working with hand-written makefiles, i think you will too.

that said, it's ultimately your decision whether to adopt it or not. it seems fairly non-invasive and the benefits are huge (at least by my experience).

> FLAGS_basename feature is documented

yes, indeed. however, if does produce weird var names like `FLAGS_foo/bar/baz` - i didn't even know slashes were allowed in var names, but i guess they are. make is weird.
it does present a problem as it generates a large number of mostly unset variables. however, given that it is not currently used, how about using per-file target flag overrides? instead of `FLAGS_foo/bar/baz = --asplode`, add a target like `foo/bar/baz.o: CFLAGS += --asplode`
it does the same thing and i think it looks more readable to casual observer, see e.g. [here](https://github.com/cesanta/mongoose-iot/blob/8768366/fw/platforms/cc3200/Makefile.build#L359).

> setting CONFIG_xxx variables to empty values seems, in my mind, to not quite match with the way Kconfig works.

yes. the way if currently behaves is by inserting a commented `# CONFIG_FOO is not set` line for each option that is not selected, whereas it would need to set it to empty. i thought about doing that in this PR, but then thought to keep it minimally invasive. if you like, i can do it.

projectgus:
Thanks for the tip regarding per-target overrides, that's another make feature I didn't know existed! A much cleaner solution.

projectgus:
> whereas it would need to set it to empty. i thought about doing that in this PR, but then thought to keep it minimally invasive. if you like, i can do it.

This seems like a good approach, because changing the configuration variable name and forgetting to change the makefile is a very real failure scenario, and this would catch it. Because the configuration makefile inclusion causes a make restart, it seems likely that there will always be some undefined warnings on first run (unless make takes this into account somehow and doesn't warn until the restart phase is over.)

rojer:
please take another look. i modified Kconfig to emit `CONFIG_FOO=` instead of `not set`

projectgus:
Hey @rojer,

Sorry for the delay in getting back to you on this.

This change doesn't seem quite right, the sdkconfig file is generated with CONFIG_FOO= instead of "# CONFIG_FOO is not set", however this isn't the file which is included into the make process. That file is auto.conf, and it is being generated with unset config variables missing entirely.

(I also noticed the changeset still has `CONFIG_SECURE_BOOT_SIGNING_KEY ?=` in one of the component makefiles, which is undesirable once this change works as-desired - it's helpful to get a warning if a makefile refers to a non-existent config variable.)

rojer:
looks i fumbled something and did not push the latest branch.
sorry about that, PTAL.

rojer:
and yes, `CONFIG_SECURE_BOOT_SIGNING_KEY` is used but not defined

projectgus:
Hmm, I see. SECURE_BOOT_SIGNING_KEY is a valid config item, but when its dependency (secure boot) is undefined it doesn't get generated...

I guess there are two possible solutions. Either have kconfig generate empty make variables for all impossible variables, not just unset ones. Or to wrap the make logic in an ifdef for secure boot.

rojer:
I think i handled it.
here's auto.conf diff with the latest commit: http://pastebin.com/HbPhwy7u

rebased onto latest master. PTAL.

projectgus:
There are still some issues here, for example empty strings that are disabled by prerequisite output like this in auto.conf:
```
CONFIG_SECURE_BOOT_SIGNING_KEY=""
```

Which resolves true for GNU make `ifdef`, because make doesn't strip quotes.

To be honest, this change seems like it will need a lot of additional workarounds to get it to work in a useful way - and if these are the obvious limitations, I worry there may be other subtle ones. It'd be a nice benefit when refactoring to know that there are aren't any undefined references in the makefile, but at the same time it doesn't seem like esp-idf has had any bugs yet that would have been prevented due to the extra warnings.

rojer:
well, i see that you don't like it for some reason. say a word and i'll close it, it's your prerogative. however, consider this (warnings in the current build system):
```
/opt/Espressif/esp-idf/make/project.mk:343: warning: undefined variable 'TEST_COMPONENT_PATHS'
/opt/Espressif/esp-idf/make/project.mk:380: warning: undefined variable 'which git'
/opt/Espressif/esp-idf/make/project.mk:380: warning: undefined variable 'which git'
/opt/Espressif/esp-idf/make/project.mk:380: warning: undefined variable 'which git'
/opt/Espressif/esp-idf/make/project.mk:380: warning: undefined variable 'which git'
```
i am almost positive that `which git` was not supposed to be a variable :)

projectgus:
Haha, well you've got me there. The TEST_COMPONENT_PATHS one is OK (this variable is only set if TEST_COMPONENTS is passed into make) but that is a proper bug that was otherwise undetected! :)

Do you think it will be ugly to modify the auto.conf generation so that the line is always assigning to empty if the prerequisite is disabled, ie:
```
CONFIG_SECURE_BOOT_SIGNING_KEY=
```
?

Otherwise, it may be fine for us to check for the prerequisite (in this case secure boot) in the Makefile, before testing for a dependent config item.

rojer:
hm. it would be possible, but undesirable i think: we can indeed distinguish between empty and disabled because of dependency. but it seems to me that this _would_ be a bit too subtle, i think.
with `""` at least it's consistent, you only need to learn one rule: empty value is not empty, must remove quotes to compare. i see there's already a handy `dequote` function to make it easy. whereas if the empty value changes depending on whether it's disabled by dependency or enabled but empty would be a ripe ground for gotchas. so my preference would be to leave it as is.

looking specifically at `CONFIG_SECURE_BOOT_SIGNING_KEY`, it looks like we're ok - it is not iused in conditionals at all. there are also no other string-valued variables with defaults.

projectgus:
OK, that makes sense.

I'm going to be away for the next week, but I'll pick this up again when I get back.

rojer:
ping?
i rebased the branch onto latest master

projectgus:
Thanks for persisting with this. I finally got a chance to look at it again today.

* After adding this update, a partial build outputs a lot of errors caused by the old-style auto.conf. These persist until you either change the config or run a clean build. Maybe this doesn't matter, as anyone updating a significant amount will get an updated config and anyone else will probably be OK with living on the bleeding edge. On the other hand, a simple solution is to rename auto.conf to something else.

> it's clean except for bootloader build
> which i actually think you should take a look at - there may be some weirdness in the way the build is invoked which causes all those vars to not be defined.

I found these:
```
idf/components/bootloader/Makefile.projbuild:113: warning: undefined variable 'SECURE_BOOTLOADER_KEY'
idf/components/bootloader/Makefile.projbuild:113: warning: undefined variable 'BOOTLOADER_DIGEST_BIN'
```

Which were a false positive. Are you seeing anything else?

* I'm still seeing a couple of oddnesses with config items that are disabled by their parents. For example, building the "ethernet" example I get 
```
/home/gus/esp/32/idf/components/esp32/component.mk:7: warning: undefined variable 'CONFIG_PHY_ENABLED'
```

Which is legitimate as CONFIG_PHY_ENABLED doesn't appear in either the sdkconfig or auto.conf files. It's :

```
config PHY_ENABLED
	bool
	default y if WIFI_ENABLED || BT_ENABLED
```

And BT_ENABLED & WIFI_ENABLED are both of type "menuconfig", which I think means this item is disabled unless one of them is set (and they're both unset in the ethernet example.)

projectgus:
Here's a patch for a few of the small things mentioned above:
https://gist.github.com/projectgus/fca095b16b199cf0390d1b948f91725c#file-pull138_tweaks-patch

(The CONFIG_PHY_ENABLED and the out-of-date-auto.conf are not solved by this.)

rojer:
ok, i ended up redefining PHY_ENABLED a little, which is logically equivalent, but ends up creating a symbol that actually gets written to the output.
i've now built all the examples and all of them build cleanly, with the exception of system/ulp, which doesn't build due to lack of ULP assembler binary.
also, system/ulp prints this warning:
```
component_wrapper.mk:228: warning: undefined variable 'GenerateBuildDirTarget'
```

this seems legit, as i cannot find definition of `GenerateBuildDirTarget` anywhere (as opposed to, say, `GenerateEmbedTarget`).

rojer:
i also make conf binary depend on its sources, so it is rebuilt when those change. verified that an example built on `master` is rebuilt cleanly on `makewarns` with a simple `make`.

projectgus:
> i ended up redefining PHY_ENABLED a little, which is logically equivalent, but ends up creating a symbol that actually gets written to the output.

OK. I was a bit uncomfortable with this change at first, but I guess following this "define everything" philosophy it makes sense to throw a warning if a kconfig item is defined without a valid default.

***

Looking the whole changeset again I noticed one more thing, which is that the replacements for `ifdef` and `ifndef` are a bit inconsistent. I think `$(ifeq "$(V)","y")`, `$(ifeq "$(V)","")`, `$(ifneq "$(V)","y")` and `$(ifneq "$(V)","")` are all used at least once.

Would you mind changing these to all use the same two patterns? I think `$(ifeq "$(V)","y")`, `$(ifeq "$(V)","")` are probably the most readable (if you can call make syntax readable, anyhow!)

On the same note, the example in the build system documentation still uses `ifdef`. Are you able to please change that, and also add a small documentation section about make warnings with some tips to work with them (ie don't use ifdef, pre-declare possible command line variables with `?=`, make sure config items always have a value default value, etc.)

rojer:
so, i amended the docs. but wrt `$(V)` -  i don't actually find that many uses of it in the IDF.
maybe my grep-fu is weak, but...
```
[rojer@nba ~/cesanta/esp/esp-idf makewarns]$ grep -rI '\$(V)' .
./make/common.mk:ifeq ("$(V)","1")
./make/project.mk:$(if $(V),$(if $(wildcard $(1)/Makefile.projbuild),$(info including $(1)/Makefile.projbuild...)))
./components/esptool_py/esptool/flasher_stub/Makefile:ifneq ("$(V)","1")
./components/bootloader/Makefile.projbuild:     V=$(V) BUILD_DIR_BASE=$(BOOTLOADER_BUILD_DIR) TEST_COMPONENTS= TESTS_ALL=
```
so, that's only one use, and its value is compared to `1`.
i'm ok with changing it, but wanted to confirm first, that we're looking at the same thing.

rojer:
i rebased the change. it works, i swear :)

projectgus:
Your patience is astounding (thanks!), but I finally have some good news. I've cherry-picked this onto our master branch, resolved conflicts, and queued it for review & merge internally.

I also made some small changes:

* Added ifdef/ifndef back in Makefiles. On reflection, I find it much more readable than `ifeq("$(SOMEVAR)","y")`. I think the chance of accidental errors due to misreading ifeq/ifneq is just as high as the chance of accidental errors due to misspelling a variable name in an ifdef/ifndef.
* Added a disable option in menuconfig for people whose projects use a lot of undefined variables.
* Added some documentation.

rojer:
that's great news! thanks a lot!

however... with the Kconfig changes, won't `ifdef` now always be true?

projectgus:
> however... with the Kconfig changes, won't ifdef now always be true?

No. For whatever reason, ifdef in Makefiles means "variable is defined and has non-empty value"

`#ifdef` in C means "macro is defined" only... but the C header generation is not effected by these changes.

If this becomes confusing for people then we can always change the style back.

rojer:
Ok, well, make is weird. just when you think you understand how it works, you get a surprise like this :)

anyway, thanks a lot! one fewer patch to maintain for me.

Issue #133, opened 2016-12-02
IPv6 packet causes exception


rojer:
I'm getting the following exception shortly (a few seconds) after connecting to WiFi:

```
ets Jun  8 2016 00:22:57

rst:0x10 (RTCWDT_RTC_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0x00
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3ffc0008,len:0
load:0x3ffc0008,len:1964
load:0x40078000,len:3668
load:0x40080000,len:260
entry 0x40080034
I (550) heap_alloc_caps: Initializing heap allocator:
I (550) heap_alloc_caps: Region 19: 3FFC0E9C len 0001F164 tag 0
I (552) heap_alloc_caps: Region 25: 3FFE8000 len 00018000 tag 1
I (561) cpu_start: Pro cpu up.
I (567) cpu_start: Single core mode
I (573) cpu_start: Pro cpu start user code
I (598) rtc: rtc v160 Nov 22 2016 19:00:05
I (609) rtc: XTAL 40M
W (696) cpu_start: failed to load RF calibration data, falling back to full calibration
I (2041) phy: phy_version: 258, Nov 29 2016, 15:51:07, 0, 2
I (2106) cpu_start: Starting scheduler on PRO CPU.
tcpip_task_hdlxxx : 3ffc4824, prio:18,stack:2048
I (2119) wifi: frc2_timer_task_hdl:3ffc635c, prio:22, stack:2048
I (2131) wifi: pp_task_hdl : 3ffc8bf4, prio:23, stack:8192
I (2132) wifi: mode : sta (24:0a:c4:00:2f:9a)
I (2133) foo: event: 2
I (3008) wifi: n:6 0, o:1 0, ap:255 255, sta:6 0, prof:1
I (3657) wifi: state: init -> auth (b0)
I (3658) foo: ping
I (3670) wifi: state: auth -> assoc (0)
I (3675) wifi: state: assoc -> run (10)
I (3686) wifi: connected with Cesanta, channel 6
I (3687) foo: event: 4
I (4658) foo: pong
I (5658) foo: ping
I (6658) foo: pong
I (6759) event: ip: 192.168.1.40, mask: 255.255.255.0, gw: 192.168.1.1
[...]

PR #149, opened 2016-12-07
Add backtrace to exception output.

rojer:
On Xtensa windowed ABI backtrace is easy to do, and it's immensely
useful for debugging, so - do it by default.
We try to be careful and not deref bogus pointers while walking the
frames.

Example output (from debugging espressif/esp-idf#133):

Backtrace: 0x1:0x3ffc51e0 0x400e9dfa:0x3ffc5210 0x400e9ebc:0x3ffc5230 0x400ec487:0x3ffc5260

With just addr2line utility from binutils, this already pinpoints the
location of the crash.

### DIFF
--- a/components/esp32/panic.c
+++ b/components/esp32/panic.c
@@ -45,7 +45,7 @@ Note: The linker script will put everything in this file in IRAM/DRAM, so it als
 
 #if !CONFIG_ESP32_PANIC_SILENT_REBOOT
 //printf may be broken, so we fix our own printing fns...
-inline static void panicPutchar(char c) {
+inline static void panicPutChar(char c) {
 	while (((READ_PERI_REG(UART_STATUS_REG(0))>>UART_TXFIFO_CNT_S)&UART_TXFIFO_CNT)>=126) ;
 	WRITE_PERI_REG(UART_FIFO_REG(0), c);
 }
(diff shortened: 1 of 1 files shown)


components/esp32/panic.c
```diff
@@ -228,6 +229,32 @@ static void disableAllWdts() {
 
 #endif
 
+static inline bool stackPointerIsSane(uint32_t sp) {
+	return !(sp < 0x3ffae010 || sp > 0x3ffffff0 || ((sp & 0xf) != 0));
+}
+static void putEntry(uint32_t pc, uint32_t sp) {
+	char buf[24];
+	if (pc & 0x80000000) pc = (pc & 0x3fffffff) | 0x40000000;
+	sprintf(buf, " %p:%p", (void *) pc, (void *) sp);
```
projectgus: Could you please modify this to use the panicPutHex function? The rest of the panic functionality has been written to avoid newlib, it would be nice to keep the same here.

rojer: done. i made a couple minor changes: panicPutchar -> panicPutChar (sorry, OCD :)) and also added `0x` to hex numbers - both addr2line and gdb take hex numbers with 0x, so this is more copy-paste friendly.

projectgus:
Thanks! This is in our internal review & merge queue now.

PR #151, opened 2016-12-08
Create temp config in the same dir as target

rojer:
Fixes case where SDK and app are on different mounts

### DIFF
--- a/tools/kconfig/confdata.c
+++ b/tools/kconfig/confdata.c
@@ -771,7 +771,7 @@ int conf_write(const char *name)
 	sprintf(newname, "%s%s", dirname, basename);
 	env = getenv("KCONFIG_OVERWRITECONFIG");
 	if (!env || !*env) {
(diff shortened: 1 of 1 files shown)


projectgus:
I'm a little lost by what this changes. From what I can see, "tmpname" is always created inside directory `dirname`, it's just a case of the file being called `.tmpconfig.PID` or `sdkconfig.tmp.PID`.

Am I missing some detail?

rojer:
no, it's not just the name of the file, obviosuly.
`dirname` is actually empty and `basename` contains the path. this seems a bit weird, but i didn't dig deeper - see how 3 lines above `newname` is constructed from `dirname` and `basename`? i basically took that and added a temp suffix to it, that's all. as long as `newname` ends up correct, `tmpname` should too.

rojer:
i made it more obvious by using `newname` directly.

projectgus:
Got it, thanks.

PR #155, opened 2016-12-09
Remove alignment reqs from spi_flash_{read,write}

rojer:


projectgus:
This looks really good, thanks for submitting it.

The host-side test program looks good, but it might be better long-term to replace it with some on-device unit tests that can run in our CI process. These run via the unit test app:
https://github.com/espressif/esp-idf/tree/master/tools/unit-test-app

Is this something you'd be interested in adding to the PR?

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,147 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
 
-        free(temp_write_buf);
-        temp_write_buf = NULL;
-        size -= 1024;
-        dest_addr += 1024;
-        src = (uint8_t*)src + 1024;
+    SpiFlashOpResult rc = SPI_FLASH_RESULT_OK;
+    COUNTER_START();
+    spi_flash_disable_interrupts_caches_and_other_cpu();
+    const char *srcc = (const char *) srcv;
```
rojer: we never actually deref the pointer as char, we only use it as such in pointer arithmetic. as long as memcpy handles alignment requirements, we should be good.

rojer:
i'll take a look at the unit test framework and see if i can get the test to run there.

components/spi_flash/flash_ops.c
tzxespressio: Hi rojer，I think if this api want to write data in Dcache to falsh,and the pointer point to memory of Dcache, but spi_flash_disable_interrupts_caches_and_other_cpu() closed cache,so when you call memcpy(((uint8_t *) &t) + (dst - left_off), srcc, left_size); I think it will crash,I'm not sure,you can test it.

igrr: I think our current code checks if the source pointer is in DCACHE somewhere at the top of spi_flash_write function, and bails out with invalid argument error if that is the case.

tzxespressio: I think current code don't need check if the source pointer is in DCACHE, it use "memcpy" to copy data of Dcache to heap

rojer: ah, i misunderstood what the comment was about. makes sense.
i don't see any checks in the [current code](https://github.com/espressif/esp-idf/blob/72422b3/components/spi_flash/flash_ops.c#L130) either, but it will indeed work because it always copied the bytes out to an external DRAM buffer before disabling the cache.
so yes, there's that side effect. i tried to avoid allocating an external buffer if possible, but do you think writing from dcache is important?

rojer: i can add a check for specific address range and make it fall through to copying via temp buffer in that case too (even if addresses are aligned).

igrr: It was there, got lost in previous changeset somehow:
https://github.com/espressif/esp-idf/blob/be53a6950c64151d98aae421008df28c15a66753/components/spi_flash/flash_ops.c#L141-L146

tzxespressio: yes,I think it a better way

projectgus: that sounds like a good approach to me (using the temporary buffer path for either DROM or unaligned data).

rojer: ok, all done. i converted the PC-based test into device-based. it seems too expensive to run a full set of combinations - it would take too long and wear out the flash, so i kept a few key test points that should test important edge cases.

added a check for source locality when writing. some remaining things:

1) similar check for read path? if we allow writing from non-dram regions, it makes sense to allow reading into them.
2) non-aligned reads and writes to and from regions that require 32-bit aligned access will actually fail, because memcpy is not load32-aware. i'm not sure how/whether do deal with this. i'd expect mem* routines to be aware of this and do the right thing for me.

projectgus:
Thanks @rojer , this looks good to me but the others may have some input once Shanghai wakes up.

I was a little curious about how much overhead the 32 byte block version has, so I made a very quick benchmark that erases 64KB and then writes 4x16KB blocks.

* This PR as-is (aligned writes), ~500ms to erase & write.
* This PR, forced to use unaligned slow path, ~600ms to erase & write.
* Slow path modified to pull enable/disabling caches out of the inner loop, ~500ms to erase & write. ie same as the "fast path".

So the only measurable overhead for large writes seems to come from flash cache & CPU coordination, rather than copying the data. Someone could potentially rewrite the spi_flash_write function a bit based on this information, but I'm perfectly happy with it as-is.

> similar check for read path? if we allow writing from non-dram regions, it makes sense to allow reading into them.

The flash cache regions (DROM/IROM) are for mapping strings and instruction code from the flash into the address space, so they're read only. Any other memory region remains accessible while flash cache is off, so it should always be safe to read into those from SPI flash.

>     non-aligned reads and writes to and from regions that require 32-bit aligned access will actually fail, because memcpy is not load32-aware. i'm not sure how/whether do deal with this. i'd expect mem* routines to be aware of this and do the right thing for me.

The only way around this (AFAIK) is to make the C library routines know quite a bit about the architecture (or to add unaligned read/write exception handlers to "fix" unaligned reads/writes when they fail). I think it's reasonable to expect people passing IROM/IRAM addresses around to know that they have to use aligned addresses and lengths, at least for now. It's a lot more specialised than calling spi_flash_write with const data (which triggers the DROM use case.)

projectgus:
>  The flash cache regions (DROM/IROM) are for mapping strings and instruction code from the flash into the address space, so they're read only. Any other memory region remains accessible while flash cache is off, so it should always be safe to read into those from SPI flash.

I was clearly half asleep so forgot about SPI RAM, good point. I think it's acceptable to leave this until the SPI RAM features are more mature, though.

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
 {
-    // Destination alignment is also checked in ROM code, but we can give
-    // better error code here
-    // TODO: add handling of unaligned destinations
-    uint8_t *temp_write_buf = NULL;
-    uint8_t pad_head = 0;
-    uint8_t pad_end = 0;
-    SpiFlashOpResult rc;
-    // Out of bound writes are checked in ROM code, but we can give better
+    // Out of bound reads are checked in ROM code, but we can give better
```
igrr: Out of bound writes

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
-        return ESP_ERR_INVALID_SIZE;
-    }
-
-    while(size >= 1024) {
-        // max need pad byte num for 1024 is 4
-        temp_write_buf = (uint8_t*)malloc(1024 + 4);
-        if(temp_write_buf == NULL) {
-            return ESP_ERR_NO_MEM;
-        }
+    if (dst + size > g_rom_flashchip.chip_size) return ESP_ERR_INVALID_SIZE;
```
igrr: This could use same formatting as in the original code — return broken on a separate line from conditional.

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
+     * - Left padding: 4 bytes up to the first 4-byte aligned destination offset.
+     * - Middle part
+     * - Right padding: 4 bytes from the last 4-byte aligned offset covered.
+     */
+    size_t left_off = dst & ~3U;
+    size_t left_size = MIN(((dst + 3) & ~3U) - dst, size);
+    size_t mid_off = left_size;
+    size_t mid_size = (size - left_size) & ~3U;
+    size_t right_off = left_size + mid_size;
+    size_t right_size = (size - mid_size - left_size);
```
igrr: No parens needed here

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
+            spi_flash_enable_interrupts_caches_and_other_cpu();
+            if (rc != SPI_FLASH_RESULT_OK) goto out;
+            COUNTER_ADD_BYTES(write, mid_size);
+        } else {
+            /*
+             * Otherwise, unlike for read, we cannot manipulate data in the
+             * user-provided buffer, so we write in 32 byte blocks.
+             */
+            while (mid_size > 0) {
+                uint32_t t[8];
```
igrr: This line could use a comment similar to the one in `spi_flash_read`.

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
+                uint32_t write_size = MIN(mid_size, sizeof(t));
+                memcpy(t, (srcc + mid_off), write_size);
```
igrr: Extra parens here

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
-    }
-    if (size % 4 != 0) {
-        return ESP_ERR_INVALID_SIZE;
-    }
     // Out of bound reads are checked in ROM code, but we can give better
     // error code here
-    if (src_addr + size > g_rom_flashchip.chip_size) {
-        return ESP_ERR_INVALID_SIZE;
-    }
+    if (src + size > g_rom_flashchip.chip_size) return ESP_ERR_INVALID_SIZE;
```
igrr: Same as in spi_flash_write, original formatting can be kept.

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
+    SpiFlashOpResult rc = SPI_FLASH_RESULT_OK;
     COUNTER_START();
     spi_flash_disable_interrupts_caches_and_other_cpu();
-    SpiFlashOpResult rc = SPIRead((uint32_t) src_addr, (uint32_t*) dest, (int32_t) size);
-    COUNTER_ADD_BYTES(read, size);
+    /* To simplify boundary checks below, we handle small reads separately. */
+    if (size < 16) {
+        uint32_t t[6]; /* Enough for 16 bytes + 4 on either side for padding. */
+        uint32_t read_src = (src & ~3U);
+        uint32_t left_off = (src & 3U);
```
igrr: Extra parens

components/spi_flash/flash_ops.c
```diff
@@ -127,121 +128,162 @@ esp_err_t IRAM_ATTR spi_flash_erase_range(uint32_t start_addr, uint32_t size)
...
+        uint32_t read_size = (left_off + size + 3) & ~3U;
+        rc = SPIRead(read_src, t, read_size);
+        if (rc != SPI_FLASH_RESULT_OK) goto out;
+        COUNTER_ADD_BYTES(read, read_size);
+        memcpy(dest, ((char *) t) + left_off, size);
+        goto out;
+    }
+    char *dst = (char *) dest;
```
igrr: You can use same `dstv`/`dstc` naming pattern here, as in `spi_flash_write`

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+
+#include <unity.h>
+#include <esp_spi_flash.h>
+#include <rom/spi_flash.h>
+#include "../cache_utils.h"
+#include "soc/timer_group_struct.h"
+#include "soc/timer_group_reg.h"
+
+/* Base offset in flash for tests. */
+#define FLASH_BASE 0x100000
```
igrr: IIRC, the factory partition ends at only at 0x110000

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+
+static void fill(char *dest, int32_t start, int32_t len)
+{
+    for (int32_t i = 0; i < len; i++)
+    {
```
igrr: This file could use the same formatting style as other files (~ `astyle --style=otbs`)

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+
+static void IRAM_ATTR test_read(int src_off, int dst_off, int len)
+{
+    uint32_t src_buf[16];
+    char dst_buf[64], dst_gold[64];
+    fprintf(stderr, "src=%d dst=%d len=%d\n", src_off, dst_off, len);
+    memset(src_buf, 0xAA, sizeof(src_buf));
+    fill(((char *) src_buf) + src_off, src_off, len);
+    ESP_ERROR_CHECK(spi_flash_erase_sector((FLASH_BASE + src_off) / SPI_FLASH_SEC_SIZE));
+    spi_flash_disable_interrupts_caches_and_other_cpu();
```
igrr: Does this need an `extern` declaration?

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+    assert(SPIWrite(FLASH_BASE, src_buf, sizeof(src_buf)) == SPI_FLASH_RESULT_OK);
+    spi_flash_enable_interrupts_caches_and_other_cpu();
+    memset(dst_buf, 0x55, sizeof(dst_buf));
+    memset(dst_gold, 0x55, sizeof(dst_gold));
+    fill(dst_gold + dst_off, src_off, len);
+
+    ESP_ERROR_CHECK(spi_flash_read(FLASH_BASE + src_off, dst_buf + dst_off, len));
+    assert(cmp_or_dump(dst_buf, dst_gold, sizeof(dst_buf)) == 0);
```
igrr: Suggest using Unity's assertion macros

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+TEST_CASE("Test spi_flash_read", "[spi_flash_read]")
+{
+#if 1
```
igrr: How about `#if CONFIG_SPI_FLASH_MINIMAL_TEST` or something similar?

components/spi_flash/test/test_read_write.c
```diff
@@ -0,0 +1,207 @@
...
+  test_read(0, 0, 0);
```
igrr: Indenting here is different from the rest of the code.

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/flash_ops.c
rojer: yes, i'm a bit liberal with parens sometimes. removed.

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/flash_ops.c
rojer: but here we're not doing any padding, this is just the block size. 8 * 4 = 32, just like noted above.

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/flash_ops.c
rojer: done

components/spi_flash/test/test_read_write.c
rojer: ok, let it be 0x120000

components/spi_flash/test/test_read_write.c
rojer: ah, that's what it is. reformatted with `astyle --style=otbs`

components/spi_flash/test/test_read_write.c
rojer: it's declared in `cache_utils.h`

components/spi_flash/test/test_read_write.c
rojer: fixed by `astyle`.

components/spi_flash/test/test_read_write.c
rojer: done. however, i couldn't identify a suitable config category to add this to (there isn't currently one for tests or spi_flash), so i'll leave it to you.

projectgus:
Thanks @rojer. I've squashed these into a single commit and put them into our internal review/merge queue.

projectgus:
Merged in d246836. Thanks!

PR #163, opened 2016-12-14
Allow gw to be null

rojer:
Allow interfaces to be configured without a default gateway, for
local-only communication.

In case of the AP interface, if gw is not set, do not offer it.


projectgus:
Looks good to me. Have submitted for review/merge.

PR #186, opened 2016-12-29
Change of comment

nkolban:
A comment had the word OUTPUT when INPUT was meant.


projectgus:
LGTM. Do you have some more cleanup-type changes you plan to add to this branch before merging?

negativekelvin:
Gpio driver just changed so this is no longer relevant

nkolban:
@projectgus I'll be delighted to follow any practices or procedures you wish.  My thinking was that as I found a source file that wanted these trivial typos changed, I would make the change on Github itself which makes a commit.  I would then create a Pull Request ... but only if I don't have one already open for previous trivial changes.   If there is one already open, then the commit of subsequent changes gets rolled into the existing pull request.

My thinking is that I will be dabbling in ESP-IDF source files and headers for a long while to come.  If I create a private branch in my repo, then would be the correct time to issue the pull-request?  My thinking is that there are likely periodic times within Espressif where someone sits down to examine pull-requests and those that are accepted, get consumed.   At that point, my subsequent changes would be part of a new pull requested started at that time.

Again ... I'm delighted to fall in line with any policy or procedures y'all wish to come up with ... if the current procedure I am following isn't good, please think through and write up one that ideally all the community members can follow.

igrr:
Part of this MR has been merged in https://github.com/espressif/esp-idf/commit/3efbbab2cfd4a6766c1d9ffe182d447ec58f247a, remaining changes are not relevant any more. Thanks for the contribution!

PR #190, opened 2016-12-30
BLE Advertisement flags not working

inwotep:
I noticed the flag parameter of the ESP API structs was not being passed through correctly.  This patch fixes it.  Additionally, some users have reported not being able to connect on certain Android devices: http://www.esp32.com/viewtopic.php?t=787.  I verified this on my Sony Xperia Z3 with Android 5.1.1.  Specifying the BR/EDR unsupported flag in the advertisement fixes it on my device, but unsure about others.


projectgus:
Thanks for contributing these fixes, and sorry for the very slow reply.

It seems @TianHao-Espressif fixed these bugs in daf58e3 &  d6fcec. Thanks again for letting us know about them.

Xplorer001:
Glad this was fixed, had struggled for 4 days with this. Was learning BLE and was thinking I was doing something wrong with the good. Am able to connect know from Moto G 2nd Gen Android 5.1 Thanks!

PR #214, opened 2017-01-10
Example 10_openssl_server should use TLSv1.2 instead of SSLv3

MalteJ:
For security reasons modern browsers, curl and openssl do not support the SSLv3 protocol anymore.
You should use TLSv1.2 instead.

Also there was a bug in the HTTP header.


projectgus:
Thanks for sending this PR, sorry for the slow reply. I've rebased it against current master and submitted it to our internal review & merge queue.

projectgus:
Merged with a minor revert in 34d0b00:

After some internal discussion, the openssl_server example uses TLS_Server_Method() rather than TLS v1.2 for compatibility with older browsers. A comment has been added recommending TLS v1.2 where possible, for better security.

PR #217, opened 2017-01-12
gatts_demo: re-enter advertising after client disconnects

mikeryan:
In the existing tree, after a connected client disconnects the device
will never re-enter the advertising state. Additional clients will not
be able to discover or connect to the device. This patch forces the
device back into the advertising state upon disconnection.


projectgus:
Thanks Mike. Sorry for the delay, I've put this fix into our internal merge queue.

mikeryan:
Thanks!

PR #249, opened 2017-01-16
Allow writes to encrypted partitions

rojer:
There is a size alignment requirement but it is checked by
spi_flash_write_encrypted. However, this check flat-out bans encrypted
writes.


projectgus:
Cherry-picked as 71ab455c8719a683eed20bddff8712069e3d0f3c. Thanks @rojer!

PR #257, opened 2017-01-17
Updated routines in intr_alloc and cross_core to stay in IRAM

nathanjel:
I got randomly a GuruMeditation with a IllegalInstruction while using SPI Flash commands, referring to line 164 of cache_utils.c - where spi_flash_enable_interrupts_caches_no_os() calls esp_intr_noniram_enable(). In theory the call is made after the spi_flash_restore_cache() call happens, but somehow it did not work well. Also I thought it might be better to have a function called from a IRAM function to also be in IRAM. After applying, the issues with spi_flash_* did not occur again.

There is also a fix to add another cross-core interrupt function to IRAM. It's called from an another IRAM function (vPortYieldOtherCore()) and I have encountered IllegalInstruction issues with this one as well.

### DIFF
--- a/components/esp32/crosscore_int.c
+++ b/components/esp32/crosscore_int.c
@@ -82,7 +82,7 @@ void esp_crosscore_int_init() {
     assert(err == ESP_OK);
 }
 
-void esp_crosscore_int_send_yield(int coreId) {
+void IRAM_ATTR esp_crosscore_int_send_yield(int coreId) {
     assert(coreId<portNUM_PROCESSORS);
     //Mark the reason we interrupt the other CPU
     portENTER_CRITICAL(&reasonSpinlock);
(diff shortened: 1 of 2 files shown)


projectgus:
Hi @nathanjel,

There have been a couple of changes on the master branch recently that relate to this (including 4676d159adc6e92ab90546a618c340b9a5814459). Can you please try with the latest master and see if these changes are still necessary?

Thanks,

Angus

nathanjel:
Hi,

Currently on business trip. Will check on 24th Jan.

projectgus:
Looks like the changes for disabling/enabling interrupts have been incorporated in another change already.

I've cherry-picked the change that moves esp32_corss_core_send_yield() to IRAM and added it to our internal review queue.

nathanjel:
Thanks, this is perfect.

projectgus:
Merged in 174c313

PR #267, opened 2017-01-19
Re-enable interrupts_caches_and_other_cpu after...

rojer:
...encrypted write.


projectgus:
Thanks for sending this, @rojer. This change should be taken care of in 6bc9cd39e9a16d5e5de7ae8ce93a4d7933ceb444 (which also adds 16-byte-aligned writes.) PTAL.

PR #269, opened 2017-01-19
gpio: Attach gpio intr to the isr_server on the right core.

thaytan:
Make sure GPIO interrupts get attached to the core that
the ISR service routine is running on, otherwise they
don't arrive.


projectgus:
Thanks Jan. I've pushed these into our internal review/merge queue, they should appear on github master branch soon.

PR #272, opened 2017-01-20
Enable lwip PPPoS support

amuzyka-grinn:
In PoC phase of my project I need to use ppp connection over GSM module, so
I successfully enabled lwip pppos api and I think it would be nice to add this feature to mainline. 

What comes with this patch:

 * Fix some lwip pppos api bugs
 * Added PPP_SUPPROT parameter to lwip Kconfig
 * Added example pppos_client

Reviews welcome!

Adrian


components/esp32/Kconfig
```diff
@@ -117,6 +117,11 @@ config SYSTEM_EVENT_TASK_STACK_SIZE
     help
         Config system event task stack size in different application.
 
+config TCPIP_TASK_STACK_SIZE
+    int "Tcp IP loop task stack size"
```
projectgus: Suggest calling this "TCP/IP Task Stack Size"

Under "help" suggest this text:

```
Configure TCP/IP task stack size, used by LWIP to process multi-threaded TCP/IP operations. The default is 2048 bytes, setting this stack too small will result in stack overflow crashes.
```

examples/protocols/pppos_client/README.md
```diff
@@ -0,0 +1,9 @@
+# PPPoS client example
```
projectgus: To make it easier for people to find this, suggest writing:
```
# PPP over Serial (PPPoS) client example
```
here.

amuzyka-grinn:
Hi,
I changed code following to your comments.
In general, I'm not going to use other ip interfaces, however, I checked if wifi can somehow disturb PPP connection, and an only thing is default ip route setting. Wifi during initialization is setting itself as the default route, so, you have to take care of initialization sequence depending what default interface you want. 
Btw: I thought about using two interfaces, but (maybe I'm wrong), I don't see in 'lwip' that multi IP source support is enabled (LWIP_HOOK_IP4_ROUTE_SRC), am I right?

loboris:
What is the current status of this pull request ?

amuzyka-grinn:
For me, OK. The code should be able to merge smoothly.

projectgus:
We're still keen to merge this into IDF, but it won't happen until after we take a branch for release v2.0 (any day now.)

Also we have to think through to what extent we (Espressif) support this. Because it's a core ESP-IDF feature, but not one we've tested ourselves or put into our internal QA process. I'm thinking we'll probably incorporate it and mark it as unsupported for now, although we may start supporting it more actively in the future (as it's an awesome feature!)

danicampora:
+1 on enabling this and making it part of the official IDF. At Pycom we are going to be using it very soon for adding cellular connectivity.

Smeedy:
Though I dislike the me-too comments, +1 on supporting PPP as this can boost ESP32 deployments in the field.

projectgus:
Thanks everyone for being patient regarding this change.

I've rebased this branch on master and pushed it into our internal review/merge queue for further discussion.

projectgus:
Thanks again for bearing with us. This PR has been cherry-picked into master as 47c722d.

PR #288, opened 2017-01-23
gatts_demo:  p_service_uuid for play with android

anakinsw:
Modifying parameter so that Android >= 5.0 can connect and see the services
Set p_service_uuid = NULL


seopyoon:
I use Android 5.0 to test with Bluetooth Gatt Server example, and I see the services without setting the parameters to NULL.  I wonder what is causing the issue in your case.

projectgus:
The current gatts_demo also works for me (Android 7.1). Able to connect, read & write the example characteristics.

Do you have any background on why "Android >= 5.0" seemed to have this requirement?

projectgus:
Closing this as it seems like most versions of Android work with the current version.

If you have more information on why/where this change is necessary, please reopen.

Issue #253, opened 2017-01-17
[TW#12229] OTA & flash encryption are currently incompatible


rojer:
I enabled flash encryption but not secure boot.
I burned a reusable key (so i can reflash in development), flashed plain text FW and booted the device, which proceeded to encrypt app partitions but not OTA data.
Subsequent boot failed due to, i think, `otadata` partition not having been encrypted.
This is not mentioned in the guide and i think it should just be encrypted unconditionally, just like app partitions, since it is accessed by the boot loader using mmap.

```
I (47) boot: ESP-IDF v1.0-374-gf9eff70 2nd stage bootloader
I (47) boot: compile time 15:55:31
I (47) boot: Enabling RNG early entropy source...
I (67) boot: SPI Speed      : 40MHz
I (79) boot: SPI Mode       : DIO
I (92) boot: SPI Flash Size : 4MB
I (104) boot: Partition Table:
I (115) boot: ## Label            Usage          Type ST Offset   Length
I (138) boot:  0 nvs              WiFi data        01 02 00009000 00004000
I (161) boot:  1 otadata          OTA data         01 00 0000d000 00002000
I (185) boot:  2 app_0            OTA app          00 10 00010000 00100000
I (208) boot:  3 fs_0             Unknown data     01 82 00110000 00020000
I (231) boot:  4 app_1            OTA app          00 11 00130000 00100000
I (254) boot:  5 fs_1             Unknown data     01 82 00230000 00020000
I (278) boot: End of partition table
I (397) boot: Checking flash encryption...
W (397) flash_encrypt: Using pre-loaded flash encryption key in EFUSE block 1
I (401) flash_encrypt: Setting CRYPT_CONFIG efuse to 0xF
I (455) flash_encrypt: Disable UART bootloader encryption...
I (456) flash_encrypt: Disable UART bootloader decryption...
I (465) flash_encrypt: Disable UART bootloader MMU cache...
I (484) flash_encrypt: Disable JTAG...
I (2378) flash_encrypt: Encrypting partition 2 at offset 0x10000...
I (40813) flash_encrypt: Encrypting partition 4 at offset 0x130000...
ts Jun  8 2016 00:22:57

rst:0x3 (SW_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0x00
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[...]

PR #293, opened 2017-01-24
Sync PART_FLAG_ENCRYPTED value with gen_esp32part.py

rojer:
Currently paritions marked as encrypted by gen_esp32part.py are not
recognized as such and encrypted writes don't work.

This is part of espressif/esp-idf#253

### DIFF
--- a/components/esp32/include/esp_flash_data_types.h
+++ b/components/esp32/include/esp_flash_data_types.h
@@ -64,7 +64,7 @@ typedef struct {
 #define PART_TYPE_END 0xff
 #define PART_SUBTYPE_END 0xff
 
-#define PART_FLAG_ENCRYPTED (1<<0)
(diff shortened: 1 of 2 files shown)


projectgus:
Oh dear, an off-by-one error. Thanks for noticing this.

I think, on balance, I'd rather we fixed this in gen_esp32part.py as I don't think we have a use for bit 0, and it means people with bootloaders already flashed (ie secure boot on) can still use the flag.

If using bit 0 doesn't cause a problem for your use cases then I'm happy to make that change myself rather than forcing you to send another one line PR. :)

rojer:
i chose to update the test because that would make new apps compatible with existing partition tables, but either way works for me. amended.

projectgus:
Thanks, I've squashed these and put them in our internal review & merge queue.

PR #295, opened 2017-01-24
Major refactoring of OTA (IDFGH-1431)

rojer:
Code is simplified: selector stores simple pointer to subpartition
instead of performing calculations based on sequence. Sequence now
serves as a simple tie-breaker.

Logic to determine boot partition is removed. Instead, boot loader now
passes the offset of the image used to boot, this means that logic used
to select image doesn't have to be in sync between boot loader and
library code.

Initial configuration now works with encrypted flash (encrypted write is
performed when needed).

Fixes espressif/esp-idf#253

### DIFF
--- a/components/esp32/cpu_start.c
+++ b/components/esp32/cpu_start.c
@@ -91,12 +91,15 @@ extern volatile int port_xSchedulerRunning[2];
 
 static const char* TAG = "cpu_start";
 
+uint32_t g_boot_image_flash_addr;
+
 /*
  * We arrive here after the bootloader finished loading the program from flash. The hardware is mostly uninitialized,
  * and the app CPU is in reset. We do have a stack, so we can do the initialization in C.
+ *
+ * Boot loader passes flash offset of our image so we can later know for sure which partition was used for boot.
  */
-
-void IRAM_ATTR call_start_cpu0()
+void IRAM_ATTR call_start_cpu0(uint32_t boot_image_flash_addr)
 {
 #if CONFIG_FREERTOS_UNICORE
     RESET_REASON rst_reas[1];
@@ -127,6 +130,8 @@ void IRAM_ATTR call_start_cpu0()
 
     //Clear BSS. Please do not attempt to do any complex stuff (like early logging) before this.
     memset(&_bss_start, 0, (&_bss_end - &_bss_start) * sizeof(_bss_start));
+
+    g_boot_image_flash_addr = boot_image_flash_addr;
 
     /* Unless waking from deep sleep (implying RTC memory is intact), clear RTC bss */
     if (rst_reas[0] != DEEPSLEEP_RESET) {
--- a/components/app_update/esp_ota_ops.c
+++ b/components/app_update/esp_ota_ops.c
@@ -26,6 +26,7 @@
 #include "esp_partition.h"
 #include "esp_spi_flash.h"
 #include "esp_image_format.h"
+#include "esp_ota_select.h"
 #include "esp_secure_boot.h"
 #include "esp_flash_encrypt.h"
 #include "sdkconfig.h"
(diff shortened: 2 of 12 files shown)


components/bootloader/src/main/esp32.bootloader.ld
```diff
@@ -21,7 +21,7 @@ MEMORY
 }
 
 /*  Default entry point:  */
-ENTRY(call_start_cpu0);
+ENTRY(boot_loader_entry);
```
projectgus: Although I agree it's a bit odd, the call_start_cpu0 function name is a a convention in two places (here and the app entry point) and it's documented in the description of the boot flow in "doc/general-notes.rst".) Would prefer keeping the existing name here.

components/esp32/cpu_start.c
```diff
@@ -88,12 +88,15 @@ extern volatile int port_xSchedulerRunning[2];
...
+
 /*
  * We arrive here after the bootloader finished loading the program from flash. The hardware is mostly uninitialized,
  * and the app CPU is in reset. We do have a stack, so we can do the initialization in C.
+ *
+ * Boot loader passes flash offset of our image so we can later know for sure which partition was used for boot.
  */
-
-void IRAM_ATTR call_start_cpu0()
+void IRAM_ATTR call_start_cpu0(uint32_t boot_image_flash_addr)
```
projectgus: There might be a more straightforward way to do this (possibly examining the MMU config), although I don't know what it is. @Spritetm may have some ideas.

projectgus:
This looks like a really good refactor to me.

I'm going to discuss merging strategy with some of my other colleagues, we're instigating a semantic versioning policy so non-backwards-compatible changes of this kind may have to wait until V3.0. I'll get back to you on this, though.

components/esp32/cpu_start.c
rojer: yeah, i actually looked at flash_mmap.c and couldn't figure it out quickly. this seemed more straightforward to me, and also more bullet-proof, because who knows what was subsequently done to mappings - we may be running from who knows where now (i can imagine user wishing to remap executable area somewhere else), what matters for OTA is what happened at boot. so i went for this approach.

components/bootloader/src/main/esp32.bootloader.ld
rojer: fair enough, done.
(my goal was to make boot loader entrypoint name distinct from the app entry point name)

rojer:
thanks. i'm trying to reduce the amount of patches in my branch, would be great if this could go in now.
i can look into making this backward-compatible if need be.

rojer:
it's still pretty early days, there can't be that many critical deployments relying on OTA so i gather slipping this in should be fine. but it's up to you of course.

projectgus:
Unfortunately we are at the point where we need to consider backwards compatibility.

Similar to the partition flag case, if secure boot is enabled then bootloader is non-updatable (also non-OTA-updatable) so having the app recognise the old OTA data format and use this instead would be ideal.

The same thing applies to the entry point ABI, unfortunately. This should definitely be possible by looking at where MMU is mapped, I'll try and come up with a sample for this.

rojer:
ok, in case of non-updatable boot loader we not only need app to recognize the format, but also emit it.
hm. that complicates the situation a bit. i can keep the old implementation of `esp_ota_{get,set}_boot_partition` and probe boot selector format to use the right one. would that suit?

projectgus:
Sorry for the lack of updates on this. We have some fixes coming for OTA support on encrypted flash as part of the upcoming V2.0 release. They should appear on github soon. They partially overlap with these changes.

V2.0 will keep the existing "old" bootloader OTA data format. We're still interested in changing this to a simpler data format, but we'll have to look at that after V2.0 (and it will involve backwards compatibility with V1.0/V2.0 bootloaders, as discussed.)

projectgus:
(Oops I didn't meant to close the PR)

projectgus:
Hi @rojer,

Sorry, we've neglected this PR and left you updating the fork for a long time now.

There are a bunch more bootloader-related changes (major performance improvements, and better robustness if OTA data or partitions are corrupt) landing in preparation for IDF V3.0, they should appear on github in the next few days.

I think they fix most of the concerns this patchset was also intended to resolve, apart changing from the OTA data partition format itself. Happy to discuss whether anything else needs doing around that (although the existing OTA data format is not great, I'd rather not make non-backwards-compatible changes if they're not going to add new functionality.)

(Also, I'm about to go on vacation so I may not see your comments. I'll be back in mid-August and will look in on this issue again then.)

rojer:
@projectgus being able to control custom flags is a must for us. ok, let's reconvene in august and see where we stand.

projectgus:
Closing as the IDF OTA system is quite different to when this PR was submitted.

PR #296, opened 2017-01-25
Allow providing custom PARTITION_TABLE_CSV_PATH

rojer:
Provide direct absolute path to CSV, without project_path


projectgus:
Thanks for this. I've added it to our internal review/merge queue.

PR #301, opened 2017-01-26
Expand environment variables in gen_esp32part

rojer:
Allows parametrizing partition table with (exported) make variables.

### DIFF
--- a/components/partition_table/gen_esp32part.py
+++ b/components/partition_table/gen_esp32part.py
@@ -5,8 +5,10 @@
 # Converts partition tables to/from CSV and binary formats.
 #
 # See the sdkng README.md file for details about how to use this tool.
-import struct
 import argparse
+import os
+import re
+import struct
 import sys
 
 MAX_PARTITION_LENGTH = 0xC00   # 3K for partition data (96 entries) leaves 1K in a 4K sector for signature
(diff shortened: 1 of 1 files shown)


components/partition_table/gen_esp32part.py
```diff
@@ -163,7 +164,9 @@ def __init__(self):
     def from_csv(cls, line):
         """ Parse a line from the CSV """
         line_w_defaults = line + ",,,,"  # lazy way to support default fields
-        fields = [ f.strip() for f in line_w_defaults.split(",") ]
+        def expand_var(f):
+            return os.environ[f[1:]] if f.startswith("$") else f
```
projectgus: Hi @rojer,

Sorry, this somehow fell off my radar for review.

I'm 100% OK with allowing environment variables in partition table definitions. I would prefer if we could allow mixing of variable references and plaintext in the same CSV field. Was hoping Python might have a function for this out of the box, but it looks like the best (simple) solution is a one-liner like in this SO answer: http://stackoverflow.com/a/30735354

rojer:
i thought about it too, didn't know about `os.path.expandvars` and decided to keep it simple.

PTAL

projectgus:
Thanks @rojer. LGTM, have put into our internal review/merge queue.

PR #302, opened 2017-01-26
Fix fopen() in “a” (append) mode

nineisk:
fopen() does not work when file is opened in “a” (append) mode


projectgus:
Thanks for sending this, @nineisk . I've added the fix to our internal queue to review & merge.

projectgus:
Cherry-picked in 7eaf1c7. Thanks again!

PR #307, opened 2017-01-28
component/bt: fix bug where uuid was not reset in btc_gatts_act_create_attr_tab

Derrrk:
This bug happens if you try and create an attribute table with UUID != ESP_UUID_LEN_16 (eg. with ESP_UUID_LEN_128). The uuid is set at the beginning of the loop if UUID == ESP_UUID_LEN_16, but will not be reset on the next iteration.


TianHao-Espressif:
Yeah, It is a bug. Thanks for fix it.

projectgus:
Thanks for this @Derrrk , and sorry for the slow reply. This has been squashed and added to our internal review/merge queue.

projectgus:
Cherry-picked in a4a7900. Thanks!

PR #320, opened 2017-02-06
Add link to UART driver

skiz:


negativekelvin:
Not all boards use silabs usb-uart chipsets so may want to qualify this or put it under troubleshooting in step 4

projectgus:
Thanks @skiz for sending this in.

This is something we're looking into internally.  We have board-specific Getting Started guides which includes links to drivers, and the ESP-IDF documentation here which doesn't have any hardware-specific stuff in it yet.

The problem with the Getting Started guides is that they can't keep up with new versions of IDF, or the differences between stable & master IDF versions.

What we want to do is add some hardware-specific getting started guides to the docs, with links to the appropriate drivers (not always silabs, as noted.) This will be the long-term fix for this.

PR #350, opened 2017-02-17
Fix hostname set in tcpip_adapter_lwip

alarruskain:
hostname por each interface is not correctly stored in "hostinfo" variable.


projectgus:
Thanks for this, @alarruskain . I've pushed it into our internal review & merge queue.

projectgus:
Cherry-picked in 3119f93. Thanks again!

PR #359, opened 2017-02-20
Clean-up the TCB at the right time.

danicampora:
In the ESP32 port the current statement about clean-up TCB does not apply:

```c
/* This call is required specifically for the TriCore port.  It must be
above the vPortFree() calls.  The call is also used by ports/demos that
want to allocate and clean RAM statically. */
portCLEAN_UP_TCB( pxTCB );
```

Because it's not a tri-core port. What is actually useful is the ability to "release" an statically allocated TCB when a task is deleted. In order to do so, ```portCLEAN_UP_TCB``` can be used, but it needs to be called after all the accesses to the TCB are performed. In the current location, releasing the TCB causes a crash because the TCB is accessed by FreeRTOS several times afterwards. The proposed patch moves the call to the end and inside the check for a statically allocated TCB (where it makes sense).

### DIFF
--- a/components/freertos/tasks.c
+++ b/components/freertos/tasks.c
@@ -3740,11 +3740,6 @@ BaseType_t xTaskGetAffinity( TaskHandle_t xTask )
 
 	static void prvDeleteTCB( TCB_t *pxTCB )
 	{
-		/* This call is required specifically for the TriCore port.  It must be
-		above the vPortFree() calls.  The call is also used by ports/demos that
-		want to allocate and clean RAM statically. */
-		portCLEAN_UP_TCB( pxTCB );
-
 		/* Free up the memory allocated by the scheduler for the task.  It is up
 		to the task to free any memory allocated at the application level. */
 		#if ( configUSE_NEWLIB_REENTRANT == 1 )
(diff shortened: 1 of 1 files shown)


danicampora:
@igrr this one is also needed for Threading to work properly on MicroPython. Related to this other one as well: https://github.com/espressif/esp-idf/pull/358

projectgus:
To be sure I understand the motivation here: the cleanup work in this case will be a callback into the Micropython core, so Micropython's thread accounting knows this task no longer exists?

dpgeorge:
> the cleanup work in this case will be a callback into the Micropython core, so Micropython's thread accounting knows this task no longer exists?

Yes, it needs to free the memory for the stack and TCB itself for the thread that is being deleted.  This memory is allocated on the uPy heap.  You can't free the TCB memory until the very last moment because it's used by FreeRTOS to do the deletion of the thread.

projectgus:
Do you mind sending this one as a new PR (or updating this PR) when you send the Kconfig item for enabling it as a feature?

(If that makes things more difficult for you in terms of maintaining forks then we can merge this earlier, it just seems like they are closely related changes.)

danicampora:
@projectgus OK, I'll create a new PR with this one and https://github.com/espressif/esp-idf/pull/358 plus the Kconfig changes included. Thanks.

danicampora:
Sorry for the long delay. Superseded by: https://github.com/espressif/esp-idf/pull/444

PR #360, opened 2017-02-20
Place sdmmc_isr() in IRAM

danicampora:
To prevent a crash if this interrupts is triggered while the flash memory is locked.


igrr:
Hi Daniel,
The interrupt is allocated without IRAM flag: https://github.com/danicampora/esp-idf/blob/983534b0100f7cb6921250914d43df68da29e5e4/components/driver/sdmmc_host.c#L246, are you seeing an illegal instruction exception happening in this interrupt?

danicampora:
Hi @igrr ,

No, sorry, I just assumed it had to be placed in IRAM. I overlooked that detail. Apologies. Closing it now. Thanks.

Cheers,
Daniel

PR #361, opened 2017-02-20
components/driver: Add width field to the slot config.

danicampora:
Therefore if the width is set to 1, you can choose to only
configure the CLK, DAT0 and CMD pins.


projectgus:
Thanks for this, @danicampora. I've added it to our internal review & merge queue.

This changes the SDMMC API, so there's a chance we'll have to make it backwards-compatible. But I think that may not be necessary as we haven't released a version with the SDMMC driver yet.

danicampora:
@projectgus thanks! looking forward to see it merged :-)

projectgus:
Cherry-picked in c08a287 with some tweaks in 6ee5a1e to keep the pre-existing default behaviour. Thanks again!

PR #366, opened 2017-02-21
Same sense of meaning SW_STRAP

ESP32DE:


projectgus:
Cherry-picked in  f4c4787 , thanks again!

PR #392, opened 2017-03-02
Bugfix/eclipse windows paths

flummigit:


projectgus:
Hi @flummigit,

Oops, sorry - I should have deleted this branch a little while ago! This problem should be taken care of in the current esp-idf master branch (although it was fixed in a slightly different way.)

There are, however, some additional steps in the new docs for Windows users:
http://esp-idf.readthedocs.io/en/latest/eclipse-setup-windows.html

If you follow these and Eclipse still isn't working correctly for you, please open an Issue and we'll look into it.

Angus

PR #398, opened 2017-03-04
Move ethernet phy support into components tree and refactor example app

RobinCutshaw:
I've started adding ethernet support to the examples and it made sense to move the phy support into one place that all of the examples could use.  This is dependent on the previous pull request adding LAN8720 support.


projectgus:
Thanks for sending this and sorry we didn't get back to you about it earlier.

As mentioned in the related PR, I've made some similar changes with the same goal as this. Probably the major philosophical difference is that the component itself doesn't have config options for the PHY - these are left to the programmer (although they're exposed in KConfig for the examples. ) This is because there are some use cases where "code as configuration" is desirable, in particular environments like Arduino & Micropython where the PHY implementation is chosen at runtime.

If you have time to take a look at the temp/ethernet_lan8720 branch at https://github.com/espressif/esp-idf/tree/temp/ethernet_lan8720 (in particular 1b8a1672c) then I'd be keen to hear any thoughts you have on it

Angus

projectgus:
Hi @RobinCutshaw,

I'm going to close this now that 90c8bd9 has landed, as I think it covers the changes here (plus a bit more.) Please let me know if I missed anything.

Angus

PR #399, opened 2017-03-05
Add ethernet version of http_request example

RobinCutshaw:
This pull request includes commits from the previous ethphyrefactor pull request upon which it relies.  Commits fa891de and 3a5b26b are the relevant changes.

### DIFF
--- a/examples/protocols/http_request_ethernet/Makefile
+++ b/examples/protocols/http_request_ethernet/Makefile
@@ -1,0 +1,9 @@
+#
+# This is a project Makefile. It is assumed the directory this Makefile resides in is a
+# project subdirectory.
+#
+
+PROJECT_NAME := http-request-ethernet
+
+include $(IDF_PATH)/make/project.mk
+
--- a/examples/protocols/http_request_ethernet/README.md
+++ b/examples/protocols/http_request_ethernet/README.md
@@ -1,0 +1,5 @@
+# HTTP Request Example
+
+Uses a POSIX socket to make a very simple HTTP request.
+
+See the README.md file in the upper level 'examples' directory for more information about examples.
--- a/examples/protocols/http_request_ethernet/main/Kconfig.projbuild
+++ b/examples/protocols/http_request_ethernet/main/Kconfig.projbuild
@@ -1,0 +1,54 @@
+menu "Example Configuration"
+
+choice PHY_MODEL
+    prompt "Ethernet PHY"
+    default CONFIG_PHY_TLK110
+    help
+        Select the PHY driver to use for the example.
+
+config PHY_TLK110
+    bool "TI TLK110 PHY"
+    help
+        Select this to use the TI TLK110 PHY
+
+config PHY_LAN8720
+    bool "Microchip LAN8720 PHY"
+    help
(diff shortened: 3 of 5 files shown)


krzychb:
You are doing pioneering work adding options and examples for ethernet support :+1:

I propose to rename the new `http_request example` to `http_request_eth` and move it to [examples/etherent](https://github.com/espressif/esp-idf/tree/master/examples/ethernet).

Original example will not gain on complexity and the new example will be better exposed in dedicated folder.

projectgus:
Hi @RobinCutshaw ,

Thanks for sending this in. I think I agree with @krzychb that it may be neater to write a separate ethernet HTTP example rather than add support for both in the WiFi HTTP example.

Now that all of the PHY changes have landed in master branch, are you interested in updating this PR? It would be a fantastic addition to IDF, if you're still keen.

Angus

RobinCutshaw:
I'll be happy to update.  Perhaps close this one and I'll submit based on the latest IDF.

I still like pulling out the common bits into a separate file that is sharable by all.

The plan is to have a modified version for all examples that currently use wifi.  Any prefs on how to separate the examples in the tree?

To be honest, I think each wifi example doesn't need to know about wifi in the base code.  That's why I separated it so that each example worked with wifi or ethernet.  Sounds like you prefer duplicate examples though.

Robin

krzychb:
> To be honest, I think each wifi example doesn't need to know about wifi in the base code.

@RobinCutshaw, you are right and this is how I would do it in an application that requires both Wi-Fi and cable connection. 

As for providing examples we would like each of them to demonstrate [one distinct functionality](https://github.com/espressif/esp-idf/tree/master/examples#contributing-examples) to make it easy / straightforward for people to test this one thing they want.

Therefore I proposed to provide a separate example instead of adding complexity to existing one. If you like to demonstrate coexistence of both Wi-Fi and cable connection,, this is perfectly fine as another / separate example.

Thank you again for contributing code!

projectgus:
Hi @RobinCutshaw,

Thanks for that, looking forward to it. :)

>  Sounds like you prefer duplicate examples though.

On balance, yes. One of the ideas for the examples is that you can take any of them, copy just that example's project directory to anywhere, and use it to start your own standalone project outside the IDF tree. So they are intended to have limited external dependencies (apart from IDF), where possible.

That said, there are some boilerplate patterns in the WiFi examples that may be useful as helper functions in IDF itself, possibly.

> I'll be happy to update. Perhaps close this one and I'll submit based on the latest IDF.

Up to you. You can force-push to this branch again in order to completely update the PR, or close and open a new one.

Angus

projectgus:
(In the particular case of HTTP examples, the other thing that we don't have yet is an HTTP/HTTPS client library. @nkolban has a port of libcurl that he maintains, but we don't have anything lighter-weight yet. That will dramatically shrink the complexity of those examples.)

RobinCutshaw:
I finally had time to resync the remote and fix the example.

mahavirj:
This (ethernet or WiFi connectivity) is now seamlessly handled through common example component at https://github.com/espressif/esp-idf/tree/master/examples/common_components/protocol_examples_common. Closing this PR.

PR #407, opened 2017-03-07
comment out (#) to (//).

hirotakaster:
Can't compile idf when (# CONFIG_WIFI_ENABLED is not set) on make menuconfig, because of this comment out "#".

### DIFF
--- a/components/esp32/include/esp_wifi.h
+++ b/components/esp32/include/esp_wifi.h
@@ -135,7 +135,7 @@ typedef struct {
     .magic = WIFI_INIT_CONFIG_MAGIC\
 };
 #else
-#define WIFI_INIT_CONFIG_DEFAULT #error Wifi is disabled in config, WIFI_INIT_CONFIG_DEFAULT will not work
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks for sending this, but this error is intentional. You can't use WIFI_CONFIG_DEFAULT in your code without enabling the WIFI stack. This check causes a compile time failure, but without it you'll still see a (more obscure) link failure when the WiFi libraries aren't available at link time.

Is there a particular configuration combination we've missed, where you have WIFI_CONFIG_DEFAULT in your code but have disabled WiFi?

hirotakaster:
Hi, I find this comment out miss when I use only the SD card without WiFi setting on "make menuconfig" and I don't use WIFI_CONFIG_DEFAULT on that code (examples/storage/sd_card/main.c). 
And I can compile/work my SD card app with disable CONFIG_WIFI_ENABLED(of course comment out this # only).

projectgus:
Hi @hirotakaster ,

Another fix for this issue has been suggested in #440, which seems like a better plan (less confusing error message, but still allows us to warn on Wifi use.)

I'm going to close this in favour of that PR. Thanks again for bringing this to our attention.

Angus

PR #425, opened 2017-03-13
Update to fix #424 - Logic error in mbedtls network error checking for non blocking sockets

### LINKED ISSUE #424: mbedtls and detecting non-blocking reads during handshaking
If we attempt to use a non-blocking socket with mbedtls we receive a failure.  I dug deeply into this and believe I have found the resolution.

If we look at this function:

https://github.com/espressif/esp-idf/blob/master/components/mbedtls/port/net.c#L204

The intent is to determine if an error reported by a previous socket call was caused because the socket is non blocking.  The return code is 0 if the last error was NOT caused by being a non-blocking socket while the return code is 1 if the last error was because we WERE a non-blocking socket. The logic is:

1. If the socket is a blocking socket , return 0
2. If the last error was because we were non-blocking, return 1.
3. Else ... return 0

This sound sensible enough, however there is perhaps a flaw in the implementation.  The we check "if the last error was because we were non-blocking" is to ASK the socket about its last error.  We don't use `errno` because we are multi-threaded and premptable.  Again, on the surface, that took makes sense.  However in test number 1, we perform an `fcntl` against the socket to determine if it is blocking or non-blocking ... and **that** resets the error condition that was present on the socket when the function was called.   What this means is that the original error is lost.

nkolban:
Updates to fix issue #424

### DIFF
--- a/components/mbedtls/port/net.c
+++ b/components/mbedtls/port/net.c
@@ -203,14 +203,14 @@ int mbedtls_net_bind( mbedtls_net_context *ctx, const char *bind_ip, const char 
  */
 static int net_would_block( const mbedtls_net_context *ctx )
 {
+    int error = mbedtls_net_errno(ctx->fd);
+    
     /*
      * Never return 'WOULD BLOCK' on a non-blocking socket
      */
     if ( ( fcntl( ctx->fd, F_GETFL, 0) & O_NONBLOCK ) != O_NONBLOCK ) {
         return ( 0 );
     }
-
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @nkolban,

Thanks for this, fix looks 100% correct. I've rewritten the commit message and put the commit into our internal review & merge queue.

Angus

projectgus:
Merged in 16e1a27. Thanks again!

PR #438, opened 2017-03-17
if esp-idf is a submodule, then everything except this check works

edmund-huber:
Instead of forking this repo, and then customizing on top of it, (and then later having to reconcile changes), I have a root repo with all my goodies and then an esp-idf submodule.

It just works! Except that esp-idf's make process does a check for `${IDF_PATH}/.git` -- which doesn't exist when esp-idf is a submodule. I don't really see the point of this check, since if esp-idf cannot be manipulated by git, any number of git commands in the makefile will fail and the build will stop.

### DIFF
--- a/make/project.mk
+++ b/make/project.mk
@@ -410,7 +410,7 @@ define GenerateSubmoduleCheckTarget
 check-submodules: $(IDF_PATH)/$(1)/.git
 $(IDF_PATH)/$(1)/.git:
 	@echo "WARNING: Missing submodule $(1)..."
-	[ -d ${IDF_PATH}/.git ] || ( echo "ERROR: esp-idf must be cloned from git to work."; exit 1)
+	[ -e ${IDF_PATH}/.git ] || ( echo "ERROR: esp-idf must be cloned from git to work."; exit 1)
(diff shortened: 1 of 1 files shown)


projectgus:
Hi Edmund,

Thanks for raising this! The way you're using esp-idf is great, and we definitely want to support this configuration.

The reason for the explicit check and error is that this may be the first error message new users get from IDF (for example, if they click "Download" on github then they get a non-functional zip file as github downloads don't include any submodules). So the intention was to make it more helpful than a git error, which would then needs to be interpeted by the new user (possibly someone who hasn't used git before).

Changing the `-d` to `-e` (from 'is directory' to 'exists') should fix the problem while keeping the helpful message, as submodules have a .git file rather than a .git directory. What do you think?

Angus

edmund-huber:
You're right, I misspoke in that `.git` does exist if you're in a submodule -- only that it's not a directory. What you're saying sounds good to me.

Side node: git didn't start making gitfiles for submodules until 1.7.8 apparently: https://github.com/git/git/blob/master/Documentation/RelNotes/1.7.8.txt#L109 . Which was 2013.

projectgus:
LGTM, have cherry-picked into our review & merge queue.

projectgus:
Cherry-picked in 5272bd8, thanks!

PR #441, opened 2017-03-19
pass more arguments to git describe so that it always works even if d…

edmund-huber:
…etached branch, etc

Without this patch, if your `$IDF_PATH` is in a detached head state, you get a bunch of `fatal: No names found, cannot describe anything.` during make and `$IDF_VER` is empty.

This change keeps the behavior of `git describe` the same when you're in a branch/tag, but also allows `git describe` to do something reasonable when in a detached head. Additionally, `--dirty` makes `git describe` make a note if the git repo had uncommitted changes.

### DIFF
--- a/make/project.mk
+++ b/make/project.mk
@@ -186,8 +186,7 @@ endif
 	@echo $(ESPTOOLPY_WRITE_FLASH) $(ESPTOOL_ALL_FLASH_ARGS)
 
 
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks @edmund-huber . I've cherry-picked this into our internal review/merge queue.

projectgus:
Cherry-picked in 6a58e17, thanks!

PR #446, opened 2017-03-20
 	add api spi flash send cmd ( read uid flash ect )

ESP32DE:
add api for spi flash send cmd 
- can used for unique flash chip id

example code:
```
..
#include "esp_spi_flash.h"
#include "string.h"
..
```

```
void uid (void) {
	
	// test
	uint8_t flash_unique_id[8];
   // Get flash unique id
   uint8_t command[13] = {0x4b, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
   uint8_t response[13];

   spi_flash_send_cmd(sizeof(command), command, response);
   memcpy(flash_unique_id, response + 5, sizeof(flash_unique_id));
   
   printf("UID: ");
   for (int i=0; i<8; i++) {
	printf("0x%02x ",flash_unique_id[i]);     
   }
   printf("\n");
}
```

### DIFF
--- a/components/spi_flash/flash_ops.c
+++ b/components/spi_flash/flash_ops.c
@@ -479,6 +479,154 @@ static esp_err_t IRAM_ATTR spi_flash_translate_rc(SpiFlashOpResult rc)
     }
 }
 
+static void IRAM_ATTR spi_master_ops(int unit, unsigned int word_size, unsigned int len, unsigned char *out, unsigned char *in) {
+ unsigned int bytes = word_size * len; // Number of bytes to write / read
+ unsigned int idx = 0;
+
+ /*
+  * SPI data buffers hardware registers are 32-bit size, so we use a
+  * transfer buffer for adapt user buffers to buffers expected by hardware, this
+  * buffer is 16-word size (64 bytes)
+  *
+  */
+ uint32_t buffer[16]; // Transfer buffer
+ uint32_t wd;         // Current word
+ unsigned int wdb;     // Current byte into current word
+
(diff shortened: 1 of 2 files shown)


igrr:
Thanks for the PR, @ESP32DE.

Unfortunately this code can not be added into the ESP-IDF as it is, because we can only accept contributions from authors of the code (first point listed [in the contribution guide](http://esp-idf.readthedocs.io/en/latest/contributing.html#before-contributing)), and this code seems to come from [Lua-RTOS](https://github.com/whitecatboard/Lua-RTOS-ESP32) project.

If you package this code addition as a reusable component, you can post the link to the component repository in this forum topic: https://esp32.com/viewtopic.php?f=2&t=1498. We are planning to make a list of components built for the ESP-IDF, so that it is easier for other developers to find components for their projects.

ESP32DE:
@igrr 
i understand - ok - no problem. thank you.

then is that the same theme with CAN driver then ivan?

[CAN driver is ready from thomas](https://github.com/ThomasBarth/ESP32-CAN-driver work) with few add works from other  and me to this "CAN Driver repo".

the work on menuconfig is ready too, but this complet from me.

then thomas must then do the pullrequest to ESP-IDF
i do the pullrequest for menuconfig

or "better" we do this as 3thrd component then?
only for know what is the best to do in future too.
and that you guys have not more work with us that sometimes not knowing ?

thank you ivan!

best wishes
rudi ;-)

PR #453, opened 2017-03-24
Enable checks for encrypted flash in OTA

rojer:
Even if firmware is compiled without CONFIG_FLASH_ENCRYPTION_ENABLED
Rayionale: CONFIG_FLASH_ENCRYPTION_ENABLED controls whether boot loader
generates keys for encryption or not, but flash encryption can be
configured externally. With this change, it's possible to have boot
loader not generate keys but still have encryption working.

### DIFF
--- a/components/app_update/esp_ota_ops.c
+++ b/components/app_update/esp_ota_ops.c
@@ -46,10 +46,8 @@ typedef struct ota_ops_entry_ {
     const esp_partition_t *part;
     uint32_t erased_size;
     uint32_t wrote_size;
-#ifdef CONFIG_FLASH_ENCRYPTION_ENABLED
     uint8_t partial_bytes;
     uint8_t partial_data[16];
-#endif
     LIST_ENTRY(ota_ops_entry_) entries;
 } ota_ops_entry_t;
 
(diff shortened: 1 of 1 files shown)


components/app_update/esp_ota_ops.c
```diff
@@ -166,7 +163,7 @@ esp_err_t esp_ota_write(esp_ota_handle_t handle, const void *data, size_t size)
                         return ESP_OK; /* nothing to write yet, just filling buffer */
                     }
                     /* write 16 byte to partition */
-                    ret = esp_partition_write(&it->part, it->wrote_size, it->partial_data, 16);
+                    ret = esp_partition_write(it->part, it->wrote_size, it->partial_data, 16);
```
edmund-huber: `&it->part` changed to `it->part`. Judging from the signature of esp_partition_write this wasn't compiling properly..

rojer: yep. a drive-by fix. it generated a compilation warning that must've been overlooked. it could not have possibly worked.

igrr:
Rebased and merged in a845f21, thanks!

PR #456, opened 2017-03-25
give the AP and STA netifs different names for ease of debugging lwip

edmund-huber:
lwip debugging while in AP+STA mode was not great since both netifs were named the same, "en". This changes the AP netif to be called "ap" and the STA netif to be called "st".


igrr:
Merged in 82b7ca9, thanks!

PR #489, opened 2017-04-04
Fix impossible check in uart_set_line_inverse

shinyquagsire23:
uart_set_line_inverse had an impossible check which would fail regardless of the desired mask. Fix now checks for any bits set outside of the mask, in addition to the desired mask not being 0.

If the inverse mask should be able to set to 0 (ie to undo an inverse line), I'd be willing to throw in a fix for that as well. Uncertain if not being able to unset inverse lines is desired or not.


igrr:
Cherry-picked in 8e89a13, thanks for the fix!

PR #492, opened 2017-04-04
Fix rmt_set_tx_thr_intr_en(): Check evt_thresh only in enable path.

### LINKED ISSUE #473: [TW#11603] rmt_driver_uninstall() reports error when executed for an RX channel
I get error logs when executing `rmt_driver_uninstall()` for an RX channel.
This happens due to the check in `rmt_set_tx_thr_intr_en()` on parameter `evt_thresh`. It is set to `0xffff` by `rmt_driver_uninstall()` which is out of range.

Happens most likely also for a TX channel (where the abort is more serious) but I haven't tried that.

devsaurus:
This PR is supposed to fix #473.

The range check on parameter `evt_thresh` of `rmt_set_tx_thr_intr_en()` is not relevant when the interrupt is *disabled*: It should be skipped in this case.


igrr:
Cherry-picked in cc164b0d, thanks for contributing!

PR #495, opened 2017-04-05
components/driver/rmt: Add const qualifier for config and tx data.

devsaurus:
Several function parameters are constant in the way the driver code uses them but aren't formally declared as such. This PR adds the `const` qualifier to enable the application code to feed in constants without compiler warnings:
- configuration set `rmt_config` for `rmt_config()`
- tx data `item` for `rmt_fill_tx_items()`
- tx data `rmt_item` for `rmt_write_items()`


igrr:
Cherry-picked in ae94fe04, thanks for contributing!

PR #500, opened 2017-04-07
fixing http_request_example request content

MalteJ:
request new lines must be `\r\n` as specified in the HTTP RFC.
Also the following logic checks for the socket to be closed by the server. This only happens with HTTP/1.0 not HTTP/1.1. So I have adjusted the protocol in the request, too.

### DIFF
--- a/examples/protocols/http_request/main/http_request_example_main.c
+++ b/examples/protocols/http_request/main/http_request_example_main.c
@@ -46,10 +46,10 @@ const int CONNECTED_BIT = BIT0;
 
 static const char *TAG = "example";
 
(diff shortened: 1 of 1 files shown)


MalteJ:
See also https://www.esp32.com/viewtopic.php?f=13&t=1620

igrr:
Thanks for the PR @MalteJ, this (and the identical change to https_request example) is now in our merge queue.

MalteJ:
Great! Thank you!

igrr:
Merged in 8bc153e, thanks!

PR #502, opened 2017-04-08
Replace to strncmp from strcmp on examples/bluetoot/gatt_client

asukiaaa:
Thanks for useful example projects.

I tried to use GATT client project but I cannot establish connection between server and the client.
I found that `strcmp` does not return expected result on my environment so I changed the value.

My environment
- OS: Ubuntu16.10 64bit
- GCC: version 6.2.0 20161005 (Ubuntu 6.2.0-5ubuntu12)

After this modification, GATT client proceed connection to found server device.

Thanks.

### DIFF
--- a/examples/bluetooth/gatt_client/main/gattc_demo.c
+++ b/examples/bluetooth/gatt_client/main/gattc_demo.c
@@ -317,7 +317,7 @@ static void esp_gap_cb(esp_gap_ble_cb_event_t event, esp_ble_gap_cb_param_t *par
             }
 
             if (adv_name != NULL) {
-                if (strcmp((char *)adv_name, device_name) == 0) {
+                if (strncmp((char *)adv_name, device_name, adv_name_len) == 0) {
                     ESP_LOGI(GATTC_TAG, "Searched device %s\n", device_name);
                     if (connect == false) {
                         connect = true;


examples/bluetooth/gatt_client/main/gattc_demo.c
```diff
@@ -314,10 +315,12 @@ static void esp_gap_cb(esp_gap_ble_cb_event_t event, esp_ble_gap_cb_param_t *par
             ESP_LOGI(GATTC_TAG, "Searched Device Name Len %d\n", adv_name_len);
             for (int j = 0; j < adv_name_len; j++) {
                 ESP_LOGI(GATTC_TAG, "%c", adv_name[j]);
+                adv_name_chars[j] = adv_name[j];
             }
+            adv_name_chars[adv_name_len] = '\0';
 
             if (adv_name != NULL) {
-                if (strcmp((char *)adv_name, device_name) == 0) {
+                if (strcmp(adv_name_chars, device_name) == 0) {
```
igrr: Am I right to think that it would be sufficient to replace this `strcmp` with `strncmp((char *)adv_name, device_name, adv_name_len)`?

Allocating a 256 byte array on the stack may cause stack overflow issues...

asukiaaa:
@igrr
Thanks for your advice.
I succeeded in connecting device with using `strncmp`.

Is ts ok that I commited 2 times for this pull request?
If it's not suit for some rule of this repository, I will amend or create other pull request that correct one line for `strcmp`.

Thanks.

igrr:
If you can squash two commits and force-push to this branch, that would be perfect.

asukiaaa:
Done, how about it?

igrr:
Thanks, looks good. Copying this into the internal review and merge queue.

asukiaaa:
Thanks.

projectgus:
Cherry-picked in a797dca. This merge took a little longer than expected because we ended up tweaking the commit a bit (for the case where device name is a prefix of adv_name, or vice versa.) Thanks for identifying the problem and sending the fix. :)

asukiaaa:
@projectgus 
Thanks you for cherry picking.

PR #519, opened 2017-04-13
components/driver: 'const' all config calls.

buserror:
Some were, some weren't. They all could/should be.

Signed-off-by: Michel Pollet <buserror@gmail.com>

### DIFF
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -270,7 +270,7 @@ esp_err_t gpio_set_direction(gpio_num_t gpio_num, gpio_mode_t mode)
     return ret;
 }
 
-esp_err_t gpio_config(gpio_config_t *pGPIOConfig)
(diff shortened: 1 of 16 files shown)


buserror:
@bvernoux I went for consistency with the other ones, as I didn't want to submit a big patch; we can always do it as a second round. I'd like to have that, as you can then also const the config struct you pass, and initializes it with c99 notation instead of using code to do it.

projectgus:
Thanks Michel, this look good. I've pushed it into our internal review/merge queue.

Regarding const-ifying all arguments (including values), to some extent this is a matter of personal preference. I used to be in the habit of doing it but got talked out of it some years ago. Pointers to const data have the most significance (due to their implications for the calling code, for storage type of the parameter, and the possible optimisations the compiler can apply.)

FWIW, our doxygen includes `@param[out]` annotations for "out" pointer arguments. It's probably not as consistent as it should be across the entire API, though. :|

Angus

PR #537, opened 2017-04-22
lwip_close should free the socket for a new connection?

edmund-huber:
I'm building an HTTP server, and so obviously it does a lot of `lwip_accept` and `lwip_close`.

I found out pretty quickly that for some reason my httpd stopped working after a while. It turned out that `lwip_accept` was failing, but why? In `lwip_accept`, I had it print the state of each of the NUM_SOCKETS to figure out why it couldn't find a free socket. It turns out that none of the sockets were free, even though I know I only had maximum 4 simultaneous connections.

Looking further, it seems that you're supposed to call `free_socket` to make `conn == NULL` so that the socket is "free". But for me `ESP_THREAD_SAFE` is some truthy value, so `free_socket` was never called... why is the code like this?

I can't comment on whether or not the check on ESP_THREAD_SAFE should be there, but when I remove it, my httpd recycles sockets fine, and it works perfectly.

Note: if I had been using just `close` instead of `lwip_close`, I would not have had this problem since `#define close(s) lwip_close_r(s)`, and it calls `free_socket` as the result of some macro expansion. All the same, this seems like a bug.

### DIFF
--- a/components/lwip/api/sockets.c
+++ b/components/lwip/api/sockets.c
@@ -864,9 +864,7 @@ lwip_close(int s)
     return -1;
   }
 
-#if !ESP_THREAD_SAFE
   free_socket(sock, is_tcp);
-#endif
 
   set_errno(0);
   return 0;


edmund-huber:
Looking more into this .. it looks like all the `.._r` functions, which rely on the ref-counting stuff in `LWIP_API_LOCK`, `LWIP_API_UNLOCK`, are not in upstream lwip. 

So, I'm hypothesizing that you had some (a lot?) of code that needed to just work, but used the POSIX API names, and said code also didn't do a good job of cleaning up after itself which is why the ref-counting stuff (and close socket when ref=0) was added? This is more for my curiosity's sake than anything.

Back to the matter at hand: why in fact are there checks for `ESP_THREAD_SAFE` in `free_socket`? If `ESP_THREAD_SAFE` were not true, I'm pretty sure the `free_socket` would be plain wrong -- it would never set `conn = NULL`. Honestly I'm a little confused.

projectgus:
The ESP_THREAD_SAFE changes are intended to allow you to close a socket that may still be blocked sending or receiving in a different task. LWIP doesn't fully support this case out of the box. Sockets are reference counted until the last task is done with it, and then it is freed.

It is possible that there are some bugs introduced for the !ESP_THREAD_SAFE case, I don't believe we've tested that.

There shouldn't a socket leak just using accept() and close() with sockets, though. Do you have some code you can post that displays this problem?

edmund-huber:
To reproduce the issue, simply try to use `lwip_close` on any socket. It won't work.

But I see your point, I will just use the `close` macro. Thank you.

projectgus:
Ah, I understand now - sorry I didn't quite follow before. Yes, calling `lwip_close_r()` or `close()` will work correctly but `lwip_close()` doesn't. There may be a way we can refactor this to make it work when called directly, but the solution you're using is the best for now.

edmund-huber:
Thank you for taking the time to consider it more. :) If I had an idea on how to refactor it, I would offer it, but it looks already quite heavily refactored and I'm not sure what I would break.

hwmaier:
I just ran into the same issue, using naively `lwip_accept()` and `lwip_close()` in my code as an experienced lwip user would do. Took me hours to figure out what was going wrong.

But calling `lwip_close()` on esp-idf will not free the sockets as explained above. After a few close calls all sockets are exhausted. This is unacceptable. An official API call should never be allowed to misbehave in such a way.

I recommend in cases when `ESP_THREAD_SAFE` is set (which is the default on the esp-idf platform) either declare the non- `lwip..._r` calls static so it is simply not possible to use those defunct calls or map those calls to `lwip..._r` calls in similar way as it is done with the non-lwip prefixed calls.

So for example if `ESP_THREAD_SAFE` is enabled:

- `close()` maps to `lwip_close_r()`   // already the case
- `accept()` maps to `lwip_accept_r()`   // already the case
- `lwip_close()`  also maps to `lwip_close_r()`
- `lwip_accept()`  also maps to `lwip_accept_r()`
- and the some for all the other `lwip_...` calls.

I do not think this bug report should be closed.

PR #559, opened 2017-05-02
Fix vfs_fat_spiflash failed to mount / f_mkfs()

nineisk:


igrr:
Thanks for the fix. Will merge this soon.
`FM_SFD` usage was a pretty late change, and while testing I had the filesystem created already.

igrr:
Thanks for the PR! Cherry-picked in b65779e.

PR #577, opened 2017-05-06
Make gen_esp32part.py compatible with both python2.7 and python3.

dmopalmer:
I ran into the issue of https://github.com/espressif/arduino-esp32/issues/236

This is an step forward to make the toolchain compatible with Python3.  It includes both a comparison issue and unicode/str/bytes issues.

This has been tested to the extent that 'make flash' of hello_world.c runs under both python2.7 and python 3.6.1 under OSX, and I don't see any reason why it shouldn't work under linux.

The test that I have not done is: under Windows, does the binary output to stdout work properly (and did it before)?

In the binary case, I use 'wb' as the mode for files, and use stdout.buffer instead of stdout as the writeable object .   So what was windows doing when it tried to write a 0x0a (which is the byte for newline)?  Did it add an 0x0d byte (to make it look like '\r\n')?

### DIFF
--- a/components/partition_table/gen_esp32part.py
+++ b/components/partition_table/gen_esp32part.py
@@ -80,7 +80,7 @@ class PartitionTable(list):
             p.verify()
         # check for overlaps
         last = None
-        for p in sorted(self):
+        for p in sorted(self, key=lambda x:x.offset):
             if p.offset < 0x5000:
                 raise InputError("Partition offset 0x%x is below 0x5000" % p.offset)
             if last is not None and p.offset < last.offset + last.size:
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks for sending this, @dmopalmer .

FWIW, this change looks good to me (I'm the original gen_esp32part.py author.) I'd been meaning to do something like this for gen_esp32part.py and the other IDF Python tools for a while.

I expect this change will mean everything works fine with Python 3 in ESP32 Arduino. esptool.py is tested working with Python 2 & 3 already.

projectgus:
Oops, when I replied before I thought this PR was on the arduino repo not the IDF one.

Have cherry-picked this into our review & merge queue, with a few additional tweaks.

projectgus:
Merged in beffcd6, thanks.

PR #585, opened 2017-05-10
bootloader: remove image verification when waking from deep sleep

CristiCimpianu:
In the application I'm building it's very important to have access to interrupts as fast as possible after deep sleep wake up.
On my dev board image verification takes ~800msec, which is way too long, and I think it's best done only on boot after flashing and it's not really needed on wakeup. The only scenario I can think of where the image would change is if before entering deep sleep there was an OTA update, but OTA images are verified before being written to flash.
This reduces the time until interrupts are again available to under 200msec on my board, which is manageable for my application.


projectgus:
Hi @CristiCimpianu ,

Sorry noone ever got back to you about this.

One of the changes which is coming for IDF V3.0 is a rewrite of the bootloader, so it verifies as it loads. Performance is substantially improved (both when waking from sleep and for normal boots.) Please take a look once these changes land (they'ren ot on github yet, buy they should be some time in the next few days.)

Angus

CristiCimpianu:
Hi Angus,

Thanks a lot for the reply. Can you provide some estimate on when the v3.0 code will land on github?

The latest changes in the booloader don't seem to bring any improvement and also overriding checksum checks doesn't seem so easy as in the fix above so I'm forced to stick with an old version of the IDF if I can't find a workaround. This will be a problem for our product since we'll be missing a lot of useful patches. However at this point with the device waking up from sleep in 1.8 seconds our device is completely unusable.

Regards,
Cristi.

projectgus:
Hi @CristiCimpianu ,

Sorry, I missed this comment when you made it. I think the changes you have should include the faster bootloader code. I'm surprised it's as slow as 1.8 seconds. Are you using secure boot? This is the only thing I know of that is likely to slow boot down anywhere near this much.

The new bootloader code (in the current master branch) combines loading & verification, to the point where the overhead of verification is quite low compared to beforehand.

If it's genuinely too slow then we can look at removing these checks, but I'd like to understand your use case & settings better (debug output, flash settings, etc.) that is making it this slow.

Angus

CristiCimpianu:
Hi @projectgus and thanks for your reply.
I'm not using secure boot. The 1.8 sec time I got before moving the flash to QIO mode. I have already done some changes to avoid verification during deep sleep wakeup.
With my changes on top of the current master branch code I get 834msec on first boot and 169msec when coming out of deep sleep. I would have loved for that time to be under 100msec, but that's as fast as I could make it. The application is an IoT button that needs to measure the button click duration and since wakeup is done on the button press I need it to be up fast enough to catch the button raise edge interrupt. Any other approach either leaves me with an incorrect click duration or gets me confused between single or multiple clicks. Within 834msec I can certainly fit 2 clicks before the cpu is up. Within 169msec it's difficult to lose the final interrupt, but with a little practice I can do it in 1 out of several tries.
The flash is a 4MB, 80 MHz W25Q32 chip connected via SPI and the boot time comes from a xthal_get_ccount() right at the start of app_main().
My changes are not as clean as I'd like them mostly because they remove the RAM obfuscation you added and that's why I didn't push them.

projectgus:
> The application is an IoT button that needs to measure the button click duration and since wakeup is done on the button press I need it to be up fast enough to catch the button raise edge interrupt. Any other approach either leaves me with an incorrect click duration or gets me confused between single or multiple clicks. 

Have you thought about writing a simple ULP program for this? It can be running while the ESP32 is asleep and while the main CPU is booting IDF.

CristiCimpianu:
> Have you thought about writing a simple ULP program for this?
Before coming up with the idea to avoid image verification I did think about the ULP, but since I'm not that good at writing assembly and it's not clear to me even now how to do it I gave up on it. I guess the ULP would need to wake up when the main cpu wakes up and do some polling on the gpio port to see when it changes state and how many times it does that. Is there any other documentation besides [ULP guide](http://esp-idf.readthedocs.io/en/latest/api-guides/ulp.html)? From that doc it seems I would need to keep the ULP waking up at a fast rate all the time and its logic would have to be not so simple because it needs to understand when to start counting and when to stop.

projectgus:
There are two ULP examples you can work from.

This one is probably the most relevant:
https://github.com/espressif/esp-idf/blob/master/examples/system/ulp/main/ulp/pulse_cnt.S

It currently waits for a pulse of a minimum duration before it wakes the main CPU, but you could modify it to communicate the pulse length to the main CPU.

projectgus:
> boot time comes from a xthal_get_ccount() right at the start of app_main().

I forgot to mention this yesterday, but this is probably not an accurate measurement of boot time. The boot process runs at multiple different clock frequencies - crystal freq to start, then 80MHz, and then the final app frequency. The CCOUNT value is not guaranteed to be stable across frequency changes.

The recommended reliable way to measure the boot time is to set a GPIO in app_main() and use an external oscilloscope or logic analyzer to measure the time between EN pin going high and this GPIO being set.

Eventually we would like the ESP_LOG() timestamps to accurately reflect time since boot, but they don't at the moment (they are accurate from app_main() forwards.) #700 has some discussion of this.

projectgus:
I'm going to close this. Bootloader loading times should be improved over the situation at the time this PR was sent, and the bootloader code has changed quite a bit.

@CristiCimpianu if you're still concerned about boot times, please open a Github issue and we can discuss the issue further.

Issue #273, opened 2017-01-21
[TW#12212] VFS and LWIP should share file/socket descriptor space


rojer:
Currently some functions of the POSIX file API are routed to VFS and those that are partof the BSD socket API are routed to LWIP. this is very confusing. for example, it is possible to `open` a file, but subsequent `read` will fail because it goes to LWIP.
of course, one can side-step the issue by using buffered file i/o (`fopen`, `fread`), but it's not ideal. it should be possible to route LWIP through VFS and make them share descriptor space.

young-mu:
Hi, @rojer. I think you can consider the LWIP build option **LWIP_SOCKET_OFFSET** to let socket descriptor starting from an offset to avoid the descriptor overlap with VFS file descriptor.

rojer:
the issue is not overlap of descriptor space, but the fact that `read` is routed to LWIP only and cannot be used for operations on VFS files.

young-mu:
Okay.

jsarrett:
I also have run in to this problem.  It's very serious because it means basically that libc is linking unreliably.  It's not at all clear which read or write will be linked to a component.  LWIP_SOCKET_OFFSET should make it easy enough to fix inside of newlib, but that call routing *has* to be done at runtime.

chanibal:
@igrr, in a [forum post from half a year ago](https://www.esp32.com/viewtopic.php?t=2317), you mentioned that you'll fix the LWIP macro issues with this ticket. As far as I see in the closing commit, that was not the case.

I know, there is the `#undef` hack, but frankly it's very annoying to have issues with such common indentifiers. Not only `bind`, but also the lack of `send` is quite annoying.

Any chance to fix this, perhaps attach it to another ticket?

projectgus:
@chanibal
> you mentioned that you'll fix the LWIP macro issues with this ticket. As far as I see in the closing commit, that was not the case.

Sorry we missed this comment on this closed issue until just now. I agree we should fix this properly. Could you please open a new issue to keep track of this one?

PR #586, opened 2017-05-10
Turn off LWIP_POSIX_SOCKETS_IO_NAMES

rojer:
Real solution to read() and write() on sockets should come as a result of proper routing through the VFS layer: espressif/esp-idf#273
Current situation produces unreliable results. Also, macros defined here tend to conflict with other identifiers called "read", "write" and "close".

### DIFF
--- a/components/coap/port/coap_io_socket.c
+++ b/components/coap/port/coap_io_socket.c
@@ -37,6 +37,8 @@
 #endif
 #include <errno.h>
 
+#include "lwip/sockets.h"
+
 #ifdef WITH_CONTIKI
 # include "uip.h"
 #endif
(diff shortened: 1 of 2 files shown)


projectgus:
Thanks for contributing this @rojer.

Unfortunately I think this has the potential to break a lot of people's code so it's probably not a change we can make until we have VFS routing in place. I realise the current situation is a mess.

In your own code, if you need to mix LWIP & POSIX I/O functions in the same compilation unit, a possible workaround is something like:

```
#include "lwip/opt.h"
#undef LWIP_POSIX_SOCKETS_IO_NAMES
#include "lwip/socket.h"
```

I realise this is also a nasty hack. We're hoping to have the proper fix in place soon.

rojer:
yeah, we really need a fix for this.
i'll close this PR, but will keep this in our fork until proper fix is in place.

PR #604, opened 2017-05-15
update nghttp2 to version v1.22.0

MrBuddyCasino:
Updated nghttp2 to v1.22.0.

New files:
	components/nghttp/include/nghttp2_debug.h
	components/nghttp/library/nghttp2_debug.c

All other files are 1:1 copies of the corresponding v1.22.0 tag files.

This version makes it possible to return NGHTTP2_ERR_PAUSE from a nghttp2_data_source_read_callback. Without this feature, it is impossible to POST data in a non-blocking way, resulting in eventual memory exhaustion.

Also see: https://github.com/nghttp2/nghttp2/pull/672/files#diff-20273fc5d9b0c6c133eb3d701444596d


projectgus:
Thanks for sending this, @MrBuddyCasino .

Rather than continuing to update these source files in our tree, I've put together a proposal to make nghttp2 into a git submodule. This will make future updates a lot more painless. Will let you know what happens.

MrBuddyCasino:
Makes sense. I had to do the work anyway because my project depends on NGHTTP2_ERR_PAUSE, so no biggie.

projectgus:
Sorry I didn't reply months ago about this, @MrBuddyCasino . As you probably noticed, nghttp2 is now a submodule (and the above-mentioned bugfix has been merged into the upstream code base, as well.)

PR #607, opened 2017-05-16
Print the name of the task that hit the stack watchpoint

rojer:
This may be enough to troubleshoot (increase stack size).

### DIFF
--- a/components/esp32/panic.c
+++ b/components/esp32/panic.c
@@ -212,7 +212,10 @@ void panicHandler(XtExcFrame *frame)
                     //debugcause if the cause is watchdog 1 and clearing it if it's watchdog 0.
                     if (debugRsn&(1<<8)) {
 #if CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK
-                        panicPutStr("Stack canary watchpoint triggered ");
+                        const char *name = pcTaskGetTaskName(xTaskGetCurrentTaskHandleForCPU(core_id));
+                        panicPutStr("Stack canary watchpoint triggered (");
+                        panicPutStr(name);
+                        panicPutStr(") ");
 #else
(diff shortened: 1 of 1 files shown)


components/esp32/panic.c
```diff
@@ -212,7 +212,10 @@ void panicHandler(XtExcFrame *frame)
                     //debugcause if the cause is watchdog 1 and clearing it if it's watchdog 0.
                     if (debugRsn&(1<<8)) {
 #if CONFIG_FREERTOS_WATCHPOINT_END_OF_STACK
-                        panicPutStr("Stack canary watchpoint triggered ");
+                        const char *name = pcTaskGetTaskName(xTaskGetCurrentTaskHandleForCPU(core_id));
```
igrr: This pointer will likely be in cached region of memory. Would it make sense to sanity-check it before trying to print it? I.e. something like
```
if (spi_flash_cache_enabled() || !((uint32_t) name >= SOC_DROM_LOW && (uint32_t) name < SOC_DROM_HIGH)) 
```

rojer: but task name is a member of the current TCB, so, is it actually possible for it to be not in RAM?

rojer: task name provided by the user is copied into [this field](https://github.com/espressif/esp-idf/blob/1e0710f1b24429a316c9c34732aa17bd3f189421/components/freertos/tasks.c#L175)

igrr: Ok, yes, you're right.

projectgus:
Thanks @rojer. I've cherry-picked this into our internal review & merge queue.

igrr:
Cherry-picked as 3ac3a78, thanks for the PR!

PR #612, opened 2017-05-16
check that the contents of partitions will fit into the partitions (IDFGH-2503)

edmund-huber:
(I expect changes to be requested. Opening this to start a discussion!)

I was looking to tweak the flash partition sizes, and I realized that apparently esp-idf isn't checking that the images will fit into their given partitions before flashing them in.

For example, I used the following `partitions.csv`:
```
nvs,      data, nvs,     0x9000,  0x3000
otadata,  data, ota,     0xd000,  0x2000
phy_init, data, phy,     0xf000,  0x1000
factory,  app,  factory, 0x10000,  1K
ota_0,    app,  ota_0,   ,         1M
ota_1,    app,  ota_1,   ,         1M
```
to successfully flash a device. (The device booted, did all the stuff it's supposed to in the program, etc.) Note the `factory` partition size: 1K. My app is 1.2MB, so I guess in fact esptool happily spilled the app image over into the `ota_0` partition.

In this pull request I'm adding a little Python script to check whether the images will fit into their partitions. This caught my error (I tested that much).

A caveat: for this to be entirely correct, it should be rounding sizes (and offsets) to dword boundaries.

### DIFF
--- a/components/esptool_py/Makefile.projbuild
+++ b/components/esptool_py/Makefile.projbuild
@@ -14,6 +14,7 @@ PYTHON ?= $(call dequote,$(CONFIG_PYTHON))
 ESPTOOLPY_SRC := $(COMPONENT_PATH)/esptool/esptool.py
 ESPTOOLPY := $(PYTHON) $(ESPTOOLPY_SRC) --chip esp32
 ESPTOOLPY_SERIAL := $(ESPTOOLPY) --port $(ESPPORT) --baud $(ESPBAUD) --before $(CONFIG_ESPTOOLPY_BEFORE) --after $(CONFIG_ESPTOOLPY_AFTER)
+CHECK_ESP32PART := $(COMPONENT_PATH)/../partition_table/check_esp32part.py
 
 # Supporting esptool command line tools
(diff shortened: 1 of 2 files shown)


edmund-huber:
It might also be good to check that everything fits into 4MB..

projectgus:
Hi Edmund,

Thanks - this looks really useful. I had some similar ideas around creating some tools to read partition information into the build process (currently the factory partition offset has to be independently set in menuconfig, and it would be good to also let the build process know the ota data offset so it can clear it on "make flash"). I might use your script as a jumping-off point for that as well.

Angus

projectgus:
> It might also be good to check that everything fits into 4MB..

esptool.py will fail if you try to write outside the configured flash size. Also the bootloader will refuse to boot if it looks like the partition table includes partitions which don't live inside the flash size. But I agree it would be good to get that error at build time, instead.

edmund-huber:
OK, let me know what you want to do with this. I'll be happy to work with you on a different/better way to do it. No offense taken if you choose to close this PR. :)

PR #617, opened 2017-05-17
Update idf_monitor.py to make --toolchain-prefix argument work

hwmaier:
--toolchain-prefix command line argument was not taken into account when specified


projectgus:
Thanks Henrik, appreciate the fix. This looks like it fixes the --eol argument as well.

Have put this into our internal review & merge queue.

hwmaier:
I don't have the gcc compiler in the path and depend on the configuration to point to the correct location. This is how I discovered this minor bug.

igrr:
Thanks for the PR @hwmaier, it has been cherry picked as b95bef4.

PR #622, opened 2017-05-18
Converting URLs from HTTP to HTTPS

JLospinoso:
URLs in `README` are unsafe. Modifying to HTTPS; I manually checked that all of these URLs are hosted on servers with SSL/TLS support.


projectgus:
Good catch, @JLospinoso . I had not actually noticed that readthedocs.io had HTTPS, very nice.

projectgus:
Cherry-picked, thanks!

PR #639, opened 2017-05-23
Add xPortGetHeapSize

rojer:
Provides a way to find out the size of the system heap, mostly informational.

Similar to xPortGetFreeHeapSize, xPortGetHeapSize returns the size of the byte-access-capable RAM,
but xPortGetHeapSizeCaps and xPortGetHeapSizetagged are provided as well.

### DIFF
--- a/components/esp32/heap_alloc_caps.c
+++ b/components/esp32/heap_alloc_caps.c
@@ -380,9 +380,7 @@ void *pvPortMallocCaps( size_t xWantedSize, uint32_t caps )
     return NULL;
 }
 
-
-size_t xPortGetFreeHeapSizeCaps( uint32_t caps )
-{
+static size_t sum_regions_with_caps( uint32_t caps, size_t (*func)(BaseType_t tag)) {
     int prio;
     int tag;
     size_t ret=0;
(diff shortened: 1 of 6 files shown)


projectgus:
Thanks for sending this, @rojer.

It hasn't been announced, but we have a new heap implementation in the pipeline for v3.0 (version after forthcoming v2.1), which deprecates the FreeRTOS-style API. So I'm thinking we may not merge this, and instead ensure that this functionality is available in the new API (it's available in my work-in-progress now via a statistics function call that returns a struct with a variety of data in it, but we can add a standalone function to get only the total size as well.)

rojer:
i'm fine either way. when the new functions comes along, i'll use it.
if you intend to land your refactoring soon, then feel free to close this.
if it's, like, weeks away still, i'd of course prefer it to go in - it seems fairly straightforward.

projectgus:
Sorry I didn't reply about this earlier, @rojer.

Using the new heap code in esp_heap_caps.h you can use something like `heap_caps_get_info(&info, MALLOC_CAP_8BIT);` and then the `info.total_free_bytes + info.total_allocated_bytes` will give you an estimate of this.

This result is minus all heap management overhead (something which is variable depending on the number of blocks allocated in the heap). We don't actually have a way to get the total fixed region size. However it would be quite easy to add it, if that's something you need.

I'm going to close this PR, but drop me a line (or send another PR) if you need that particular functionality.

Angus

PR #640, opened 2017-05-24
add FATFS_TINY menuconfig item

edmund-huber:
Enabling `_FS_TINY` saves `4096 bytes * max_files` of heap. This makes a big difference for us!

### DIFF
--- a/components/fatfs/Kconfig
+++ b/components/fatfs/Kconfig
@@ -105,4 +105,19 @@ config FATFS_MAX_LFN
    help
       Maximum long filename length. Can be reduced to save RAM.
 
+choice FATFS_TINY
+   prompt "Use per-file caches"
+   default FATFS_TINY_NO
+   help
+      By default, FatFs keeps a 4096-byte (sector size) cache per open file to
+      reduce reads and writes to the underlying media. Disabling this forces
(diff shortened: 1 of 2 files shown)


components/fatfs/Kconfig
```diff
@@ -105,4 +105,19 @@ config FATFS_MAX_LFN
    help
       Maximum long filename length. Can be reduced to save RAM.
 
+choice FATFS_TINY
+   prompt "Use per-file caches"
```
projectgus: Good catch noticing this option!

Could you please make this a "bool" option CONFIG_FATFS_PER_FILE_CACHES rather than a "prompt"? Can do the necessary inverting of logic (tiny mode == !per file caches) in the header file if necessary.

projectgus:
@edmund-huber are you still working on this? Are you able to make the requested change?

edmund-huber:
Thanks for checking. I am on vacation abroad currently and don't have the means to make or test any change. I'll be back later this week and can do it then.

projectgus:
Support for FATFS_TINY has been added  in 07c44d7f0 (as "Use separate cache for each file"). Thanks for bringing this feature to our attention, @edmund-huber .

PR #650, opened 2017-05-30
Remove some redundant code line from uart.c

hwmaier:
This PR removes some lines of code which are redundant. Just to save some valuable bytes in flash...

### DIFF
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -366,6 +366,7 @@ esp_err_t uart_set_pin(uart_port_t uart_num, int tx_io_num, int rx_io_num, int r
 
     int tx_sig, rx_sig, rts_sig, cts_sig;
     switch(uart_num) {
+        default:
         case UART_NUM_0:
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @hwmaier,

Thanks for considering our precious flash size, and thanks for sending this change.

Unfortunately these changes don't make a significant difference to compiled size. In debug mode, the difference is 4 bytes (because of trying to keep the compiled code in line with the source for debugging, I think). In release mode, the difference is zero. Modern C compilers are good at recognising things like redundant writes and identical branches, and merging them.

So the question becomes, do the changes improve readability? The UART driver can certainly benefit from more readability! The first change, I think, does not - having the explicit case there (and no fallthroughs) is easier to see what's going on. The second change does though.

I've cherry-picked the second change into our internal review & merge queue.


Angus

projectgus:
Cherry-picked in part, thanks!

PR #661, opened 2017-06-01
Add missing spiflash function pointers

rojer:


projectgus:
Hi @rojer,

Looks like we overlapped on this one and I added these in b428d28 around the same time you sent the PR. Sorry it took us so long to respond.

Angus

PR #667, opened 2017-06-03
Add RS-485 RTS line driver control to UART driver

hwmaier:
This PR adds an initial implementation for RS-485 half-duplex line driver control using the RTS control line.

It uses the UART[uart_num]->rs485_conf.en bit to track that this mode is enabled.

### DIFF
--- a/components/driver/include/driver/uart.h
+++ b/components/driver/include/driver/uart.h
@@ -636,6 +636,19 @@ esp_err_t uart_disable_pattern_det_intr(uart_port_t uart_num);
  *     - ESP_FAIL Parameter error
  */
 esp_err_t uart_enable_pattern_det_intr(uart_port_t uart_num, char pattern_chr, uint8_t chr_num, int chr_tout, int post_idle, int pre_idle);
+
+/**
+ * @brief   UART enable RS-485 mode and RS-485 hald-duplex line driver control using the RTS control line
+ *
+ * @param uart_num UART port number.
+ *
+ * @return
+ *     - ESP_OK Success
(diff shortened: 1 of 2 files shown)


projectgus:
Hi @hwmaier ,

Thanks for sending this. Sorry noone got back to you earlier.

I've asked the maintainer of the IDF peripheral drivers to take a look, then we can look to merging this.

Angus

components/driver/include/driver/uart.h
```diff
@@ -636,6 +636,19 @@ esp_err_t uart_disable_pattern_det_intr(uart_port_t uart_num);
  *     - ESP_FAIL Parameter error
  */
 esp_err_t uart_enable_pattern_det_intr(uart_port_t uart_num, char pattern_chr, uint8_t chr_num, int chr_tout, int post_idle, int pre_idle);
+
+/**
+ * @brief   UART enable RS-485 mode and RS-485 hald-duplex line driver control using the RTS control line
```
projectgus: Typo: half

x-logLT:
One thing wrong with this implementation - according to TM, UART2 does not have fifo reset capability, also resetting UART1 fifo might affect UART2 (and it does), this is also ignored in uart_reset_fifo function & rx_ovf interrupt handler. SW workaround would be to manually ofload RX buffer before enabling reception, something like:
`while(uart_reg->status.rxfifo_cnt)
					dummy = uart_reg->fifo.rw_byte;`

alisitsyn:
Sorry for the long delay. Espressif has implemented RS-485 support based on the implementation provided here, it will be merged soon.

PR #669, opened 2017-06-03
components/driver/rmt: keep ISR registered in case other channels are still active

### LINKED ISSUE #486: [TW#12119] rmt_driver_uninstall() unconditionally de-registers ISR for all other channels
The following use case involving several RMT channels apparently does not work with the current RMT driver implementation.

Channel 0 is used persistently. I.e. after initial configuration and driver installation, the channel is operated as required and is supposed to be usable until the chip is reset.

Channel 1 is used temporarily. I.e. it cycles through the sequence of configuration, installation and uninstallation from time to time.

I observe that channel 0 becomes unresponsive once channel 1 was temporarily active for the first time. Channel 1 is not affected and it will operate as expected each time it's activated even when channel 0 is already broken.

Looking at `components/driver/rmt.c` I suspect that the first call to `rmt_driver_uninstall()` will de-register the ISR even though another channel is still installed and relies on the interrupts being serviced. This is contrary to my understanding that channels can be installed & uninstalled independently from each other since the driver API requires a channel number for both of these functions (especially for `rmt_driver_uninstall()`).
Please clarify the intended API use model.

devsaurus:
Fixes #486 as proposed in https://github.com/espressif/esp-idf/issues/486#issuecomment-292050885.

### DIFF
--- a/components/driver/rmt.c
+++ b/components/driver/rmt.c
@@ -635,8 +635,19 @@ esp_err_t rmt_driver_uninstall(rmt_channel_t channel)
     }
     free(p_rmt_obj[channel]);
     p_rmt_obj[channel] = NULL;
-    s_rmt_driver_installed = false;
-    return rmt_isr_deregister(s_rmt_driver_intr_handle);
+
+    rmt_channel_t idx;
+    for (idx = 0; idx < RMT_CHANNEL_MAX; idx++) {
+        if (p_rmt_obj[idx]) {
+            break;
+        }
+    }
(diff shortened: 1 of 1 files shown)


igrr:
Thanks for the patience with this @devsaurus, a fix which is similar to the proposed one has been merged recently:
https://github.com/espressif/esp-idf/commit/c77b6994639f3a537ee4d23abd968cfee779417e

Thank you for contributing.

devsaurus:
Thank you!

PR #670, opened 2017-06-03
Update external-ram.rst

ESP32DE:
see the [Getting Starting Guide of Wrover Kit](http://espressif.com/sites/default/files/documentation/esp-wrover-kit_getting_started_guide_en.pdf)

wrover_get_started

"GPIO16 and GPIO17 are used as the CS and clock signal for PSRAM."
 
means
GPIO16 = CS
GPIO17 = CLK

btw: 
PSRAM SIO[2] (pin 3) - flash WP
perhabs you can describe - for what we need this pin and how we handle it.
not sure why here is named WP
WP is usually used in SPI Flash - - do we need this in SPI RAM too?

### DIFF
--- a/docs/api-guides/external-ram.rst
+++ b/docs/api-guides/external-ram.rst
@@ -22,11 +22,11 @@ pin to a high signal level on bootup, or program the fuses in the ESP32 to alway
 damaging the PSRAM and/or flash chip.
 
 To connect the ESP-PSRAM chip to the ESP32, connect the following signals:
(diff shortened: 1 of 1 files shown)


ESP32DE:
and yes, if you see here
https://github.com/espressif/esp-idf/blob/feature/psram_malloc/components/esp32/psram.c#L50
the pins swapped to the describtion

ESP32DE:
you allready done it now without merge so this can be closed
new open PR is here [#686](https://github.com/espressif/esp-idf/pull/686)

Issue #459, opened 2017-03-25
[TW#12147] Unable to switch on C++ exception handling -fexceptions


nkolban:
I am working in C++ and wish to switch on exception handling.  I compiled my C++ code by enabling the GCC flag called `-fexceptions`.  I did this by editing my local component.mk and adding:

```
CXXFLAGS+=-fexceptions
```

The code seems to compile cleanly however when I run "make" for the ESP-IDF build system, we abort at the end with the following:

```
$ make
CXX BLEUtils.o
/home/kolban/esp32/esptest/apps/workspace/bt_test/components/cpp_utils/./BLEUtils.cpp:181:20: 
AR libcpp_utils.a
LD app-template.elf
esptool.py v2.0-beta2
Traceback (most recent call last):
  File "/home/kolban/esp32/esptest/esp-idf/components/esptool_py/esptool/esptool.py", line 2294, in <module>
    main()
  File "/home/kolban/esp32/esptest/esp-idf/components/esptool_py/esptool/esptool.py", line 2096, in main
    operation_func(args)
  File "/home/kolban/esp32/esptest/esp-idf/components/esptool_py/esptool/esptool.py", line 1732, in elf2image
    image.save(args.output)
  File "/home/kolban/esp32/esptest/esp-idf/components/esptool_py/esptool/esptool.py", line 1306, in save
    assert (f.tell() + 8) % IROM_ALIGN == segment.addr % IROM_ALIGN
AssertionError
/home/kolban/esp32/esptest/esp-idf/components/esptool_py/Makefile.projbuild:49: recipe for target '/home/kolban/esp32/esptest/apps/workspace/bt_test/build/app-template.bin' failed
make: *** [/home/kolban/esp32/esptest/apps/workspace/bt_test/build/app-template.bin] Error 1
```

Unfortunately, I don't know how to make progress from here.

nkolban:
Attached to this post is the request app-template.elf which was added to a ZIP on linux using the "zip" command.

[app-template.elf.zip](https://github.com/espressif/esp-idf/files/870548/app-template.elf.zip)

projectgus:
Thanks @nkolban . `xtensa-esp32-elf-objdump -h` shows that the gcc exception structures are being linked in to the default "flash" address space as separate sections:

```
  7 .gcc_except_table._Z41__static_initialization_and_destruction_0ii 000001b5  4015db5d  4015db5d  000cc93d  2**0
[...]

PR #681, opened 2017-06-08
Add missing support for C++ exceptions

yaqwsx:
Specifying -fexceptions for the compiler is not enough.

- add necessary zero padding after .eh_frame section
- link .gcc_except_table_table in a way flash script does not complain
- call __registrer_frame_info before global constructors

Kudos jcmvbkbc for the necessary help.

This (partially) solves #459

### DIFF
--- a/components/esp32/ld/esp32.common.ld
+++ b/components/esp32/ld/esp32.common.ld
@@ -151,7 +151,7 @@ SECTIONS
     *(.rodata1)
     __XT_EXCEPTION_TABLE_ = ABSOLUTE(.);
     *(.xt_except_table)
-    *(.gcc_except_table.*)
+    *(.gcc_except_table .gcc_except_table.*)
     *(.gnu.linkonce.e.*)
     *(.gnu.version_r)
     . = (. + 3) & ~ 3;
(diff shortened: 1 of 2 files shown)


components/esp32/ld/esp32.common.ld
```diff
@@ -151,11 +151,13 @@ SECTIONS
     *(.rodata1)
     __XT_EXCEPTION_TABLE_ = ABSOLUTE(.);
     *(.xt_except_table)
-    *(.gcc_except_table)
+    *(.gcc_except_table.*)
```
jcmvbkbc: Are you sure you don't want to have both, i.e. ```*(.gcc_except_table .gcc_except_table.*)```?

yaqwsx: Why should I need both?

jcmvbkbc: In case -ffunction-sections switch isn't used LSDAs will be in the .gcc_except_table, but it will not be matched by the ```*(.gcc_except_table.*)```.

yaqwsx: I haven't thought about that - thanks! I fixed it.

igrr:
Cherry-picked as f698a1b2, thanks! Toolchain update coming up.

PR #682, opened 2017-06-08
Change esp_vfs_t.write return value to ssize_t

rojer:
write() should return ssize_t, not size_t.


edmund-huber:
I think you could go even further: for example `esp_vfs_write` uses `int ret` when it really wants `ssize_t ret`.

rojer:
good point. fixed type of the local `ret` variable for all of `esp_vfs_{read,write,lseek}`.

PR #701, opened 2017-06-15
Fix distinction between shared and non-shared interrupts in docs

pathob:
Hi, I think there is an error in the description what shared and non-shared interrupts are.

For the case that I understood this correctly, here is the fix.


projectgus:
Thanks @pathob. I've cherry-picked this into our review & merge queue.

projectgus:
Cherry-picked, thanks!

Issue #728, opened 2017-06-23
esp_ble_gatts_create_attr_tab regression


rojer:
@jack0c - after https://github.com/espressif/esp-idf/commit/491932b80cb27239fca4f8e07a36623afafd7a7f `esp_ble_gatts_create_attr_tab` stopped working - characteristics are just not registered anymore.
I isolated it to [this added return statement](https://github.com/espressif/esp-idf/commit/491932b80cb27239fca4f8e07a36623afafd7a7f#diff-f12d21a86d653101bdd60f613897b0a7R575) - if i remove it, everything goes back to normal.

There's another issue with this change: check for "one primary or one secondary service per table" is done twice: once in `btc_gatts_check_valid_attr_tab` and then again while creating services, via `btc_creat_tab_env.is_use_svc`. Maybe there's some reason to doing it twice, but i don't see it...

heyinling:
@rojer could you share your application code for this issue? I didn't meet your issue in my own application. I can create service by `esp_ble_gatts_create_attr_tab ` successfully on latest master (3bd3640d).

TianHao-Espressif:
@rojer , we do a more test, and there's no problem as you said. After this commit, the check of `esp_ble_gatts_create_attr_tab` is more strict. So you need to check the corresponding event status.

rojer:
i extracted a test standalone app from my current in-progress code it reproduces the issue - see attached archive.
basically, after https://github.com/espressif/esp-idf/commit/491932b80cb27239fca4f8e07a36623afafd7a7f there are no more `ADD_CHAR` events following the `CREAT_ATTR_TAB`.
app archive is attached. this is how it works with the change:

```
[Jun 26 21:37:07.629] I (1743) system_api: Base MAC address is not set, read default base MAC address from BLK0 of EFUSE
[Jun 26 21:37:07.894] I (2013) phy: phy_version: 355.1, 59464c5, Jun 14 2017, 20:25:06, 0, 0
[Jun 26 21:37:07.945] I (2043) main: BT init ok
[Jun 26 21:37:07.945] I (2053) main: LOCAL_IR
[Jun 26 21:37:07.945] I (2053) main: REG if 3 st 0 app 0
[Jun 26 21:37:07.945] I (2053) main: esp_ble_gap_config_adv_data 0
[Jun 26 21:37:07.945] I (2053) main: esp_ble_gatts_create_attr_tab 0
[Jun 26 21:37:07.945] I (2053) main: LOCAL_ER
[...]

PR #729, opened 2017-06-23
Fix esp_ble_gatts_create_attr_tab

rojer:
- Remove premature return from btc_gatts_inter_cb
 - Remove extra check for "only one pri/sec service per tab", it was already performed in btc_gatts_check_valid_attr_tab.

Fixes https://github.com/espressif/esp-idf/issues/728


rojer:
worked out on #728

PR #733, opened 2017-06-26
Use IPC for low-level LwIP API calls in mDNS server

e-asphyx:
Low level LwIP calls are not thread safe and must not be used outside of TCP/IP thread


me-no-dev:
Great! I'm in the process of rewriting the whole service to be RFC compliant and this edit comes right on time! Thanks for the contribution!

projectgus:
@e-asphyx Sorry for the delay with this one. We've had it in review for a while but it got held up behind other things, should hopefully get some traction soon.

PR #746, opened 2017-06-29
l2cu_no_dynamic_ccbs() fix

SiarheiVolkau:
l2cu_no_dynamic_ccbs() can cause assert in freertos's timers.c when SMP_INCLUDED == FALSE due to timeout stays equal to 0.


projectgus:
Thanks for sending this, @SiarheiVolkau . Sorry noone got back to you earlier.

I think this issue has been taken care of on the master branch now, as of 5b9a89f04d. Please reopen (or open an issue) if that doesn't seem to be the case.

SiarheiVolkau:
Yes now this case handled in master branch.

PR #757, opened 2017-07-01
Using param, which corresponds to handled event

SiarheiVolkau:
In the handlers of ESP_GATTS_ADD_CHAR_DESCR_EVT was used incorrect param, which corresponding to ESP_GATTS_ADD_CHAR_EVT.
 It was work because both params have same type and position in union.


projectgus:
Thanks @SiarheiVolkau . I've pushed this into our review & merge queue.

projectgus:
Cherry-picked in 1922f6c with tweaks in d5301bda .

PR #767, opened 2017-07-04
Use relative submodules

APokorny:
This allows users to make a deep mirror of the repositories to safe bandwith.

Signed-off-by: Andreas Pokorny <andreas.pokorny@siemens.com>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,31 +1,31 @@
 [submodule "components/esp32/lib"]
 	path = components/esp32/lib
-	url = https://github.com/espressif/esp32-wifi-lib.git
+	url = ../esp32-wifi-lib.git
 
 [submodule "components/esptool_py/esptool"]
 	path = components/esptool_py/esptool
-	url = https://github.com/espressif/esptool.git
+	url = ../esptool.git
 
 [submodule "components/bt/lib"]
 	path = components/bt/lib
-	url = https://github.com/espressif/esp32-bt-lib.git
+	url = ../esp32-bt-lib.git
 
 [submodule "components/micro-ecc/micro-ecc"]
 	path = components/micro-ecc/micro-ecc
-	url = https://github.com/kmackay/micro-ecc.git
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,31 +1,31 @@
...
 
 [submodule "components/nghttp/nghttp2"]
 	path = components/nghttp/nghttp2
-	url = https://github.com/nghttp2/nghttp2.git
+	url = ../../nghttp2/nghttp2.git
 
 [submodule "components/libsodium/libsodium"]
 	path = components/libsodium/libsodium
-	url = https://github.com/jedisct1/libsodium.git
+	url = ../..//jedisct1/libsodium.git
```
igrr: There's an extra slash here, which makes the clone fail with a 'The requested URL returned error: 400' error.

APokorny:
sorry.. please try again

igrr:
Thank you for the PR, it does solve the mirroring problem in a nice way! We are having some internal discussions about this approach. I will update when we reach some conclusion.

APokorny:
Besides bandwidth the other problem this would solve is that our CI system might loose internet connectivity during the build.

projectgus:
Hi @APokorny ,

Sorry for the slow reply getting back to you.

We were looking at some similar approaches to solve this as well, because we have internal Git repos inside China (our "primary" internal repo is not github), and also experience some performance problems dealing with Github at times.

In the end, we resolved this in a different way. You can take a look through how our CI system handles it:
https://github.com/espressif/esp-idf/blob/master/.gitlab-ci.yml#L44
https://github.com/espressif/esp-idf/tree/master/tools/ci

... we use "git config submodule.<name>.url" a lot to override the paths of submodules. You can see the "mirror-submodule-update.sh" script which does this.

Unfortunately, this approach (and our non-github main repo) is not compatible with relative paths of this kind. So we can't accept this PR.

Sorry about that (and the slow reply),

Angus

PR #800, opened 2017-07-13
Add quotes to PATH declarations

lf-:
This is necessary in case someone has a space in their PATH variable (which is within the realm of possibility).


projectgus:
Seems prudent, thanks @lf- . Have cherry-picked into our review & merge queue.

projectgus:
Cherry-picked, thanks!

PR #810, opened 2017-07-16
Added xTaskGetHandle FreeRTOS method

rovercoder:
**Merged minimal changes from FreeRTOS 9.0.0 (no personal changes)** > [patch.zip](https://github.com/espressif/esp-idf/files/1151341/patch.zip)

Added optional xTaskGetHandle FreeRTOS method + FreeRTOSConfig.h entry
- Useful to check if a task is running and stop it respectively (without manually storing reference)

Renamed pcTaskGetTaskName to pcTaskGetName
- made backwards compatible
- removed from optional
- refactored all usages


components/freertos/tasks.c
```diff
@@ -2273,18 +2273,142 @@ UBaseType_t uxTaskGetNumberOfTasks( void )
...
+
+	TaskHandle_t xTaskGetHandle( const char *pcNameToQuery ) /*lint !e971 Unqualified char types are allowed for strings and single characters only. */
+	{
+	UBaseType_t uxQueue = configMAX_PRIORITIES;
+	TCB_t* pxTCB;
+
+		/* Task names will be truncated to configMAX_TASK_NAME_LEN - 1 bytes. */
+		configASSERT( strlen( pcNameToQuery ) < configMAX_TASK_NAME_LEN );
+
+		vTaskSuspendAll();
```
igrr: I don't think this is correct — calling `vTaskSuspendAll` will suspend the scheduler on the current CPU only. The scheduler running on the other CPU may modify `pxReadyTasksLists` while the following procedure runs.

rovercoder: Missed that. No wonder it was omitted from the final version. You should drop this pull request if it's not worth the trouble.

Cheers for the correction.

rovercoder:
Missed that. No wonder it was omitted from the final version. You should drop this pull request if it's not worth the trouble.

Cheers for the correction. The latest commit should fix the issue.

Spritetm:
Looks okay. Can I ask you to finalize the conversion to muxes entirely? At the moment, there's still a commented out vTaskSuspend as well as a reference to that function somewhere in the comments. These are somewhat misleading re. what's really happening.

Spritetm:
Now your PR includes a lot of (probably unrelated) submodule changes...

rovercoder:
Here is the full patch for xTaskGetHandle method.
[full_patch.zip](https://github.com/espressif/esp-idf/files/1300972/patch.zip)

PR #822, opened 2017-07-20
RTC-GPIO pin calculation incorrect

husigeza:
components/esp32/deep_sleep.c: Modifying esp_deep_sleep_get_ext1_wakeup_status so it calculates the GPIO pins from RTC pins correctly. Previously the function composed the return value incorrectly, e.g. when the returned RTC pin was Pin 0.

### DIFF
--- a/components/esp32/deep_sleep.c
+++ b/components/esp32/deep_sleep.c
@@ -318,8 +318,14 @@ uint64_t esp_deep_sleep_get_ext1_wakeup_status()
     }
     uint32_t status = REG_GET_FIELD(RTC_CNTL_EXT_WAKEUP1_STATUS_REG, RTC_CNTL_EXT_WAKEUP1_STATUS);
     // Translate bit map of RTC IO numbers into the bit map of GPIO numbers
-    uint64_t gpio_mask = 0;
-    for (int gpio = 0; gpio < GPIO_PIN_COUNT; ++gpio) {
+    uint64_t gpio_mask = 1;
+    uint64_t gpio_return = 0;
+    for (unsigned int gpio = 0; gpio < GPIO_PIN_COUNT; ++gpio) {
+
+        if(gpio > 0) {
+            gpio_mask <<= 0x1;
+        }
+
         if (!RTC_GPIO_IS_VALID_GPIO(gpio)) {
             continue;
         }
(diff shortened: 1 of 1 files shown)


components/esp32/deep_sleep.c
```diff
@@ -318,18 +318,24 @@ uint64_t esp_deep_sleep_get_ext1_wakeup_status()
...
+        }
+
         if (!RTC_GPIO_IS_VALID_GPIO(gpio)) {
             continue;
         }
         int rtc_pin = rtc_gpio_desc[gpio].rtc_num;
         if ((status & BIT(rtc_pin)) == 0) {
             continue;
         }
-        gpio_mask |= BIT(gpio);
```
igrr: Would it be sufficient to replace this line with `gpio_mask |= 1ULL << gpio;` to fix the issue with GPIOs higher than 32

husigeza: Hello,
According to my understanding the gpio_mask should show that which PIN caused the wakeup, with setting a relevant bit to 1. E.g.: the interrupt comes through GPIO Pin3, then gpio_mask should look like: 0x0000000000001000, because 4th bit for from the right (LSB) is associated with GPIO Pin 3. However with the current solution, the gpio_mask would look like the following: 0x0000000000001100. Is this the value the function should return ? If yes then sorry, I misunderstood the concept.
Even if I misunderstood the interpretation of the gpio_mask, a problem still exists with GPIO Pin 0: gpio_mask would be 0, even the interrupt has came through Pin 0, or even no interrupt came at all. 

Cheers,
Geza

negativekelvin: Why? If gpio = 0, BIT(0) = 1, and gpio_mask = 1.  Are you sure you aren't misunderstanding the macro BIT?  The patch does not seem like it would work.  Igrr's suggestion seems right to fix gpio >= 32.

husigeza: Yes, you are right, sorry I did not check what BIT() does when I replied to the first comment.
Actually my change exactly does the same what Igrr suggests just without BIT() macro.
So in summary: I agree that the only change is necessary is which proposed by Igrr.

husigeza:
The requested change is done, new commit is pushed.

projectgus:
Thanks @husigeza . I've squashed this commit and cherry-picked into our review & merge queue.

projectgus:
Cherry-picked, thanks!

PR #835, opened 2017-07-25
Fixes timer example using timer_idx_t etc.

lacyrhoades:
Not totally sure I completely understand hardware timers yet but with these changes it does compile and upload! Otherwise it complains about mismatched types `int` vs `timer_idx_t` etc.  😄


lacyrhoades:
Disregard for now. I needed to make these changes when running the sample inside PlatformIO. Ran the code with plain esp-idf + `make` just now, and everything works as-is on `master`. Will continue to investigate thanks!

PR #842, opened 2017-07-25
Check for SDKCONFIG_DEFAULTS at recipe execution

rojer:
This way, defaults need not exist yet at the time of structure evaluation (e.g. can be provided by earlier targets).


projectgus:
Hi @rojer,

Sorry for the super slow reply.

We have a general (unwritten, sorry) rule to try and avoid shell-isms in our Makefiles where we can. Is there any other way you can get around this issue for the use case you have?

Angus

rojer:
in the end, providing SDKCONFIG_DEFAULTS was not enough, i had to override menuconfig and defconfig to make deps on external files work properly, so this is no longer required.

PR #848, opened 2017-07-27
Fix compile error when using timer in C++

imxieyi:
In file included from E:/timer_test/main/main.cpp:11:0:
E:/esp-idf/components/soc/esp32/include/soc/timer_group_struct.h: At global scope:
E:/esp-idf/components/soc/esp32/include/soc/timer_group_struct.h:193:19: error: 'timg_dev_t TIMERG0', declared using anonymous type, is used but never defined [-fpermissive]
 extern timg_dev_t TIMERG0;
                   ^
E:/esp-idf/components/soc/esp32/include/soc/timer_group_struct.h:192:3: note: 'typedef volatile struct<anonymous> timg_dev_t' does not refer to the unqualified type, so it is not used for linkage
 } timg_dev_t;

### DIFF
--- a/components/soc/esp32/include/soc/gpio_sd_struct.h
+++ b/components/soc/esp32/include/soc/gpio_sd_struct.h
@@ -44,5 +44,11 @@ typedef volatile struct {
         uint32_t val;
     } version;
 } gpio_sd_dev_t;
+#ifdef __cplusplus
+extern "C" {
+#endif
 extern gpio_sd_dev_t SIGMADELTA;
+#ifdef __cplusplus
+}
+#endif
 #endif  /* _SOC_GPIO_SD_STRUCT_H_ */
--- a/components/soc/esp32/include/soc/gpio_struct.h
+++ b/components/soc/esp32/include/soc/gpio_struct.h
@@ -200,5 +200,11 @@ typedef volatile struct {
         uint32_t val;
     } func_out_sel_cfg[40];
 } gpio_dev_t;
+#ifdef __cplusplus
+extern "C" {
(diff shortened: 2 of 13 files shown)


yaqwsx:
All the SoC header files miss the `#ifdef __cplusplus extern "C" { #endif` guards. Is this a bug or is there a reason for that? If it is a bug I think it should be corrected everywhere, not only for the timer...

imxieyi:
I don't think all header files should have such guards. As you can see nearly all .h files in soc directory have only macros and structures. Those things don't need 'extern'. What 'extern "C"' does is to use C linker instead of C++ linker. The timer_group_struct.h has two external constants which are defined in a .c file. So it should be linked with C linker otherwise C++ linker cannot find them.

igrr:
I guess what @yaqwsx means is that all the soc/xxx_struct.h files define `extern` global structures, not just the timer_group one.
Your change looks good though, thank you for sending this in. I can either add C++ guards to the rest of the _struct header files myself or wait for you to do this.

yaqwsx:
@igrr Yes, that's exactly what I meant.

imxieyi:
@igrr  I think I have fixed all the remaining extern global structures in soc include directory.

projectgus:
Hi @imxieyi ,

Thanks for sending this. Sorry we didn't get back to you earlier.

As it turns out, we accidentally duplicated effort and there's a change coming which adds these guards. The commit is c7173e0d08. It's not on github right now, but it should appear in the next day or so.

Sorry for the mixup,

Angus

PR #889, opened 2017-08-08
Updated missing includes file.  Fixes  #271.

### LINKED ISSUE #271: Problem including regex.h from newlib
Hi,

I am trying to validate some regular expressions and I found the regex library into the newlib include folder. When I try to include this header in my code, I have some unknown type errors:

~/esp-idf/components/newlib/include/regex.h:43:1: error: unknown type name 'off_t'
 typedef off_t regoff_t;

Also I have tried to find the regex symbols into the newlib build directory but I couldn't find anything.

Can we use this library? Am I missing any link flags or something like that? I use the time.h and works properly.

Regards.

psiphi75:
`#include <sys/stat.h>` was missing.  When compiling without this include you would get the `error: unknown type name 'off_t'` message.

Fixes  #271.

### DIFF
--- a/components/newlib/include/regex.h
+++ b/components/newlib/include/regex.h
@@ -38,6 +38,7 @@
 #define	_REGEX_H_
 
 #include <sys/cdefs.h>
+#include <sys/stat.h>
 
 /* types */
 typedef off_t regoff_t;


psiphi75:
I actually see that the regex library has been removed, so this PR won't do much.  Hence closing.

PR #890, opened 2017-08-08
Add uart api call to use xon/xoff flow control

Ealdwulf:
The uart driver has support for hardware flow control, but we (Kano Computing) are using a flow control pin to control device reset, which I guess is a common use case. We'd therefore like to use software flow control. Since the hardware has this build in, I've made this PR to switch it on. 

We are not especially attached to this API and would be fine with anything which does the same job.


projectgus:
Hi @Ealdwulf ,

Thanks for sending this. Are you able to click-through and sign our CLA as prompted, please? Once CLA is signed we can look at merging the PR.

Angus

Ealdwulf:
Thanks @projectgus, my manager confirmed this was okay, so I have signed it.

projectgus:
Thanks @Ealdwulf . Have cherry-picked this into our internal queue for review & merging.

projectgus:
Cherry-picked as 23f10e1.

Ealdwulf:
Cheers!

PR #903, opened 2017-08-11
Add missing debug log message for out-of-sockets case

slcasner:
This log message is printed in two parts.  For the case of  not being able to allocate a socket because they are all used, the second part of the message was missing.


projectgus:
Thanks @slcasner . Agree this seems to keep consistency. Have cherry-picked it into our review & merge queue.

projectgus:
Cherry-picked, thanks!

Issue #888, opened 2017-08-07
[TW#14607] bta_gattc_register_service_change_notify() causes GATT_Write failures


jkhax0r:
I've tracked down a possible bug/regression that I believe is coming from commit: https://github.com/espressif/esp-idf/commit/c5debed62106e533b740cfae6458f2f55fba8853

If I connect a peripheral device to the ESP32, and then reset the peripheral, when it comes back up and reconnects, I can't write a characteristic anymore.  The first characteristic write I do silently fails with no callback - neither success nor failure.  Only a message in the logs that says "GATTC_Write: GATT_BUSY" (see below logs).  I tracked down that the stack is originally getting into that "busy state" from a GATTC_Write call initiated by the service change discovery code added by this commit.  When I remove all changes related to that commit, it works every time.  

Hopefully I'm not doing something stupid with my peripherals service change service, but best I can tell this is implemented in the peripheral's stack and I couldn't change any code here (it is the CC2540 chipset).

Logs below with GATT VERBOSE logging enabled. 

```
I (19446) ble_rmt: Opening gattc connection for gattc_if=4 to 20:c3:8f:ad:62:50
E (19447) BT: GATT_Connect gatt_if=4
E (19447) BT: gatt_get_ch_state: ch_state=0
E (19447) BT: gatt_set_ch_state: old=0 new=2
E (19449) BT: gatt_update_app_use_link_flag  is_add=1 chk_link=0
E (19449) BT: gatt_update_app_hold_link_status found=1[1-found] idx=0 gatt_if=4 is_add=1
E (19449) BT: gatt_get_ch_state: ch_state=2
E (19450) BT: GATT_GetConnIdIfConnected status=0

I (19450) ble_rmt: REG FOR_NOTIFY: status 0, srvc_id = 1d17, char_id = 1d18
E (19749) BT: GATT   ATT protocol channel with BDA: 20c38fad6250 is connected
E (19749) BT: gatt_is_bda_in_the_srv_chg_clt_list :20-c3-8f-ad-62-50
E (19750) BT: gatt_get_ch_state: ch_state=2
E (19750) BT: gatt_set_ch_state: old=2 new=4
E (19750) BT: gatt_connect_cback: from 20c38fad6250 connected:1 conn_id=1 reason = 0x0000
E (19751) BT: gatt_get_ch_state: ch_state=4
E (19751) BT: GATT_GetConnIdIfConnected status=1

[...]

PR #907, opened 2017-08-14
Failed GATTC_* calls not result in ...

SiarheiVolkau:
... call of corresponding app callback.
For example when GATTC_Write fails with GATT_BUSY error, the callback in app never be called.
This excerpt:
`if (p_clcb->p_q_cmd == NULL) {`
`APPL_TRACE_ERROR("No pending command");`
`return;`
`}`
from `bta_gattc_op_cmpl()` prevents to execute callbacks without `p_clcb->p_q_cmd`.

### DIFF
--- a/components/bt/bluedroid/bta/gatt/bta_gattc_act.c
+++ b/components/bt/bluedroid/bta/gatt/bta_gattc_act.c
@@ -908,11 +908,6 @@ void bta_gattc_cfg_mtu(tBTA_GATTC_CLCB *p_clcb, tBTA_GATTC_DATA *p_data)
 
         /* if failed, return callback here */
         if (status != GATT_SUCCESS && status != GATT_CMD_STARTED) {
-            /* Dequeue the data, if it was enqueued */
-            if (p_clcb->p_q_cmd == p_data) {
-                p_clcb->p_q_cmd = NULL;
-            }
-
             bta_gattc_cmpl_sendmsg(p_clcb->bta_conn_id, GATTC_OPTYPE_CONFIG, status, NULL);
         }
     }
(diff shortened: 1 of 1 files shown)


jkhax0r:
@SiarheiVolkau .  Thanks for the update!  Please see this issue as I'm sure your commit "fixes" it, and I will be anxious to try.  (https://github.com/espressif/esp-idf/issues/888)

However, I'm a bit afraid that this is sort of masking a bigger underlying problem.  And it may be related to the CC2540 since I remember you also had another issue common to me with the CC2540 and overall I seem to have some connecting flakiness that others mostly don't seem to have (except you). 

To summarize that issue 888, this GATTC_Write failed issue was happening for me 100% on the second connect.  Basically I could connect once reliably, kill peripheral power, and on reconnect this would happen every time.  My temporary "fix" was to remove all the code that was doing the service change characteristic discovery that was added by the commit referenced in that issue (https://github.com/espressif/esp-idf/issues/888).

Thanks again for the commit!  I'm anxious to try and hopefully get back to the master branch again soon

SiarheiVolkau:
@jkhax0r this not related to CC2540 or my other PRs or issues.

Im just write firmware that collect messages from external sources and send it via BLE to other device. For this task was picked guaranteed delivery in order by using GATT "Write with response" command and freertos's queue of messages.  I was expect to receive callback for each message i sent, but some writes (with small time in between 2 consecutive writes) not cause callback to be called, since BLE stack still process first message and for second one GATTC_Write() returns GATT_BUSY. Lost callback hungs queue processing.

This patch restore lost callbacks as you see, and potentially can fix #888 with writing special callback handler in user's code.

jkhax0r:
Okay thanks.  I see how https://github.com/espressif/esp-idf/issues/888 has a different cause, but I think this will fix the symptom.

The issue https://github.com/espressif/esp-idf/issues/888 was caused by the stack internally issuing a GATTC_Write() call for the service change characteristic notification, which I believe would fail on my hardware for some reason, and then never clear the pending status.  So then any GATTC_write from the app would lead to the GATT_BUSY error message, but quietly fail and never return a failure callback.  Sounds like your change may fix the internal state hang, but most definitely will fix the missing failure callback.

TianHao-Espressif:
Maybe we need to look for another way to fix this problem.

SiarheiVolkau:
@TianHao-Espressif do what you want.
I just propose change that enlights problem.
This patch works for me and im not see any side effects with it.

TianHao-Espressif:
@SiarheiVolkau , thanks for your kind. We will do something about it.

projectgus:
I'm going to close this PR as I believe this is now fixed in the master branch. @SiarheiVolkau please let us know if that's not the case.

PR #910, opened 2017-08-15
Provide a default for CONFIG_BT_RESERVE_DRAM

rojer:
...when BT is not built

### DIFF
--- a/components/esp32/ld/esp32.ld
+++ b/components/esp32/ld/esp32.ld
@@ -16,6 +16,11 @@
 */
 #include "sdkconfig.h"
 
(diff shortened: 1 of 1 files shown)


rojer:
@igrr for apps that don't use BT we want to be able to not even compile BT (takes a long time and ends up being eliminated anyway). this provides a default for origin computation in that case.

projectgus:
Thanks @rojer, seems reasonable. Have cherry-picked into our review & merge queue.

projectgus:
Cherry-picked, thanks!

PR #950, opened 2017-09-01
Added monkey patch GATTS_INCLUDED guards

Smeedy:
Fix for no member named 'sr_cmd' compile error. Added this missing guard for recent change couple of days ago to the gatt_utils.c. I only BLE scan for adv data in a small project. So no need for GATT server or client in the BT component. This error popped up. Thx


TianHao-Espressif:
Hi，@smeedy, thanks for your PR. 
We fixed it on August 24th.

Issue #906, opened 2017-08-14
[TW#14794] When an additional task calls void esp_task_wdt_feed() only once it never triggers


TomIowaSoft:
When there is only one task in wdt_task_list, and that task only called esp_task_wdt_feed() once, the watchdog triggers but no task name is displayed.

When there is more than one task in wdt_task_list, and an additional task calls  esp_task_wdt_feed() once, no watchdog is triggered.

The reason turned out to be that when a task is added to the list in void esp_task_wdt_feed(), it initializes  newtask->fed_watchdog=true; This is not correct, it should be false. The reason is because it did not participate in the decision to feed.

Dazza0:
@TomIowaSoft task watchdog timer has been overhauled, please refer to commit 9d63e1da4a233f8b3043157f58894fb9b2c2668d

PR #982, opened 2017-09-10
esp_task_wdt_feed() did not reset fed_watchdog flag

honza-klu:
esp_task_wdt_feed() did not reset fed_watchdog flag properly after feeding real watchdog. It skipped last task in the list. Issue is fixed now.
I think that this error can be related to issue #906


negativekelvin:
I think you meant to only remove `->next` and leave `!=`

components/esp32/task_wdt.c
```diff
@@ -131,7 +131,7 @@ void esp_task_wdt_feed() {
         TIMERG0.wdt_feed=1;
         TIMERG0.wdt_wprotect=0;
         //Reset fed_watchdog status
-        for (wdttask=wdt_task_list; wdttask->next!=NULL; wdttask=wdttask->next) wdttask->fed_watchdog=false;
+        for (wdttask=wdt_task_list; wdttask=NULL; wdttask=wdttask->next) wdttask->fed_watchdog=false;
     }
```
PerMalmberg: Surely `wdttask=NULL` should be  `wdttask != NULL`

components/esp32/task_wdt.c
```diff
@@ -131,7 +131,7 @@ void esp_task_wdt_feed() {
         TIMERG0.wdt_feed=1;
         TIMERG0.wdt_wprotect=0;
         //Reset fed_watchdog status
-        for (wdttask=wdt_task_list; wdttask->next!=NULL; wdttask=wdttask->next) wdttask->fed_watchdog=false;
+        for (wdttask=wdt_task_list; wdttask=NULL; wdttask=wdttask->next) wdttask->fed_watchdog=false;
```
PerMalmberg: Surely `wdttask=NULL` should be  `wdttask != NULL`

honza-klu:
Yes, I did, It is fixed now.

Spritetm:
Looks good, but could you please squash your two commits into one?

Spritetm:
Oops, a colleague just pointed out we have some rework on the WDT code going on that also fixes this. WIll keep this open until that is merged, but we may not need this fix. Still thanks for pointing it out, though!

PR #1006, opened 2017-09-16
Add RS485 Half-Duplex Echo example

pctj101:
Tested with 3 node RS485 bus.

Requires Half-Duplex driver code from #667


alisitsyn:
Sorry for the long delay. Espressif has implemented RS-485 support based on the implementation provided here, it will be merged soon.

Issue #1017, opened 2017-09-19
[TW#15678] PPPoS client Crash 


llewellynnu:
Hi, 
So I updated the IDF to the latest commit as of today and also updated the tool chain. There is no change in any of my code. I am running the code on ESP32-WROVER. I get the following crash trace 

0x400d9c5c: ip4_route at /Users/llewellyn/esp/esp-idf/components/lwip/core/ipv4/ip4.c:229 (discriminator 2)

A2      : 0x3ffb6138  A3      : 0x3ffdc248  A4      : 0x00000001  A5      : 0x162af5ad  
A6      : 0x00000000  A7      : 0x00000005  A8      : 0x0000f01d  A9      : 0x3ffda550  
A10     : 0x00000000  A11     : 0x00000000  A12     : 0xf41ea90a  A13     : 0xffffffff  
A14     : 0x00000000  A15     : 0x00000000  SAR     : 0x00000018  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x0000f0d2  LBEG    : 0x4000c28c  LEND    : 0x4000c296  LCOUNT  : 0x00000000  

Backtrace: 0x400d9c5c:0x3ffda570 0x400d9c93:0x3ffda590 0x400e1241:0x3ffda5b0 0x400df45c:0x3ffda5d0 0x400df4d9:0x3ffda610 0x400df8f0:0x3ffda630 0x400dff61:0x3ffda660 0x401115d4:0x3ffda680 0x400dece1:0x3ffda6a0
0x400d9c5c: ip4_route at /Users/llewellyn/esp/esp-idf/components/lwip/core/ipv4/ip4.c:229 (discriminator 2)

0x400d9c93: ip4_route_src at /Users/llewellyn/esp/esp-idf/components/lwip/core/ipv4/ip4.c:159

0x400e1241: udp_sendto at /Users/llewellyn/esp/esp-idf/components/lwip/core/udp.c:569 (discriminator 4)

0x400df45c: dns_send at /Users/llewellyn/esp/esp-idf/components/lwip/core/dns.c:787

0x400df4d9: dns_check_entry at /Users/llewellyn/esp/esp-idf/components/lwip/core/dns.c:984

0x400df8f0: dns_enqueue at /Users/llewellyn/esp/esp-idf/components/lwip/core/dns.c:1388

0x400dff61: dns_gethostbyname_addrtype at /Users/llewellyn/esp/esp-idf/components/lwip/core/dns.c:1493

[...]

PR #1028, opened 2017-09-22
Make PPPoS run (again) after latest lwip changes

loboris:
There was a problen running PPPoS after latest changes to lwip (https://github.com/espressif/esp-idf/issues/1017).
This simple patch makes it run again.

### DIFF
--- a/components/lwip/core/ipv4/ip4.c
+++ b/components/lwip/core/ipv4/ip4.c
@@ -172,9 +172,6 @@ ip4_route_src(const ip4_addr_t *dest, const ip4_addr_t *src)
 struct netif *
 ip4_route(const ip4_addr_t *dest)
 {
-#if ESP_LWIP
-  struct netif *non_default_netif = NULL;
-#endif
   struct netif *netif;
 
 #if LWIP_MULTICAST_TX_OPTIONS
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @loboris,

I haven't had time to look at this properly yet (sorry), but this seems to revert some changes that required to allow APSTA modes to coexist & route properly (ie allow the ESP32 to route to the internet via a STA interface while sending/receiving packets on the AP interface).

Is it possible there is some other way to configure routes & addresses in LWIP to make the PPPoS interface work correctly?

loboris:
I'll try some other solutions.

Usualy, we use PPPoS when we have no WiFi Internet connection and STA mode is not used (and canno't be used), so APSTA mode is also not used. In that case, we connect to the Internet via GSM/PPPoS and (eventualy) use AP mode to connect to the ESP32 (and possibe route the traffic to the Internet).

danicampora:
@loboris I think this should be made to work also when APSTA is used. I'll have a look as well.

loboris:
In PPPoS gw address is set to **10.64.64.64**, we can check it in _ip4_route_ function (ip4.c) and return the _netif_ if other checks fail.
This shouldn't affect APSTA (not tested yet).

projectgus:
Hmm, that's an interesting fix but I don't think hardcoding interface-specific addresses in the IP stack routing layer is a good idea.

It should be possible to externally configure LWIP's routes so that it knows what your preferences for routing are.

loboris:
I'm avare that it is a dirty fix, for now it's working for me until I (or someone else) find the better solution.

@danicampora
It should be written in hex: 0x4040400A (gw address 10.64.64.64)

loboris:
Better solution found:
https://github.com/espressif/esp-idf/pull/1100

PR #1029, opened 2017-09-22
linux-setup.rst: Add note about adding key

hackrid:
for me it was necessary to add the key manually to the key ring. Otherwise I could not have installed the packages


projectgus:
Thanks @hackrid . I must have done this when I installed these AUR packages myself, and forgot.

I've cherry-picked this for merging, with a slight change - rather than listing the command here I referred readers back to the comments in the AUR links. It's less immediately useful, but it should be more future-proof than listing a particular maintainer's public key.

PR #1033, opened 2017-09-24
changed makefile to give correct example

tuskiomi:


projectgus:
Hi @tuskiomi ,

As per discussion here, actual /cygpath/ style here probably won't work unless make is a MINGW32 make:
https://esp32.com/viewtopic.php?f=2&t=2173&p=10320

Maybe it's more accurate to  describe the the /c/xyz style as MINGW32-style?

Angus

PR #1039, opened 2017-09-26
Fix multi_heap_get_info alias

rojer:


projectgus:
Thanks @rojer, sorry for the delay due to the holiday. Queued for merging now.

PR #1040, opened 2017-09-26
Use closesocket instead of close

rojer:
Prevents conflict with close() when it's not provided by lwip.


projectgus:
Hi @rojer,

Sorry for the slow turnaround responding to this. We (finally) have a better fix for this (ability to use unistd.h functions for files & sockets) that should be merged very soon.

Angus

rojer:
no longer relevant

PR #1055, opened 2017-09-30
Use localtime_r instead of gmtime to get timezone

luc-github:
impact on fatfs


projectgus:
Thanks @luc-github . This is now pending review & merge in our internal queue.

PR #1078, opened 2017-10-05
Implemented AVRCP metadata and notification register commands

pufstudio:
Added metadata support so A2DP sink can show track title, artist, album and various other data.


components/bt/bluedroid/api/esp_avrc_api.c
```diff
@@ -70,6 +70,76 @@ esp_err_t esp_avrc_ct_deinit(void)
...
+    btc_msg_t msg;
+    msg.sig = BTC_SIG_API_CALL;
+    msg.pid = BTC_PID_AVRC;
+    msg.act = BTC_AVRC_STATUS_API_SND_META_EVT;
+
+    btc_avrc_args_t arg;
+    memset(&arg, 0, sizeof(btc_avrc_args_t));
+
+    arg.md_cmd.tl = tl;
+    arg.md_cmd.attr_list = attr_list;
```
mywang-espressif: I suggest to transfer data value of attr_list(copy into btc_avrc_args_t) and not just the data pointer, thus application can use local variable and do not need to worry about managing the storage of attribute_list

components/bt/bluedroid/btc/profile/std/a2dp/btc_media_task.c
```diff
@@ -179,11 +179,11 @@ void btc_a2dp_sink_reg_data_cb(esp_a2d_data_cb_t callback)
     bt_av_sink_data_callback = callback;
 }
 
-static inline void btc_a2d_data_cb_to_app(const uint8_t *data, uint32_t len)
+static inline void btc_a2d_data_cb_to_app(const uint8_t *data, uint32_t len, uint32_t sample_rate)
 {
     // todo: critical section protection
     if (bt_av_sink_data_callback) {
-        bt_av_sink_data_callback(data, len);
```
mywang-espressif: For SBC codec, the sample frequency information can be provided in Codec Specific Information Element through ESP_A2D_AUDIO_CFG_EVT, so there is no need to transfer this parameter in the data callback function.

pufstudio: I understand, however A2DP official spec says the following:
>The codec information that determines the bit rate is contained in the SBC frame
header and repeatedly sent to the SNK associated with audio data stream. The SRC
is capable of changing the bit rate dynamically by changing the bitpool parameter
without suspending.

Both 44.1 kHz and 48 kHz sample rate support is mandatory for audio sink
and sample rate can be changed dynamically without pausing the stream
so I saw no other way than to include it in the data callback.

components/bt/bluedroid/api/esp_avrc_api.c
pufstudio: Ah, I see.

There are currently maximum of 7 attributes to use and I just realised that I
should have created a bit mask instead.

Are you ok with that solution?

components/bt/bluedroid/btc/profile/std/avrc/btc_avrc.c
```diff
@@ -390,6 +471,131 @@ static void btc_avrc_ct_deinit(void)
     LOG_INFO("## %s ## completed", __FUNCTION__);
 }
 
+static bt_status_t btc_avrc_ct_send_set_player_value_cmd(uint8_t tl, uint8_t attr_id, uint8_t value_id)
+{
+    tAVRC_STS status = BT_STATUS_UNSUPPORTED;
+
+    if (tl >= 16 || attr_id > ESP_AVRC_PS_MAX_ATTR - 1) {
```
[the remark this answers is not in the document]
pufstudio: Ok. I will fix it.

components/bt/bluedroid/api/esp_avrc_api.c
mywang-espressif: Yes, that's a good idea.

components/bt/bluedroid/btc/profile/std/a2dp/btc_media_task.c
mywang-espressif: I think "bit rate" and "sampling frequency" have different meanings in the A2DP spec. "Sampling Frequency" is the raw sampled data stream which can be configured to be 44.1KHz, 48KHz, etc. "bit rate" indicates the encoded media stream(SBC frames) format which is transmitted over the air, this as you say, can be altered by A2DP source and is related to "bitpool" value as well as "Sampling Frequency."
Sampling frequency and Max/Min SBC bitpool value are determined in the AVDTP signaling procedure, they will not be changed unless reconfigured. Sampling frequency is already provided through "ESP_A2D_AUDIO_CFG_EVT", Since the output of ESP A2DP data callback is the decoded audio stream (PCM), and we don't need to output the bitrate here.

components/bt/bluedroid/btc/profile/std/avrc/btc_avrc.c
```diff
@@ -250,6 +258,75 @@ static void handle_rc_disconnect (tBTA_AV_RC_CLOSE *p_rc_close)
...
+
+    param[i].meta_rsp.attr_id = vendor_msg->p_vendor_data[3 + attr_index] |
+      vendor_msg->p_vendor_data[2 + attr_index] << 8 | vendor_msg->p_vendor_data[1 + attr_index] << 16 |
+      vendor_msg->p_vendor_data[attr_index] << 24;
+
+    btc_avrc_ct_cb_to_app(ESP_AVRC_CT_METADATA_RSP_EVT, &param[i]);
+
+    attr_index += (int) vendor_msg->p_vendor_data[7 + attr_index] + 8;
+  }
+}
```
mywang-espressif: I suggest to remove the static buffer "meta_text_buffer". Instead, in the function "handle_rc_attributes_rsp", to "malloc" and "free" a 128Byte memory is enough, as the attribute text are transfered one by one for each attribute. Or we can use the p_vendor_data[] itself and thus the 128Byte malloc-ed buffer may not be necessary.
In this way we can save the static buffer, and the application layer is responsible to copy the data in the callback function, the example code should be changed accordingly(Implement a "deep copy" function used in bt_app_rc_ct_cb when invoke "bt_app_work_dispatch", and free the buffer after it is used.)

components/bt/bluedroid/btc/profile/std/a2dp/btc_media_task.c
pufstudio: >Sampling frequency is already provided through "ESP_A2D_AUDIO_CFG_EVT".

It is? I did not realise that, my mistake.

pufstudio:
Done. I made fixes you suggested @mywang-espressif.

components/bt/bluedroid/stack/avrc/avrc_bld_ct.c
```diff
@@ -175,6 +175,70 @@ static BT_HDR *avrc_bld_init_cmd_buffer(tAVRC_COMMAND *p_cmd)
...
+**
+** Description      This function builds the Set Player Application Value command.
+**
+** Returns          AVRC_STS_NO_ERROR, if the command is built successfully
+**                  Otherwise, the error code.
+**
+*******************************************************************************/
+static tAVRC_STS avrc_bld_set_player_value_cmd(tAVRC_SET_APP_VALUE_CMD *p_cmd, BT_HDR *p_pkt)
+{
+   int i;
```
mywang-espressif: There is a build warning <unused variable "i">, please fix this.

pufstudio: oops, fixed.

components/bt/bluedroid/btc/profile/std/avrc/btc_avrc.c
```diff
@@ -250,6 +252,77 @@ static void handle_rc_disconnect (tBTA_AV_RC_CLOSE *p_rc_close)
...
+    esp_avrc_ct_cb_param_t param[attr_count];
+    memset(&param[0], 0, sizeof(esp_avrc_ct_cb_param_t) * attr_count);
+
+  for(int i = 0; i < attr_count; i++){
+    attr_length = (int) vendor_msg->p_vendor_data[7 + attr_index] | vendor_msg->p_vendor_data[6 + attr_index] << 8;
+
+    //Received attribute text is not null terminated, so it's useful to know it's length
+    param[i].meta_rsp.attr_length = attr_length;
+    param[i].meta_rsp.attr_text = &vendor_msg->p_vendor_data[8 + attr_index];
+
```
mywang-espressif: The code indentation need to be fixed. You can use the script $IDF_PATH/tools/format.sh to fix the typos.

pufstudio: I ran the script on all the files I modified in this PR, should be ok now.

mywang-espressif:
I don't have more comments for now. Thank you very much for the work, I will forward the PR to my collaborator so that it can be merged.

projectgus:
Cherry-picked in 86fa182 , thank you @pufstudio ! :)

PR #1086, opened 2017-10-07
Add note about esp-idf-git in AUR

jose1711:


projectgus:
Hi @jose1711 ,

Thanks for creating an AUR package for esp-idf! I'm sympathetic to this idea (I'm an Arch user and I use a ton of stuff from AUR.) However I'm afraid we can't really link to a third party package that we don't support, even if it's a simple wrapper around our git repo.

It's also much easier for us, support-wise, if everyone follows the same instructions to clone esp-idf into their working directory. When we get issue reports we already have to navigate a lot of variables about what platform users are running on, and this adds another variable..

I took a quick look at the PKGBUILD as written, and FWIW I also have a few concerns there. Although we try to avoid it, the IDF build system assumes that the user running the build has write access to the IDF_PATH directory. For example, when you run "menuconfig" the conf binaries are built in `tools/kconfig`. I'm not sure how you can handle this while putting the IDF working directory into a global directory like /opt (unless it's chown-ed to one particular user, or world writeable, neither of which seem like a sensible practice.)

Sorry that we can't support this particular change.

Angus

PS FWIW, in case you're interested, an AUR package that installed the xtensa-esp32-elf toolchain in a correct way, similar to other Arch cross-toolchains, might be a different story.

PR #1092, opened 2017-10-09
fix ULP makefile for multiple source files

tomtor:
The original does not handle ULP projects with multiple .S sources.

Only the first file was linked into the final binary.


igrr:
Thank you for the PR @tomtor, it was cherry-picked as 58b411a5fe5345b3f3ad1d2d4c20d909b0ffbfa1.

PR #1100, opened 2017-10-11
Fix type conversion error in components/lwip/api/pppapi.c

Dmitry4Bh:
Pointer tcpip_api_call* m  should be converted to pppapi_msg* instead of pppapi_msg_msg*
in pppapi_do_ppp_set_default(), pppapi_do_ppp_free() and so on.

It solve this issue https://github.com/espressif/esp-idf/pull/1028
so there is no need to patch ip4.c because now netif_defauilt is setted correctly.
Also it prevents memory corruption when pppapi_free() is called.

### DIFF
--- a/components/lwip/api/pppapi.c
+++ b/components/lwip/api/pppapi.c
@@ -48,9 +48,9 @@
 static err_t
 pppapi_do_ppp_set_default(struct tcpip_api_call *m)
 {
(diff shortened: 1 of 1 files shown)


igrr:
The fix was cherry-picked as a0cedb1f442f79e1c87a37c2653226abdd293756, thank you @Dmitry4Bh!

PR #1109, opened 2017-10-12
Potential variable shadowing in ESP_ERROR_CHECK

jeanleflambeur:
Renamed the internal rc to __err_rc to avoid clashes with local variables.
This code would not do the expected thing with the original ESP_ERROR_CHECK macro:

esp_err_t my_func(esp_err_t x)
{
  assert(x == 23);
}

esp_err_t rc = 23; //some value that is important fo the user
ESP_ERROR_CHECK(my_func(rc));

The macro will expand to:
esp_err_t rc = (my_func(rc));

And the code will assert, as my_func will receive a random value - whatever is in the internal macro rc temp variable. This is due to the C weirdness of allowing this code:

int x = x; //x has a random value.

### DIFF
--- a/components/esp32/include/esp_err.h
+++ b/components/esp32/include/esp_err.h
@@ -64,14 +64,14 @@ void _esp_error_check_failed(esp_err_t rc, const char *file, int line, const cha
  */
 #ifdef NDEBUG
 #define ESP_ERROR_CHECK(x) do {                                         \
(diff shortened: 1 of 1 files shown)


igrr:
Thanks for spotting this issue and for the PR, @jeanleflambeur! It was cherry-picked as 8712fd3ccfd6edde17305b1c8ff51b8d3ecc11d0.

PR #1138, opened 2017-10-19
components/openssl: Add support for SNI with SSL_set_tlsext_host_name

ajs124:
I'm not very familiar with this project and its codebase (yet), but I tried sending some API requests with esp-request, which did not work, as the server required SNI, so I added it.

### DIFF
--- a/components/openssl/include/internal/ssl_methods.h
+++ b/components/openssl/include/internal/ssl_methods.h
@@ -31,7 +31,8 @@
                     set_fd, get_fd, \
                     set_bufflen, \
                     get_verify_result, \
-                    get_state) \
+                    get_state, \
+                    set_hostname) \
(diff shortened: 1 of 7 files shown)


projectgus:
Thanks for sending this, @ajs124 . I'm sincerely sorry noone has gotten back to you earlier.

In a piece of poor coordination, it looks like we've recently implemented this in-house as b65f47c5869. I think the functionality we have now covers the functionality that's in this PR, but you're very welcome to reopen & update the PR if that's not the case.

ajs124:
No problem @projectgus, thanks for responding anyways. As long as this made its way into the codebase somehow, I'm happy.

Although I switched away from the OpenSSL API wrapper, since it does not verify certificates in any way, as far as I can tell and therefore allows for more or less trivial man-in-the-middle attacks.

PR #1146, opened 2017-10-20
When pressing a special key in idf_monitor, the monitor crashes

chanibal:
When running `make monitor` and pressing a key that doesn't translate to ASCII well (for example up arrow), `idf_monitor.py` exits with an exception:

```
  File "(...)/tools/idf_monitor.py", line 568, in <module>
    main()
  File "(...)/tools/idf_monitor.py", line 502, in main
    monitor.main_loop()
  File "(...)/tools/idf_monitor.py", line 267, in main_loop
    self.handle_key(data)
  File "(...)/tools/idf_monitor.py", line 292, in handle_key
    self.serial.write(codecs.encode(key))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe0' in position 0: ordinal not in range(128)
make: *** [(...)/components/esptool_py/Makefile.projbuild:96: monitor] Error 1
```

Added pull request fixing the issue.

Windows 10, msys32 from latest Espressif's bundle.

### DIFF
--- a/components/driver/include/driver/rmt.h
+++ b/components/driver/include/driver/rmt.h
@@ -119,6 +119,8 @@ typedef struct {
 
 typedef intr_handle_t rmt_isr_handle_t;
 
+typedef void (*rmt_on_tx_end_t)(rmt_channel_t channel);
+
 /**
  * @brief Set RMT clock divider, channel clock is divided from source clock.
  *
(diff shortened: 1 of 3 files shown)


igrr:
We have cherry-picked the idf_monitor fix as 5827fd8c715990a21088d1c6e7a9fe517f908486, thank you! The second commit seems to be related to your other PR.

chanibal:
Yes, of course I meant only the first one. Didn't know github adds successive commits to pull requests after it was posted to the forked repo. Looks like every pull request from now on will have to be in it's own branch.

PR #1165, opened 2017-10-26
Ensure that C++ and C compilers get the same preprocessor setup

APokorny:
CPPFLAGS is applied for both languages while CFLAGS only for C. Without that users integrating C++ components into the esp-idf build system will run into weird runtime behavior when interacting with mbedtls, because structure sizes will differ.

Signed-off-by: Andreas Pokorny <andreas.pokorny@siemens.com>


projectgus:
Ouch! Good catch, thank you @APokorny .

PR #1170, opened 2017-10-27
components/mdns: wrong Message compression detect

SiarheiVolkau:
This patch fix wrong Message compression detection.

Old behavior assumes message compressed when any of 2 most significant bits are set,
But in fact Message compressed only when both those bits are set to 1.

Also there fixed maximal label length check (see https://tools.ietf.org/html/rfc6762#appendix-C).


SiarheiVolkau:
Recreated with clean patchset: https://github.com/espressif/esp-idf/pull/1171

PR #1171, opened 2017-10-27
components/mdns: wrong Message compression detect

SiarheiVolkau:
This patch fix wrong Message compression detection.

Old behavior assumes message compressed when any of 2 most significant bits are set,
But in fact Message compressed only when both those bits are set to 1.

Also there fixed maximal label length check (see https://tools.ietf.org/html/rfc6762#appendix-C).

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -555,9 +555,9 @@ static const uint8_t * _mdns_read_fqdn(const uint8_t * packet, const uint8_t * s
             return NULL;
         }
         uint8_t len = start[index++];
-        if ((len & 0xC0) == 0) {
-            if (len > 64) {
-                //length can not be more than 64
+        if (len < 192) {
+            if (len > 63) {
+                //length can not be more than 63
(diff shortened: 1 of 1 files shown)


components/mdns/mdns.c
```diff
@@ -555,9 +555,9 @@ static const uint8_t * _mdns_read_fqdn(const uint8_t * packet, const uint8_t * s
             return NULL;
         }
         uint8_t len = start[index++];
-        if ((len & 0xC0) == 0) {
-            if (len > 64) {
-                //length can not be more than 64
+        if (len < 192) {
```
me-no-dev: Can we keep this to ```0xC0``` for clarity?

SiarheiVolkau: Of course, do what you want.

igrr:
Cherry-picked as 6e24566186c52dc5432b6b25c81abda577c21e85, made the change suggested by @me-no-dev. Thanks @SiarheiVolkau!

Issue #1167, opened 2017-10-26
[TW#16117] Cannot disable displaying some logs (ESP_EARLY_LOGx)


chanibal:
One annoying example is RMT with it's "D (377817) rmt: RMT INTR : TX END" that can repeat tens of times per second on some uses.

`ESP_EARLY_LOGD` does not seem to listen to `esp_log_level_set`, it can be controlled by the define `LOG_LOCAL_LEVEL` - but this usually requires editing IDF's source to change this.

Is there any, cleaner way to disable the logs?

chanibal:
These logs have been moved to verbose level. Probably will not get a better fix than this, closing.

PR #1174, opened 2017-10-27
rmt: Adds a simple handler that can be called when transmission ends.

chanibal:
Adds an easy way to know when RMT has finished transmitting, by adding a handler that will be executed in interrupt context.

Also removes a debug message that might spam the logs when TX has finished (see #1167), but only if the new functionality is used - because you can easily add this log using the newly added handle.

### DIFF
--- a/components/driver/rmt.c
+++ b/components/driver/rmt.c
@@ -75,6 +75,9 @@ typedef struct {
 
 rmt_obj_t* p_rmt_obj[RMT_CHANNEL_MAX] = {0};
 
+// Event called when transmission is ended
+static rmt_tx_end_t rmt_tx_end_callback = NULL;
+
 static void rmt_set_tx_wrap_en(rmt_channel_t channel, bool en)
 {
     portENTER_CRITICAL(&rmt_spinlock);
@@ -540,7 +543,7 @@ static void IRAM_ATTR rmt_driver_isr_default(void* arg)
                 switch(i % 3) {
                     //TX END
                     case 0:
-                        ESP_EARLY_LOGD(RMT_TAG, "RMT INTR : TX END");
+                        ESP_EARLY_LOGV(RMT_TAG, "RMT INTR : TX END");
                         xSemaphoreGiveFromISR(p_rmt->tx_sem, &HPTaskAwoken);
                         if(HPTaskAwoken == pdTRUE) {
                             portYIELD_FROM_ISR();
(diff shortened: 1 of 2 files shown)


components/driver/rmt.c
```diff
@@ -538,7 +541,8 @@ static void IRAM_ATTR rmt_driver_isr_default(void* arg)
                 switch(i % 3) {
                     //TX END
                     case 0:
-                        ESP_EARLY_LOGD(RMT_TAG, "RMT INTR : TX END");
+                        if(rmt_on_tx_end_handler == NULL)
+                            ESP_EARLY_LOGD(RMT_TAG, "RMT INTR : TX END");
```
projectgus: I appreciate your consideration for not changing the existing behaviour, but I think it's OK in this case to downgrade this message to ESP_EARLY_LOGV so it only gets logged at the highest "verbose" level. What do you think?

components/driver/rmt.c
```diff
@@ -547,6 +551,8 @@ static void IRAM_ATTR rmt_driver_isr_default(void* arg)
                         p_rmt->tx_len_rem = 0;
                         p_rmt->tx_offset = 0;
                         p_rmt->tx_sub_len = 0;
+                        if(rmt_on_tx_end_handler != NULL)
+                            rmt_on_tx_end_handler(channel);
```
projectgus: We try to use curly braces around single-line if statements, could you please reformat?

components/driver/rmt.c
chanibal: You mean downgrading to LOGV and keeping it only when `rmt_on_tx_end_handler ` is used - or keeping as LOGV and always executing (removing the if in line 544)?

components/driver/rmt.c
chanibal: Sorry, didn't notice that. I'll fix this with the above.

components/driver/rmt.c
projectgus: I meant downgrading to LOGV and always executing - my take is that this is a fairly low-level "verbose" debugging statement. Maybe there's a use case I'm missing, though?

chanibal:
Fixed, and also took the liberty to rename the newly added method from <del>`rmt_on_tx_end`</del> to `rmt_register_tx_end_callback`, because all functions in IDF that do these kind of things have a verb in them. Also this is not a handler as it doesn't have to handle anything, it's a way to add functionality, so callback is the better term. Please correct me if I'm wrong with this.

Also changed the type from <del>`rmt_on_tx_end_t`</del> to `rmt_tx_end_t` and private/static variable to `rmt_tx_end_callback`.

Also, if LOGD has been changed to LOGV in the TX part, shouldn't it also be changed in the RX part? Not exactly the topic of this patch, but might be a good idea to change both of them to the same level.

projectgus:
> Also, if LOGD has been changed to LOGV in the TX part, shouldn't it also be changed in the RX part? Not exactly the topic of this patch, but might be a good idea to change both of them to the same level.

Seems like a good idea to me, are you able to add it in the scope of this PR?

Also, if you can [squash the PR to a single commit](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit/) that would be great (github provides this feature in their UI but we merge PR branches separately to github).

chanibal:
Commits squashed, RX message changed to LOGV.

projectgus:
Cherry-picked as 1066c74, with a small tweak (adding a callback argument) in 2227c44. Thanks!

PR #1187, opened 2017-10-29
freertos: minor issue in documentation snippet of queue.

chanibal:
`taskYIELD` was used in ISR context, but `portYIELD_FROM_ISR` should instead.


igrr:
Thanks for the PR, cherry-picked as 4da90f0.

PR #1192, opened 2017-10-30
monitor: WLS workaround for ConsoleReader cancel

enrikb:
This patch addresses the fact that TIOCSTI is not available on WSL /
bash on windows. It avoids calling getkey() when no data is available
very much like the workaround already available for Windows.

(Note that pthread_kill() is not available before python 3.3 and serial
read with timeout (VMIN = 0, VTIME = 1) does not work on WSL, either.
Therefore, the workaround was implemented using peek/wait.)

### DIFF
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -134,6 +134,14 @@ class ConsoleReader(StoppableThread):
         super(ConsoleReader, self).__init__()
         self.console = console
         self.event_queue = event_queue
+        self.posix_is_wsl = False
+        if os.name == 'posix':
+            try:
+                with open('/proc/version', 'r') as pv:
+                    if 'microsoft' in pv.read().lower():
+                        self.posix_is_wsl = True
+            except:
+                pass
 
     def run(self):
         self.console.setup()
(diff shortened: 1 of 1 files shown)


tools/idf_monitor.py
```diff
@@ -159,7 +179,9 @@ def run(self):
             self.console.cleanup()
 
     def _cancel(self):
-        if os.name == 'posix':
+        if self.wsl:
+            pass
```
projectgus: What do you think about `if os.name == 'posix' and not self.is_wsl` ?

enrikb: I would prefer to keep the WSL workaround separate from the original implementation, as the original will not introduce extra latency when typing on the console (once stdin from console will be implemented ...).

enrikb: Uh, sorry, I misunderstood your suggestion.
I hope I got it now. I have slightly reworked the variable naming, changed the condition like you suggested and already squashed/rebased the commit.

enrikb:
Unfortunately, the method used in my change does not work at all due to stdin etc. buffering. For example, escape sequences will starve in the input buffer. Most probably select() has been removed from the original miniterm for the same reason, as mentioned in the comments.

At this time, I will cancel the pull request until I will hopefully find a better solution.

PR #1206, opened 2017-10-31
DHCP Server: option to configure lease time multiplier and number of max stations

chanibal:
Currently the two values are hardcoded, and I had an issue where I had to change them, and it was not possible in current IDF.

The values are settable in `make menuconfig` and have their defaults set to the original values.


components/lwip/Kconfig
```diff
@@ -111,6 +111,24 @@ config LWIP_DHCP_DOES_ARP_CHECK
         Enabling this option performs a check (via ARP request) if the offered IP address
         is not already in use by another host on the network.
 
+config LWIP_DHCPS_LEASE_UNIT
```
projectgus: Even though there's only two items here, what do you think about grouping these in a submenu?

components/lwip/Kconfig
```diff
@@ -111,6 +111,24 @@ config LWIP_DHCP_DOES_ARP_CHECK
...
+    int "DHCP server: Multipiler for lease time, in seconds"
+    range 1 3600
+    default 60
+    help
+        The DHCP server is calculating lease time multiplying the sent 
+        and received times by this number of seconds per unit. 
+        The default is 60, that equals one minute.
+
+config LWIP_MAX_STATION_NUM
```
projectgus: Could you please namespace this as LWIP_DHCPS_, similar to the other new config item?

components/lwip/Kconfig
```diff
@@ -111,6 +111,24 @@ config LWIP_DHCP_DOES_ARP_CHECK
         Enabling this option performs a check (via ARP request) if the offered IP address
         is not already in use by another host on the network.
 
+config LWIP_DHCPS_LEASE_UNIT
+    int "DHCP server: Multipiler for lease time, in seconds"
```
projectgus: Typo: Multiplier

chanibal: Fixed

components/lwip/Kconfig
chanibal: Fixed

components/lwip/Kconfig
chanibal: Added a submenu

chanibal:
Fixed all issues, @projectgus. Frankly, the multiplier option shouldn't be needed at all, IMHO. It should be hardcoded as one second - but configuring it this way gives backwards compatibility and if someone needs to find this, they will.

projectgus:
Thanks @chanibal . Hava squashed these into one commit and cherry-picked into our merge queue.

projectgus:
Cherry-picked in 626ad5f, thanks!

PR #1209, opened 2017-11-01
monitor: make Ctrl-] work when running on WSL (v2)

enrikb:
Because the TIOCSTI termios ioctl is not supported on WSL (bash on
windows) at this time, another way for cancelling a blocking read on
stdin is needed. Also, setting a timeout using termios.c_cc[VMIN] = 0
with termios.c_cc[VTIME] = 1 does not currently work for WSL.

A first approach to mimic the Windows kludge based on kbhit() by using
ioctl(TIOCINQ) for peeking at the input queue before calling read()
basically worked but failed badly for fast multi character input like
escape sequences sent back by the terminal. This can be easily
reproduced using the console example (examples/system/console). The root
cause was identified to be the buffering involved reading and decoding
stdin in python.

The current solution uses pthread_kill() via ctypes (python 2) to
unblock getkey() in ConsoleReader which seems to be working well.
However, even in this case stdin must have been re-opened without
buffering. Otherwise, the read() may be restarted internally after
seeing EINTR.

This change has been successfully tested on WSL with the console
example, escape sequences (UP, DOWN keys), Ctrl-] to leave the monitor
and Ctrl-T/Ctrl-A for re-flashing. Unfortunately, testing on other
systems was not possible at this time.

### DIFF
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -175,6 +175,8 @@ class ConsoleReader(StoppableThread):
                     if not self.alive:
                         # expected during _cancel when killed
                         break
+                    else:
+                        raise
                 if c is not None:
                     self.event_queue.put((TAG_KEY, c), False)
         finally:
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -135,9 +135,8 @@ class ConsoleReader(StoppableThread):
         super(ConsoleReader, self).__init__()
         self.console = console
         self.event_queue = event_queue
+        self.running_on_wsl = running_on_wsl
         self.tid = None
-        self.pthread_self = None
-        self.pthread_kill = None
         if running_on_wsl:
             libpthread = ctypes.CDLL("libpthread.so.0")
             self.pthread_self = libpthread.pthread_self
@@ -152,7 +151,7 @@ class ConsoleReader(StoppableThread):
         pass
 
     def run(self):
-        if self.pthread_self != None:
+        if self.running_on_wsl:
             self.tid = self.pthread_self()
         self.console.setup()
         try:
@@ -183,8 +182,8 @@ class ConsoleReader(StoppableThread):
             self.console.cleanup()
 
     def _cancel(self):
-        if self.pthread_kill != None:
-            if self.tid != None:
+        if self.running_on_wsl:
+            if self.tid is not None:
                 self.pthread_kill(self.tid, signal.SIGRTMIN)
         elif os.name == 'posix':
             # this is the way cancel() is implemented in pyserial 3.3 or newer,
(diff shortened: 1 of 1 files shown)


enrikb:
Please note that I'm a python beginner. If there are better ways e.g. to 'unbuffer' stdin etc., please let me know.

tools/idf_monitor.py
```diff
@@ -130,12 +131,29 @@ class ConsoleReader(StoppableThread):
...
+            self.pthread_kill = libpthread.pthread_kill
+            self.pthread_kill.argtypes = [ ctypes.c_void_p, ctypes.c_int ]
+            self.pthread_kill.restype = ctypes.c_int
+            signal.signal(signal.SIGRTMIN, self._handler)
+
+    def _handler(self, signum, frame):
+        pass
 
     def run(self):
+        if self.pthread_self != None:
```
projectgus: Typical Python style is to use "is not None" instead of "!= None".

In this case (and in the same check below) I'd be more comfortable using `if running_on_wsl` everywhere as this makes it easier to trace all the of the WSL workarounds (which can hopefully some day be removed!)

tools/idf_monitor.py
```diff
@@ -153,13 +171,20 @@ def run(self):
                     c = self.console.getkey()
                 except KeyboardInterrupt:
                     c = '\x03'
+                except IOError:
+                    if not self.alive:
+                        # expected during _cancel when killed
+                        break
```
projectgus: This code swallows the IOError if self.alive is true. Does it make more sense to re-raise it?

tools/idf_monitor.py
```diff
@@ -220,6 +245,20 @@ class Monitor(object):
     """
     def __init__(self, serial_instance, elf_file, make="make", toolchain_prefix=DEFAULT_TOOLCHAIN_PREFIX, eol="CRLF"):
         super(Monitor, self).__init__()
+        posix_is_wsl = False
```
projectgus: What do you think about naming this running_on_wsl to match the field used in the Console instance? Another change that makes it easier to search for all of the WSL-related hacks at once. :)

tools/idf_monitor.py
projectgus: (Also if you don't need to test whether `self.pthread_self` is None, you can remove the `self.pthread_self = None` default value settings from the constructor)

enrikb:
Thank you very much for the valuable feedback! I will take care ASAP, but not before 2017-11-26 as I will be travelling till Sunday.

enrikb:
Hi, I tried to implemented the requested changes.
What is stille really unclean is the signal handler registration in ConsoleReader __init__.
Wouldn't it be better to do this in main() to have a better defined lifetime of the handler?

enrikb:
Anyway, I will try to find a select() based solution, now that it is clear that `unbuffered stdin' seems to be key. I would also be more comfortable without having to use pthread_kill() ;-)

projectgus:
> What is stille really unclean is the signal handler registration in ConsoleReader init.
> Wouldn't it be better to do this in main() to have a better defined lifetime of the handler?

This seems like a good idea. You can probalby make the `running_in_wsl` test a module-level variable, and set it at import time (ie run the test at the top level of the module source.)

> Anyway, I will try to find a select() based solution, now that it is clear that `unbuffered stdin' seems to be key. I would also be more comfortable without having to use pthread_kill() ;-)

OK, that sounds great. I will hold off on doing anything more with this PR until I hear how you get on. :)

projectgus:
Closing in favour of new approach in #1324.

PR #1217, opened 2017-11-02
Don't disable UART1 CLK when UART1 is console (should fix #1131)

### LINKED ISSUE #1131: [TW#15931] No custom log output on UART1 GPIO16,17 after boot
## Expected Behavior
When selecting custom log output UART1, RX: GPIO16, TX: GPIO17, all stuff loged via ESP_LOG macros shall be visible via serial.

## Observed behavior
Since a couple of weeks (sorry, can't pinpoint it), after pulling latest idf: All boatloader messages are visible, then when the app starts, no log output is visible. There are three ways to mitigate this right now:

- choose UART0 (not an option)
- On native esp-idf run
```cpp
void app_main(void)
{
    DPORT_SET_PERI_REG_MASK(DPORT_PERIP_CLK_EN_REG, DPORT_UART1_CLK_EN);
    DPORT_CLEAR_PERI_REG_MASK(DPORT_PERIP_RST_EN_REG, DPORT_UART1_RST);
    // ....
}
```
- On Arduino run 
```cpp
    HardwareSerial serialUART1(1);
    serialUART1.begin(115200, SERIAL_8N1, 16, 17);
    serialUART1.println("Hello on UART1");
    serialUART1.flush(); 
```

Both of the last hacks work, but it would be nice to find the root cause of this. IMHO ESP_LOG should always work when the correct settings in `sdkconfig `are set.

## Steps to Reproduce

- In `sdconfig `set:
```bash
CONFIG_CONSOLE_UART_CUSTOM=y
CONFIG_CONSOLE_UART_CUSTOM_NUM_1=y
CONFIG_CONSOLE_UART_NUM=1
CONFIG_CONSOLE_UART_TX_GPIO=17
CONFIG_CONSOLE_UART_RX_GPIO=16
CONFIG_CONSOLE_UART_BAUDRATE=115200
```
- Compile, flash an reboot device
- No log output from app visible
- Try hacks from [Observed behavior](#observed-behavior)
- No it works again

## Notes
This is also mentioned in https://www.esp32.com/viewtopic.php?f=13&t=3166

Curclamas:
This fixes #1131

### DIFF
--- a/components/esp32/clk.c
+++ b/components/esp32/clk.c
@@ -186,7 +186,15 @@ void esp_perip_clk_init(void)
     else {
         common_perip_clk = DPORT_WDG_CLK_EN |
                               DPORT_I2S0_CLK_EN |
+#if CONFIG_CONSOLE_UART_NUM != 0
+                              DPORT_UART0_CLK_EN |
+#endif
+#if CONFIG_CONSOLE_UART_NUM != 1
                               DPORT_UART1_CLK_EN |
+#endif
+#if CONFIG_CONSOLE_UART_NUM != 2
+                              DPORT_UART2_CLK_EN |
+#endif
                               DPORT_SPI_CLK_EN |
                               DPORT_I2C_EXT0_CLK_EN |
                               DPORT_UHCI0_CLK_EN |
@@ -205,7 +213,6 @@ void esp_perip_clk_init(void)
                               DPORT_PWM1_CLK_EN |
                               DPORT_I2S1_CLK_EN |
                               DPORT_SPI_DMA_CLK_EN |
-                              DPORT_UART2_CLK_EN |
                               DPORT_PWM2_CLK_EN |
                               DPORT_PWM3_CLK_EN;
         hwcrypto_perip_clk = DPORT_PERI_EN_AES |


negativekelvin:
Should DPORT_UART0_CLK_EN be disabled instead?

Curclamas:
IMHO, it could be, but I am not sure about possible side effects and whether this would constitute any benefit. Possibly someone from espressif can shead light on this?

igrr:
I think @negativekelvin is right and you should be able to clock gate UART peripheral which is not used for console output. So possibly something like
```c++
#if CONFIG_CONSOLE_UART_NUM == 0
                               DPORT_UART1_CLK_EN |
#endif
#if CONFIG_CONSOLE_UART_NUM == 1
                               DPORT_UART0_CLK_EN |
#endif
should cover both cases.

Curclamas:
@igrr One thing just for completeness: is it safe, that UART2 will never be used as log output? The docs [mention](http://esp-idf.readthedocs.io/en/latest/api-reference/kconfig.html#console-uart-num) a ROM-bug preventing UART2 from being used. Is that fixed in Rev1 (or possibly later)?

One more thing just out of curiosity: What is the benefit of disabling the clocks? Power saving, more stable operation, CPU load reduction?

igrr:
The issue is that ROM code was compiled with incorrect base address for UART2 peripheral. We can still provide our own hooks which output to UART, the main problem is what hooks to use when bootloader is transferring control to the application. There will be some moment when bootloader IRAM code is already not accessible, but application has not yet installed its own hooks. Perhaps we can install UART hooks as one of the first things in application startup code. Will look into this. For now it is safe to disable UART2 on startup, if UART2 is added as console option, we will modify this part accordingly.

Disabling all peripheral clocks saves a few mA from current consumption.

Curclamas:
@igrr Thanks for the explanation! I've pushed accordingly. For now I've included a case for UART2, I think that makes the code more obvious, covers the possible future case and probably won't hurt.

igrr:
@Curclamas  could you please squash the two commits into one? Thanks.

Curclamas:
@igrr Check!

igrr:
Cherry-picked in 20ccec10619e7f2993f4fed9cdd853ec6088216c, thank you for the PR @Curclamas!

negativekelvin:
https://github.com/espressif/esp-idf/issues/1301

Oops I think we got the name wrong

PR #1229, opened 2017-11-05
Fix: Use ESP_LOGI instead of ESP_LOGE to log the start of scanning process

DominikPalo:
In the `spp_client_demo` example, the information about the start of scanning process was incorrectly logged as an error.


projectgus:
Thanks @DominikPalo . I've cherry-picked this into our internal review & merge queue.

igrr:
Thanks for the PR @DominikPalo, it was cherry-picked as ad8ebe5.

PR #1238, opened 2017-11-07
Call heap_caps_enable_nonos_stack_heaps() earlier

rojer:
Just before starting the scheduler, since starting the scheduler involves allocating significant amount of memory for tasks.
It should be safe to do, since no other allocations take place and ROM functions are not being used.

### DIFF
--- a/components/esp32/cpu_start.c
+++ b/components/esp32/cpu_start.c
@@ -73,13 +73,20 @@
 #define STRINGIFY(s) STRINGIFY2(s)
 #define STRINGIFY2(s) #s
 
+#define CPU_STARTUP_STACK_SIZE 2048
+
 void start_cpu0(void) __attribute__((weak, alias("start_cpu0_default"))) __attribute__((noreturn));
 void start_cpu0_default(void) IRAM_ATTR __attribute__((noreturn));
+static uint32_t *cpu0_startup_stack;
 #if !CONFIG_FREERTOS_UNICORE
 static void IRAM_ATTR call_start_cpu1() __attribute__((noreturn));
 void start_cpu1(void) __attribute__((weak, alias("start_cpu1_default"))) __attribute__((noreturn));
 void start_cpu1_default(void) IRAM_ATTR __attribute__((noreturn));
-static bool app_cpu_started = false;
+
+/* This variable is used to coordinate startup between PRO and APP CPUs. */
+enum app_cpu_phase { APP_CPU_START, APP_CPU_STARTED, APP_CPU_HEAP_OK, APP_CPU_STACK_OK };
+static volatile enum app_cpu_phase app_cpu_phase = APP_CPU_START;
+static uint32_t *cpu1_startup_stack;
 #endif //!CONFIG_FREERTOS_UNICORE
 
 static void do_global_ctors(void);
@@ -101,6 +108,10 @@ static const char* TAG = "cpu_start";
 struct object { long placeholder[ 10 ]; };
 void __register_frame_info (const void *begin, struct object *ob);
 extern char __eh_frame[];
+
+/* In switch_stack.S */
+void switch_stack_and_jump(void *stack, uint32_t size, void (*target)(void))
+    __attribute__((noreturn));
 
 /*
  * We arrive here after the bootloader finished loading the program from flash. The hardware is mostly uninitialized,
@@ -179,9 +190,7 @@ void IRAM_ATTR call_start_cpu0()
     }
     ets_set_appcpu_boot_addr((uint32_t)call_start_cpu1);
 
-    while (!app_cpu_started) {
-        ets_delay_us(100);
-    }
+    while (app_cpu_phase != APP_CPU_STARTED);
 #else
     ESP_EARLY_LOGI(TAG, "Single core mode");
     DPORT_CLEAR_PERI_REG_MASK(DPORT_APPCPU_CTRL_B_REG, DPORT_APPCPU_CLKGATE_EN);
@@ -206,8 +215,13 @@ void IRAM_ATTR call_start_cpu0()
        fail initializing it properly. */
     heap_caps_init();
 
-    ESP_EARLY_LOGI(TAG, "Pro cpu start user code");
-    start_cpu0();
+#if !CONFIG_FREERTOS_UNICORE
+    app_cpu_phase = APP_CPU_HEAP_OK;
+#endif
+
+    cpu0_startup_stack = (uint32_t *) heap_caps_malloc(CPU_STARTUP_STACK_SIZE, MALLOC_CAP_8BIT);
+    ESP_EARLY_LOGD(TAG, "%s cpu start user code, stack at %p", "PRO", cpu0_startup_stack);
+    switch_stack_and_jump(cpu0_startup_stack, CPU_STARTUP_STACK_SIZE, start_cpu0);
 }
 
 #if !CONFIG_FREERTOS_UNICORE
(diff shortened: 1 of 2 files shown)


rojer:
@projectgus this one is for you

projectgus:
I share the concern @igrr has.

We are aware of this general problem (especially if there's a lot of statically allocated RAM). The solution, we think, is going to be moving the startup stack somewhere else in memory so that we can reduce the amount of memory reserved for startup stacks. I have a non-working patch for this somewhere if you want to see it, but I haven't got around to finishing it yet (you'd be welcome to start from there and get something going, though.)

rojer:
yeah, good point @igrr. @projectgus - show me the patch, maybe i can finish it. this prevents us from updating the IDF.

projectgus:
Here you go:
https://gist.github.com/projectgus/29e9f3b3d344bfc4d4a0371cc314683c#file-0001-wip-allow-first-128kb-of-dram-to-be-allocated-static-patch

 This is a fairly old patch and a WIP (predates new heap management code I think), so it probably doesn't apply cleanly to the github master. There are a bunch of things missing, but it may give you something to work from...

projectgus:
(I should be clear, I don't remember if this patch even works for the part which is implemented here...)

rojer:
@projectgus PTAL
this is mostly based on your change, but i went one step further and actually allocate temp stacks on the heap - hey, why not? :)

rojer:
```
[Nov 10 14:47:34.195] I (408) cpu_start: Pro cpu up.
[Nov 10 14:47:34.195] I (409) heap_init: Initializing. RAM available for dynamic allocation:
[Nov 10 14:47:34.217] I (413) heap_init: At 3FFAE6E0 len 00001920 (6 KiB): DRAM
[Nov 10 14:47:34.218] I (419) heap_init: At 3FFB3D50 len 0002C2B0 (176 KiB): DRAM
[Nov 10 14:47:34.218] I (426) heap_init: At 3FFE0440 len 00003BC0 (14 KiB): D/IRAM
[Nov 10 14:47:34.239] I (432) heap_init: At 3FFE4350 len 0001BCB0 (111 KiB): D/IRAM
[Nov 10 14:47:34.239] I (438) heap_init: At 40089758 len 000168A8 (90 KiB): IRAM
[Nov 10 14:47:34.240] I (445) cpu_start: Starting app cpu, entry point is 0x400806a4
[Nov 10 14:47:34.240] I (0) cpu_start: APP cpu start user code, stack at 0x3ffae7b4
[Nov 10 14:47:34.252] I (458) cpu_start: PRO cpu start user code, stack at 0x3ffaefb8
[Nov 10 14:47:34.270] I (142) cpu_start: Starting scheduler on PRO CPU.
[Nov 10 14:47:34.270] I (0) cpu_start: Starting scheduler on APP CPU.
[Nov 10 14:47:34.270] I (145) cpu_start: Entering app_main
```

components/esp32/cpu_start.c
```diff
@@ -102,6 +102,15 @@ struct object { long placeholder[ 10 ]; };
 void __register_frame_info (const void *begin, struct object *ob);
 extern char __eh_frame[];
 
+uint32_t *cpu0_stack;
+#if !CONFIG_FREERTOS_UNICORE
+uint32_t *cpu1_stack;
```
projectgus: Can these two be declared static?

components/esp32/cpu_start.c
```diff
@@ -161,6 +170,16 @@ void IRAM_ATTR call_start_cpu0()
 
     ESP_EARLY_LOGI(TAG, "Pro cpu up.");
 
+    /* Initialize heap allocator. WARNING: This *needs* to happen *after* the app cpu has booted.
```
projectgus: I haven't looked closely yet, but I'm a little concerned about the point mentioned in this comment.

components/esp32/cpu_start.c
```diff
@@ -237,9 +247,10 @@ void IRAM_ATTR call_start_cpu1()
 #endif
 
     wdt_reset_cpu1_info_enable();
-    ESP_EARLY_LOGI(TAG, "App cpu up.");
+    cpu1_stack = (uint32_t *) heap_caps_malloc(2048, MALLOC_CAP_8BIT);
```
projectgus: If you only need 8-bit memory then you should be able to call plain malloc() here

components/esp32/cpu_start.c
```diff
@@ -415,15 +442,16 @@ static void main_task(void* args)
     // Now that the application is about to start, disable boot watchdogs
     REG_CLR_BIT(TIMG_WDTCONFIG0_REG(0), TIMG_WDT_FLASHBOOT_MOD_EN_S);
     REG_CLR_BIT(RTC_CNTL_WDTCONFIG0_REG, RTC_CNTL_WDT_FLASHBOOT_MOD_EN);
+
+    heap_caps_free(cpu0_stack);
```
projectgus: Similarly, I think you can call free() for these.

projectgus:
I left some minor comments but generally this looks good to me, much simpler than what I was playing with.

I'm travelling at the moment so I won't get a chance to run the code and test it until next week, sorry.

components/esp32/cpu_start.c
rojer: done

components/esp32/cpu_start.c
rojer: gee, i should really read comments better, shouldn't i? sorry.
ok, a bit more tap-dancing is required, but it's done now.

components/esp32/cpu_start.c
rojer: i tried that t first, but for some reason if i change these to malloc/free i get a crash.

rojer:
ok, all comments addressed.

projectgus:
Thanks @rojer.

I've had a closer look at this now, and I also remembered the full details of the use case I had. This PR doesn't quite meet the use case I had in mind. I think this use case is a bit different to the use case you have, can you please explain your use case? I'm guessing it's heap allocation during static constructors, or something like that?

The use case that I was looking at is removing the ~192KB limit on statically allocated DRAM (.bss+.data). At the moment CPU0 (call_start_cpu0) starts with its stack around 0x3FFE3C40 and CPU1 (call_start_cpu1) is around 0x3FFE7D90. 

Relocated the stack at the end of call_start_cpuX doesn't quite fix this, because there's already stack used by call_start_cpuX and this can overwrite data/bass. Relocating the stack at the beginning of call_start_cpu0 isn't quite enough, because the software bootloader uses stack as well.

So the solution I had in mind was something like:

* Bootloader should relocate its stack before loading the .data section (or should put its stack in the .bss section).
* call_stack_cpu1 should relocate its stack when CPU1 is first started, before any stack frames are written at the default stack address.

Moving the switch_stack assembler code you've written into a place that it can be linked into both the bootloader and the user app looks like it will make it fairly simple to achieve this goal.

I'm not necessarily asking you to do this work. If this PR currently solves the use cases that you have then I'm happy to merge this now and I'll get back to the 192KB static DRAM limit later on. But I figured I'd explain what I'm thinking, as well.

rojer:
well, the case was that with a bit more static data scheduler wouldn't start because it couldn't allocate stack for tasks. but certainly nowhere near 192K of it.
now it actually turns out that this was mostly caused by the bug i had with c++ exception init, which ate 20K of heap, so strictly speaking, this change is no longer needed.

so, if you think it's useful in its current form, let's get it in and i'll let you take it from here to do the bootloader part.

projectgus:
Hi @rojer ,

Just wanted to let you know I've been looking at removing the 192KB static limit this week, using this PR as a basis. Still working on ith, though.

rojer:
ack. i'll close this then. there's no real urgency, with c++ exception init fixed i don't actually need this, at the moment.

PR #1248, opened 2017-11-09
Add esp_intr_set_in_iram

rojer:
A function to change "in IRAM" status for an existing interrupt handle, without going through free/allocate of the interrupt.
mOS HW timer implementation requires this to make HW timers safe to re-set from an IRAM ISR.

### DIFF
--- a/components/esp32/intr_alloc.c
+++ b/components/esp32/intr_alloc.c
@@ -685,6 +685,25 @@ esp_err_t esp_intr_alloc(int source, int flags, intr_handler_t handler, void *ar
     return esp_intr_alloc_intrstatus(source, flags, 0, 0, handler, arg, ret_handle);
 }
 
+esp_err_t IRAM_ATTR esp_intr_set_in_iram(intr_handle_t handle, bool is_in_iram)
+{
+    if (!handle) return ESP_ERR_INVALID_ARG;
+    vector_desc_t *vd = handle->vector_desc;
+    if (vd->flags & VECDESC_FL_SHARED) {
+      return ESP_ERR_INVALID_ARG;
+    }
+    portENTER_CRITICAL(&spinlock);
+    uint32_t mask = (1 << vd->intno);
+    if (is_in_iram) {
+        vd->flags |= VECDESC_FL_INIRAM;
+        non_iram_int_mask[vd->cpu] &= ~mask;
+    } else {
+        vd->flags &= ~VECDESC_FL_INIRAM;
+        non_iram_int_mask[vd->cpu] |= mask;
+    }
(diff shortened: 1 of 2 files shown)


rojer:
@Spritetm this is a bit esoteric, but it's the small bit i need to make hwttimers in mOS re-schedulable from the ISR without putting entire IDF's `timer_*` API in IRAM.

rapzak:
Very nice 👍

Spritetm:
Rojer: Can you explain a bit more on how that would work?

rojer:
@Spritetm basically, as far as IDF is concerned the handler remains the same (a function in mos), but the user handler we dispatch to may change, and its IRAM/non-IRAM preference too. by tweaking just this bit, i'll be able to avoid freeing and reallocating the interrupt, which involves a lot of manipulations and uses functions that are not currently in IRAM.

rapzak:
I like this very much :)

rapzak:
Could we please proceed with this topic :)

projectgus:
@rojer Internal review has identified that "this code will set the entire shared interrupt to in-iram given only one handler that shares that interrupt."

Can you add a check for shared interrupts, please?

rojer:
@projectgus done.

i think it might be possible to support mixed iram/non-iram shared ints with some effort and modest additional runtime cost at the on/off (esp_intr_noniram_{en,dis}able) time, but this is not currently an issue for me, so i'm just going to disallow this function for shared ints for the time being.

components/esp32/intr_alloc.c
```diff
@@ -638,21 +638,17 @@ esp_err_t esp_intr_alloc_intrstatus(int source, int flags, uint32_t intrstatusre
...
-    }
-    if (source>=0) {
-        intr_matrix_set(cpu, source, intr);
-    }
 
     //Fill return handle data.
     ret->vector_desc=vd;
     ret->shared_vector_desc=vd->shared_vec_info;
 
+    esp_intr_set_in_iram(ret, (flags & ESP_INTR_FLAG_IRAM) != 0);
```
projectgus: I could be wrong, but it looks like this refactor breaks shared interrupts - as esp_intr_set_in_iram will (silently) return ESP_ERR_INVALID_ARG and do nothing.

rojer: yes, sorry!
ok, removed changes to `esp_intr_alloc_intrstatus`.

rojer:
@projectgus apologies, you're correct. fixed.

projectgus:
Sorry, this PR has dragged a bit but it looks like it will probably be merged soon.

rojer:
no problem, thanks!

PR #1272, opened 2017-11-14
Fix pppapi_set_default()

rojer:
Most of the calls were fixed in 47c722d6742265c2c0260eca7ef17e71000f39ea, except this one.


rojer:
oh, fixed already. my bad.

PR #1279, opened 2017-11-16
Fixed broken performance example (tcp_perf, udp_perf)

petermccloud:
Both example needed nvs_flash_init to run successfully. Added function and required header file to both examples


projectgus:
Thanks @petermccloud . For a while the WiFi libraries were calling this internally (which wasn't desirable as you can't change the storage mode), and we obviously came to rely on this without noticing it.

Have cherry-picked this into our review & merge queue.

igrr:
Thanks for the PR, cherry-picked as 3b84c1e.

PR #1286, opened 2017-11-18
Function for obtain current log handler before redirect

phonec:
Please add simple function esp_log_get_vprintf() to get context of the current log output function in order to return the output to the current handler, because variable s_log_print_function is static.

### DIFF
--- a/components/log/include/esp_log.h
+++ b/components/log/include/esp_log.h
@@ -55,22 +55,14 @@ void esp_log_level_set(const char* tag, esp_log_level_t level);
  * @brief Set function used to output log entries
  *
  * By default, log output goes to UART0. This function can be used to redirect log
- * output to some other destination, such as file or network.
+ * output to some other destination, such as file or network. Returns the original
+ * log handler, which may be necessary to return output to the previous destination.
  *
- * @param func Function used for output. Must have same signature as vprintf.
+ * @param func new Function used for output. Must have same signature as vprintf.
+ *
+ * @return func old Function used for output.
  */
-void esp_log_set_vprintf(vprintf_like_t func);
-
-/**
- * @brief Get function used to current output log handler
- *
- * This function can be used to obtain current log handler before redirect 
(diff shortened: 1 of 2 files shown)


components/log/include/esp_log.h
```diff
@@ -61,6 +61,17 @@ void esp_log_level_set(const char* tag, esp_log_level_t level);
  */
 void esp_log_set_vprintf(vprintf_like_t func);
 
+/**
+ * @brief Get function used to current output log handler
+ *
+ * This function can be used to obtain current log handler before redirect 
+ * log output to other destination, with later time return output to the 
+ * previous destination.
```
projectgus: Rather than this sentence, what do you think about something like:

`Returns a function previously set via esp_log_set_vprintf(), or the default log output handler.`
?

projectgus:
Changes look good, thanks for sending this!

igrr:
May I suggest modifying `esp_log_set_vprintf` to return the old handler, instead of introducing a getter? Using the getter and setter can cause non-thread-safe behaviour, in the hypothetical case of two components installing hooks around the same time. On the other hand, `esp_log_set_vprintf` can be made thread-safe (take lock, save old handler into a variable, set new handler, release the lock, return old handler from a variable).

components/log/include/esp_log.h
phonec: Yes, this is the best solution to immediately return a pointer to the current function when set a new in esp_log_set_vprintf ()

phonec:
I think it's better to immediately return a pointer to the current function when installing the new in esp_log_set_vprintf() and make it thread-safe as you wrote. Does not need increase the number of functions. Please modify esp_log_set_vprintf to return the old handler.

components/log/include/esp_log.h
phonec: Returns a function previously set

projectgus:
Cherry-picked as 08ba79c, thanks for sending this!

PR #1298, opened 2017-11-21
freertos: enable use of pxTaskGetStackStart

chanibal:
Having `pxTaskGetStackStart` allows for measuring of free stack space - not only the high water mark (the lowest amount ever), but also for current stack use. On a platform where you can so easily run into low memory, having a more precise way to measure how much of it is left is a must.

As a use case, here is a screenshot of my tool for measuring stack using `pxTaskGetStackStart` for stack start and `uxTaskGetSnapshotAll` for stack end and current position.
image


projectgus:
Cherry-picked as 969911b, thank you!

PR #1310, opened 2017-11-23
Bugfix I_DELAY macro

robotrovsky:
When compiling 

> const ulp_insn_t program[] = {
> I_DELAY(1)
> };

with the xtensa-esp32-elf-g++ compiler i always got the error:

> sorry, unimplemented: non-trivial designated initializers not supported
> 
>        };

This was due to the different order in the macro and the struct. The struct has another order of the fields (opcode, unused, cycles) vs (cycles, unused, opcode):
>    struct {
>        uint32_t cycles : 16;       /*!< Number of cycles to sleep */
>        uint32_t unused : 12;       /*!< Unused */
>        uint32_t opcode : 4;        /*!< Opcode (OPCODE_DELAY) */
>    } delay;                        /*!< Format of DELAY instruction */

After updating the order in the macro it is possible to compile with the g++ compiler.

### DIFF
--- a/components/ulp/include/esp32/ulp.h
+++ b/components/ulp/include/esp32/ulp.h
@@ -266,9 +266,9 @@ _Static_assert(sizeof(ulp_insn_t) == 4, "ULP coprocessor instruction size should
  * Delay (nop) for a given number of cycles
  */
 #define I_DELAY(cycles_) { .delay = {\
-    .opcode = OPCODE_DELAY, \
+    .cycles = cycles_, \
     .unused = 0, \
-    .cycles = cycles_ } }
+    .opcode = OPCODE_DELAY } }
 
 /**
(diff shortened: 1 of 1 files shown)


projectgus:
Thanks @robotrovsky . I've cherry-picked this into our review & merge queue.

robotrovsky:
Thanks to you @projectgus and the ESP32 team, you made an incredible platform!! Anyways i think this is more a compiler issue than a IDF issue.

projectgus:
Yes, C++ is backwards-compatible with C except for a few additional rules like this one for PODs.

We try to catch these at time of writing, but because most of our code is in C some can slip through.

projectgus:
Cherry-picked as 6a51a13. Thanks!

PR #1318, opened 2017-11-26
Add *.cc files to list of file extensions compiled by default

paulreimer:
`*.cc` is also a common file extension for C++, I think it should be supported automatically (alongside `*.cpp`).


projectgus:
Thanks @paulreimer , this looks like a useful addition. I've squashed the two commits and cherry-picked them for internal review & merging.

projectgus:
Cherry-picked as 7885521. Thanks!

PR #1320, opened 2017-11-26
BLE SM Security Flags not set in all situations

BTEngineer:
I found a situation where the proper BLE security flags were not set in all pairing situations.   In the scenario I found the ESP was acting as Slave to a remote Win 10 master.  In this scenario during SM pairing only the slave exchanges the LTK (master doesn't send an LTK).   Because of this the BLE security flags for the link were not properly set to indicate that a link key was known (and if MITM was used if applicable).   This caused a ATT write request from the Master to be improperly rejected to an attribute that requires encrypted writes.


TianHao-Espressif:
@Yulong-espressif , please help to review it

projectgus:
Hi @BTEngineer ,

Sorry for the delay in reviewing this. It's now been cherry-picked for merging.

Angus

projectgus:
This has been cherry-picked into master and also backported to the v3.0/release branch. Thanks again for sending this fix and your patience.

PR #1352, opened 2017-12-05
Make spi_master structures available to user application (IDFGH-2504)

loboris:
This simple PR enables some of spi_master structures to be visible from the user application.

This makes it possible to use the spi_master driver to initialize the bus, add devices, and then use the returned _spi_device_handle_t_ variable for some lower level functions.

This way it is possible to create some spi functions which don't use the transactions and/or DMA.
For example, setting large amount  individual pixels on spi display using the transactions is **very** slow. Using direct transfer which is possible with this changes, the transfer can be up to 10x faster.

### DIFF
--- a/components/driver/include/driver/spi_master.h
+++ b/components/driver/include/driver/spi_master.h
@@ -105,42 +105,6 @@ typedef struct {
 
 typedef struct spi_device_t* spi_device_handle_t;  ///< Handle for a device on a SPI bus
 
-typedef struct spi_device_t spi_device_t;
-
-#define NO_CS 3     //Number of CS pins per SPI host
-
-
-/// struct to hold private transaction data (like tx and rx buffer for DMA).
-typedef struct {        
-    spi_transaction_t   *trans; 
-    uint32_t *buffer_to_send;   //equals to tx_data, if SPI_TRANS_USE_RXDATA is applied; otherwise if original buffer wasn't in DMA-capable memory, this gets the address of a temporary buffer that is;
-                                //otherwise sets to the original buffer or NULL if no buffer is assigned.
-    uint32_t *buffer_to_rcv;    // similar to buffer_to_send
-} spi_trans_priv;
-
-typedef struct {
-    spi_device_t *device[NO_CS];
-    intr_handle_t intr;
-    spi_dev_t *hw;
-    spi_trans_priv cur_trans_buf;
-    int cur_cs;
-    lldesc_t *dmadesc_tx;
-    lldesc_t *dmadesc_rx;
-    bool no_gpio_matrix;
-    int dma_chan;
-    int max_transfer_sz;
-#ifdef CONFIG_PM_ENABLE
-    esp_pm_lock_handle_t pm_lock;
-#endif
-} spi_host_t;
-
-struct spi_device_t {
-    QueueHandle_t trans_queue;
-    QueueHandle_t ret_queue;
-    spi_device_interface_config_t cfg;
-    spi_host_t *host;
-};
-
 /**
  * @brief Initialize a SPI bus
  *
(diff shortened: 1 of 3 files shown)


projectgus:
Hi @loboris ,

This looks like a worthwhile thing to expose, although to avoid confusion about the API (especially in the generated docs) maybe we need a new spi_master_internal.h or similar header?

@ginkgm @Spritetm PTAL as well.


Angus

loboris:
@projectgus
You are right, _spi_master.h_ dous not to be changed at all.
I've added the new header _spi_master_internal.h_ so that if one needs the access to the internal structures/function it is only necessary to add the new header.

ginkgm:
For the case send seperate pixels to the LCD, I would say in many systems, a buffer between the app and the I/O operations is usually needed to improve the communicate effciency. I strongly suggest you to think about this.

To force enable or disable DMA in a specific transaction is maybe a good idea. Now there's somewhere our DMA doesn't work properly, while the non-DMA mode doesn't support over 32-byte transactions. Do you have clear examples where we will sometimes use DMA mode while sometimes use non-DMA mode?

I do not approve of expose much things inside the API, this is very like to change members from protected to public in OO languages : we do not guarantee it will works well.

loboris:
@ginkgm
Of course, I'm aware that the best approach for working with spi LCD is to use the frame fuffer and transfer it to the display in one transaction.
It is ideal for small displays or simple applications, but, in a large application, you can easily be left with less than 100K of free RAM. For example, I'm using 320x480 display which works with 24-bit color and would require 460800 bytes for framebuffer. I am buffering some operations (like drawing the font characters, horizontal or vertical lines, decoded image lines etc) but, if I need to draw the circle with 100 pixels radius, rotated font character etc, the full size frame buffer must be used or individual pixels must be drawn.

The non-DMA transfer function is not included in this PR, it can be easily added in user app if the related structures are exposed.
One example of using non-DMA transfer would be transfering large frame buffer which can be placed in psRAM, or transfering raw images directly from flash. Not ideal, as you are blocked during the transfer, but usable. In many cases you will wait for the operation to finish anyway, and well implemented non-DMA transfer function is only couple of % slower than DMA one.

loboris:
Only 3 devices can be added to spi host even if if external CS is used (`spics_io_num = -1`).
This simple change allows more than 3 devices to be added (defined by `#define NO_CS`) if thay are using externally handled CS.

Spritetm:
So... essentially, the higher goal here is a way to transfer bytes without the irq latency by idle-waiting for the transfer to complete, right? Would you also be served if the esp-idf driver was expanded to also provide this, or is there more you need?

Spritetm:
Also, while I appreciate your enthousiasm, it's probably a better idea to have a separate MR for the software-CS thing as that essentially is a different issue,

loboris:
The main goal is to give the user the possibility to use the spi peripheral in more flexible way.
While queued transactions are working very well for many (most) uses, they can be a limiting factor for some spi usage (driving the spi display without the frame buffer for example).
This goal can be accomplished simply ba exposing the spi_master structures to the user (as in this PR) or by expanding the spi_master api with more functions for spi transfers, not queue transactions/interrupt  oriented.

The CS issue is acctually a bug in spi_master driver. There is no reason to limit the number of devices which can be attached if the external CS handling is used.
As only HSPI_HOST is available to the user if psRAM is used at 80 MHz, it can be a real problem.

Spritetm:
Hmm. The main issue I'm having with merging this is that it makes internals of the SPI driver external. This essentially means that it becomes part of the API, and that we're bound to rules wrt keeping it the same within point releases (3.0->3.1, for example) and in general, it makes us much more hesitant to change things in this struct because we don't want to randomly break peoples code.

(If any, I do agree that the clock calculation/setting code could be moved to the spi_common file and given a public API. I'll do that regardless of what happens to the rest of the changes you propose here.)

Spritetm:
As an alternative, would you be okay with adding support for your non-DMA, interrupt-less transfers to the esp-idf spi_master drivers? We could use a per-transaction flag (transaction.flags|=SPI_TRANS_POLLING or something) to invoke your routine instead of the normal, DMA-based one. At this moment, given the code quality is okay etc, I see no reason why we can't integrate a mode like that.

For the patch to add multiple software CS lines: please do that as a separate PR, as they're two separate issues.

ginkgm:
@loboris 

Since we already have https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/spi_master.html#polling-transactions, will close this PR.

Generally speaking, we don't want to break the encapsulation of the drivers.

PR #1371, opened 2017-12-11
Add ported httpd component

mjahamel:
This is the httpd server from the ESP8266 world, adapted to the ESP32. I have added a couple of minor features to support redirects and server shutdown, with the aim of making a captive portal configuration server.


projectgus:
Hi @mjahamel ,

Thanks for sending this PR. Sorry it took so long for anyone to get back to you.

This looks like a very useful component! For large pieces of additional functionality, like a web server, we prefer not to merge them into the core ESP-IDF.

As an alternative suggestion, would you be interested in setting up your own repo to host this component? Then people can git clone (or copy) it into the "components" folder of their own projects (you can have a second "components" folder, in your project, that is separate from the ESP-IDF components.)

This way you can continue to provide this web server, and support it, without needing to send PRs to ESP-IDF whenever something needs to change.

We are planning a "package manager" of some kind to make finding & installing components of this kind easier, although that's still in the works. In the meantime, if you publish this component up somewhere then posting to https://esp32.com forum would be great - I think lots of people would find it useful.

Cheers,

Angus

projectgus:
I'm going to close this PR, but @mjahamel please let us know what your plans are with this component.

mjahamel:
I hope to put it up on a public repo when I have some time, along with the rest of the captive portal configuration work. I'll post here when that happens.

projectgus:
Sounds great, please do!

agmuntianu:
Hi,
I have found the root cause of my crash TW#18509, it was in the httpd module I used from you,
In function `http_recv(void *arg, struct tcp_pcb *pcb, struct pbuf *p, err_t err)`, the 'hs' retries counter is being set to 0 after a possible freeing of the 'http_state' structure, this happens on refresh , In my issue ticket I could not reproduce it on ethernet due to lower latency with it.
So basically `hs->retries = 0;` needs to be executed before:
```
        if (err == ERR_CLSD) {
            LWIP_DEBUGF(HTTPD_DEBUG, ("received close connection %p %p\n",pcb,hs));
            http_close_conn(pcb, hs);
        }
```
BR. Andrei.

PR #1376, opened 2017-12-11
Bugfix: Use bit_num instead of duty_resolution

Deadolus:
ledc_timer_config_t no longer has a member duty_resolution.
Instead it is now called bit_num.

Signed-off-by: Simon Egli <deadolus@gmail.com>

### DIFF
--- a/examples/peripherals/ledc/main/ledc_example_main.c
+++ b/examples/peripherals/ledc/main/ledc_example_main.c
@@ -60,7 +60,7 @@ void app_main()
      * that will be used by LED Controller
      */
     ledc_timer_config_t ledc_timer = {
-        .duty_resolution = LEDC_TIMER_13_BIT, // resolution of PWM duty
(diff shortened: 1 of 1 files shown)


krzychb:
Hi @Deadolus,

The name `bit_num` has been changed in d488bb7bacca80e6089bdf070bb1e10324d31dbb to `duty_resolution` to provide more descriptive name. of this member.  This PR actually reverts `duty_resolution` back to `bit_num` in one of examples. 

What is the purpose of that?

BTW:

For backward compatibility both names are supported:

https://github.com/espressif/esp-idf/blob/a7a861962daacec17bdb0f34fd59f16651b50b0f/components/driver/include/driver/ledc.h#L101-L109

Deadolus:
The example did not compile for me and I found that the member was renamed. 
But I thought the other way around. 
I checked this again with a clean checkout of esp-idf master and I can no longer reproduce the problem. 
So it was a problem on my side - sorry about that. 
Closing this.

PR #1378, opened 2017-12-12
Ethernet: Let applications specify DMA buffers (IDFGH-2876)

aykevl:
By default, ethernet requires 2x 16kB of statically allocated RAM for the DMA buffers, whether ethernet is actually used by the application or not.

This pull request enables applications to specify the buffer manually. They can either specify it as a statically allocated buffer (just like it was before) or allocate it dynamically when needed.

See: https://github.com/micropython/micropython-esp32/issues/222#issuecomment-350582604

### DIFF
--- a/components/ethernet/include/esp_eth.h
+++ b/components/ethernet/include/esp_eth.h
@@ -45,6 +45,11 @@ typedef enum {
     ETH_MODE_FULLDUPLEX,
 } eth_duplex_mode_t;
 
+#define ETH_DMA_RX_BUF_NUM CONFIG_DMA_RX_BUF_NUM
+#define ETH_DMA_TX_BUF_NUM CONFIG_DMA_TX_BUF_NUM
+#define ETH_DMA_RX_BUF_SIZE 1600
+#define ETH_DMA_TX_BUF_SIZE 1600
+
 typedef enum {
     PHY0 = 0,
     PHY1,
@@ -108,6 +113,8 @@ typedef struct {
     bool flow_ctrl_enable;                      /*!< flag of flow ctrl enable */
     eth_phy_get_partner_pause_enable_func  phy_get_partner_pause_enable; /*!< get partner pause enable */
     eth_phy_power_enable_func  phy_power_enable;  /*!< enable or disable phy power */
+    uint8_t *dma_rx_buf;                        /*!< DMA-capable buffer of size ETH_DMA_RX_BUF_SIZE * ETH_DMA_RX_BUF_NUM, may be NULL */
+    uint8_t *dma_tx_buf;                        /*!< DMA-capable buffer of size ETH_DMA_TX_BUF_SIZE * ETH_DMA_TX_BUF_NUM, may be NULL */
 
 } eth_config_t;
 
--- a/components/ethernet/emac_common.h
+++ b/components/ethernet/emac_common.h
@@ -103,12 +103,7 @@ struct emac_close_cmd {
     int8_t err;
 };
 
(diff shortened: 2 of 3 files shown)


components/ethernet/include/esp_eth.h
```diff
@@ -118,12 +123,14 @@ typedef struct {
...
  * @param[in] config  mac init data.
+ * @param[in] dma_rx_buf  DMA-capable buffer of size DMA_RX_BUF_SIZE * DMA_RX_BUF_NUM
+ * @param[in] dma_tx_buf  DMA-capable buffer of size DMA_TX_BUF_SIZE * DMA_TX_BUF_NUM
  *
  * @return
  *      - ESP_OK
  *      - ESP_FAIL
  */
-esp_err_t esp_eth_init(eth_config_t *config);
+esp_err_t esp_eth_init(eth_config_t *config, uint8_t *dma_rx_buf, uint8_t *dma_tx_buf);
```
igrr: I suggest that we keep this function backwards compatible. To do so, could you please move these new options into `eth_config_t` structure?
In `esp_eth_init`, check if `dma_rx_buf` is NULL — in this case, allocate `CONFIG_DMA_RX_BUF_NUM` buffers. Same with TX buffers.

This way, applications which have previously zero-initialized `eth_config_t` structure will still work without any extra changes.

components/ethernet/include/esp_eth.h
```diff
@@ -45,6 +45,11 @@ typedef enum {
     ETH_MODE_FULLDUPLEX,
 } eth_duplex_mode_t;
 
+#define DMA_RX_BUF_NUM CONFIG_DMA_RX_BUF_NUM
+#define DMA_TX_BUF_NUM CONFIG_DMA_TX_BUF_NUM
+#define DMA_RX_BUF_SIZE 1600
+#define DMA_TX_BUF_SIZE 1600
```
igrr: Since these are now in global namespace, please prefix these defines with `ETH_`.

aykevl: Fixed.
I've also updated Kconfig - I think the config vars `DMA_RX_BUF_NUM` and `DMA_TX_BUF_NUM` also need to be prefixed.

components/ethernet/Kconfig
```diff
@@ -10,7 +10,7 @@ config DMA_RX_BUF_NUM
         Larger number of buffers increases throughput.
         If enable flow ctrl, the num must be above 9 .
 
-config DMA_TX_BUF_NUM
+config ETH_DMA_TX_BUF_NUM
```
igrr: While I agree that prefixing the configuration variables is a good idea, renaming existing ones will result in loss of configuration for project which had these vars set to non-default values.

aykevl: Ok, I'll revert this change.

aykevl:
I think I've applied all requested changes.

components/ethernet/include/esp_eth.h
```diff
@@ -45,6 +45,11 @@ typedef enum {
     ETH_MODE_FULLDUPLEX,
 } eth_duplex_mode_t;
 
+#define ETH_DMA_RX_BUF_NUM CONFIG_DMA_RX_BUF_NUM
```
projectgus: This header file should include sdkconfig.h so that the CONFIG_xxx values are guaranteed to be defined.

projectgus:
@aykevl Sorry for the slow follow-up. Are you still interested in working on this? Have you had a chance to take a look at the remaining review comment I left?

PR #1390, opened 2017-12-15
I2S-driver: apll_predefine was ignored

michaelwgnr:
In i2s.c, there are predefined values for audio PLL configuration in the `apll_predefine` array. In `i2s_apll_calculate`, the sampling rate supplied was compared to the bits_per_sample value in the array. This is a fix of this comparison.


michaelwgnr:
Never mind, this seems to break the i2c driver.

PR #1393, opened 2017-12-15
I2S-driver: Fix for ignored apll_predefine values

michaelwgnr:
In i2s.c, there are predefined values for audio PLL configuration in the `apll_predefine` array. In `i2s_apll_calculate`, the sampling rate supplied was compared to the bits_per_sample value in the array. This is a fix of this comparison. Furthermore, if the check had succeeded, there would have been an off-by-one error in the assignments to the output variables. This patch fixes this as well.

### DIFF
--- a/components/driver/i2s.c
+++ b/components/driver/i2s.c
@@ -249,10 +249,10 @@ static esp_err_t i2s_apll_calculate(int rate, int bits_per_sample, int *sdm0, in
     i = 0;
     while (apll_predefine[i][0]) {
         if (apll_predefine[i][0] == bits_per_sample && apll_predefine[i][1] == rate) {
-            *sdm0 = apll_predefine[i][1];
-            *sdm1 = apll_predefine[i][2];
-            *sdm2 = apll_predefine[i][3];
-            *odir = apll_predefine[i][4];
+            *sdm0 = apll_predefine[i][2];
+            *sdm1 = apll_predefine[i][3];
+            *sdm2 = apll_predefine[i][4];
+            *odir = apll_predefine[i][5];
             return ESP_OK;
         }
         i++;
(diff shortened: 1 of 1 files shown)


tuanpmt:
In the next update, I think that these predefined values will not be needed any more, because the PLL calculation time is pretty fast and accurate.

michaelwgnr:
> In the next update, I think that these predefined values will not be needed any more, because the PLL calculation time is pretty fast and accurate.

Fair enough... Will there also be fixes to PDM mode? Obviously, the clocks are calculated wrong there. (see https://esp32.com/viewtopic.php?f=13&t=3956)

tuanpmt:
To speed up the APLL calculation, the parameters are kept fixed:
```
I2S[i2s_num]->clkm_conf.clkm_div_num = 1;
I2S[i2s_num]->clkm_conf.clkm_div_b = 0;
I2S[i2s_num]->clkm_conf.clkm_div_a = 1;
I2S[i2s_num]->sample_rate_conf.tx_bck_div_num = 8;
I2S[i2s_num]->sample_rate_conf.rx_bck_div_num = 8;
```
Meanwhile, when you use PDM mode, the parameters (clkmDecimals, clkmInteger, bck)  are dependent and is calculated in a way that does not apply to APLL

```
I2S[i2s_num]->clkm_conf.clka_en = 0;
I2S[i2s_num]->clkm_conf.clkm_div_a = 63;
I2S[i2s_num]->clkm_conf.clkm_div_b = clkmDecimals;
I2S[i2s_num]->clkm_conf.clkm_div_num = clkmInteger;
I2S[i2s_num]->sample_rate_conf.tx_bck_div_num = bck;
I2S[i2s_num]->sample_rate_conf.rx_bck_div_num = bck;
```
Apparently the i2s driver is having problems there.
I will find the appropriate way to calculate that parameters

projectgus:
Thanks for your patience while we worked through this. As of commit 63ddae508 in the IDF master branch, the apll_predefine values are no longer used.

Issue #1345, opened 2017-12-03
[TW#16842] RMT device affected by log output / std::condition_variable logging


PerMalmberg:
Hi,

I've got a task with the following responsibilities: 
* Communicate with three I2C devices over two different channels.
* Control five WS281B RGB LEDs using RMT (each command waits for completation).

This is working splendidly - _but only as long as logging is set to "Info" or lower_. At debug and especially at verbose level the amount of logging apparently become to much to handle which affects the performance of the RMT, resulting in the timing being wrong and in turn results in corrupt data being sent to the five WS2812B.

As I understand it, the RMT should not be affected by the CPU load and thus be able to do signaling with very tight timing demands, right?

Assuming my understanding is correct, I think there is an issue with how the RMT interacts with the log system, based on the fact that this issue appeared after updating to [edb2400](https://github.com/espressif/esp-idf/commit/edb2400742d2fded2a2a19f349cf8842c1f77f36) in which _std::condition_variable_ continously spits out a huge amount of log messages like this:

```
V (427591) esp_pthread: pthread_cond_timedwait 0x3ffc09dc 0x3ffc09d8 0x3ffc07b8
V (427597) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427612) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427619) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427627) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427634) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427641) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427649) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427656) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
V (427664) esp_pthread: pthread_cond_timedwait 0x3ffb7fbc 0x3ffb7fb8 0x3ffc93a8
[...]

PR #1443, opened 2017-12-28
Fix regression in i2c_master_read() which rejected all data lentghs.

PerMalmberg:

PR #1444, opened 2017-12-28
Removed logging from pthread conditional variable

PerMalmberg:
Removed logging from pthread conditional variable which completely drowned out any other logging and also had negative effects on the RMT due to sharing the same log system. Related to #1345 / TW#16842"
[bug/phtread_condititon_variable e02cfbe] Removed logging from pthread conditional variable which completely drowned out any other logging and also had negative effects on the RMT due to sharing the same log system. Related to #1345 / TW#16842


projectgus:
Closing as this commit was (maybe accidentally) added to #1443 and we can merge it there this time.

PR #1448, opened 2017-12-29
components: i2c: fix erroneous sanity check

tbnobody:
The predicate for the I2C_CHECK macro describes what you expect to be true.
i2c_master_read used this macro in the wrong way. Fix that.

Signed-off-by: Thomas Basler <github.thomas@familie-basler.net>
Signed-off-by: Ralf Ramsauer <ralf@ramses-pyramidenbau.de>


PerMalmberg:
@tbnobody See #1443

projectgus:
Thanks for submitting this, @tbnobody . This regression has been fixed in #1443.

PR #1450, opened 2017-12-29
Removed logging in RMT driver which skews the timing. Issue #1345

PerMalmberg:
As discussed in #1345, RMT is affected by the logging calls it performs. This PR removes the logging that happens during normal operation.


projectgus:
Thanks @PerMalmberg , seems reasonable. Cherry-picked for merging.

projectgus:
Cherry-picked in f7f5241. Thanks!

PR #1451, opened 2017-12-29
GPIO2 note updated

Pomezny:
Updated note on GPIO2 with informations about possible pull-down resistor and where to find in on DOIT board.


projectgus:
Hi @Pomezny ,

Thanks for sending this PR.

Warning users to check for GPIO 2 is a really good idea, thanks for adding this.

Unfortunately we can't start add notes/warnings to the docs for specific third party development board designs. These are hard to keep track of, and documentating limitations is the board manufacturer's responsibility. Maybe DOIT has a wiki or something that this info could be put on? Or you could post on esp32.com?

The documentation could say something like "Check the schematic of your development board for anything connected to GPIO 2", though.

Pomezny:
@projectgus I updated file as you advised.

projectgus:
Thanks @Pomezny, looks good!

projectgus:
Cherry-picked in e7693b8. Thanks!

PR #1455, opened 2017-12-30
Position file stream at end of the file when fopen(file_name, 'a') is executed

loboris:
The file stream position should be set to the end of the file then `fopen("filename", 'a')` is executed.
>        a      Open for appending (writing at end of file).  The file is
>               created if it does not exist.  The stream is positioned at the
>               end of the file.
> 
[fopen](http://man7.org/linux/man-pages/man3/fopen.3.html)

### DIFF
--- a/components/fatfs/src/vfs_fat.c
+++ b/components/fatfs/src/vfs_fat.c
@@ -307,6 +307,16 @@ static int vfs_fat_open(void* ctx, const char * path, int flags, int mode)
         goto out;
     }
 out:
+    if ((fd >= 0) && (flags & O_APPEND)) {
+        res = f_lseek(&fat_ctx->files[fd], f_size(&fat_ctx->files[fd]));
+        if (res != FR_OK) {
+            ESP_LOGD(TAG, "%s: fresult=%d", __func__, res);
+            errno = fresult_to_errno(res);
+            res = f_close(&fat_ctx->files[fd]);
+            file_cleanup(fat_ctx, fd);
+            fd = -1;
+        }
+    }
     _lock_release(&fat_ctx->lock);
     return fd;
 }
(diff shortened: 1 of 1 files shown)


components/fatfs/src/vfs_fat.c
```diff
@@ -307,6 +307,16 @@ static int vfs_fat_open(void* ctx, const char * path, int flags, int mode)
         goto out;
     }
 out:
+    if ((fd >= 0) && (flags & O_APPEND)) {
```
projectgus: The `out:` block is intended for cleanup (common C pattern), so it reads better if this `if (flags & O_APPEND)` block sits above it. At that point fd will always be valid so you can remove the `>= 0` test (as the function skips to `out` if there's been an error).

loboris: Of course, done.

projectgus:
Thanks @loboris. Have squashed these commits & cherry-picked for merging.

igrr:
Hi @loboris, did you see some case where file pointer is not correct after opening the file in append mode? I have done a simple test with the current master branch and found that the file pointer is correctly positioned at the end of the file, when opening the file in append mode.
This is part of `fopen` implementation here: https://github.com/espressif/newlib-esp32/blob/master/newlib/libc/stdio/fopen.c#L162-L163.

loboris:
@igrr 
You are right about `fopen`.

I had a problem with opening the file in append mode, but the **open()** function was used, not **fopen()**.

Nevertheless, `int open(const char *pathname, int flags, mode_t mode);` should also open the file and position the file file pointer at the end of the file, but it does not.
Maybe the `vfs_fat_open()` is not the right place for append check and it should be checked elsewhere. I'm not shure which function is called by `open()`, maybe `esp_vfs_open`.

igrr:
You are right about `open` @loboris. Actually I found that IDF behaviour for O_APPEND mode is also incorrect when `lseek` is called. For files opened with `O_APPEND`, writes always go at the end of this file, and `lseek` only affects the write pointer. So what `fopen` implementation in newlib does is moving the read pointer to the end of the file by default. IDF implementation of `open` and `write` needs to be updated to work properly in `O_APPEND` mode, for all VFS implementations.
I'm now looking into how to do this properly.

PR #1458, opened 2018-01-01
added notes on where to change XTAL freq

gherlein:
new users will be confused by the actual config variable name so I added a note on where you can find that change in make menuconfig.

### DIFF
--- a/docs/get-started/index.rst
+++ b/docs/get-started/index.rst
@@ -285,7 +285,14 @@ To exit the monitor use shortcut ``Ctrl+]``.
         e���)(Xn@�y.!��(�PW+)��Hn9a؅/9�!�t5��P�~�k��e�ea�5�jA
         ~zY��Y(1�,1�� e���)(Xn@�y.!Dr�zY(�jpi�|�+z5Ymvp
 
-    or monitor fails shortly after upload, your board is likely using 26MHz crystal, while the ESP-IDF assumes default of 40MHz. Exit the monitor, go back to the :ref:`menuconfig <get-started-configure>`, change :ref:`CONFIG_ESP32_XTAL_FREQ_SEL` to 26MHz, then :ref:`build and flash <get-started-build-flash>` the application again.
+    or monitor fails shortly after upload, your board is likely using
+    26MHz crystal, while the ESP-IDF assumes default of 40MHz. Exit
+    the monitor, go back to the :ref:`menuconfig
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @gherlein ,

Thanks for taking the time to author this PR. Having a non-confusing new user experience is important for us!

The ESP-IDF docs used to have a lot of these type of instructions in them (where to find certain options). We removed a lot of them when we added the config items help page, because the content there (which is autogenerated) includes a description of the item.

ie if you look at the docs here and scroll down to the Note, you'll see a clickable link to the "menuconfig" section and a link to the ESP32_XTAL_FREQ_SEL item which has the menu hierarchy mentioned under the heading:
http://esp-idf.readthedocs.io/en/latest/get-started/index.html#monitor

Do you think this is too confusing? Is there any other way you think we could make it clearer (without having to copy the hierarchy into the configuration text)?

(I'm not saying that we won't merge this - but I'd like to understand if our attempt to clean up configuration documentation is counter-intuitive in some way).

Thanks,

Angus

projectgus:
@krzychb interested in your thoughts on this as well

krzychb:
@projectgus, I like this addition as for the "Getting Started" section targeted to the first time users.

gherlein:
I think this is a difficult documentation problem.  I'm a long-time kernel guy so seeing these kinds of variables discussed for the first time in a doc and I think "oh, I just have to go grep for that and find the section" and it's mildly annoying.  But for a new user - like some high school kids I am moving from Arduino to here - this would be a cognitive block.  This is the first time the docs switch from talking about using menuconfig to change the variable into using the actual name of the variable - a name that is not visible in the menuconfig by the way - so it's really confusing.  

Yet the right thing to do is to start using the variable names, especially to explain early how to use them to modify the sdkconfig.defaults file and keep that in their source control instead of constantly editing the sdkconfig file.

I propose we accept this PR as a stopgap and add some more discussion earlier about the config system.  I volunteer to take a stab a that since I have an active (new) user community that I hope to be newbie users.

projectgus:
Hi @gherlein ,

Thanks for your comments. I think I agree all around. Will cherry-pick this change into our internal review & merge queue.

Any discussions you wish to spearhead about the config system would be most welcome, thanks. We have some thoughts about this ourselves as well (such as the difference between app-level, host system-level and user-level config) but I don't think we've published them anywhere.

Angus

projectgus:
Cherry-picked in a834c06. Thanks!

PR #1468, opened 2018-01-05
DRV8825 Stepper Motor Library for ESP32

akbarhash:
This is a simple library for DRV8825 Stepper Motor Control using ESP32.
It cycles through 0->50->100->150->200->0 steps per second as an example.
This is the first time I am contributing like this. So any comments will be helpful.
Could even work for other based on pulse control like a4988 from pololu.
Output pin is GPIO18.

### DIFF
--- a/examples/peripherals/stepper_motor_control/Makefile
+++ b/examples/peripherals/stepper_motor_control/Makefile
@@ -1,0 +1,9 @@
+#
+# This is a project Makefile. It is assumed the directory this Makefile resides in is a
+# project subdirectory.
+#
+
+PROJECT_NAME := steppercontrol
+
+include $(IDF_PATH)/make/project.mk
+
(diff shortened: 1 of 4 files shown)


projectgus:
Hi @akbarhash ,

Thanks for publishing this example and sending it to us.

Unfortunately we have a policy of not merging drivers or support for third-party functionality like this. It's difficult to maintain it as part of ESP-IDF itself.

Perhaps you could start a new repository with this example app (& associated library code) only, by itself? Then you can post a link to it on https://esp32.com for people to see. 

You can find an example of the kind of app repository layout here:
https://github.com/espressif/esp-idf-template/

This way you can also keep maintaining & updating the code yourself, as well.

Thanks,

Angus

costaud:
Another choice is to add this driver as a component in [https://github.com/espressif/esp-iot-solution](https://github.com/espressif/esp-iot-solution)

akbarhash:
thanks for the inputs

PR #1469, opened 2018-01-05
"Wrong" ULP clock datasheet

urbanze:
In the ESP32 datasheet in the ULP section (https://i.imgur.com/SV0ZDUo.png), it is said that the frequency is 8MHz, however, in several places it is said about clock calibration or visualization (and this suggests that it is not exactly 8MHz).

`// calibrate 8M/256 clock against XTAL, get 8M/256 clock period
uint32_t rtc_8md256_period = rtc_clk_cal(RTC_CAL_8MD256, 100);
uint32_t rtc_fast_freq_hz = 1000000ULL * (1 << RTC_CLK_CAL_FRACT) * 256 / rtc_8md256_period;`

Here this variable return's ~8.833MHz, which is a bit far from 8MHz.
They could make a correction in the datasheet about this "detail" or add a note about it, if true. That can make some projects unstable because the programmer does not pay attention to this by calculating wrong delay's in "high frequences".

from "rtc.h"
"/* With the default value of CK8M_DFREQ, 8M clock frequency is 8.5 MHz +/- 7% */
#define RTC_FAST_CLK_FREQ_APPROX 8500000"


negativekelvin:
Don't do a pull request against release branch into master branch.

igrr:
Thanks for pointing this out, I will forward the issue to the documentation team. For future questions related to documentation, please use https://esp32.com/viewtopic.php?f=5&t=905 forum topic.

PR #1474, opened 2018-01-07
Add space between -I and path for CPPFLAGS

roamingthings:
In order for the Eclipse CDT GCC Build Output Parser to be able to
pick up the include paths for C++/cpp files a space is required
between the compiler option (-I) and the actual path.

### DIFF
--- a/make/component_wrapper.mk
+++ b/make/component_wrapper.mk
@@ -244,12 +244,12 @@ $(1)/%.o: $$(COMPONENT_PATH)/$(1)/%.c $(COMMON_MAKEFILES) $(COMPONENT_MAKEFILE) 
 
 $(1)/%.o: $$(COMPONENT_PATH)/$(1)/%.cpp $(COMMON_MAKEFILES) $(COMPONENT_MAKEFILE) | $(COMPONENT_SRCDIRS)
 	$$(summary) CXX $$(patsubst $$(PWD)/%,%,$$(CURDIR))/$$@
-	$$(CXX) $$(CXXFLAGS) $$(CPPFLAGS) $$(addprefix -I ,$$(COMPONENT_INCLUDES)) $$(addprefix -I,$$(COMPONENT_EXTRA_INCLUDES)) -I$(1) -c $$< -o $$@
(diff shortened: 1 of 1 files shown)


make/component_wrapper.mk
```diff
@@ -244,12 +244,12 @@ $(1)/%.o: $$(COMPONENT_PATH)/$(1)/%.c $(COMMON_MAKEFILES) $(COMPONENT_MAKEFILE)
 
 $(1)/%.o: $$(COMPONENT_PATH)/$(1)/%.cpp $(COMMON_MAKEFILES) $(COMPONENT_MAKEFILE) | $(COMPONENT_SRCDIRS)
 	$$(summary) CXX $$(patsubst $$(PWD)/%,%,$$(CURDIR))/$$@
-	$$(CXX) $$(CXXFLAGS) $$(CPPFLAGS) $$(addprefix -I,$$(COMPONENT_INCLUDES)) $$(addprefix -I,$$(COMPONENT_EXTRA_INCLUDES)) -I$(1) -c $$< -o $$@
+	$$(CXX) $$(CXXFLAGS) $$(CPPFLAGS) $$(addprefix -I ,$$(COMPONENT_INCLUDES)) $$(addprefix -I,$$(COMPONENT_EXTRA_INCLUDES)) -I$(1) -c $$< -o $$@
```
igrr: There are another two `addprefix -I,` in this line (and the second one), should they be updated as well?

igrr:
Thanks for submitting this, @roamingthings! Do you have a reference why this is only needed for C++ files but not for C?

make/component_wrapper.mk
roamingthings: The part that is responsible for C has already been correct.

roamingthings: I've just reviewed my changes and it seems as if I've missed to places

roamingthings:
The part that is responsible for C has already been correct.

projectgus:
Thanks @roamingthings . I've squashed these two commits and cherry-picked into our internal merge queue.

projectgus:
Thanks again @roamingthings . Cherry-picked in 77adf85, and @igrr noticed the component directory also may need a  space so this was added in 3f5d6cf.

PR #1487, opened 2018-01-10
Add task tracking option for heap usage monitoring

slcasner:
This is an update of PR 1476 to incorporate feedback from @projectgus. Sorry for the duplicate commits here; I am not a git expert and must not have done the rebase correctly.

If you don't think that be6f699 is appropriate for inclusion in the system, that's OK, we can just keep it on our fork.

### DIFF
--- a/components/freertos/tasks.c
+++ b/components/freertos/tasks.c
@@ -87,6 +87,7 @@ task.h is included from an application file. */
 #include "StackMacros.h"
 #include "portmacro.h"
 #include "semphr.h"
+#include "multi_heap_poisoning.h"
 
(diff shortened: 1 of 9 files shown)


slcasner:
I have also added commit ae5a434 that I would not expect you to pull.  It would be nice to have uxTaskGetSystemState return both the total stack block size and the current amount free (or the amount used), but since the pxStackBase member is not being set at all in the current code for some reason, perhaps you would not want to do this.

slcasner:
I messed up this PR.  I have redone it cleanly from a new branch in another fork in PR 1498.

PR #1498, opened 2018-01-12
Feature/heap task tracking

slcasner:
This is an update of PR 1476 from openvehicles to incorporate feedback from @projectgus.  I tried to do this update first with PR 1487 from openvehicles, but I didn't do that correctly and cleanly so I am trying again from a nice clean branch here.

Add back a feature that was available in the old heap implementation in release/v2.1 and earlier: keep track of which task allocates each block from the heap. The task handle is conditionally added as another word in the heap poisoning header under this configuration option CONFIG_HEAP_TASK_TRACKING.  As requested in the feedback, the platform-specific parts of this feature (config macro and FreeRTOS API call) are now in multi_heap_platform.h and helper functions were added to the multi_heap and multi_heap_poisoning APIs to allow this feature to be used by additional heap debugging functions such as what I've added in cdf32aa.  However, I did not check that host-side tests still compile & pass because I don't know how to do that.

If you don't think that cdf32aa is appropriate for inclusion in the system, that's OK, we can just keep it on our openvehicles fork.  The esp_heap_debug_dump_totals() API implemented there is used by C++ application level code in our project to produce output like this:

    OVMS > module memory tiT AsyncConsole
    Free 8-bit 74492/282256, 32-bit 17836/45008, SPIRAM 0/0
    --Task--     Total DRAM D/IRAM   IRAM SPIRAM   +/- DRAM D/IRAM   IRAM SPIRAM
    no task            5276      0      0      0      +5276     +0     +0     +0
    esp_timer         49792      0    644      0     +49792     +0   +644     +0
    main              37376      0      0      0     +37376     +0     +0     +0
    ipc0              11096      0      0      0     +11096     +0     +0     +0
    Housekeeping      26044  33676      0      0     +26044 +33676     +0     +0
    tiT                 112   1412      0      0       +112  +1412     +0     +0
    eventTask            16  13156      0      0        +16 +13156     +0     +0
    ipc1                 12      0      0      0        +12     +0     +0     +0
    NetManTask            0   2272      0      0         +0  +2272     +0     +0
    wifi                  0   1772      0      0         +0  +1772     +0     +0
    AsyncConsole          0     20  26404      0         +0    +20 +26404     +0
    ============================ blocks dumped = 32
    + t=tiT s=  16 a=0x3ffcc780  3FFCC7A4 00000190 401B90EC 00000000 BAAD5678 3FFCC7B8 ABBA1234 3FFCD71C
    + t=tiT s=  16 a=0x3ffcc7a4  3FFCC810 00000000 401B90D4 00000000 BAAD5678 3FFCC7DC ABBA1234 3FFCD71C
    + t=tiT s=  16 a=0x3ffcc7c8  3FFE5FF8 3FFE607C BAAD5601 BAAD5678 3FFCC840 3FFCC800 ABBA1234 3FFCD71C
    + t=tiT s=  16 a=0x3ffcc7ec  3FFCC780 00000000 401B905C 00000000 BAAD5678 3FFCC824 ABBA1234 3FFCD71C
    + t=tiT s=  16 a=0x3ffcc810  3FFE94EC 00000000 401B9098 00000000 BAAD5678 3FFCC840 ABBA1234 3FFBE2A0
    + t=tiT s=  16 a=0x3ffcc850  3FFCC874 00000000 401B9070 00000000 BAAD5678 3FFCC888 ABBA1234 3FFCD71C
    + t=tiT s=  16 a=0x3ffcc874  00000000 00000000 401B90C0 00000000 BAAD5678 3FFCC891 3FFCDE84 3FFCC8B0
    + t=tiT s=  16 a=0x3ffe37fc  3FFCC7EC 00000064 401B90AC 00000000 BAAD5678 3FFE3844 ABBA1234 3FFE782C
    + t=tiT s=  96 a=0x3ffe3d0c  2BA41009 00000000 01000A01 00000000 00000000 00000044 0000A8C0 00012750
    + t=tiT s=  20 a=0x3ffe3e80  3FFE5FD0 3FFE5C98 FB0000E0 00000101 01300301 BAAD5678 3FFE3EA9 3FFE3FF8
    + t=tiT s=  92 a=0x3ffe3ee0  00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
    + t=tiT s= 232 a=0x3ffe5c98  3FFB8E98 5401A8C0 00000000 00000000 00000000 00000000 00FFFFFF 00000000
    + t=tiT s=  16 a=0x3ffe5e6c  3FFEAB24 3FFEABA8 00000001 BAAD5678 3FFE5F4C 3FFE5EA8 ABBA1234 3FFD7D98
    + t=tiT s=  20 a=0x3ffe5fd0  00000000 3FFE5C98 010000E0 00000200 3FFE5F01 BAAD5678 3FFE606C ABBA1234
    + t=tiT s= 112 a=0x3ffe5ff8  3FFE6050 3FFE6068 3FFE6050 3FFE6064 00000000 3FFE6010 FFFFFFFF 3FFE6010
    + t=tiT s=  88 a=0x3ffe607c  00000000 00000000 3FFE607C 00000000 00000000 3FFE6094 FFFFFFFF 3FFE6094
    + t=tiT s=  88 a=0x3ffe79ac  5401A8C0 00000000 00000000 00000000 00000000 FB0000E0 00000000 00000000
    + t=tiT s=  68 a=0x3ffe93bc  00000000 496D3233 30332820 20293835 33707300 66697732 57203A69 20694669
    + t=tiT s=  16 a=0x3ffe94ec  3FFCC850 00000000 401B9084 00000000 BAAD5678 3FFE950D 3FFEA384 BAAD5678
    + t=tiT s=  88 a=0x3ffea98c  00000000 00000000 3FFEA98C 00000000 00000000 3FFEA9A4 FFFFFFFF 3FFEA9A4
    + t=tiT s=  68 a=0x3ffeaacc  00000000 5C015C00 5C035C02 5C055C04 5C095C00 5C0B5C0A 5C0D5C0C 5C195C18
    + t=tiT s= 112 a=0x3ffeab24  3FFEAB7C 3FFEAB94 3FFEAB7C 3FFEAB90 00000000 3FFEAB3C FFFFFFFF 3FFEAB3C
    + t=tiT s=  88 a=0x3ffeaba8  00000000 00000000 3FFEABA8 00000000 00000000 3FFEABC0 FFFFFFFF 3FFEABC0
    + t=tiT s=  12 a=0x3ffeac5c  3FFEDDD4 3FFEA98C 00000001 BAAD5678 3FFEAC75 3FFEDE48 3FFEACE0 ABBA1234
    + t=tiT s= 112 a=0x3ffeddd4  3FFEDE2C 3FFEDE44 3FFEDE2C 3FFEDE40 00000000 3FFEDDEC FFFFFFFF 3FFEDDEC
    + t=tiT s=  68 a=0x3ffedf1c  00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
    ============================
    + t=AsyncConsole s=  20 a=0x3ffe3738  00000001 3FFB41E0 00000000 00000000 3FFE147C BAAD5678 3FFE37B0 ABBA1234
    + t=AsyncConsole s=12000 a=0x400952c0  FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE
    + t=AsyncConsole s=12000 a=0x400981b4  3FFCD71C 3FFCC780 80000010 3FFCD71C 3FFCC7A4 80000010 3FFCD71C 3FFCC7C8
    + t=AsyncConsole s=1080 a=0x4009b0a8  3FFE32E8 3FFE3320 00000012 00000001 00000005 00000005 00000000 14001108
    + t=AsyncConsole s=  40 a=0x4009b4f4  3FFCD71C 3FFE32E8 FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE FEFEFEFE
    + t=AsyncConsole s=1284 a=0x4009b530  00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
    ============================


projectgus:
Thanks for this. I like the multi_heap changes, very clean.

Am pondering what the best option for the esp_heap_debug stuff you have is. This is a really nice function that I'd like for us to incorporate as well, just need to figure out the best way to do it.

slcasner:
Just a poke to see how the pondering is going.

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+
+    SLIST_FOREACH(reg, &registered_heaps, next) {
+	multi_heap_handle_t heap = reg->heap;
+	if (heap == NULL) {
+	    continue;
+	}
+	uint32_t caps = get_all_caps(reg);
+	enum region_types type = DRAM;
+	if (caps & MALLOC_CAP_EXEC)
+	    type = D_IRAM;
```
projectgus: We try to always use braces around single line blocks (applies in a few other places as well). 

There's a style guide here, and a guide to using the "astyle" tool to reformat code automatically:
https://esp-idf.readthedocs.io/en/latest/contribute/style-guide.html

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+	if (!(caps & MALLOC_CAP_8BIT))
+	    type = IRAM;
+	if (caps & MALLOC_CAP_SPIRAM)
+	    type = SPIRAM;
+	multi_heap_block_handle_t b = multi_heap_get_first_block(heap);
+	multi_heap_internal_lock(heap);
+	while (b) {
+	    if (multi_heap_is_free(b)) {
+		b = multi_heap_get_next_block(heap, b);
```
projectgus: Indenting seems to switch to tabs somewhere around here (IDF code is 4 spaces), and the indenting here is missing.

(astyle tool, mentioned above, is the quickest way to fix this.)

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+		    if (btask == tasks[i])
+			break;
+		}
+		if (i == ntasks) {
+		    b = multi_heap_get_next_block(heap, b);
+		    continue;
+		}
+	    }
+	    if (remaining > 0) {
+		buffer->task = btask;
```
projectgus: Same here re:indenting

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// For each task that has allocated memory from the heap, return totals for
+// allocations of each type of heap region (8-bit DRAM, 8-bit D/IRAM, 32-bit
+// IRAM, SPIRAM) into provided array of heap_dump_totals_t structs 'totals'.
+
+// If one or more task handles is provided in array 'tasks', for each block
+// allocated by one of the tasks a heap_dump_block_t struct is written into
+// provided array 'buffer' to give the task, address, size and region type for
+// the block.
```
projectgus: This usage description is excellent, but would be better placed in the header file. Can you also please reformat it as Doxygen syntax (the other header descriptions will give some clues about how to do this).

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+
+size_t esp_heap_debug_dump_totals(heap_dump_totals_t* totals, size_t* ntotal, size_t max,
```
projectgus: This is really useful functionality. I think we can more or less merge the implementation as-is, but I'd like to work on the naming a bit so that it's clearer how to use it (and also try to bring it into line with the existing heap_caps_xxx API).

For a name, how about heap_caps_get_per_task_info()? (With renaming esp_heap_debug.c -> heap_task_info.c, and merging the header into the main esp_heap_caps.h header?)

For heap_dump_totals_t, maybe rename to heap_task_total_t?

I'm not sure how I feel about using the region_info_t enum to split up capabilities, as this isn't how we've treated heap capabilities everywhere else. Instead, would attaching a "uint32_t caps" parameter to this function, so you get aggregate per-task info for a given capabilities mask? In most use cases I think people want to get info on a certain type of memory use (8-bit, or external, or internal) so I think would serve most purposes, and it means this API will match the usage pattern for the other heap metadata APIs. What do you think?

For the (tasks != NULL) behaviour, this seems like it's almost a separate piece of functionality. How would you feel about splitting it into a different function? Something like `heap_caps_get_task_allocations( uint32_t caps, TaskHandle_t *tasks, size_t ntasks, heap_block_info *blocks, size_t nblocks)`?

components/heap/include/esp_heap_debug.h
```diff
@@ -0,0 +1,39 @@
...
+typedef struct {
+    TaskHandle_t task;
+    int size[NUM_USED_TYPES];
+} heap_dump_totals_t;
+
+typedef struct {
+    TaskHandle_t task;
+    void* address;
+    int size: 24;                   // The size of the allocated block.
+    int type: 7;                    // Type of block's region
```
projectgus: In keeping with the suggestions I made above, I'd prefer if this type included a `uint32_t caps` member rather than creating a new enum for expressing the memory type.

components/heap/multi_heap_internal.h
```diff
@@ -28,6 +31,7 @@ void multi_heap_get_info_impl(multi_heap_handle_t heap, multi_heap_info_t *info)
 size_t multi_heap_free_size_impl(multi_heap_handle_t heap);
 size_t multi_heap_minimum_free_size_impl(multi_heap_handle_t heap);
 size_t multi_heap_get_allocated_size_impl(multi_heap_handle_t heap, void *p);
+void* multi_heap_get_block_address_impl(multi_heap_block_handle_t block);
```
projectgus: This is minor, but in the rest of the codebase we usually write `void *` not `void* `. (I see we're not 100% consistent on this ourselves, though!) Don't mind in the source files, but do you mind changing the headers?

slcasner:
Thanks for all the comments.  I will reply to them soon.

components/heap/esp_heap_debug.c
```diff
@@ -0,0 +1,102 @@
...
+#include "multi_heap_internal.h"
+//#include "multi_heap_poisoning.h"
+#include "heap_private.h"
+#include "esp_heap_debug.h"
+
+#ifdef CONFIG_HEAP_TASK_TRACKING
+
+// return all possible capabilities (across all priorities) for a given heap
+
+inline static uint32_t get_all_caps(const heap_t *heap)
```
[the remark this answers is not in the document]
slcasner: I'm fine with this.  I was just minimizing the changes I made to existing files.  Also, in my file I did not need the initial NULL check, but that is no big deal.

components/heap/esp_heap_debug.c
slcasner: Certainly.

components/heap/esp_heap_debug.c
slcasner: I always try to comply with coding conventions, so no problem.  I use emacs which is also quite capable of reformatting.

My personal preference on this one is to omit them because I try to conserve vertical space to maximize the context within a screen.

components/heap/esp_heap_debug.c
slcasner: Oops, that was just a straight-up oversight.  My own file convention includes tabs and I neglected to set the no-tabs setting when working on this file.

components/heap/multi_heap_internal.h
slcasner: I've done more C++ than C for a while, hence the preference for void*, but I'm happy to change this.

components/heap/esp_heap_debug.c
slcasner: Changing the function and struct names is not a problem.

Regarding the enum for capabilities, I would prefer not to add another 4 bytes to the per-block info struct because the array to store these needs to be quite large and RAM on ESP32 is very tight.  I made my code capable of storing these blocks in 32-bit-only IRAM so that this code intended for debugging heap usage would not place too much burden on the heap itself.

One possibility would be to take an array of pairs of capability masks and values, rather than a single caps value, as an additional parameter and then use the index in that array in place of the enum.  That would allow callers to collect one or more regions as they wish.

Since the parameter list is already longer than one would like, adding more might not be good.  What do you think of the idea of defining a struct to hold all the parameters and having that struct as a single parameter to the function?

I see your point that the per-task totals and the list of individual blocks could be considered separate functionalies.  However, both require a full scan through the heap, which is somewhat expensive and requires locking, and if the heap is active then the results between the totals and the block list could be inconsistent.  Therefore I would prefer to keep this as one function.

components/heap/include/esp_heap_debug.h
slcasner: Addressed in the reply above.

components/heap/esp_heap_debug.c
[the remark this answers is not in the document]
slcasner: Addressed in 5683a7c.

components/heap/esp_heap_debug.c
slcasner: Addressed in 00fa388.

components/heap/esp_heap_debug.c
slcasner: Partially addressed, pending my comments, in 00fa388.

components/heap/esp_heap_debug.c
slcasner: Addressed in 5683a7c and 00fa388.

components/heap/esp_heap_debug.c
slcasner: Addressed in 5683a7c and 00fa388.

components/heap/include/esp_heap_debug.h
slcasner: Still pending further discussion.

components/heap/multi_heap_internal.h
slcasner: Addressed in 5683a7c.

slcasner:
Please see my replies to your initial comments for some further discussion.

projectgus:
Hi @slcasner ,

Thanks for the comprehensive reply and thorough addressing of comments!

Regarding the region type enum vs caps mask:

>  I would prefer not to add another 4 bytes to the per-block info struct because the array to store these needs to be quite large and RAM on ESP32 is very tight. 

I understand the wish not to grow the block buffer usage by 25%. That said, are you certain that 25% additional buffer overhead (optionally in IRAM, as you say) will be too much for this to be useful? If you say it is then I'll take your word for it.

> One possibility would be to take an array of pairs of capability masks and values, rather than a single caps value, as an additional parameter and then use the index in that array in place of the enum. That would allow callers to collect one or more regions as they wish.

Sounds acceptable to me. It's a little fiddly but it means we don't mix complicate our API's idea about how to express "types of memory". 

> Since the parameter list is already longer than one would like, adding more might not be good. What do you think of the idea of defining a struct to hold all the parameters and having that struct as a single parameter to the function?

Yes, I agree this is a good idea.

slcasner:
@projectgus, I think I have now addressed all the changes that we've discussed.  I came to realize that it is not necessary to return any indication of the capabilities with the details for individual blocks because the block address determines the heap region which determines the capabilities.  My app code did not use the "type" field that I had before; this was simply a carryover from my first implementation based on the old heap implementation which packed the size and tag into 32 bits.  The reason I don't want to add the capabilities info unnecessarily is that my app code allocates two arrays of 1000 block structs each in order to have enough room to collect all the blocks and to be able compare the previous collection of blocks with the new collection to see what has been freed or newly allocated.

Although this function could be split into two and it could be simplified to collect totals for only one capability value at a time, I believe it is important to be able to collect all this information in one pass through the heaps in order to minimize the amount of time we hold the lock and to get a complete and consistent snapshot at one point in time when the heap state is dynamic.  Having the pair of the caps value and the mask for comparing it is required to be able to group together some regions that share a capability while excluding others. Four <caps,mask> pairs is sufficient for my needs; we could make that constant larger if you think it would be more appropriate, at a cost of 8 bytes per pair times the number of tasks the caller allows for.

I changed the name of the header file to include "esp_" because I think that makes it fit the naming convention.  I added Doxygen documentation for the data structures and function.  I did not add copyright headers because that would be presumptuous.

projectgus:
Sorry for the slow turnaround, I've been out sick this week. Changes LGTM!

>  I did not add copyright headers because that would be presumptuous.

Good point. Do you mind adding these to the new files, based on the other IDF source files? You can add your own Copyright or assign Copyright to Espressif, your call.

slcasner:
Copyright headers added.

projectgus:
Hi @slcasner ,

Thanks! I've cherry-picked this and squashed the commits with a couple of minor tweaks for compilation warnings in certain configurations. Then put it into our review & merge queue.

Final merge may be delayed a little due to the Chinese New Year holiday at the moment.

Thanks for working through this with me!

Angus

slcasner:
Thanks for merging.

PR #1530, opened 2018-01-21
wget

hetlelid:


loboris:
Why? There are other methods to download the file on Linux except `wget`

projectgus:
Hi @hetlelid  ,

Thanks for thinking of us and sending this change. However, can you please explain your reasoning a bit? The URLs in this document are rendered to HTML as part of the documentation - as can be seen here:
https://esp-idf.readthedocs.io/en/latest/get-started/linux-setup.html

So for most users following the getting started guide in the docs, I think the most likely outcome is that they'll click the download link rather than run a command line download tool. For users who want to run a command line tool like wget (or curl, or something else), they will probably know about this command themselves.

What do you think?

Angus

projectgus:
Closing, but @hetlelid if you think we've missing something here please reopen. I hope this doesnt discourage from submitting future PRs in the future if you find problems with docs or code. They are much appreciated.

PR #1557, opened 2018-01-28
This is a fix for the esp_log_level_set function. The problem is when…

dashxdr:
… this

function is called but NOT withe the same 'c' string constant that the LOG*
calls used in each module, the cache check doesn't match, so the cached
entry won't get updated. There's no point in optimizing this function
anyway because it is only called rarely.

### DIFF
--- a/components/log/log.c
+++ b/components/log/log.c
@@ -164,7 +164,7 @@ void esp_log_level_set(const char* tag, esp_log_level_t level)
 #ifdef LOG_BUILTIN_CHECKS
         assert(i == 0 || s_log_cache[(i - 1) / 2].generation < s_log_cache[i].generation);
 #endif
-        if (s_log_cache[i].tag == tag) {
+        if (strcmp(s_log_cache[i].tag,tag) == 0) {
             s_log_cache[i].level = level;
             break;
         }


slcasner:
Thanks.  This is the same problem that I asked about in the forum.
https://www.esp32.com/viewtopic.php?f=14&t=4234&sid=9d985b33b88062cac905a13db1b8295a

projectgus:
Thanks @dashxdr . I've pushed this into our internal review & merge queue.

@slcasner Thanks for linking these two discussions. This is a much simpler fix than the diff you posted. I think in this case it's OK to make the simpler strcmp() change only though. As @dashxdr notes, this is an infrequently called function compared to the log lines (which will tend to be clustered in a single source file and therefore share the same exact string literal.) Please let us know if you think there's something we're missing here.

slcasner:
I agree that the simpler fix will take care of the problem.  @dashxdr notes that optimization is not important here, and I agree; my longer fix includes the same non-optimization as this simpler one.  Since setting the level for a tag is like an instance of accessing the tag, my code updates the generation in the cache like the access code does.  But that is certainly not critical to the performance of this call since if the cache entry were flushed then by definition the stale level setting would be gone!

dashxdr:
If I might offer a suggestion... I think the entire fast cache solution ought to be eliminated. The way it is implemented I believe it will actually reduce overall performance. Assume we start with an initial esp_log_level_setl("*", whatever). This will clear out both the cached and uncached tables. Now there will only be 0 or a few calls additional to the esp_log_level_set function to enable debugging (during development) where we have a tag other than "*". As such the uncached table will contain at most a few entries (and more often just 1). Example: esp_log_level_set("*", NONE) then esp_log_level_set("wifi", VERBOSE). So we want to shut up everything but wifi, which we want to investigage...

So if we just do a naive strcmp of all entries in the uncached list performance is fine, since the list only has one element. Then when we're done developing we don't have any in the uncached list at all, just the default "*" level set once.

That's how it ought to work. Except now how it works is EVERY single LOG* call adds a new tag to the cached list. And they all get assigned the default log level from the "*" log level set call, which doesn't help us at all anyway. So the cached list keeps growing and growing and growing. And now we need to worry about efficiently searching the cached list, so we implement a complicated heap sort.. and then since the list is of finite size we need to implement a most-often-seen-tag solution.

This is insane! The simpler naive implementation would offer better performance, is smaller, is easier to maintain... and is just plain better. The added  complexity is actually harmful.

igrr:
The implementation without cache ends up being slightly faster when no `esp_log_level_set` calls are used to set non-default log levels for tags, and the application uses only a small number of tags. For example, if application only logs with 4 different tags, then each call will take 7us on average without the cache and 8us on average with cache. With larger number of tags being used, and larger number of tags which have non-default log level set, the implementation without the cache becomes slightly slower that the one with cache (for example, 12us vs 9us when the application uses 64 different tags, and 32 tags have non default log levels set).
I agree that the difference is not significant, and given that cache logic adds complexity, it makes sense to remove it.

Issue #1136, opened 2017-10-18
[TW#15969] `make monitor` Error 1


sigwinch28:
After following the install guide for Windows on a Windows 10 64-bit machine, I receive an error when attempting to run `make monitor` after flashing the `hello_world`.

I installed the IDF via the "from scratch" method for Windows and the monitor was _previously working_, but I can't figure out what's changed:

```
$ make monitor
MONITOR
--- idf_monitor on COM3 115200 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x13 (S
Traceback (most recent call last):
  File "C:/msys32/home/Joe/esp/esp-idf/tools/idf_monitor.py", line 568, in <module>
    main()
  File "C:/msys32/home/Joe/esp/esp-idf/tools/idf_monitor.py", line 502, in main
    monitor.main_loop()
  File "C:/msys32/home/Joe/esp/esp-idf/tools/idf_monitor.py", line 269, in main_loop
    self.handle_serial_input(data)
  File "C:/msys32/home/Joe/esp/esp-idf/tools/idf_monitor.py", line 300, in handle_serial_input
    self.console.write_bytes(b)
  File "C:/msys32/mingw32/lib/python2.7/site-packages/serial/tools/miniterm.py", line 63, in write_bytes
    self.byte_output.write(byte_string)
  File "C:/msys32/home/Joe/esp/esp-idf/tools/idf_monitor.py", line 560, in write
    self.output.write(b)
IOError: [Errno 0] Error
make: *** [/home/Joe/esp/esp-idf/components/esptool_py/Makefile.projbuild:96: monitor] Error 1
```

projectgus:
The IOError is a Windows-level OS error. I get similar errors if the USB connection is flaky, or the USB/Serial chip is browning out and resetting.

If the same board has previously worked in the same situation, try swapping out the USB cable.

sigwinch28:
Tried a different board, different USB cable, and different USB port. The problem still occurs with different boards, too.

adamt80:
Same issue here on Win 10 64bit.  Worked fine about a week ago with the same board, USB port, and cable. Now I get this error 100% of the time with any code for make monitor.  The serial monitor in Arduino IDE and Putty work fine for monitoring.

sigwinch28:
[...]

PR #1567, opened 2018-01-30
Additional Fix for Windows 10 IO-Error in make monitor

FredrikFornstad:
It appears that the fix provided to https://github.com/espressif/esp-idf/issues/1136 do not cover all situations: There is another write output in idf_monitor.py that also can go wrong... When running the factory lws app, it seems to trigger this one.

How to reproduce:
Compile and flash https://github.com/warmcat/lws-esp32-factory
Run make monitor, access esp32 with web browser, select an AP, enter password
make monitor (on Windows 10)  will now crash within 40 seconds but normally just after 5-10 seconds. Refering to IO-Error at row 588.

Proposal for fix:
In my pull request, I have duplicated the fix (second try to write if first one fail) that was introduced to the other write statement in idf_monitor.py as the original solution to issue 1136.

Testing:
After applying my fix and have had it running for 30 minutes, I have had no crash,


projectgus:
Thanks @FredrikFornstad , we missed that spot the first time!

Now this is needed in two places I'd prefer to do a little refactoring here, maybe add a _persistent_write(self, data) method to the ANSIColorConverter class and call that from the two places.

Is that change you'd be interested in making to this PR?

FredrikFornstad:
Fully understand that you prefer a nicer solution. Honestly think it is better if you do the refactoring as I think the result will be better then. I totally new to esp-idf...

PR #1589, opened 2018-02-07
fix type mismatch of intr_alloc_flags

h3ndrik:
intr_alloc_flags is of type int. or do we throw away the flags in the upper byte on purpose?


projectgus:
Thanks for noticing this and sending the fix, @h3ndrik . This has been cherry-picked into our review and merge queue (merging may be delayed by the Chinese New Year holidays at the moment, but the PR will update once the fix is merged).

projectgus:
Cherry-picked as dcc466a, thanks!

PR #1594, opened 2018-02-08
Set direction of SMI pins MDC and MDIO correctly.

sauttefk:
Some `GPIOs` do not have the correct default settings for `SMI`.
This PR solves a problem with e.g. `MDIO` on `GPIO32` https://github.com/espressif/arduino-esp32/issues/744#issuecomment-363797043


sauttefk:
@igrr : please merge this little patch so MDC and MDIO can be used on all GPIO pins

projectgus:
Cherry-picked in cb3f094, thanks @sauttefk !

PR #1606, opened 2018-02-09
[api-guides] Remove the THEN keyword from IF

SmallRoomLabs:
Neither the TinyBasicPlus nor the ESP32 variant use the THEN keyword in a IF statement

screen shot 2018-02-09 at 18 03 15


krzychb:
Merged in ef080c0841920c0c73e386be977d445ba5b61049
Thanks!

PR #1638, opened 2018-02-18
esp_ping - Document that timeout units are in milliseconds

chmorgan:
Remove 1000 multiplier from esp_ping_set_target() parameters that define time

### DIFF
--- a/components/lwip/apps/ping/esp_ping.c
+++ b/components/lwip/apps/ping/esp_ping.c
@@ -49,11 +49,11 @@ esp_err_t esp_ping_set_target(ping_target_id_t opt_id, void *opt_val, uint32_t o
         break;
     case PING_TARGET_RCV_TIMEO:
         ESP_PING_CHECK_OPTLEN(opt_len, uint32_t);
-        ping_option_info->ping_rcv_timeout = (*(uint32_t *)opt_val) * 1000;
+        ping_option_info->ping_rcv_timeout = (*(uint32_t *)opt_val);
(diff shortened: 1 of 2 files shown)


projectgus:
Thanks for bringing this to our attention, @chmorgan .

I agree that milliseconds seems to make more sense here. The dilemma is whether to document the current units (seconds) and leave the behaviour unchanged, or change the behaviour and risk breaking existing code. I've asked my colleague who originally wrote this API for their opinion on this.

chmorgan:
Thanks for taking a look st it.

Seconds isn’t a very good unit for a network operation like this. I’m
guessing there are few users of this code. The example I found on the forum
looked like it was assuming milliseconds so perhaps they didn’t test it.

liuzfesp:
LGTM

projectgus:
Thanks @liuzfesp . I've pushed this into our internal review & merge queue.

chrta:
@igrr @projectgus Any chance this change gets cherry-picked into the branch release/v3.0? A ping timeout and delay time in milliseconds make a lot more sense, e.g. during production testing.

projectgus:
Hi @chrta ,

Sorry I didn't reply to you earlier. Unfortunately backporting this to the release branch would change the behaviour of the ping API, which we're reluctant to do in a bugfix point release.

It will be available in the v3.1 release (or you can cherry-pick it yourself if needed, I don't think it will lead to any problems updating IDF in the future.)

Angus

PR #1663, opened 2018-02-28
Enable lwip PPPoS server listen support

ep00ch:
Modified pppapi_listen and pppapi_do_ppp_listen to be consistent with ppp_connect. They no longer use: undefined TCPIP_PPPAPI macro; msg.function which is not a member of pppapi_msg structure; msg.err, which is not a member of pppapi_msg_msg structure. 
ESP32 lwip will now compile and run as PPP server with PPP_SERVER defined.

### DIFF
--- a/components/lwip/api/pppapi.c
+++ b/components/lwip/api/pppapi.c
@@ -270,11 +270,12 @@ pppapi_connect(ppp_pcb *pcb, u16_t holdoff)
 /**
  * Call ppp_listen() inside the tcpip_thread context.
  */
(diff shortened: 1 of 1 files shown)


igrr:
@danicampora @loboris since you are using PPPoS in micropython, could you please review this PR?

loboris:
It would certainly be good to have this enabled.
Unfortunately, I can't test it because I don't have a SIM card with fixed IP at the moment and my device cannot be accessed from Internet.

PR #1668, opened 2018-03-01
Extension to lwip sntp to provide a weak callback notifying us the ti…

markwj:
Regarding:

https://www.esp32.com/viewtopic.php?f=13&t=4833
and
https://www.esp32.com/viewtopic.php?f=2&t=4695

A few users (vonnieda, loboris, and myself) have asked for a way of overriding, or getting notified of, the SNTP time being set. The use case here is to be able to delay the main application code until we have an accurate clock. The application flow then becomes:

1] Wait for network to come up
- sntp_setoperatingmode(SNTP_OPMODE_POLL);
- sntp_setservername(0, (char*)"pool.ntp.org");
- sntp_init();

2] Wait for SNTP time to be set
- Continue with application code

Following forum discussions with ESP_Sprite, and ESP_Angus, this patch implements a mechanism for [2], using weak linked overridable functions.

User code can then override
  void sntp_setsystemtime_us(u32_t t, u32_t us)
to be notified of, and set, system time.

For compatibility and completeness, we have also provided a void sntp_setsystemtime(u32_t t), although this does not seem to be required for the standard ESP32 environment.

### DIFF
--- a/components/lwip/apps/sntp/sntp.c
+++ b/components/lwip/apps/sntp/sntp.c
@@ -199,6 +199,19 @@ static u32_t sntp_last_timestamp_sent[2];
 #endif /* SNTP_CHECK_RESPONSE >= 2 */
 
 /**
+ * Weak callbacks to implement setting system time
+ */
+extern void __attribute__((weak)) sntp_setsystemtime_us(u32_t t, u32_t us)
+{
+  SNTP_SET_SYSTEM_TIME_US(t, us);
+}
+
+extern void __attribute__((weak)) sntp_setsystemtime(u32_t t)
+{
+  SNTP_SET_SYSTEM_TIME(t);
+}
+
+/**
  * SNTP processing of received timestamp
  */
 static void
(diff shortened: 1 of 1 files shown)


igrr:
Hi @markwj, at the moment we are working on a similar feature that allows installing custom callback for the SNTP update function, at run time.
Would run-time registration satisfy your requirements?

projectgus:
@igrr Oops, that might be on me - I'd recommended @markwj use weak linking.

markwj:
Runtime registration of custom callback would also be fine. That was my original approach, but @projectgus suggested weak linking may be more appropriate. Both work just fine for me. So long as I can (a) get notified when the time is set, and (b) be allowed to set the time myself.

igrr:
Ok, sorry for the extra work you had to do @markwj. The change to LwIP SNTP module which allows runtime registration of an update hook (and also smooth time adjustment via adjtime) is in review at the moment.

mwick83:
I'm glad to sse, I'm not the only one needing this feature. I was thinking of implementing something similar myself. @igrr could you provide a link to the runtime registration review/pull request?

igrr:
@mwick83 that would be on our internal code review system, sorry.

mwick83:
@igrr Oh, I see. Do you have an ETA when we can expect to see it on master?

X-Ryl669:
What about a callback in esp/settimeofday function, that way, it does not depends on LWIP version at all ?
You set a pointer to the callback, and it's checked in settimeofday and called if not null. That's enough for what you need, IMHO:
```
static void (*esp_time_updated_cb)(void *) = 0; static void * esp_time_updated_cb_arg = 0;
void set_time_updated_cb(void(*callback)(void*), void * arg) 
{
    esp_time_updated_cb = callback;
    esp_time_updated_cb_arg = arg;
}

int settimeofday(const struct timeval *tv, const struct timezone *tz)
{
    (void) tz;
#if defined( WITH_FRC ) || defined( WITH_RTC )
    if (tv) {
        adjtime_corr_stop();
        uint64_t now = ((uint64_t) tv->tv_sec) * 1000000LL + tv->tv_usec;
        uint64_t since_boot = get_time_since_boot();
        set_boot_time(now - since_boot);
        if (esp_time_updated_cb != NULL) (*esp_time_updated_cb)(esp_time_updated_cb_arg);
    }
    return 0;
#else
    errno = ENOSYS;
    return -1;
#endif
}

```

olegantonyan:
Any news here?

pctj101:
nice! https://github.com/espressif/esp-idf/commit/7e5be1b58cab05cb9c45f06d519f8845175938c8

PR #1671, opened 2018-03-02
Update rmt.c - Fixes exceptions LoadProhibited in RMT driver

jedi7:
Fixes exceptions LoadProhibited, when rmt tries read variable p_rmt on RX END which can be 0.
It happens after esp_reset()  (OTA) and the rmt is not probably stopped by it.
In other words the routine rmt_driver_isr_default was called before second rx channel was registered.

scenario:
register tx channel
register rx channel
reboot
register tx channel
-> exception (because rx channel is accessed)


projectgus:
Cherry-picked and merged in 3d89c8b.

PR #1675, opened 2018-03-03
Place libc gettimeofday function in IRAM

danicampora:
When calling ``gettimeofday()`` from the IDF, the whole call chain resides in IRAM, except for gettimeofday from libc which makes impossible to use this functionality inside interrupt context when the cache is disabled. Adding this line in the ld script fixes the problem.


danicampora:
Hello Angus,

Yes, I’m happy with it. Please close this. Thanks.

Cheers,
Daniel

PR #1694, opened 2018-03-09
fix: fatfs link

XXXXzzzz000:
i think add a link of Virtual filesystem component or instead of fafts is better .

### DIFF
--- a/docs/api-reference/storage/wear-levelling.rst
+++ b/docs/api-reference/storage/wear-levelling.rst
@@ -3,7 +3,7 @@
 See also
 --------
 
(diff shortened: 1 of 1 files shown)


krzychb:
Indeed it looks like a copy and paste error. 
I will add this correction to review queue as well.
Thank you for help to make the documentation better!

igrr:
Thanks @XXXXzzzz000, this has been cherry-picked as https://github.com/espressif/esp-idf/commit/bf2a7fd2c91b124c5a34231dabef91a086ee34ea.

PR #1735, opened 2018-03-17
Update rtc_module.c

AfshinZlfgh:
core panic when using adc1_get_raw inside an interrupt when logging level set to debug, fixed

### DIFF
--- a/components/driver/rtc_module.c
+++ b/components/driver/rtc_module.c
@@ -1287,7 +1287,7 @@ esp_err_t adc1_adc_mode_acquire()
     //lazy initialization
     //for adc1, block until acquire the lock
     _lock_acquire( &adc1_i2s_lock );
-    ESP_LOGD( RTC_MODULE_TAG, "adc mode takes adc1 lock." );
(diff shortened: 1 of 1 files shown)


AfshinZlfgh:
for some uses user may need to reading adc inside an interrupt (in my case to achieve fixed 500 sample rate per second sampling), so if logging level be on debug level,commented logs cause core panic, so i commented 3 logs from this file and issue fixed

igrr:
Hi @AfshinZlfgh, I'm afraid this modification is not going to work reliably because `_lock_acquire` calls can not be done in ISR. If you only have one concurrent call to `_lock_acquire`, done from an ISR, it will be successful. However if there is some task which is holding the same lock when `_lock_acquire` is called, then the code in ISR will fail. This is because `_lock_acquire` has to block until the lock is available, but blocking in an ISR is not possible.

If you need to do ADC sampling at regular intervals, please check the [i2s_adc_dac](https://github.com/espressif/esp-idf/tree/master/examples/peripherals/i2s_adc_dac) example in ESP-IDF, which shows how to achieve this using the DMA block of I2S peripheral.

PR #1760, opened 2018-03-22
Update system_api.c

Mixerito:
out_info->model was zeroed by memset


projectgus:
Cherry-picked in 9f39411. Thanks again!

PR #1761, opened 2018-03-22
Try to merge tail with next block when splitting

rojer:
When splitting a memory block, check if the next block is free.
If it is, then just extend it upwards instead of creating a new block.
This fixes a bug where when shrinking existing allocations would result in irreversible free space fragmentation.


projectgus:
Thanks very muchfor sending this @rojer, good catch.

A couple of small questions/requests. If you don't have time to do them then I can do them on this end:

* Do you think it may be possible to call the existing `merge_adjacent()` function for this case (after testing if the next block is free). This may cut down on code size (and number of individual code paths) a little.
* Could you please write a test for this case in the "test_multi_heap_host" directory?

rojer:
@projectgus 
 * i thought about it too but constructing a block header to have it immediately merged and rewritten seemed a bit silly. the actions seem straightforward enough, wdyt?
 * as noted on IRC (before you posted this), i thought the new check for no adjacent free blocks should be enough (it's invoked multiple times in tests) - but fair enough, will do.

projectgus:
Oops, sorry too many concurrent discussion at the moment. Will take this to IRC for now.

rojer:
@projectgus PTAL

it was indeed not triggering because the default and only test mode for host tests was to check with full poisoning, which masked the problem (by implementing realloc as malloc-copy-free and not trimming the existing allocation).
i added two test cases specifically for this issue and also changed the test to be run in all three configurations.

small side note: the randomized test ends up running the same sequence of operations each time because RND is not seeded. this may be intentional or an oversight, not sure which, so i left it as is for now.

projectgus:
Thanks @rojer, much appreciated.

> small side note: the randomized test ends up running the same sequence of operations each time because RNG is not seeded. this may be intentional or an oversight, not sure which, so i left it as is for now.****

Intentional, but possibly we should just log the (random) seed each time. I wonder if there are seeds which would trigger this bug (with the extra check for adjacent free blocks), or if the random tests aren't wide-ranging enough in their scope.

rojer:
the check is triggered as expected if the test is run with poisoning mode other than comprehensive.

projectgus:
Sorry for the delay, @rojer. This pending merge now.

PR #1770, opened 2018-03-24
fix : gettimeofday

XXXXzzzz000:


docs/en/api-reference/system/esp_timer.rst
```diff
@@ -38,7 +38,7 @@ Obtaining Current Time
 
 ``esp_timer`` also provides a convenience function to obtain the time passed since start-up, with microsecond precision: :cpp:func:`esp_timer_get_time`. This function returns the number of microseconds since ``esp_timer`` was initialized, which usually happens shortly before ``app_main`` function is called.
 
-Unlike `gettimeofday` function, values returned by :cpp:func:`esp_timer_get_time`:
+Unlike :cpp:func:`gettimeofday` function, values returned by :cpp:func:`esp_timer_get_time`:
```
igrr: Does this actually work? As far as I know, we don't have Doxygen-compatible documentation for newlib functions...

XXXXzzzz000: I just found it when I was reading the document.maybe is need be fixed later

krzychb: Hi @XXXXzzzz000, all functions, types, etc. tracked with Doxygen are in [index](https://esp-idf.readthedocs.io/en/latest/genindex.html).

XXXXzzzz000: ok,thanks

Issue #1815, opened 2018-04-04
[TW#20018] RMT ISR null pointer dereference when initialising driver with multiple RMT channels in use after an unexpected reset


DavidAntliff:
Refer https://esp32.com/viewtopic.php?f=13&t=5252&p=22750#p22750

Applies to release/v3.0 f8515688cbc9f7227f4c5ef1706eb3f60c696b13

When two or more RMT channels are in use, and an unexpected reset occurs (unrelated to RMT), the RMT peripheral will continue to run and generate interrupts. There is a race condition between the interrupt handler installed by the first call to rmt_driver_install responding to an interrupt for the other (uninitialised) channel, and the second call to rmt_driver_install which would actually allocate the data structure for the second channel.

The following before a call to `rmt_config()` will avoid the issue:
```
    periph_module_disable(PERIPH_RMT_MODULE);
    periph_module_enable(PERIPH_RMT_MODULE);
```

However this might affect programs that sleep, since it will reset the peripheral. So it seems to me that the better approach is for the ISR to _not_ assume that `p_rmt_obj[channel]` is always non-NULL, and ignore interrupts for uninitialised channels.

I'll submit a PR shortly that ignores interrupts when the channel data structure has not been allocated yet. One side effect of this is that a "case 2" error occurs shortly after the second channel is installed, however the peripheral does appear to function correctly.

koobest:
Hi, @DavidAntliff 
    We reset some peripherals at system startup, including rmt, https://github.com/espressif/esp-idf/blob/64b56beff5a4d6a165bd88027f585bfe51fbb990/components/esp32/clk.c#L240
 please try the latest IDF.
thanks !!

DavidAntliff:
Hi @koobest, thanks for letting me know, I'll try it out and report back if I see any issues.

guirespi:
Hi, I was in ESP IDF v4.4.2 and in my code, i made a loop that constantly turn on and turn off rgb leds. At certain moment the ISR sets the NULL pointer. Can you tell me how to fix this?

guirespi:
I have to add that I only use RMT channel 0 with 4 blocks of the rmt buffer.

PR #1816, opened 2018-04-04
Avoid null pointer dereference in RMT ISR (IDFGH-2502)

DavidAntliff:
Refer #1815.

### DIFF
--- a/components/driver/rmt.c
+++ b/components/driver/rmt.c
@@ -540,52 +540,54 @@ static void IRAM_ATTR rmt_driver_isr_default(void* arg)
             if(intr_st & BIT(i)) {
                 channel = i / 3;
                 rmt_obj_t* p_rmt = p_rmt_obj[channel];
-                switch(i % 3) {
-                    //TX END
-                    case 0:
-                        xSemaphoreGiveFromISR(p_rmt->tx_sem, &HPTaskAwoken);
-                        if(HPTaskAwoken == pdTRUE) {
-                            portYIELD_FROM_ISR();
-                        }
-                        p_rmt->tx_data = NULL;
-                        p_rmt->tx_len_rem = 0;
-                        p_rmt->tx_offset = 0;
-                        p_rmt->tx_sub_len = 0;
-                        if(rmt_tx_end_callback.function != NULL) {
-                            rmt_tx_end_callback.function(channel, rmt_tx_end_callback.arg);
-                        }
-                        break;
-                        //RX_END
-                    case 1:
-                        RMT.conf_ch[channel].conf1.rx_en = 0;
-                        int item_len = rmt_get_mem_len(channel);
-                        //change memory owner to protect data.
-                        RMT.conf_ch[channel].conf1.mem_owner = RMT_MEM_OWNER_TX;
-                        if(p_rmt->rx_buf) {
-                            BaseType_t res = xRingbufferSendFromISR(p_rmt->rx_buf, (void*) RMTMEM.chan[channel].data32, item_len * 4, &HPTaskAwoken);
-                            if(res == pdFALSE) {
-                                ESP_EARLY_LOGE(RMT_TAG, "RMT RX BUFFER FULL");
-                            } else {
-
-                            }
+                if(p_rmt != NULL) {
+                    switch(i % 3) {
+                        //TX END
+                        case 0:
+                            xSemaphoreGiveFromISR(p_rmt->tx_sem, &HPTaskAwoken);
                             if(HPTaskAwoken == pdTRUE) {
                                 portYIELD_FROM_ISR();
                             }
-                        } else {
-                            ESP_EARLY_LOGE(RMT_TAG, "RMT RX BUFFER ERROR\n");
-                        }
-                        RMT.conf_ch[channel].conf1.mem_wr_rst = 1;
-                        RMT.conf_ch[channel].conf1.mem_owner = RMT_MEM_OWNER_RX;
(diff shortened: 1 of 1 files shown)


DavidAntliff:
The diff is actually only four lines - turn off whitespace diff :)

However, I'm not sure your suggestion would prevent the race, however I may have misunderstood you.  It's not the interrupt corresponding to the first call to `rmt_driver_install` that causes the crash, but the _other_ channel, that isn't yet initialised, but is still generating interrupts. So that would require the first call to also disable interrupts on all other channels too? But then the second call needs to know not to do that.

Sorry, I've probably misunderstood you - can you elaborate please?

DavidAntliff:
> Would it be easier to clear interrupt status bits before registering RMT interrupt in rmt_driver_install, after clearing the interrupt enable bits?

Sorry, I'm re-reading this sentence over and over but I'm not really understanding. Are you saying that, if `s_rmt_driver_channels == 0`, the code should clear the interrupt _enable_ bits for _all_ channels, then clear the interrupt _status_ bits for _all_ channels, then register the RMT interrupt handler, then just enable the specified channel's interrupt _enable_ bit? Then subsequent calls (`s_rmt_driver_channels != 0`) would just enable their respective enable bit?

Aside, in my limited experience it's wise to avoid assuming a pointer will never be NULL unless you can formally prove it. Unfortunately in this case, it's not a good assumption to make. So maybe we need both the NULL check and some sort of interrupt flag management?

jedi7:
Hi guys, isn't it connected also with this? https://github.com/espressif/esp-idf/pull/1671

igrr:
@DavidAntliff Sorry that I wasn't clear. I meant clearing interrupt status bits in the `if` block which @negativekelvin has linked to:

```c++
    if(s_rmt_driver_channels == 0) { // first RMT channel using driver
        rmt_clr_intr_enable_mask(0xffffffff);
        RMT.int_clr.val = 0xffffffff;
        err = rmt_isr_register(rmt_driver_isr_default, NULL, intr_alloc_flags, &s_rmt_driver_intr_handle);
    }
```

However as @negativekelvin pointed out, interrupts for the current channel are enabled a few lines before this block. So i think we need to move those `rmt_set_rx_intr_en` and `rmt_set_err_intr_en` calls after the `if` block.

> But what I'm not sure about is how this interacts with software that sleeps while RMT is operating. 

Do you mean deep sleep or light sleep here? In both cases RMT peripheral is not working, because the digital part of the chip is powered down (in deep sleep) or clock gated (in light sleep). Or do you mean something else?

There is also a change to the driver being reviewed at the moment which forces `periph_module_reset(PERIPH_RMT_MODULE);` after software reset. This also has the effect of clearing all pending interrupt status bits.

DavidAntliff:
@jedi7 yes, I believe it's the same issue.

I'm tempted to close this PR as it's not a complete fix for the original issue, however I still think it's a good idea to check a pointer for NULL before dereferencing it, especially if it can't be proven formally that it is never NULL. Since that is all this PR actually does, maybe I'll leave it open...

DavidAntliff:
@igrr thanks for the clarification. I actually use `periph_module_disable/enable(PERIPH_RMT_MODULE);` in my project as it does the job.

In terms of my comment about _sleep_: I'm not very familiar with how that works so I just wanted to make sure there wouldn't be an issue with an approach such as peripheral reset.

In addition, as @jedi7 says, there's also #1671 which is relevant.

I don't think I'm in a good position to sort out all of these related issues and come up with a consolidated fix, so I'll defer to you and the good folks at Espressif to figure out the best approach. However I hope it's now clear that there's definitely a problem with RMT when using multiple channels.

chmorgan:
Espressif guys, any updates on this? I'm looking at David's code for measuring frequency and see that it is using rmt to count pulses and now it looks like I'll be joining the crowd here worrying about null pointer issues...

DavidAntliff:
The merged PR from #1671 technically conflicts with this, but it does something very similar (avoids dereferencing a null pointer, but also clears the interrupt that led to it), so I'm tempted to close this as no longer needed, however I haven't tested whether the alternative fix works for me, so I'll do that first.

Currently I just call `periph_module_disable(PERIPH_RMT_MODULE); periph_module_enable(PERIPH_RMT_MODULE);` right at the start of my program to make this issue "go away", but that's a workaround, not a fix.

@chmorgan if you're seeing this issue then please try the latest 3.0rc1 which contains #1671 and see if it prevents crashing after a crash. Note that your first crash is probably due to something else.

chmorgan:
Ahh ok. I’ll check that I have 1671 in my esp-idf here. Haven’t tried the
code yet so not sure if I’m seeing the same issue I just recalled the other
user working with your led library maybe that was seeing a crash, went
looking into that and found that this was still open. Thanks for the info
and I’m glad you ran it down and there is a fix in place.

projectgus:
@DavidAntliff @chmorgan Are either of you still seeing the underlying issue with v3.0 or current master branch?

DavidAntliff:
@projectgus I've been using v3.0 with #1671 included, but I've also been using `periph_module_disable/enable(PERIPH_RMT_MODULE);` so I'll need to remove this and check to see if the issue has been resolved.

projectgus:
Sorry for the long gap between responses, @DavidAntliff .

 I think this potential null pointer dereference is taken care of in the RMT driver now, at least in master and in the release branches I checked. If you see another place that there may be a race when disabling the driver, please let us know by opening and issue or submitting another PR.

PR #1832, opened 2018-04-06
Allow custom types as suggested by the partition table documentation (IDFGH-2882)

rluba:
The [partition table documentation](http://esp-idf.readthedocs.io/en/latest/api-guides/partition-tables.html#type-field) explicitly recommends defining custom partition types:

> If your application needs to store data, please add a custom partition type in the range 0x40-0xFE.

But there’s _no way_ to use these custom partition types with the partition API. Both `esp_partition_find(…)` and `esp_partition_find_first(…)` demand a `esp_partition_type_t` specifying the partition type.

This PR changes the `type` parameters of these two functions to `uint8_t` to allow custom types.

### DIFF
--- a/components/spi_flash/include/esp_partition.h
+++ b/components/spi_flash/include/esp_partition.h
@@ -96,7 +96,7 @@ typedef struct esp_partition_iterator_opaque_* esp_partition_iterator_t;
  * However, this is the format used by this API.
  */
 typedef struct {
-    esp_partition_type_t type;          /*!< partition type (app/data) */
+    uint8_t type;			            /*!< partition type (app/data) */
     esp_partition_subtype_t subtype;    /*!< partition subtype */
     uint32_t address;                   /*!< starting address of the partition in flash */
     uint32_t size;                      /*!< size of the partition, in bytes */
@@ -119,7 +119,7 @@ typedef struct {
  *         Iterator obtained through this function has to be released
  *         using esp_partition_iterator_release when not used any more.
  */
-esp_partition_iterator_t esp_partition_find(esp_partition_type_t type, esp_partition_subtype_t subtype, const char* label);
+esp_partition_iterator_t esp_partition_find(uint8_t type, esp_partition_subtype_t subtype, const char* label);
 
 /**
  * @brief Find first partition based on one or more parameters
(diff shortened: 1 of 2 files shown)


components/spi_flash/include/esp_partition.h
```diff
@@ -96,7 +96,7 @@ typedef struct esp_partition_iterator_opaque_* esp_partition_iterator_t;
  * However, this is the format used by this API.
  */
 typedef struct {
-    esp_partition_type_t type;          /*!< partition type (app/data) */
+    uint8_t type;			            /*!< partition type (app/data) */
```
projectgus: Can you please do tabs->spaces here so it lines up with the other comments

igrr:
I suppose we would also need to change `subtype` argument in a similar fashion?

How about something like
```
typedef union {
    esp_partition_wellknown_type_t type;
    uint8_t custom_type;
} esp_partition_type_t;
```
and similar with subtype?

rluba:
Sorry for the long delay. I’ve missed the notifications for this thread.

@igrr I did not change the `subtype` because [the documentation](https://esp-idf.readthedocs.io/en/latest/api-guides/partition-tables.html#data-subtypes) states that
> Other data subtypes are reserved for future esp-idf uses.
Therefore I didn’t want to encourage custom subtypes shouldn’t be encouraged.

We _could_ do the union you suggested for `esp_partition_type_t`, but I don’t see the need for adding any type complexity to this simple API.

@projectgus I’ve rebased the branch and changed the tabs to spaces.

projectgus:
Hi @rluba ,

Sorry this has sat around for an extraordinary long time without further feedback.

As noted in an earlier comment, this is only a problem when using C++ not C as C enums are interchangeable with integers. Casting on C++ will work around the problem there (according to the C++ standard it's Undefined Behaviour, but our compiler - and most compilers - will do the right thing without complaint or risk).

I'd rather not totally lose the enum type from the API, as this takes away any association between the two.

The union that Ivan suggested seems like a good idea in that it preserves the enum association but allows C++ users to set any value. However it's a breaking API change, so we'd have to save it for a major version update now.

After all this time, I think what we will do is add a note in the docs that C++ users will need to cast to the enum type(s). Also will update the docs to resolve ambiguity around what values of subtypes are reserved (subtypes are only intended to be specified when the partition type is one of the ESP-IDF reserved partition types, we're happy for applications to use any subtype values when the type is an application-defined type if this is useful for the application.)

Sorry again for ignoring this for so long in between updates.

PR #1878, opened 2018-04-21
esp-tls: fix memory leak when using CA certification validation

redchenjs:
I tested esp-tls with CA certification validation, and I found `cacert` in `struct esp_tls` was used but never be freed. This commit solved the problem.


projectgus:
Thanks @redchenjs for submitting this fix. I've pushed this into our internal review & merge queue.

PR #1880, opened 2018-04-23
Initial support for uart RS485 half duplex

mtx512:
This is my attempt at creating a cleaner fix for adding RS485 half duplex support. An simple echo example is also povided. 

Compared to a previous [PR 667](https://github.com/espressif/esp-idf/pull/667)  the benefits are:

1. Data direction control via the RTS pin can be done in the application code dependent on the RS485 transceiver code deployed.
2. Spurious break characters shouldn't occur in the RX FIFO.
3. The collision detection interrupt is handled.


projectgus:
Similar to comment on [#667](https://github.com/espressif/esp-idf/pull/667#issuecomment-392758389), RS-485 support should finally be merged (based on contributions here, thank you, plus some additional work and unit tests) in the next week or so.

PR #1882, opened 2018-04-25
PKCS#7: initial support

viktorTarasov:
needed on the way to implement the SCEP certificate enrollment.

The ultimate intention is  to implement the CISCO's enrollments protocols (SCEP and EST) for ESP32 with the primary support of secure element (smart card) in the limits needed to enroll the certificate and use it for SSL/TLS .

### DIFF
--- a/components/mbedtls/include/mbedtls/config.h
+++ b/components/mbedtls/include/mbedtls/config.h
@@ -2195,6 +2195,19 @@
 #define MBEDTLS_PKCS5_C
 
 /**
+ * \def MBEDTLS_PKCS7_C
+ *
+ * Enable PKCS#7 functions.
+ *
+ * Module:  library/pkcs7.c
+ *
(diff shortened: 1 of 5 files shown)


projectgus:
Hi @viktorTarasov ,

Thanks for sending this contribution and for your patience while someone got back to you.

As you've probably noticed, last month we refactored mbedTLS to be hosted from a submodule which is a fork of the [upstream project](https://github.com/ARMmbed/mbedtls/) with a few minor fixes/tweaks (that we intend to contribute back).

For major new non-ESP32-specific funcionality like this, we'd very much prefer that it went via the upstream project first and then we can update once mbedTLS releases a new version. Have you considered sending this to the mbedTLS project directly?

Angus

mahavirj:
As @projectgus mentioned, it will be great if you can raise this as PR on `mbedtls` project, we try to keep in sync with their latest release with minimal or no-changes from our side. For now I will close from here. Thanks.

Issue #1917, opened 2018-05-04
esp_pm_dump_locks command causing interrupt watchdog timeout crash (IDFGH-231)


mikaelkanstrup:
## Environment

- Development Kit:      [NodeMCU-32S]
- Core (if using chip or module): [ESP32D0WDQ6 (revision 1)]
- IDF version (``git rev-parse --short HEAD`` to get the commit id.): 0d65f3b
- Development Env:		[Make]
- Operating System:		[Ubuntu]
- Power Supply:         [USB]

## Problem Description

Function esp_pm_dump_locks writes to a file descriptor under critical section. This results in an interrupt watchdog timeout if the underlying FS driver is interrupt based. An example where this problem is seen is when calling the function with stdout file descriptor and stdout being registered to a UART with esp_vfs_dev_uart_use_driver.

### Expected Behavior

No crash expected

### Actual Behavior

Watchdog crash seen

### Steps to reproduce

1. Apply attached patch to esp-idf
2. Enable the following config flags and build examples/system/console
CONFIG_PM_ENABLE=y
CONFIG_PM_PROFILING=y
3. Run the added command 'pm_lock'
4. Notice interrupt watchdog crash

### Code to reproduce this issue

Patch file attached.

## Debug Logs

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0018,len:4
load:0x3fff001c,len:4592
ho 0 tail 12 room 4
load:0x40078000,len:0
load:0x40078000,len:13580
entry 0x40078574
I (260) cpu_start: Pro cpu up.
I (260) cpu_start: Starting app cpu, entry point is 0x40081174
0x40081174: call_start_cpu1 at /esp/esp-idf/components/esp32/cpu_start.c:224

I (0) cpu_start: App cpu up.
I (263) heap_init: Initializing. RAM available for dynamic allocation:
I (270) heap_init: At 3FFAE6E0 len 00001920 (6 KiB): DRAM
I (276) heap_init: At 3FFB9468 len 00026B98 (154 KiB): DRAM
I (282) heap_init: At 3FFE0440 len 00003BC0 (14 KiB): D/IRAM
[...]

PR #1919, opened 2018-05-04
Avoid writing to file under critical section in esp_pm_dump_locks

mikaelkanstrup:
Function esp_pm_dump_locks writes to a file descriptor under critical section.
This results in an interrupt watchdog timeout if the underlying FS driver is
interrupt based. An example where this problem is seen is when calling the function
with stdout file descriptor and stdout being registered to a UART with
esp_vfs_dev_uart_use_driver.

Fix the problem by only copying needed data under critical section and do the
actual file writing with interrupts enabled.

Fixes https://github.com/espressif/esp-idf/issues/1917

Signed-off-by: Mikael Kanstrup <mikael.kanstrup@sony.com>

### DIFF
--- a/components/esp32/pm_locks.c
+++ b/components/esp32/pm_locks.c
@@ -174,26 +174,29 @@ esp_err_t esp_pm_dump_locks(FILE* stream)
 
     fprintf(stream, "Lock stats:\n");
     esp_pm_lock_t* it;
+    esp_pm_lock_t lock;
     SLIST_FOREACH(it, &s_list, next) {
         portENTER_CRITICAL(&it->spinlock);
-        if (it->name == NULL) {
+        lock = *it;
+        portEXIT_CRITICAL(&it->spinlock);
+        if (lock.name == NULL) {
             fprintf(stream, "lock@%p ", it);
         } else {
-            fprintf(stream, "%-15s ", it->name);
+            fprintf(stream, "%-15s ", lock.name);
         }
 #ifdef WITH_PROFILING
-        pm_time_t time_held = it->time_held;
-        if (it->count > 0) {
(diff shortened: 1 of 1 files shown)


components/esp32/pm_locks.c
```diff
@@ -174,26 +174,35 @@ esp_err_t esp_pm_dump_locks(FILE* stream)
...
     esp_pm_lock_t* it;
+    esp_pm_lock_t lock;
+    char* name;
     SLIST_FOREACH(it, &s_list, next) {
         portENTER_CRITICAL(&it->spinlock);
-        if (it->name == NULL) {
+        lock = *it;
+        name = NULL;
+        if (it->name) {
+            name = strdup(it->name);
```
igrr: I'm not sure it it is actually necessary to `strdup` the lock name. Documentation for esp_pm_lock_create requires that names are strings with lifetime no less than the lifetime of the lock itself. Since you're holding s_list_lock, the lock is not going to be deleted until the function returns, so using `lock.name` outside of critical section should be fine.

mikaelkanstrup: Ok nice. Even better. I'll upload a new version.

igrr:
Thanks for the PR @mikaelkanstrup! Once you click through the CLA, I can put this into our internal merge queue.

mikaelkanstrup:
@igrr Sorry for the delay. Unfortunately the Individual CLA caused me troubles contributing code. I will try and sort it out but it might take some time. I'm closing the pull request for now and ask you not to bring in my patch at the moment. It's a simple change so you may choose to handle the bug on your own. Otherwise I'll reopen the pull request once everything's sorted.

PR #1965, opened 2018-05-16
Fix compile error using asm

Pro:
In function 'xPortGetCoreID':
error: expected ')' before ':' token
         :"=r"(id));

It is recommended to use `__asm__` instead of `asm`

See:
https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords

### DIFF
--- a/components/freertos/include/freertos/portable.h
+++ b/components/freertos/include/freertos/portable.h
@@ -205,7 +205,7 @@ BaseType_t xPortInterruptedFromISRContext();
 /* Multi-core: get current core ID */
 static inline uint32_t IRAM_ATTR xPortGetCoreID() {
     int id;
-    asm (
+    __asm__ (
         "rsr.prid %0\n"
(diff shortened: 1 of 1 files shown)


igrr:
Hi @Pro, can you please describe the case in which you get this error? Are you passing any custom CFLAGS when compiling the source files?

Pro:
I do not know which flag exactly caused the compile error, but I'm using the open62541 library in combination with the ESP32 and it has very strict compile flags.

Is there a specific reason why you would like to use `asm` instead of `__asm__`? (Please note the description in the linked documentation)

projectgus:
I'm guessing @Pro is either compiling some source files in their project with either `-ansi` or `std=c99` or similar, and this file is including FreeRTOS headers.

I don't think there is a downside to us using `__asm__` in public headers. Have forwarded this to our internal review queue, if a downside comes up then I'll update this PR.

PR #1990, opened 2018-05-23
Update eclipse-setup.rst

wnoliveira:
This is workaround for eclipse indexer problems. When I disable that option, my problems with eclipse and header files are gone.


projectgus:
Thanks for noticing this, @wnoliveira . Have cherry-picked into our review & merge queue.

PR #2002, opened 2018-05-25
Enable usage of escape char '\' in string values

SonnyX:
Allows the use of escape characters in kconfig and sdkconfig string values, allowing the use of unicode characters, etc.

### DIFF
--- a/tools/kconfig/confdata.c
+++ b/tools/kconfig/confdata.c
@@ -161,18 +161,12 @@ static int conf_set_sym_val(struct symbol *sym, int def, int def_flags, char *p)
 	case S_STRING:
 		if (*p++ != '"')
 			break;
-		for (p2 = p; (p2 = strpbrk(p2, "\"\\")); p2++) {
-			if (*p2 == '"') {
-				*p2 = 0;
-				break;
-			}
-			memmove(p2, p2 + 1, strlen(p2));
-		}
-		if (!p2) {
+		if(p[strlen(p)-1]!='"') {
 			if (def != S_DEF_AUTO)
 				conf_warning("invalid string found");
 			return 1;
 		}
+		p[strlen(p)-1]=0;
 		/* fall through */
 	case S_INT:
 	case S_HEX:
@@ -623,6 +617,7 @@ static void conf_write_symbol(FILE *fp, struct symbol *sym,
 			      struct conf_printer *printer, void *printer_arg)
 {
 	const char *str;
+	char *str2;
 
 	switch (sym->type) {
 	case S_OTHER:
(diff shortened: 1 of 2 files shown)


negativekelvin:
https://patchwork.kernel.org/patch/6357771/

projectgus:
Hi @SonnyX ,

Thanks for sending the PR. Can you explain a bit about the use case for this?

Normally I would be all for merging a change like this. However we actually maintain 4 copies of the kconfig parser right now (this copy, a copy in [kconfig-frontends](http://github.com/espressif/kconfig-frontends) which is for Windows users on the CMake branch, a copy of kconfiglib in the master branch and a different copy of kconfiglib in the CMake branch). This change would need to be applied in all four places to be consistent.

Once we merge the CMake preview branch, at least this will be back down to 2 places not 4.

I'm not saying we can't merge this, but I would like to understand the rationale for this and how important it is to your use case.

Angus

projectgus:
Oh, also - I only just noticed the target branch. We only accept PRs against master. For major bug fixes and the like we can also backport to the v3.0 release branch after merging to master.

SonnyX:
@projectgus 
The use case for this would be allowing the use of special characters in the sdkconfig file: the use case for most users would be the usage of newline and return characters (`\n` and `\r`).
However since a string is just a (unsigned) char array, we use this to initialize a array with device-specific data so that when we change hardware later-on we can just select the right sdkconfig file for the job. (dynamic pin configurations for example)
As an example, the equivelant of:
`unsigned char modulePins[4] = { 17, 18, 21, 22 };` would be `unsigned char modulePins[4] = "\x1a\x1b\x20\x21";`
Then dropping in a variable defined through the kconfig we can turn this into  `unsigned char modulePins[CONFIG_MODULE_AMOUNT] = CONFIG_MODULE_PINS;`

projectgus:
Hi @SonnyX ,

Thanks for explaining. Normally we'd recommend one of the following:

* Setting a single `int` config item for each pin. Better if you have a small number of pins.
* Setting a  "choice" config item for device type and then having some `device.h` which can do `#ifdef CONFIG_DEVICE_MODEL_A`, etc, etc. Better if you have a large number of pins. This will give less obscure errors than having the preprocessor inject a hex string (ie you get the error at the place the array is mis-typed, rather than having to navigate from the macro usage back to the macro definition and then into the sdkconfig file).

Are either of these approaches suitable for your use case?

SonnyX:
Hi @projectgus ,
I agree with you upon these approaches, don't know how I missed the idea of using preprocessor device models.
Regarding the use of special characters in preprocessor defined strings, the changes could ofcourse still be useful for some end-users.

With the above suggestion I am closing this pull request.

PR #2009, opened 2018-05-27
lwip: add support for IPv6 RA RDNSS (RFC 8106)

### LINKED ISSUE #2008: [TW#23133] Support RFC 8106 IPv6 RA Recursive DNS server configuration
IPv6 Router Advertisment messages can contain Recursive DNS servers as defined in [RFC 8106](https://tools.ietf.org/html/rfc8106), so that DNS can be configured without DHCP(v4/v6). This is currently not supported in LWIP, but is a requirement for ESP32 devices to work in IPv6-only networks.

praseodym:
Closes #2008

### DIFF
--- a/components/lwip/core/ipv6/nd6.c
+++ b/components/lwip/core/ipv6/nd6.c
@@ -45,17 +45,20 @@
 
 #if LWIP_IPV6  /* don't build if not configured for use in lwipopts.h */
 
+#include "lwip/dns.h"
 #include "lwip/nd6.h"
 #include "lwip/pbuf.h"
 #include "lwip/mem.h"
 #include "lwip/memp.h"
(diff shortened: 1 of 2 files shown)


projectgus:
@praseodym a little bit of extra context on top of @liuzfesp's comment on #2008: We're in the process of updating the ESP-IDF LWIP to V2.0.3, and probably won't merge any changes to LWIP functionality until after this is done (planning for a matter of weeks not months).

It looks like these patches should apply to upstream V2.0.3 as well though, so we can take a look at this some time after this change lands.

Thanks very much for sending these improvements to us.

praseodym:
Sounds good 👍 Feel free to close this PR unmerged when the LWIP merge is done.

PR #2018, opened 2018-05-31
Fix PartitionDefinition comparison in Python 3.6

yawor:
Even though the #577 has been merged last year, there's still a problem with comparison of `PartitionDefinition` objects. The comparison protocol requires all `__lt__`, `__le__`, `__gt__` and `__ge__` methods to be implemented.


projectgus:
Hi @yawor ,

Thanks for sending this. Please note that right now we still only support Python 2 for ESP-IDF. 2 & 3 support is planned for later this year, though. But we can merge this now. Have cherry-picked into our review & merge queue.

Angus

PR #2024, opened 2018-06-03
eclipse build command is missing a space

wushifu32:


projectgus:
Thanks for sending this, @wushifu32 !

PR #2029, opened 2018-06-04
initial FreeBSD support (IDFGH-1657)

trombik:
been building some codes with SDK 2.x, pre-3.0, and master, I haven't encountered any issues so far.

### DIFF
--- a/tools/cmake/kconfig.cmake
+++ b/tools/cmake/kconfig.cmake
@@ -34,6 +34,11 @@ function(__kconfig_init)
             set(MCONF "\"${WINPTY}\" \"${MCONF}\"")
         endif()
     endif()
+    if(${CMAKE_HOST_SYSTEM_NAME} MATCHES "FreeBSD")
+        set(MAKE_CONMMAND "gmake")
+    else()
+        set(MAKE_CONMMAND "make")
+    endif()
 
     if(NOT MCONF)
         # Use the existing Makefile to build mconf (out of tree) when needed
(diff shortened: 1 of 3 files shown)


projectgus:
Hi @trombik ,

Sorry noone got back to you originally when you submitted this MR.

It seems like the PR Changes feature has gotten a bit confused and is listing a lot of changes from IDF commits now.

What are the commits which are needed for FreeBSD to work? This one: https://github.com/espressif/esp-idf/commit/8d80f81868e5d3357d82a57fc83a6b575fb8c284 - are there others?

trombik:
@projectgus updated the esp-idf tree to master, and cleaned up the branch as well. `menuconfig` and `all` target (both with `make` and `idf.py`) work just fine with my patches. there are more bashism here and there. in most cases, bash is not required. I hope new code would stick with "#!/bin/sh" instead of "#!/bin/bash".

tools/cmake/kconfig.cmake
```diff
@@ -34,6 +34,11 @@ function(__kconfig_init)
             set(MCONF "\"${WINPTY}\" \"${MCONF}\"")
         endif()
     endif()
+    if(${CMAKE_HOST_SYSTEM_NAME} MATCHES "FreeBSD")
+        set(MAKE_CONMMAND "gmake")
```
projectgus: Typo: `MAKE_COMMAND`

trombik: a good catch! fixed, thanks

trombik:
as for no official support, I am fine with it. that's a norm for us. as long as the code is written with portability in mind, which you guys have done well btw, I appreciate.

projectgus:
@trombik Thanks! I've cherry-picked these commits into our internal review & merge queue. This PR will be updated when it's pushed out to master.

projectgus:
@trombik Unfortunately changing the shebang line for check-lxdialog.sh to `/bin/sh` broke macOS builds ("clang: error: unknown argument: '-n' for echo -n.") So we will have to revert this part (commit will be merged soon).

BSD users will probably need to use a bash workaround, sorry.

dobairoland:
Here is the related issue for future reference: https://github.com/espressif/esp-idf/issues/3975

CyanoFresh:
This means I cannot compile for esp32 on freebsd?

Issue #1260, opened 2017-11-11
FreeRTOS xTaskList should be extended to show xCoreID for SMP


dmcnaugh:
Since ESP-IDF has added SMP features to FreeRTOS it would be very helpful to see what core each task is running on.

I have made this change with a quick and dirty mod and it has helped me debug some task conflict problems very effectively as I had wrongly assumed all tasks were on the PRO core unless I had pinned them to the APP core, but this was not the case.

chanibal:
@dmcnaugh, maybe post the dirty mod here, so it can be a starting point?

dmcnaugh:
OK, here's a patch file for the quick and dirty mod I made.

[Issue#1260.patch.zip](https://github.com/espressif/esp-idf/files/1484329/Issue.1260.patch.zip)

dmcnaugh:
I have create a Pull-Request #2064 Added: display xCoreID in vTaskList that has a proposed resolution to this issuie

PR #2064, opened 2018-06-14
Added: display xCoreID in vTaskList

dmcnaugh:
Proposed solution for [TW#16439] FreeRTOS xTaskList should be extended to show xCoreID for SMP #1260

### DIFF
--- a/components/freertos/include/freertos/task.h
+++ b/components/freertos/include/freertos/task.h
@@ -181,7 +181,7 @@ typedef struct xTASK_STATUS
 	uint32_t ulRunTimeCounter;		/*!< The total run time allocated to the task so far, as defined by the run time stats clock.  See http://www.freertos.org/rtos-run-time-stats.html.  Only valid when configGENERATE_RUN_TIME_STATS is defined as 1 in FreeRTOSConfig.h. */
 	StackType_t *pxStackBase;		/*!< Points to the lowest address of the task's stack area. */
 	uint32_t usStackHighWaterMark;	/*!< The minimum amount of stack space that has remained for the task since the task was created.  The closer this value is to zero the closer the task has come to overflowing its stack. */
-	BaseType_t xCoreID;
+	BaseType_t xCoreID;				/*!< Core this task is pinned to */
 } TaskStatus_t;
 
 /**
--- a/components/freertos/include/freertos/task.h
+++ b/components/freertos/include/freertos/task.h
@@ -181,6 +181,7 @@ typedef struct xTASK_STATUS
 	uint32_t ulRunTimeCounter;		/*!< The total run time allocated to the task so far, as defined by the run time stats clock.  See http://www.freertos.org/rtos-run-time-stats.html.  Only valid when configGENERATE_RUN_TIME_STATS is defined as 1 in FreeRTOSConfig.h. */
 	StackType_t *pxStackBase;		/*!< Points to the lowest address of the task's stack area. */
 	uint32_t usStackHighWaterMark;	/*!< The minimum amount of stack space that has remained for the task since the task was created.  The closer this value is to zero the closer the task has come to overflowing its stack. */
+	BaseType_t xCoreID;
 } TaskStatus_t;
 
 /**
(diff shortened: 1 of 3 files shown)


components/freertos/include/freertos/task.h
```diff
@@ -181,6 +181,7 @@ typedef struct xTASK_STATUS
 	uint32_t ulRunTimeCounter;		/*!< The total run time allocated to the task so far, as defined by the run time stats clock.  See http://www.freertos.org/rtos-run-time-stats.html.  Only valid when configGENERATE_RUN_TIME_STATS is defined as 1 in FreeRTOSConfig.h. */
 	StackType_t *pxStackBase;		/*!< Points to the lowest address of the task's stack area. */
 	uint32_t usStackHighWaterMark;	/*!< The minimum amount of stack space that has remained for the task since the task was created.  The closer this value is to zero the closer the task has come to overflowing its stack. */
+	BaseType_t xCoreID;
```
igrr: This line needs a comment similar to the lines above

components/freertos/tasks.c
```diff
@@ -4449,7 +4450,7 @@ For ESP32 FreeRTOS, vTaskExitCritical implements both portEXIT_CRITICAL and port
 				pcWriteBuffer = prvWriteNameToBuffer( pcWriteBuffer, pxTaskStatusArray[ x ].pcTaskName );
 
 				/* Write the rest of the string. */
-				sprintf( pcWriteBuffer, "\t%c\t%u\t%u\t%u\r\n", cStatus, ( unsigned int ) pxTaskStatusArray[ x ].uxCurrentPriority, ( unsigned int ) pxTaskStatusArray[ x ].usStackHighWaterMark, ( unsigned int ) pxTaskStatusArray[ x ].xTaskNumber );
+				sprintf( pcWriteBuffer, "\t%c\t%u\t%u\t%u\t%hd\r\n", cStatus, ( unsigned int ) pxTaskStatusArray[ x ].uxCurrentPriority, ( unsigned int ) pxTaskStatusArray[ x ].usStackHighWaterMark, ( unsigned int ) pxTaskStatusArray[ x ].xTaskNumber, ( int ) pxTaskStatusArray[ x ].xCoreID );
```
igrr: Some existing code may rely on the format of the string returned by `vTaskList`. Suggest introducing a new sdkconfig option, e.g. CONFIG_FREERTOS_VTASKLIST_INCLUDE_COREID, disabled by default. Then put the modifications under `ifdef`, so that by default the old behavior is preserved.

components/freertos/include/freertos/task.h
weyoui: After looking BaseType_t definition in portmacro.h, BaseType_t type is int. I think uint32_t is more suitable for xCoreID.
```
#define portBASE_TYPE	int
typedef portBASE_TYPE			BaseType_t;
```

components/freertos/tasks.c
weyoui: And here int could be "unsigned int" or "uint32_t" following above comment.

components/freertos/include/freertos/task.h
dmcnaugh: No, I have taken made xCoreID have type BaseType_t because this is how the value is typed in the tskTaskControlBlock (tasks.c) where the value is sourced from.

dmcnaugh: Comment now added

components/freertos/tasks.c
dmcnaugh: Added the option CONFIG_FREERTOS_VTASKLIST_INCLUDE_COREID as requested. Please se the revised Pull Request.

dmcnaugh:
Thanks for approving this.
What happens from here: when and/or how does this PR show up in the **master** branch"?

igrr:
Hi @dmcnaugh, sorry for the slow turnaround. I've put this into review/merge queue, the changes should be in master branch in a few days.

igrr:
Cherry-picked in 0fb1945, thanks!

PR #2065, opened 2018-06-14
doc/wifi: small fix

avrahamshukron:
The WiFi station connect sequence diagram does not have any "successful" ending.
It looks like a copy-paste error, because there are two consequent `SYSTEM_EVENT_STA_DISCONNECTED` events firing with nothing happening in between.

I think the original meaning was that the last event should be `SYSTEM_EVENT_STA_CONNECTED` to signify that if all goes well, that is the event that will be fired.


igrr:
Thanks for the fix @avrahamshukron, this has been cherry-picked as https://github.com/espressif/esp-idf/commit/0cab3cecfdafffcc4bfd3aa90b4a523e455b2d7a.

Issue #2081, opened 2018-06-20
[TW#23627] no output from make monitor, but platformio's monitor works.


ritualz:
- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

## Environment

- Development Kit:      [dfrobot firebeetle esp32]
- Core (if using chip or module): [ESP-WROOM32]
- IDF version: 9a55b42f
- Development Env:		[Make]
- Operating System:		[Linux]
- Power Supply:         [USB | Battery]

## Problem Description
I just got a new board, which connect to /dev/ttyACMx, in the past I've only used esp32 modules that uses ttyUSBx.  Uploading and flashing works fine, but when trying to monitor with `make monitor` the only output is:
`
MONITOR
--- idf_monitor on /dev/ttyACM0 115200 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
` 
This didn't happen when I'm using the other boards. I'm has chmod the device and are in the dialout group. The strange thing is that the platformio monitor works, using the same baudrate. 

I've added the relevant section of the sdkconfig file.  
`
CONFIG_ESPTOOLPY_PORT="/dev/ttyACM0"
CONFIG_ESPTOOLPY_BAUD_115200B=y
CONFIG_ESPTOOLPY_BAUD_230400B=
CONFIG_ESPTOOLPY_BAUD_921600B=
CONFIG_ESPTOOLPY_BAUD_2MB=
CONFIG_ESPTOOLPY_BAUD_OTHER=
CONFIG_ESPTOOLPY_BAUD_OTHER_VAL=115200
CONFIG_ESPTOOLPY_BAUD=115200
CONFIG_ESPTOOLPY_COMPRESSED=y
CONFIG_FLASHMODE_QIO=
CONFIG_FLASHMODE_QOUT=
CONFIG_FLASHMODE_DIO=y
CONFIG_FLASHMODE_DOUT=
CONFIG_ESPTOOLPY_FLASHMODE="dio"
CONFIG_ESPTOOLPY_FLASHFREQ_80M=
CONFIG_ESPTOOLPY_FLASHFREQ_40M=y
[...]

PR #2090, opened 2018-06-22
Added RST and DTR options to idf-monitor

mmikalsen:
Fixed the issues I experienced with the idf-monitor (#2081), which was related to the RST and DTR pulled lowed by default.

### DIFF
--- a/components/esptool_py/Kconfig.projbuild
+++ b/components/esptool_py/Kconfig.projbuild
@@ -216,7 +216,7 @@ config MONITOR_DTR
 		bool "DTR value of idf-monitor"
 
 
-config MONITOR_RST
-		bool "RST value of idf-monitor"
+config MONITOR_RTS
+		bool "RTS value of idf-monitor"
 
 endmenu
--- a/components/esptool_py/Kconfig.projbuild
+++ b/components/esptool_py/Kconfig.projbuild
@@ -212,4 +212,11 @@ config MONITOR_BAUD
     default 2000000 if MONITOR_BAUD_2MB
     default MONITOR_BAUD_OTHER_VAL if MONITOR_BAUD_OTHER
 
+config MONITOR_DTR
+		bool "DTR value of idf-monitor"
+
+
+config MONITOR_RST
+		bool "RST value of idf-monitor"
+
 endmenu
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -617,6 +617,22 @@ def main():
         help="Filtering string",
         default=DEFAULT_PRINT_FILTER)
 
+    parser.add_argument(
+        '--dtr',
+        help='Serial port DTR value',
+        const=False,
+        nargs='?',
+        type=bool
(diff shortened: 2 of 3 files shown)


components/esptool_py/Kconfig.projbuild
```diff
@@ -212,4 +212,11 @@ config MONITOR_BAUD
     default 2000000 if MONITOR_BAUD_2MB
     default MONITOR_BAUD_OTHER_VAL if MONITOR_BAUD_OTHER
 
+config MONITOR_DTR
+		bool "DTR value of idf-monitor"
+
+
+config MONITOR_RST
```
igrr: Here and below: the signal name is RTS ("request to send"), not RST.

projectgus:
Hi @ritualz ,

Thanks for sending this PR! Before we review and merge, I'd like to understand the exact problem that is being solved here:

> which was related to the RST and DTR pulled lowed by default.
...
>-  serial_instance.dtr = False
>-  serial_instance.rts = False

"False" here should mean "not asserted" which for 3.3V serial is usually the same as "not pulled low" (these are active-low signals on every 3.3V serial chip I've come across).

What USB/serial interface are you using, and what wiring configuration to the ESP32 where de-asserting DTR & RTS causes problems?

Thanks,


Angus

projectgus:
@ritualz Apologies, I missed the linked issue. Will take a look there and then reply again.

projectgus:
I found a schematic for the Firebeetle and see that it's a STM32F042 with two GPIOs wired directly to GPIO0 and EN.

There are some soft USB/serial firmwares around which use DTR as the "enable serial" signal, so my guess is that's also what is happening here.

Out of interest, is esptool.py able to flash this board successfully? Does it auto-reset the board into flashing more successfully?

Angus

projectgus:
> Out of interest, is esptool.py able to flash this board successfully? Does it auto-reset the board into flashing more successfully?

Reason I ask this is that while esptool.py is uploading, both DTR and RTS should be de-asserted (set False). I can't explain why this would work and monitor wouldn't work...

tools/idf_monitor.py
```diff
@@ -617,6 +617,22 @@ def main():
         help="Filtering string",
         default=DEFAULT_PRINT_FILTER)
 
+    parser.add_argument(
+        '--dtr',
+        help='Serial port DTR value',
+        const=False,
+        nargs='?',
+        type=bool
```
weyoui: I am not sure but feeling we need to add default value for dtr and rts.
```
default = False
```

LucaCerina:
+1 on this pull request. The DF-Robot ESP32 FireBeetle requires a different configuration DTS/RTS to have the monitor working properly

chaosman123:
+1 on this pull request. The DF-Robot ESP32 FireBeetle requires a different configuration DTS/RTS to have the monitor working properly.
Just ran into exactly the same problem. Easy resolved by locally applying the suggested change.

PR #2093, opened 2018-06-23
esp_http_client: add head method support

jkoelker:


weyoui:
I think it would be more helpful to add some description on this PR.

components/esp_http_client/esp_http_client.c
```diff
@@ -690,6 +691,11 @@ static int esp_http_client_get_data(esp_http_client_handle_t client)
     if (client->state < HTTP_STATE_RES_COMPLETE_HEADER) {
         return ESP_FAIL;
     }
+
+    if (client->connection_info.method == HTTP_METHOD_HEAD) {
+        return 0;
```
weyoui: 0 means ESP_OK? It would be better to use more meaningful definition instead of 0.

examples/protocols/esp_http_client/main/esp_http_client_example.c
```diff
@@ -136,6 +136,18 @@ static void http_rest()
...
+    esp_http_client_set_url(client, "http://httpbin.org/get");
+    esp_http_client_set_method(client, HTTP_METHOD_HEAD);
+    err = esp_http_client_perform(client);
+    if (err == ESP_OK) {
+        ESP_LOGI(TAG, "HTTP HEAD Status = %d, content_length = %d",
+                esp_http_client_get_status_code(client),
+                esp_http_client_get_content_length(client));
+    } else {
+        ESP_LOGE(TAG, "HTTP HEAD request failed: %s", esp_err_to_name(err));
+    }
```
weyoui: :+1: Good test case!

components/esp_http_client/esp_http_client.c
jkoelker: No it means 0 bytes read. The return of the function is the number of bytes read. The above `ESP_FAIL` return seems to rely on the fact that currently `ESP_FAIL` is `#define`'d to `-1`.  I wavered in putting the `HEAD` check below in `esp_http_client_perform` and just skipping looking at the body at all, but then the state check above would need to move as well, and the correctly data length for a `HEAD` request is `0` since the RFC states that it "must not return a message-body in the response", so the `esp_http_client_get_data` seemed more appropriate.

tuanpmt:
Thanks for the useful method, 

LGTM

igrr:
Cherry-picked as 483c3d7, thanks @jkoelker!

PR #2097, opened 2018-06-24
Doc: alias should add in .bashrc

cubicwork:
Doc: alias should add in .bashrc in CentOS7.


igrr:
Thanks for the PR @cubicwork, this has been cherry-picked as fab55e3 and 7ccc28d.

Issue #2056, opened 2018-06-13
Eclipse can't find includes for c++ file


jeremyherbert:
## Environment
- IDF version: 7aa7b35a 
- Development Env:		Eclipse
- Operating System:		Xubuntu 18.04

- Eclipse version: Oxygen.3a Release (4.7.3a)

## Problem Description

If I open the "hello_world" example in eclipse, everything works correctly after setting up as per the documentation. If I then rename the `hello_world_main.c` file to `hello_world_main.cpp` and add the following code in:

```
extern "C" {
	void app_main(void);
}
```

the build will still succeed, but eclipse will no longer be able to resolve the include directories at the top of the file. For example, `#include "freertos/FreeRTOS.h"` has the orange squiggle underline.

### Expected Behavior

Eclipse should resolve the include paths for both C and C++

### Actual Behavior

It only works for C.

## Debug Logs

Eclipse indexer log for C++:
```
Project:               hello_world
File:                  file:/home/jeremy/esp/hello_world/main/hello_world_main.cpp
Language:              GNU C++
Index Version:         211.0
Build Configuration:   Default
Context:               file:/home/jeremy/esp/hello_world/main/hello_world_main.cpp
   C++, {}
Versions in Index:     1
   C++: {}; 0 macros, 5 includes, 13 names;

Include Search Path (option -I):
   /home/jeremy/esp/xtensa-esp32-elf/xtensa-esp32-elf/include/c++/5.2.0
   /home/jeremy/esp/xtensa-esp32-elf/xtensa-esp32-elf/include/c++/5.2.0/xtensa-esp32-elf
   /home/jeremy/esp/xtensa-esp32-elf/xtensa-esp32-elf/include/c++/5.2.0/backward
   /home/jeremy/esp/xtensa-esp32-elf/lib/gcc/xtensa-esp32-elf/5.2.0/include
   /home/jeremy/esp/xtensa-esp32-elf/lib/gcc/xtensa-esp32-elf/5.2.0/include-fixed
   /home/jeremy/esp/xtensa-esp32-elf/xtensa-esp32-elf/include
   /home/jeremy/esp/xtensa-esp32-elf/xtensa-esp32-elf/sysroot/usr/include
```

Eclipse indexer log for C:
```
Project:               hello_world
                 file:/home/jeremy/esp/hello_world/main/hello_world_main.c
Language:              GNU C
Index Version:         211.0
Build Configuration:   Default
[...]

PR #2104, opened 2018-06-26
Fix documentation; Eclipse CDT GCC build output parser regex was incorrect

jeremyherbert:
Fix as per https://github.com/espressif/esp-idf/issues/2056


projectgus:
Thanks for the patch @jeremyherbert . As per update on Github issue, this was independently fixed in 75dc2f58005808fcaa86b7f560f45fa22b7884e8 based on your report.

PR #2108, opened 2018-06-26
Add espmqtt Component to IDF

nicola-lunghi:
Add Mqtt library to ESP-IDF

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -41,3 +41,6 @@
 [submodule "components/mbedtls/mbedtls"]
 	path = components/mbedtls/mbedtls
 	url = https://github.com/espressif/mbedtls.git
+[submodule "components/espmqtt"]
+	path = components/espmqtt
+	url = https://github.com/tuanpmt/espmqtt.git


nkolban:
I'm wondering out loud if there is a relationship between this request (inclusion of MQTT) and the current inclusion of the Amazon AWS IoT component.  It is my understanding that AWS IoT also provides an MQTT client.  I am not sure if this is dedicated or limited to AWS.  I am also not sure if a generic MQTT could be used instead of AWS IoT client.  It might however be worthwhile examining the notion of having multiple MQTT clients and seeing if there isn't value in having just one or the other.

nicola-lunghi:
@tuanpmt

nicola-lunghi:
My main problem with the AWS IoT is that doesn't support MQTT without TLS.
EspMQTT support multiple transport and is more generic (e.g. you can even connect to AWS).

In my opinion in the framework is better to have a more generic MQTT library not related to any specific implementation.

stewy42:
@nicola-lunghi @tuanpmt 
The three items you listed of mine are are certainly ones I would like to see implement.

Though my preference would not have 
- the implementation of custom heap allocation routines as a stretched goal
     An alternative to this would be to have a CONFIG preference that allows the publish/subscribe buffers to be assigned to SPIRAM space 

For completeness the  fini_message   routine should be updated to allow longer messages. This would relatively easy to do by increasing MQTT_MAX_FIXED_HEADER_SIZE  to say 5 and adding additional code in fini_message   .  No other code I believe needs be changed

I should perhaps point out that I am using the current code base , unmodified, with great success.  I just work withing the limitations outlined .

tuanpmt:
Thank you for your interest in the espmqtt library. Currently, the library is being improved, fix known bugs, adds testing, and integrates into the esp-idf soon. Your comments are useful and realistic, we will try to make the release close to what you propose

PR #2112, opened 2018-06-27
add new example

nobotro:


weyoui:
I am not sure we need to have this new example in our esp-idf source tree. If this example is useful, we could make another git repository for the new examples. I think we need to have essential example to show the each esp-idf components.

PR #2200, opened 2018-07-17
Enable subprojects (make bootloader case more generic)

nathanjel:
This allows practical usage of the "subproject" idea as introduced for bootloader to deliver other subprojects. My use case is OTA, where OTA_0 is a simple application that allows re-programming only and OTA_1 is the actual application. This allows to have a fallback if programming fails, and then have a significantly large actual application, as it is larger than 50% of the flash size. As defining `IS_BOOTLOADER_BUILD` both enables subproject in project.mk and at the same time disables multiple needed functions in spi_flash, it makes sense to split this to be controlled thru different variables. So `IS_SUBPROJECT_BUILD` controls the skip of building configuration in project.mk to enable subproject, and `IS_BOOTLOADER_BUILD` remains to control the rest. This then requires setting of `IS_SUBPROJECT_BUILD`  in bootloader Makefile.

### DIFF
--- a/components/bootloader/subproject/Makefile
+++ b/components/bootloader/subproject/Makefile
@@ -22,11 +22,13 @@ CFLAGS += -I $(IDF_PATH)/components/esp32/include
 # IS_BOOTLOADER_BUILD tells the component Makefile.projbuild to be a no-op
 IS_BOOTLOADER_BUILD := 1
 export IS_BOOTLOADER_BUILD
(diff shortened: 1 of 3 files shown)


nathanjel:
I have updated with additional commit, to remove the bogus dependency on sdkconfig.h from esp32 component, that jeopardized subproject build.

nathanjel:
Squashed to one commit

PR #2209, opened 2018-07-19
authenticate the correct key so bonds can endure across esp32 resets (IDFGH-2879)

### LINKED ISSUE #2036: [TW#23304] BLE SMP:LE secure connection (passkey) - Bug when set characteristic permissions to ENC_MITM.
## Environment

- Development Kit:      [ESP32-Wrover-Kit and others]
- Kit version (for WroverKit/PicoKit/DevKitC): [v3]
- IDF version (``git rev-parse --short HEAD`` to get the commit id.): b2ff235b 
- Development Env:		[Make]
- Operating System:		[Ubuntu and MacOS]
- Power Supply:         [external 5V]

## Problem Description

Example Code: examples/bluetooth/gatt_security_server

Client Device: Various iPhones (6,6+,7,8,X) and Android. Same behaviour with ALL of these devices.

If I change the code to use passkey pairing with MITM:

```cpp
esp_ble_auth_req_t auth_req = ESP_LE_AUTH_REQ_SC_MITM_BOND;   
esp_ble_io_cap_t iocap = ESP_IO_CAP_OUT;
```
Everything works as expected. After Rebooting the ESP, my client device is still bonded and can connect/read/write without another pairing. 

HOWEVER:

If I additionally change the security permissions of the characteristics to 'xxx_ENC_MITM' e.g.

```cpp
    // Heart Rate Control Point Characteristic Value
    [HRS_IDX_HR_CTNL_PT_VAL]             =
    {{ESP_GATT_AUTO_RSP}, {ESP_UUID_LEN_16, (uint8_t *)&heart_rate_ctrl_point, ESP_GATT_PERM_WRITE_ENC_MITM | ESP_GATT_PERM_READ_ENC_MITM,
      sizeof(uint8_t), sizeof(heart_ctrl_point), (uint8_t *)heart_ctrl_point}},
```

On the first connect, passkey pairing (le secure) is triggered as expected, and the characteristic can afterwards be accessed.

**But then, after rebooting the ESP, the client device CAN NOT access those characteristics any more.**

The connection works, but somehow, the connection does not have the MITM flag set and therefore there is an auth error.

The lowlevel error ist:
```
I (16062) SEC_GATTS_DEMO: ESP_GATTS_CONNECT_EVT
E (18472) BT_GATT: GATT_INSUF_AUTHENTICATION
```

This worked perfectly with release V2 and release V3 RC1. 

I checked the connection with OSX' packet-logger and it seems the connection does really use MITM and is encrypted. (This is due to the fact, that security is switched on in the GATT CONNECT handler in the example)

Does anybody know/can confirm if this is a bug? 

Thank you very much!
Boris

tomx4096:
Hi,
This is an attempt to resolve #2036 which caused a bonded esp32 to incorrectly set characteristic permissions with ENC_MITM set after a reset. The incorrect key sec_level was checked, so the BTM_SEC_LE_LINK_KEY_AUTHED bit was never set in the security flags. I had the same problem described in #2036 and this appeared to fix it.
Thanks,
Tom

### DIFF
--- a/components/bt/bluedroid/stack/btm/btm_ble.c
+++ b/components/bt/bluedroid/stack/btm/btm_ble.c
@@ -1225,7 +1225,7 @@ void btm_sec_save_le_key(BD_ADDR bd_addr, tBTM_LE_KEY_TYPE key_type, tBTM_LE_KEY
 
             /* Set that link key is known since this shares field with BTM_SEC_FLAG_LKEY_KNOWN flag in stack/btm_api.h*/
             p_rec->sec_flags |=  BTM_SEC_LE_LINK_KEY_KNOWN;
-            if ( p_keys->pcsrk_key.sec_level == SMP_SEC_AUTHENTICATED) {
+            if ( p_keys->lenc_key.sec_level == SMP_SEC_AUTHENTICATED) {
                 p_rec->sec_flags |= BTM_SEC_LE_LINK_KEY_AUTHED;
             } else {
                 p_rec->sec_flags &= ~BTM_SEC_LE_LINK_KEY_AUTHED;


tomx4096:
@Weijian-Espressif

AxelLin:
This should be closed since the fix is already in https://github.com/espressif/esp-idf/commit/163dbc32007c5f2afc8ada81ed6371fce20c7774.

Alvin1Zhang:
@AxelLin Thanks for the reminder. @tomx4096 Thanks for your contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/163dbc32007c5f2afc8ada81ed6371fce20c7774.

PR #2226, opened 2018-07-26
Added Reason-Phrase to status line

kluverp:
Just checking for the string "200" can lead to false-positives. For example my server responded with a 404 but included the following header:
Strict-Transport-Security: max-age=63072000; includeSubdomains
Which lead to a false-positive.
Adding the Reason-Phrase seems like a solution.


mahavirj:
@kluverp Thanks for your contribution. OTA examples in IDF have been substantially refactored now, we also now have simplified abstraction layer in the form of https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/esp_https_ota.html. Please feel free to raise PR if you still observe any issues, I will go ahead and close this PR as it will not directly apply.

PR #2237, opened 2018-07-27
Adds a Kconfig option for mbedtls' MBEDTLS_PLATFORM_MEMORY define.

vonnieda:
This makes it possible to override the mbedtls allocator with your own.

More information available in this forum thread:
https://esp32.com/viewtopic.php?f=2&t=6550


projectgus:
Hi @vonnieda ,

I've replied to you on the forum again: https://esp32.com/viewtopic.php?f=2&t=6550&p=28208#p28208

As written there, I'm not convinced this is the best solution to the problem you encountered.

The downside with enabling options like this in kconfig is that we have to support the inevitable second-order problems that appear when people misuse them.

vonnieda:
Hi @projectgus,

I responded in the thread. 

I think it's important to be able to manage such a large allocation. A single TLS connection is nearly 10% of the total internal RAM, and if you take WiFi and BLE into account it's actually more like 16%. If you have to juggle more than one TLS connection, as I do, it becomes a real issue finding internal RAM for it. 

Either way, my specific issue is solved by making a copy of mbedtls in my components directory, but it would be nice not to have to do that.

Thanks,
Jason

components/mbedtls/Kconfig
```diff
@@ -1,5 +1,27 @@
 menu "mbedTLS"
 
+config MBEDTLS_PLATFORM_MEMORY
+   bool "Enable mbedTLS memory allocation layer."
```
projectgus: Suggest "Enable custom mbedTLS memory allocation layer", so it's clear this is more than just having mbedTLS support memory allocation.

components/mbedtls/Kconfig
```diff
@@ -1,5 +1,27 @@
...
+   default n
+   help
+      By default mbed TLS uses the system-provided calloc() and free().
+      This allows different allocators (self-implemented or provided) to be
+      provided to the platform abstraction layer.
+
+      Enabling MBEDTLS_PLATFORM_MEMORY without the
```
projectgus: Most of this description doesn't make sense in the ESP-IDF context, where the user isn't editing the mbedTLS config header and can't easily define macros. Suggest something like:

> If this option is disabled, mbed TLS uses the default system calloc() and free() functions.
> 
> If this option is enabled, the mbed TLS config macro MBEDTLS_PLATFORM_MEMORY will be defined. The function mbedtls_platform_set_calloc_free() must be called at runtime to provide custom calloc() and free() function pointers for use by mbedTLS.
>
> This option allows fine-grained control over how mbedTLS allocates heap memory.

vonnieda:
Thanks @projectgus, updated the help and option name as you suggested.

projectgus:
Thanks @vonnieda . I've squashed these to a single commit and cherry-picked them into our internal review & merge queue.

vonnieda:
Great, thank you @projectgus!

projectgus:
Cherry-picked as  d7a17ac, thanks!

vonnieda:
Thank you!

PR #2261, opened 2018-08-03
Fix small parameter issue in bt_a2dp_sink example

Piocky:
The wrong parameter was used to print the error if something failed during the bt init.


krzychb:
Hi @Piocky,

This is a good catch! 
Thank you for taking time to issuing the MR with resolution.

How about declaring at the beginning of`app_main()` a single `esp_err_t err` and using the same `err` variable across  the whole `app_main()` ? Chances are that the same bug may sneak again, if somebody is  adding another error check.

Piocky:
Hi @krzychb,

So in this case, using only the `ret` variable?

krzychb:
Hi @Piocky,

Sorry for slow reply.  I meant using just one variable, either `err` or `ret`.
I think that `err` goes well with `esp_err_t err` or `esp_err_to_name(err)`, but if you retain `ret` this should be also fine :)

Piocky:
Ok, do you want me to make the change?

krzychb:
Yes, please doe so.

Piocky:
Done, but there is a conflict now.

krzychb:
This conflict would show up anyway because of updates to examples/bluetooth/a2dp_sink/main/main.c that happened after your original commit in https://github.com/espressif/esp-idf/commit/1df85e0039bf444eec58e246f3e8336105bec72d#diff-3516b30c8e3250d6006229ab65e1647d

To resolve it is necessary to rebase this PR on the master.

Piocky:
I corrected the conflict, let me know if it's okay for you right now, or if I need to change something.

krzychb:
Looks good to me. I will take it for internal review. Thank you!

igrr:
Cherry-picked in 23a510c, thanks!

Issue #2246, opened 2018-07-31
esp_http_client exception on 401 without Www-Authenticate header


kluverp:
## Environment

- Development Kit:      [none]
- Kit version (for WroverKit/PicoKit/DevKitC): -
- Core (if using chip or module): [ESP-WROOM3]
- IDF version: be81d2c16d7f4caeea9ceb29fece01510664caf3
- Development Env:		[Eclipse]
- Operating System:		[Ubuntu 16.04]
- Power Supply:         [external 3.3V]

## Problem Description
When the server gives back a 401 response and the "Www-Authenticate" header is missing, a null pointer Exception occurs in the function "http_utils_trim_whitespace()".

On line 70  in "/components/esp_http_client/lib/http_utils.c" a NULL pointer may be passed in, when header is missing:

```cpp
...
char *end;
    char *start = *str;

    // Trim leading space
    while (isspace((unsigned char)*start)) start ++;
...
```
### Expected Behavior
Maybe issue warning?

## Debug Logs
I (12025) HTTP_CLIENT: Posting to https://192.168.1.130/
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.
Core 1 register dump:
PC      : 0x4010d4dc  PS      : 0x00060930  A0      : 0x8010c01f  A1      : 0x3ffcabe0  
0x4010d4dc: http_utils_trim_whitespace at /home/peter/esp/esp-idf-master/components/esp_http_client/lib/http_utils.c:74 (discriminator 1)

A2      : 0x3ffcac00  A3      : 0x00000000  A4      : 0x3ff96354  A5      : 0x0000ca11  
A6      : 0x00000191  A7      : 0x00000000  A8      : 0x00000000  A9      : 0x00000000  
A10     : 0x00000000  A11     : 0x3ffd4b04  A12     : 0x00000010  A13     : 0x00000000  
A14     : 0x3ffcb230  A15     : 0xff000000  SAR     : 0x00000010  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000000  LBEG    : 0x4000c2e0  LEND    : 0x4000c2f6  LCOUNT  : 0x00000000  

Backtrace: 0x4010d4dc:0x3ffcabe0 0x4010c01c:0x3ffcac00 0x4010c392:0x3ffcac30 0x400d6aef:0x3ffcac50 0x400d6b87:0x3ffcace0
[...]

PR #2282, opened 2018-08-08
esp_http_client: don't trim auth_header if it's NULL

rbino:
esp_http_client crashed when the server replied with 401 since I was manually setting an authorization header with
```
esp_http_client_set_header(client, "Authorization", "Bearer <jwttoken>");
```
and auth_header was NULL.


tuanpmt:
@rbino This issue has been fixed and is waiting for sync to the public

Ref: https://github.com/espressif/esp-idf/issues/2246

Thanks

rbino:
Thanks, I'll close the PR and wait for the fix

PR #2293, opened 2018-08-13
Add example for i2c tca9535 I/O port expander (IDFGH-2881)

Gordon01:


xiongyumail:
Hi @Gordon01,
I think `i2c_tca9535` can be put into components.
Just like this directory tree
<https://esp-idf.readthedocs.io/en/latest/api-guides/build-system.html#example-project>

xiongyumail:
In addition, we welcome you to submit application code in esp-iot-solution.
<https://github.com/espressif/esp-iot-solution>
Thanks!

mythbuster5:
Thanke for the contributino, ESP-BSP already has similar stuff.

PR #2346, opened 2018-08-29
gen_esp32part.py changes and fixes (IDFGH-1429)

negativekelvin:
Allow non-zero default subtypes, create non-zero default subtype for type data, fixes

### DIFF
--- a/components/partition_table/gen_esp32part.py
+++ b/components/partition_table/gen_esp32part.py
@@ -57,6 +57,7 @@ SUBTYPES = {
         "esphttpd" : 0x80,
         "fat" : 0x81,
         "spiffs" : 0x82,
+        "default" : 0xFE,
     },
 }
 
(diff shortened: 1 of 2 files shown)


projectgus:
Fixes LGTM. I see the motivation not to default to data subtype 0 (ota_data).

I'm unsure if adding a new 0xFE "unspecified" subtype is the best approach, maybe erroring out here is the best fix? @negativekelvin is there a particular reason you need a default data subtype, or do you just want to avoid setting "ota"?

negativekelvin:
Definitely want to avoid defaulting to ota_data so that leaves either forcing a subtype to be specified or defaulting to a subtype that implies no specific binary format (raw).  In either case a new subtype is required but it doesn't have to be 0xfe.

negativekelvin:
That is fine. I still think it would be useful if a raw binary data subtype was provided.

projectgus:
It looks like this has been in limbo for an extremely long time (I think maybe we each thought the other was going to make the requested changes, and then time passed...)

Both the changes here have been addressed as of 6f6b4c3983298f55b7b86f8224fe6705cc7c8057 :

* If an invalid value is contributed for type or subtype, the tool will now error out with something like "gen_esp32part.InputError: Error at line 5: Value 'invalid_subtype' is not valid. Known keywords: factory, test, ota_0, ota_1, ota_2, ota_3, ota_4, ota_5, ota_6, ota_7, ota_8, ota_9, ota_10, ota_11, ota_12, ota_13, ota_14, ota_15"
* The subtype parsing issue is fixed

So I'm going to close this.

PR #2354, opened 2018-09-01
wpa2_enterprise fixes

negativekelvin:


PaulFreund:
With this commit I can connect to our WPA2 Enterprise network without client certificates (Cisco)

PR #2362, opened 2018-09-03
unity test multiple fixes

nicola-lunghi:


tools/unit-test-app/components/unity/test/test_setup_teardown.c
```diff
@@ -0,0 +1,26 @@
+#include <stdbool.h>
+#include "unity.h"
+
+static bool setup_runned = false;
+static bool teardown_runned = false;
+
+void setUp()
```
igrr: Since these setUp and tearDown are global, does this mean that tests for this component can not be built along with any other component which defines setUp and tearDown? We build our CI tests with `TESTS_ALL=1`, so this would be a problem.

Do you think that file-level (static) setup and teardown functions would be more useful? It would not be too hard to extend TEST_CASE macro to pick them up, if they exist. And at least for the tests we have in IDF, it would allow to reduce code duplication in a few places.

nicola-lunghi: > Since these setUp and tearDown are global, does this mean that tests for this component can not be built along with any other component which defines setUp and tearDown? We build our CI tests with `TESTS_ALL=1`, so this would be a problem.

Ops

> Do you think that file-level (static) setup and teardown functions would be more useful? It would not be too hard to extend TEST_CASE macro to pick them up, if they exist. And at least for the tests we have in IDF, it would allow to reduce code duplication in a few places.

Yes please could be really useful.

nicola-lunghi:
Hi

sorry I miss that the function were global.

PR #2367, opened 2018-09-04
Fixing possible memory leak, not freeing "run_time" prior to returning

mharizanov:


mahavirj:
Thanks for contribution, fix is merged.

Issue #2383, opened 2018-09-06
[TW#26175] wifi: wpa2_enterprise: DES-CBC3-SHA is incorrectly presented as supported cipher suite


mikaelkanstrup:
## Environment

- Development Kit:      [none]
- Kit version (for WroverKit/PicoKit/DevKitC): [-]
- Core (if using chip or module): [ESP32]
- IDF version (``git rev-parse --short HEAD`` to get the commit id.): 599da58
- Development Env:		[Make]
- Operating System:		[Ubuntu]
- Power Supply:         [external 5V]

## Problem Description

When connecting to an wpa2 enterprise network ESP32 claim support for cipher suites that are not really supoorted. During EAP handshake ESP32 presents its list of supported cipher suites in the client hello eap message. Currently the list of (claimed) supported ciphers looks like this:
```
Cipher Suites (7 suites)
    Cipher Suite: TLS_RSA_WITH_AES_256_CBC_SHA256 (0x003d)
    Cipher Suite: TLS_RSA_WITH_AES_256_CBC_SHA (0x0035)
    Cipher Suite: TLS_RSA_WITH_AES_128_CBC_SHA256 (0x003c)
    Cipher Suite: TLS_RSA_WITH_AES_128_CBC_SHA (0x002f)
    Cipher Suite: TLS_RSA_WITH_3DES_EDE_CBC_SHA (0x000a)
    Cipher Suite: TLS_RSA_WITH_RC4_128_SHA (0x0005)
    Cipher Suite: TLS_RSA_WITH_RC4_128_MD5 (0x0004)
```
Connecting to networks using TLS_RSA_WITH_3DES_EDE_CBC_SHA ciphers however does not work. Looking at ESP-IDF sources it appears this non-support is intentional.

3DES initialization is removed at compile time for both HW and SW crypto implementations (function crypto_cipher_init in fast_crypto_internal-cipher.c and crypto_internal-cipher.c). The relevant code is protected by `#ifdef CONFIG_DES3` which is not set. As code won't build if CONFIG_DES3 is set I suppose the non-support is intentional.

Either way to avoid interop problems ESP32 shall not present cipher suites that it cannot handle. 

### Expected Behavior
1. Connect to an enterprise network that only supports  TLS_RSA_WITH_3DES_EDE_CBC_SHA and some weaker ciphers.
2. During EAP handshake ESP32 shall not present TLS_RSA_WITH_3DES_EDE_CBC_SHA as a supported cipher.
[...]

Issue #2381, opened 2018-09-06
[TW#26182] wifi: wpa2_enterprise with AES128-SHA cipher suite fails with HW crypto


mikaelkanstrup:
## Environment

- Development Kit:      [none]
- Kit version (for WroverKit/PicoKit/DevKitC): [-]
- Core (if using chip or module): [ESP32]
- IDF version (``git rev-parse --short HEAD`` to get the commit id.): 599da58
- Development Env:		[Make]
- Operating System:		[Ubuntu]
- Power Supply:         [external 5V]

## Problem Description

Connecting to wpa2_enterprise networks that selects TLS_RSA_WITH_AES_128_CBC_SHA cipher suite does not work when HW crypto/fast_crypto is used. Only software crypto works.

### Expected Behavior
1. Connect to wpa2_enterprise network
2. Notice that network selects TLS_RSA_WITH_AES_128_CBC_SHA cipher suite
3. Successful EAP handshake
4. Successful 4-way handshake
5. Wifi connected

### Actual Behavior
1. Connect to wpa2_enterprise network
2. Notice that network selects TLS_RSA_WITH_AES_128_CBC_SHA cipher suite
3. During EAP handshake server side rejects connection attempt due to decryption failure.
4. Wifi disconnected

### Steps to repropduce

1. Setup a enterprise test network with hostapd
2. Configure network to only allow cipher suite TLS_RSA_WITH_AES_128_CBC_SHA. i.e. hostapd.conf option 'openssl_ciphers=AES128-SHA'
3. Connect to network
4. Notice connect failure

### Code to reproduce this issue

Use the esp-idf wpa2_enterprise example code to connect to test network

## Debug Logs

Relevant ESP32 logs
```
D (605) event: SYSTEM_EVENT_STA_START
V (615) event: enter default callback
V (615) tcpip_adapter: check: local, if=0 fn=0x400e5090
0x400e5090: tcpip_adapter_start_api at esp-idf/components/tcpip_adapter/tcpip_adapter_lwip.c:1082


V (615) tcpip_adapter: call api in lwip: ret=0x0, give sem
V (625) tcpip_adapter: check: remote, if=0 fn=0x400e5090
0x400e5090: tcpip_adapter_start_api at esp-idf/components/tcpip_adapter/tcpip_adapter_lwip.c:1082


V (635) event: exit default callback
I (3045) wifi: n:3 0, o:1 0, ap:255 255, sta:3 0, prof:1
[...]

PR #2386, opened 2018-09-07
more wpa2_enterprise fixes

negativekelvin:
Fixes from https://github.com/espressif/esp-idf/issues/2383 & https://github.com/espressif/esp-idf/issues/2381


igrr:
Thanks @negativekelvin, this was cherry-picked as https://github.com/espressif/esp-idf/commit/171f54bd7ebcd44db76286db232fa9f1a5f1928f with a minor change: we want to keep enum values the same, regardless of this config option. So removed some ifdefs from enum definitions.

PR #2398, opened 2018-09-10
rtc_clk_init fix

negativekelvin:
Actually set cpu frequency

### DIFF
--- a/components/soc/esp32/rtc_clk_init.c
+++ b/components/soc/esp32/rtc_clk_init.c
@@ -119,7 +119,7 @@ void rtc_clk_init(rtc_clk_config_t cfg)
 
     bool res = rtc_clk_cpu_freq_mhz_to_config(cfg.cpu_freq_mhz, &new_config);
     assert(res && "invalid CPU frequency value");
-    rtc_clk_cpu_freq_set_config_fast(&new_config);
+    rtc_clk_cpu_freq_set_config(&new_config);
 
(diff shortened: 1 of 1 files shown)


components/soc/esp32/rtc_clk_init.c
```diff
@@ -119,6 +119,7 @@ void rtc_clk_init(rtc_clk_config_t cfg)
 
     bool res = rtc_clk_cpu_freq_mhz_to_config(cfg.cpu_freq_mhz, &new_config);
     assert(res && "invalid CPU frequency value");
+    rtc_clk_cpu_freq_set_config_fast(&new_config);
```
igrr: Thanks! This should probably be non-fast version, since _fast can't enable PLL, if needed?

negativekelvin: I thought _fast has fallback to full version if required?

igrr: It does, but that relies on static variables which are not set at that point yet

igrr:
@negativekelvin Good point on the possible optimization. To resolve the issue faster, I'll take the current version of the fix, and then confirm about using I2C_READREG_RTC.

igrr:
Squashed and cherry-picked as https://github.com/espressif/esp-idf/commit/8e2856b83d2356600137d68da0b9b0398ce609a1, thanks @negativekelvin!

PR #2404, opened 2018-09-12
bugfix: incorrect divider setting in rtc_clk_cpu_freq_to_config()

shawwwn:
Fix a bug where `out_config->source_freq_mhz` is undefined

`rtc_clk_cpu_freq_set()` now can set cpu to 2MHz

PR #2405, opened 2018-09-13
bugfix: switching from lower xtal freq to 40MHz doesn't have effect

shawwwn:
The if-statement prevents XTAL freq from being set back to 40MHz.
If the CPU runs on a ultra low frequency set by the XTAL divider, say 2MHz, then it won't be able to get back to 40Mhz.

Please advise because I'm not sure if this is a design decision to not allow xtal-driven CPU to switch below 40Mhz.

### DIFF
--- a/components/soc/esp32/rtc_clk.c
+++ b/components/soc/esp32/rtc_clk.c
@@ -641,9 +641,7 @@ void rtc_clk_cpu_freq_set_config(const rtc_cpu_freq_config_t* config)
         rtc_clk_bbpll_disable();
     }
     if (config->source == RTC_CPU_FREQ_SRC_XTAL) {
-        if (config->div > 1) {
             rtc_clk_cpu_freq_to_xtal(config->freq_mhz, config->div);
-        }
     } else if (config->source == RTC_CPU_FREQ_SRC_PLL) {
         rtc_clk_bbpll_enable();
         rtc_clk_wait_for_slow_cycle();


igrr:
You are right, this is a bug in the logic. A few lines above we should be switching to XTAL frequency not only if soc_clk_sel != RTC_CNTL_SOC_CLK_SEL_XTL, but also if running from XTAL but divider is not 1. Reason is, we'd like to get through this code path faster.

Will fix and add a test case.

igrr:
By the way, given this and your other PR, it seems like you are trying to use rtc_clk functions to (?) set CPU frequency to 2 MHz. Please be advised that these functions should not normally be called from the application, as they don't adjust FreeRTOS tick frequency and esp_timer frequency. Also at 2MHz APB frequency REF_TICK of 1MHz can not be achieved, which will cause issues with UART output.

shawwwn:
Thank you for the reply! I will be sure to take that into account.

With #2405 and #2404, I was able to apply a 40X divider to get the CPU frequency down to 1MHz.
UART output seems fine so far.

Will it be possible to reduce the the frequency below 1MHz?

negativekelvin:
> Also at 2MHz APB frequency REF_TICK of 1MHz can not be achieved

Why is this, if cpu clk is derived from xtal then ref tick should also be direct divided xtal clk 40/40=1?

PR #2410, opened 2018-09-14
add_path.sh: cmake: add tools folder to the path

nicola-lunghi:
the ${IDF_PATH}/tools/" is needed for cmake to work. add it to the add_path.sh script


projectgus:
Thanks @nicola-lunghi ! LGTM

PR #2428, opened 2018-09-18
Downgrade python-future required version to 0.15

rojer:
It works fine and this way python-future from Ubuntu Bionic can be used.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ setuptools
 # Please keep it as the first item of this list.
 #
 pyserial>=3.0
-future>=0.16.0
+future>=0.15.2


dobairoland:
Hi @rojer. Thanks for the PR. I was thinking about the downgrade but in my opinion it is not a good way to go. Probably we don't need to have anything special from 0.16 but it is hard to tell that whether "it would work fine" or not because there are Python scripts inside ESP-IDF which are not run by users/developers directly. 0.15 is three years old and 0.16 is two. I assume that the opposite (upgrade) will be necessary in the near future because of new versions of Python (3.7, 3.8) and deprecated older versions, like 2.6 (just see the recent changelog of the python-future repository). I emphasize that this is my own personal opinion.

If package provided by your OS is not sufficient then you still can do `pip install --user -r $IDF_PATH/requirements.txt`. You can even setup a virtual python environment and use that with ESP-IDF.

rojer:
i know about PIP but it's always preferable to use distro's package manager if possible.
all other dependencies can be satisfied with package from the main repository, and this seemed like a small enough change, so i made it and seems to work ok in my day to day use.
i don't claim it to be comprehensive testing though. a point can also be made that 0.15.2 is good enough for such pyhton-heavy distro like ubuntu (they use python a lot for their stuff), so it should be ok.
do you have specific requirements that 0.16 satisfies and 0.15 does not?

dobairoland:
Allright @rojer. After taking another look at the changelog and diff I agree with the downgrade. But please make it >=0.15.2. That version will be satisfied even by the Ubuntu 16.04 LTS package. Thanks!

PR #2429, opened 2018-09-18
Remove trailing spaces from parttool.py output

rojer:


components/partition_table/parttool.py
```diff
@@ -107,7 +107,7 @@ def main():
         return 1  # nothing found
 
     if args.offset:
-        print('0x%x ' % (found_partition.offset))
+        print('0x%x' % (found_partition.offset))
```
igrr: This will result in space missing if --offset and --size are both specified. How about something like
```
results = []
if args.offset:
    results.append('0x%x' % (found_partition.offset))
if args.size:
    results.append('0x%x' % (found_partition.size))
print(' '.join(results))
```

components/partition_table/Makefile.projbuild
```diff
@@ -65,8 +65,8 @@ all_binaries: $(PARTITION_TABLE_BIN) partition_table_get_info
 partition_table_get_info: $(PARTITION_TABLE_BIN)
 	$(eval PHY_DATA_OFFSET:=$(shell $(GET_PART_INFO) --type data --subtype phy --offset $(PARTITION_TABLE_BIN)))
 	$(eval APP_OFFSET:=$(shell $(GET_PART_INFO) --default-boot-partition --offset $(PARTITION_TABLE_BIN)))
-	$(eval OTA_DATA_SIZE := $(shell $(GET_PART_INFO) --type data --subtype ota --size $(PARTITION_TABLE_BIN) || echo 0) )
-	$(eval OTA_DATA_OFFSET := $(shell $(GET_PART_INFO) --type data --subtype ota --offset $(PARTITION_TABLE_BIN)) )
+	$(eval OTA_DATA_SIZE := $(shell $(GET_PART_INFO) --type data --subtype ota --size $(PARTITION_TABLE_BIN) || echo 0))
+	$(eval OTA_DATA_OFFSET := $(shell $(GET_PART_INFO) --type data --subtype ota --offset $(PARTITION_TABLE_BIN)))
```
igrr: just out of curiosity, where did this issue show up?

components/partition_table/parttool.py
rojer: `print` adds a newline, so it works in that case too

PR #2446, opened 2018-09-20
cmake -DCMAKE_${X}_COMPILER problem fixed (IDFGH-2877)

p5-vbnekit:
- bootloader external project: CMAKE_[C+CXX+ASM]_COMPILER parameters forwarding enabled;
- esp32 toolchain: "isset" check before "set" enabled.

If we run cmake with the "-DCMAKE_ASM_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc -DCMAKE_CXX_COMPILER:STRING=/path/to/xtensa-esp32-elf-g++ -DCMAKE_C_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc" instead of changing PATH, then we'll get the following errors:

> The ASM compiler identification is unknown
> Found assembler: xtensa-esp32-elf-gcc
> The C compiler identification is unknown
> The CXX compiler identification is unknown
> CMake Error at /home/nikita/work/esp/esp-idf/tools/cmake/project.cmake:99 (_project):
>   The CMAKE_ASM_COMPILER:
>     xtensa-esp32-elf-gcc
>   is not a full path and was not found in the PATH.
>   Tell CMake where to find the compiler by setting either the environment
>   variable "ASM" or the CMake cache entry CMAKE_ASM_COMPILER to the full path
>   to the compiler, or to the compiler name if it is in the PATH.
> Call Stack (most recent call first):
>   CMakeLists.txt:6 (project)
> Warning: Did not find file Compiler/-ASM
>
> CMake Error at /home/nikita/work/esp/esp-idf/tools/cmake/project.cmake:99 (_project):
>   The CMAKE_C_COMPILER:
> ... blah blah blah ...

It happens because:
 - tools/cmake/toolchain-esp32.cmake override user definitions;
 - bootloader external project executor doesn't forward theese options.

This commit fixes it.


projectgus:
Hi @p5-vbnekit ,

Thanks for bringing this up. I'm a little curious what the use cases are for specifying the toolchain on the command line like this.

Because of [CMake's cross compiling behaviour](https://cmake.org/cmake/help/v3.10/manual/cmake-toolchains.7.html#cross-compiling), for future compatibility it may be preferable to support passing `CMAKE_TOOLCHAIN_FILE=` on the command line to load a custom toolchain file, rather than overriding parts of the toolchain file. Do you think this could work for your use case?

(Ping @igrr as he may have some thoughts about the best approach here, as well.)

p5-vbnekit:
> Hi @p5-vbnekit ,
> 
> Thanks for bringing this up. I'm a little curious what the use cases are for specifying the toolchain on the command line like this.
> 
> Because of [CMake's cross compiling behaviour](https://cmake.org/cmake/help/v3.10/manual/cmake-toolchains.7.html#cross-compiling), for future compatibility it may be preferable to support passing `CMAKE_TOOLCHAIN_FILE=` on the command line to load a custom toolchain file, rather than overriding parts of the toolchain file. Do you think this could work for your use case?
> 
> (Ping @igrr as he may have some thoughts about the best approach here, as well.)

Hi! Thanks for reply. Yes it is possible, but in my case it will require more actions for configuring build profile...
In my case i use QTCreator (or "dockered" jenkins for automatic builds) & standard esp32 toolchain but non-standard toolchain path. QTCreator overrides theese cmake variables. Overriding CMAKE_${X}_COMPILER is thinnest and compatible way to build automation (in my opinion).
Reasons:
- not need to modify system environment
- not need for additional files
- it is legal cmake way too: https://gitlab.kitware.com/cmake/community/wikis/FAQ#how-do-i-use-a-different-compiler.

May be add both modes?

renzbagaporo:
Hi @p5-vbnekit, it seems I am unable to reproduce the issue that prompted you to submit this patch:

```
If we run cmake with the "-DCMAKE_ASM_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc -DCMAKE_CXX_COMPILER:STRING=/path/to/xtensa-esp32-elf-g++ -DCMAKE_C_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc" instead of changing PATH, then we'll get the following errors:
```
What I did is made a copy of the toolchain somewhere in my system not in PATH, and executed `idf.py` and a bare CMake build with `-DCMAKE_ASM_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc -DCMAKE_CXX_COMPILER:STRING=/path/to/xtensa-esp32-elf-g++ -DCMAKE_C_COMPILER:STRING=/path/to/xtensa-esp32-elf-gcc` passed. The app and bootloader build used the override toolchain binaries.

Is there something missing from my attempt to reproduce the issue?

renzbagaporo:
Deleted my previous reply since I was able to reproduce the issue you mention shortly thereafter.

About the patch, I'm not sure, since these would mean the patch would also be required for every other official toolchain-file ESP-IDF will deliver.

It doesn't seem to be standard practice for toolchain files as well: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html.

PR #2456, opened 2018-09-22
Redundant code removed / mux protection added (IDFGH-1795)

dirkx:
Removed some register sets that where already set in functions; with the added benefit that the latter has a mux protecting the register.


Alvin1Zhang:
@dirkx Thanks for the contribution and sorry for the slow turnaround. We will look into this PR. Thanks.

koobest:
Hi, 
thanks for you PR, the driver refactoring has solved this issue.

Alvin1Zhang:
@dirkx Thanks for your contribution, will close the PR now. Thanks.

PR #2458, opened 2018-09-22
Remove Content-Length from chunked encoding

shmuelzon:
Content-Length header should not be included when using chunked encoding.
Some browsers will close the socket after seeing the 0 sized response instead of
reading the chunks.

Safari on iOS will not display pages sent with chunked encoding and Wireshark also marks the HTTP response as malformed.


anurag-kar:
@shmuelzon Thanks for the PR. It's been merged

PR #2509, opened 2018-10-03
add: esp_log_vwrite(), supports va_list parameter forwarding (IDFGH-1794)

neta79:
Rework esp_log_write() to introduce esp_log_vwrite(), a variant that accepts va_list parameters.

Trivial, but important to have in case other logging frameworks are already in place in the codebase; they can easily funnel data to esp_log_vwrite(), and use esp_log as a sink.

### DIFF
--- a/components/log/include/esp_log.h
+++ b/components/log/include/esp_log.h
@@ -106,6 +106,17 @@ uint32_t esp_log_early_timestamp(void);
  */
 void esp_log_write(esp_log_level_t level, const char* tag, const char* format, ...) __attribute__ ((format (printf, 3, 4)));
 
+/**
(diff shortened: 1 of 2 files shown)


MightyPork:
this would be very useful to have...

Alvin1Zhang:
@neta79 Thanks for the contribution and sorry for the very slow turnaround. We will look into this PR. Thanks.

igrr:
Sorry that we didn't update the PR in time. These changes with some tweaks got rolled into a commit where we have optimized the Wi-Fi driver code size: 7569e34e8989ca38564cae17e7249221856cfa18. We should have kept this commit separate. I apologize for the sloppy handling of your contribution.

PR #2534, opened 2018-10-09
Fix infinite loop on esp_http_client_fetch_headers

R-Damian:
error must also be returned if esp_tls_conn_read return 0 because socket  was closed


mahavirj:
Thanks for your contribution, change is now merged.

PR #2558, opened 2018-10-13
CMake Auto App Signing Support (IDFGH-1792)

Snipesy:
This fixes the previous issue with partition table signing (was making file not found errors). It also adds support for automatic signing of the app.bin using the key specified from menuconfig.

## Steps to test

1. Run make menuconfig
2. Enable 'Require signed app images'
3. Enable 'Sign binaries during builds'
4. Set signing key to yours.
5. Attempt a build using cmake.
6. verify signature is present at end of partiton table and binary.

A somewhat related unsolved issue is enabling bootloader checks causes an error.

Example console outputs...
[944/945] Generating UNSIGNED_proj.bin
esptool.py v2.5.1
[945/945] Generating proj.bin
espsecure.py v2.5.1
Signed 1008624 bytes of data from UNSIGNED_proj.bin with key /Users/snipesy/CLionProjects/proj/secure_boot_signing_key.pem`

### DIFF
--- a/components/esptool_py/project_include.cmake
+++ b/components/esptool_py/project_include.cmake
@@ -28,18 +28,15 @@ if(CONFIG_ESP32_PHY_INIT_DATA_IN_PARTITION)
     set(PHY_PARTITION_BIN_FILE "esp32/phy_init_data.bin")
 endif()
 
-
-
 #
-#  App Signing
+#  Checks if app signing is enabled. Signs app binary if enabled.
+#  Otherwise converts from .elf directly with no signing.
 #
 if(CONFIG_SECURE_BOOT_BUILD_SIGNED_BINARIES AND NOT BOOTLOADER_BUILD)
 
-    #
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@Snipesy Thanks for the contribution and sorry for the very slow turnaround. We will look into this PR. Thanks.

projectgus:
Hi @Snipesy ,

Sorry again for all the delays on this, it's very unreasonable for us to make you wait this long for a proper response.

ESP-IDF v4.0 and newer should now have all the CMake support for Secure Boot, at the time you wrote this PR some critical pieces were missing. We never officially supported Secure Boot and CMake in the v3.x previews, and in recent versions it's actually disable in the project configuration menu.

I can't work out why the CLA bot says you haven't signed off on the CLA license - it seems like the first commits are passing but the second commit isn't ,and all three commits have the same author. I haven't seen this before, maybe a bug in the CLA bot so doubly sorry about that.

Angus

PR #2577, opened 2018-10-16
Fix component interdependencies, faster build time, smaller binary, fix #2571 (IDFGH-1793)

### LINKED ISSUE #2571: Limit interdependencies (IDFGH-410)
In order to get faster build time, and smaller binaries, I've spot many changes to the core file that could save this.

First, the minimal component list (found in your main Makefile) I'm trying to build is:
```
COMPONENTS=freertos esp32 soc heap newlib esptool_py nvs_flash spi_flash log tcpip_adapter lwip main xtensa-debug-module driver bt vfs micro-ecc nghttp wpa_supplicant smartconfig_ack app_trace mbedtls libesphttpd esp_http_client esp-tls pthread cxx
```

Doing so, leads to compilation errors in those files:
```
esp-idf/components/esp32/clk.c:34
  The include for bootloader_clock.h is not required (commenting it solves the error)

esp-idf/components/esp32/cpu_start.c:66
  The include for esp_efuse.h is not required (commenting it solves the error)

esp-idf/components/esp32/event_default_handlers.c:25
  Ethernet should be conditional
(to be continued)
```

Those are inter-module dependencies. Looking at the compilation line, there is no define whatsoever that tell one module what are the other module configuration (or presence). There isn't any shared header either (like sdkconfig.h or something).
I can add `CXXFLAGS := -DSOMETHING` and `CFLAGS:= -DSOMETHING` in my makefile, but it would be a PITA to maintain.

Do you think it would be possible to add in the general makefile rules that each component add a `-D_componentname` to the `CFLAGS` and `CXXFLAGS` to all modules ?
If so, it would be possible to remove/simplify inter component dependencies in the code.

X-Ryl669:
This fixes #2571.
Basically, it's a 2 step process. First, the project_common.mk is modified to add a vector of `-D_DECL_component_name` to the C and CXX flags, so each component is aware of the other selected components in the main makefile.
Typically, this does not change anything for default build, but make build considerably faster and produce a smaller binary when you specify only the components you need in the Makefile.
In my case, in minimalist build, I have:
```
COMPONENTS=freertos esp32 soc heap newlib esptool_py nvs_flash spi_flash log tcpip_adapter lwip main xtensa-debug-module driver bt vfs micro-ecc nghttp wpa_supplicant smartconfig_ack app_trace mbedtls libesphttpd esp_http_client esp-tls pthread cxx
 ```

Then, I've fixed all the dependencies that were cross-linked between components so we don't build useless stuff. 
I've added a I2C_SLAVE_MODE option in Kconfig of driver component to disable I2C slave code.
I've also added a RMT_RECEIVE option to disable RMT receive.

Here are my results:
# Without specifying any components in main makefile (usual default):
```
Complete (no component specifications)
Bin size: 591168
Build time: 4m52
```

# Without app_update, ethernet, bootloader_support
```
Bin size: 579872
Build time: 3m12
```

# Without app_update, bootloader_support
```
Bin size: 580176
Build time: 3m22
```

# Without ethernet, I2C slave, RMT receive
```
Bin size: 585632
Build time: 3m30
```

Please notice that binary size saving is usually 3x larger since you've a factory plus 2 OTA partitions (so 3 times the binary size). So in the end, the change gives 33kB of flash saving.

### DIFF
--- a/components/bt/host/bluedroid/btc/profile/std/spp/btc_spp.c
+++ b/components/bt/host/bluedroid/btc/profile/std/spp/btc_spp.c
@@ -22,7 +22,6 @@
 #include "osi/allocator.h"
 #include "esp_spp_api.h"
 #include "osi/list.h"
-#include "freertos/ringbuf.h"
 #include "osi/mutex.h"
 #include <sys/errno.h>
 #include <sys/lock.h>
@@ -31,6 +30,10 @@
 #include "esp_vfs_dev.h"
 
 #if (defined BTC_SPP_INCLUDED && BTC_SPP_INCLUDED == TRUE)
+#ifndef _DECL_esp_ringbuf
+#error "BTC_SPP_INCLUDED requires esp_ringbuf component"
+#endif
+#include "freertos/ringbuf.h"
 
 typedef struct {
     uint8_t serial;
--- a/components/driver/Kconfig
+++ b/components/driver/Kconfig
@@ -21,6 +21,27 @@ menu "Driver configurations"
 
     endmenu  # ADC Configuration
 
+    menu "I2C configuration"
+
+        config I2C_SLAVE_MODE
+            bool "Enable I2C slave mode"
+            default y
+            help
+                If this is set, the I2C driver will include support for being used as a slave I2C device.
+
+    endmenu # I2C Configuration
+
+    menu "RMT configuration"
+
+        config RMT_RECEIVE
+            bool "Enable RMT receive code"
+            default y
+            help
+                If this is set, the RMT driver will include support for receiving data.
+
+    endmenu # RMT Configuration
+
+
     menu "SPI configuration"
 
         config SPI_MASTER_IN_IRAM
--- a/components/driver/i2c.c
+++ b/components/driver/i2c.c
@@ -402,7 +402,6 @@ esp_err_t i2c_driver_delete(i2c_port_t i2c_num)
         p_i2c->tx_ring_buf = NULL;
         p_i2c->tx_buf_length = 0;
     }
-(??)
 #endif
 #if CONFIG_SPIRAM_USE_MALLOC
     if (p_i2c_obj[i2c_num]->evt_queue_storage) {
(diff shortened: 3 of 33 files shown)


Alvin1Zhang:
@X-Ryl669 Thanks for the contribution and sorry for the very slow turnaround. We will look into this PR. Thanks.

X-Ryl669:
If you intend to merge it, I can try to fix the conflict first. I made other changes to reduce the binary size by defining out any unwanted piece, but it's not included here.

X-Ryl669:
That was a real pain to update, the repository does not build out of the box on MacOS. Anyway, please merge the other PR (#4020 #4021 #4019) first as without those, you can't even build the basic components.

Please notice that updating to the lastest version increased the binary size by 6kB! (without this patch, just same code).

I guess some diet is required first... Anyway, this patch goes into that direction.

Alvin1Zhang:
@X-Ryl669 Thanks for the patience and the contribution. We will look into the PR. Thanks.

projectgus:
> That was a real pain to update, the repository does not build out of the box on MacOS. Anyway, please merge the other PR (#4020 #4021 #4019) first as without those, you can't even build the basic components.

Hi @X-Ryl669 ,

Thanks for the updates and I'm genuinely sorry noone has gotten back to you for so long.

This is a really impressive achievement, I would not have thought this was possible to get such a good result so I'm overall very impressed.

Unfortunately around the same time you would have started working on this, we made the decision to move away from GNU Make and towards the CMake-based build system, and we've tried to incorporate some similar concepts in the new build system (CMake components declare their dependencies, the build system can expand a minimal build from the project's list of initial dependencies, etc).

GNU Make build system has been labelled "Legacy" for V4.0 and although we still support it, we're keen to keep it more or less as-is and do future work on the CMake side.

For these reasons we won't be merging this, I apologise for the extra effort you've put into updating it recently if that was for our benefit. 

We will of course look into merging the other smaller bugfix PRs, really appreciate you submitting those as well.

If you do look into moving to the CMake build system and you have any suggestions for what can/should be simplified on that side to make it easier to have a faster build and smaller final binary size, please let us know. There are still some additional features pending from our side, such as the ability to have "soft" dependencies - ie if component Y is included in the build then expose its headers to component X, otherwise don't.

projectgus:
Closing due to reasons shown above. Kudos again for this work @X-Ryl669 , it's not something I would have thought possible!

PR #2578, opened 2018-10-17
SSL for HTTP server

MightyPork:
This PR adds HTTPS capability to the HTTP server component. It is implemented using the already existing API, with some necessary tweaks. It also fixes a few bugs and issues I encountered, see below.

HTTP server changes:
- add global user context (void *) in http_server config struct + getter routine
- modify send/recv function signature to include server handle
- add pending check function (overridable per socket)
- remove mistaken trailing semicolon in HTTPD_DEFAULT_CONFIG()
- add configurable "socket open handler" that can set up session context
- add support for length = -1 in httpd_resp_send() -> use strlen()
- add function to set session context by FD
- add functions to set session send/recv/pending overrides by FD
- FIX assert() if select fails - recovert by closing invalid sockets
- Change shutdown waiting loop to use only 100ms wait time instead of 1s
- make httpd_sess_get() non-static, it is now needed outside httpd_sess.c

HTTPS server component
- new component that initializes the HTTP server with TLS support, taking
  advantage of the new APIs and tweaks
- based on the OpenSSL server example
- added example


mahavirj:
@MightyPork 

Thanks for your contribution. This PR involves API change, we are discussing ways in which that can be avoided at this moment (for getting web server handle, maybe we can use thread local storage instead). Will keep you posted on this. 

Also just to give you heads up, `component` and `header` names for `http_server` component being too generic, resulting in trouble for some other users, so change to add `esp` prefix is already in our pieline. This will require rebase on this PR as well. Will let you know once we have this change integrated.

components/http_server/include/http_server.h
```diff
@@ -70,6 +70,69 @@ typedef void* httpd_handle_t;
...
+ * @param[in] hd      : server instance
+ * @param[in] sockfd  : session socket file descriptor
+ * @param[in] buf     : buffer with bytes to send
+ * @param[in] buf_len : data size
+ * @param[in] flags   : flags for the send() function
+ * @return
+ *  - Bytes : The number of bytes sent successfully
+ *  - -VE   : In case of error
+ */
+typedef int (*httpd_send_func_t)(httpd_handle_t hd, int sockfd, const char *buf, size_t buf_len, int flags);
```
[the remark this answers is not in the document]
MightyPork: this can be done, but it seems needlessly convoluted to me. Is there a reason to worry about API changes, if the http_server component is still unreleased in a stable version? (this is why I'm using master instead of the stable release)

while we're discussing it, what was the original purpose of the send / recv override functions? Because if they are used for SSL handling, they become unavailable for anything else.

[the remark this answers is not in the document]
anurag-kar: > Is there a reason to worry about API changes, if the http_server component is still unreleased in a stable version?

Your are right. But my only concern is that people might have already started using the API. Still let me discuss this and get back to you.

> while we're discussing it, what was the original purpose of the send / recv override functions?

The purpose was to allow for extension/encryption/encapsulation of HTTP packets as per requirement, just like you are using it for SSL

> Because if they are used for SSL handling, they become unavailable for anything else.

That is alright. They are intended to allow one layer of extension only. Now in case someone requires something fancy with SSL as well as some form of encapsulation then she will have to start implementing on top of http_server.

components/https_server/src/https_server.c
```diff
@@ -0,0 +1,203 @@
...
+ * @param server
+ * @param sockfd
+ * @param buf
+ * @param buf_len
+ * @param flags
+ * @return bytes read, negative on error
+ */
+static int httpdssl_recv(httpd_handle_t server, int sockfd, char *buf, size_t buf_len, int flags)
+{
+    SSL *ssl = httpd_sess_get_ctx(server, sockfd);
```
anurag-kar: Session context is passed along with the httpd_req_t structure when URI handler is executed. It's useful for session persistence (See the `persistent_sockets` example).

Could you use a different member variable (and corresponding APIs) for saving and accessing the pointer to SSL object. Maybe call it `priv_ctx` or `internal_ctx`?

MightyPork: sure. in that case, do you want to keep the functions `httpd_sess_get_ctx` / `httpd_sess_set_ctx` that I added for this purpose as well as adding new ones for this 'internal_ctx'?

anurag-kar: You can keep them. They may come handy sometime.

components/http_server/include/http_server.h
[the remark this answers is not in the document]
anurag-kar: @MightyPork After some discussion we have concluded that changing the API should be alright in this case, like you said

anurag-kar:
There seem to be conflicts in
```
components/http_server/include/http_server.h
components/http_server/src/httpd_txrx.c
```
Could you please resolve them?

MightyPork:
> There seem to be conflicts in
> 
> ```
> components/http_server/include/http_server.h
> components/http_server/src/httpd_txrx.c
> ```
> 
> Could you please resolve them?

sure, I plan to have a look at it tonight (GMT+1) and fix this + the other issues

components/http_server/include/http_server.h
```diff
@@ -180,11 +229,6 @@ esp_err_t httpd_stop(httpd_handle_t handle);
  * @{
  */
 
-/**
- * @brief Function type for freeing context data (if any)
- */
-typedef void (*httpd_free_sess_ctx_fn_t)(void *sess_ctx);
-
```
MightyPork: this was moved up, because it is now used in the init struct as well

components/http_server/include/http_server.h
```diff
@@ -360,13 +404,18 @@ esp_err_t httpd_unregister_uri(httpd_handle_t handle, const char* uri);
...
+ * @param[in] buf     : buffer with bytes to send
+ * @param[in] buf_len : data size
+ * @param[in] flags   : flags for the send() function
  * @return
  *  - Bytes : The number of bytes sent successfully
  *  - HTTPD_SOCK_ERR_INVALID  : Invalid arguments
  *  - HTTPD_SOCK_ERR_TIMEOUT  : Timeout/interrupted while calling socket send()
  *  - HTTPD_SOCK_ERR_FAIL     : Unrecoverable error while calling socket send()
  */
-typedef int (*httpd_send_func_t)(int sockfd, const char *buf, size_t buf_len, int flags);
```
MightyPork: API change of these functions approved by anurag-kar

components/http_server/include/http_server.h
```diff
@@ -435,6 +507,64 @@ esp_err_t httpd_set_recv_override(httpd_req_t *r, httpd_recv_func_t recv_func);
...
+ * @see httpd_set_pending_override()
+ *
+ * @param[in] hd  HTTPD instance handle
+ * @param[in] fd  session socket FD
+ * @param[in] pending_func The receive function to be set for this session
+ *
+ * @return status code
+ */
+esp_err_t httpd_set_sess_pending_override(httpd_handle_t hd, int sockfd, httpd_pending_func_t pending_func);
+
```
MightyPork: new setters were needed to allow using them in the open_fn (where no request exists)

components/http_server/src/httpd_priv.h
```diff
@@ -169,6 +172,23 @@ struct httpd_data {
...
+struct sock_db *httpd_sess_get(struct httpd_data *hd, int sockfd);
+
+/**
+ * @brief Delete sessions whose FDs have became invalid.
+ *        This is a recovery strategy e.g. after select() fails.
+ *
+ * @param[in] hd    Server instance data
+ */
+void httpd_sess_delete_invalid(struct httpd_data *hd);
+
```
MightyPork: helper func for cleaning up invalid sockets (called from the main server loop)

components/http_server/src/httpd_sess.c
```diff
@@ -61,6 +61,12 @@ esp_err_t httpd_sess_new(struct httpd_data *hd, int newfd)
             hd->hd_sd[i].handle = (httpd_handle_t) hd;
             hd->hd_sd[i].send_fn = httpd_default_send;
             hd->hd_sd[i].recv_fn = httpd_default_recv;
+
+            /* Call user-defined session opening function */
+            if (hd->config.open_fn) {
+                esp_err_t ret = hd->config.open_fn(hd, hd->hd_sd[i].fd);
+                if (ret != ESP_OK) return ret;
+            }
```
MightyPork: calling 'open' so we can set up the SSL session

components/http_server/src/httpd_sess.c
```diff
@@ -98,6 +147,21 @@ void httpd_sess_set_descriptors(struct httpd_data *hd,
...
+
+void httpd_sess_delete_invalid(struct httpd_data *hd)
+{
+    for (int i = 0; i < hd->config.max_open_sockets; i++) {
+        if (hd->hd_sd[i].fd != -1 && !fd_is_valid(hd->hd_sd[i].fd)) {
+            httpd_sess_delete(hd, hd->hd_sd[i].fd);
+        }
+    }
+}
+
```
MightyPork: this is the invalid sockets fix

MightyPork:
@anurag-kar can you please review the changes? it's rebased and should be ready to merge
I added some github comments to the code, but it made quite a mess here so i closed them again lol

it now uses 'transport_ctx' instead of 'ctx'

anurag-kar:
@MightyPork I'm in the process of reviewing, but because I am supposed to travel for the upcoming 2-3 days, so it may take some time.

I see the github comments you've put and I find them helpful.

`transport_ctx` is an appropriate name actually.

There was another `http_server` related commit that got merged which changes the component path and name to `esp_http_server`. So you will have to resolve the conflicts again. Please bear with us.

We are looking forward to have your pull request so I'll hold of any more changes to `esp_http_server` from our side till this is merged.

Also, could you add a `close_fn` callback (just like the `open_fn`) which gets invoked when a session/socket is closed. This may not be useful for HTTPS but there are some of our applications where having this will help.

MightyPork:
All good, the bug is fixed, close_fn added and I also added license comments. from my side I'm done with this feature, hope it's good like this

components/esp_http_server/include/esp_http_server.h
```diff
@@ -39,7 +39,7 @@ extern "C" {
         .lru_purge_enable   = false,                    \
         .recv_wait_timeout  = 5,                        \
         .send_wait_timeout  = 5,                        \
-};
+}
```
anurag-kar: Would the global context and function pointers need to be set to NULL under HTTPD_DEFAULT_CONFIG?

components/esp_http_server/include/esp_http_server.h
```diff
@@ -70,6 +70,35 @@ typedef void* httpd_handle_t;
  */
 typedef enum http_method httpd_method_t;
 
+/**
+ * @brief  Prototype for freeing context data (if any)
+ * @param[in] ctx : object to free
+ */
+typedef void (*httpd_free_sess_ctx_fn_t)(void *ctx);
```
anurag-kar: Since we have agreed to modify the APIs, I guess we can also rename this to `httpd_free_ctx_fn_t` since this is being used for both global and session contexts

components/esp_http_server/include/esp_http_server.h
MightyPork: not needed, when you initialize a struct with {}, the unspecified members are set to zero automatically

components/esp_http_server/include/esp_http_server.h
MightyPork: Sure, but it'll take about 12 hours before I can do this. If you want, please feel free to rename it yourself

components/esp_http_server/include/esp_http_server.h
igrr: Please note though that if this header is used from a C++ source file, compiler will complain about the missing initializers, or if the order of initializers does not match the order of structure fields. So it is still desirable to initialize all the fields.

anurag-kar:
@MightyPork Except the small changes suggested, this looks really good to me. Thanks for putting all the effort in making this feature possible.

examples/protocols/https_server/main/certs/gencert.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+openssl req -newkey rsa:2048 -nodes -keyout prvtkey.pem -x509 -days 3650 -out cacert.pem -subj "/CN=ESP32 HTTPS server example"
```
anurag-kar: Could you instead put this inside README.md

MightyPork: I can add it there, but why is this a problem? 
Here you just run the script and it generates a new certificate for the example

anurag-kar: If you wish to keep the executable

* either you need to remove the executable flag (and thereby add an extra step in README to enable the executable flag)
* or you would have to modify `esp-idf/tools/ci/executable-list.txt` (in which you would find no example related scripts till now)

Not doing either of the above will cause our CI tests to fail.

This is asserted because we wish to keep the number of executable in IDF to a minimum (only those which are essential, like scripts related to build system)

MightyPork: I see. I'd favor removing the exacutable flag, but it's ultimately up to you, what matters is to keep the example command.

The OpenSSL server example (https://github.com/espressif/esp-idf/tree/master/examples/protocols/openssl_server), that I used to learn about openSSL, references a document called "ESP8266_SDKSSL_User_Manual_EN_v1.4.pdf". That document doesn't exist, I didn't find anything at all on Google. The linked page has a file called "5a-esp8266_sdk_ssl_user_manual_en.pdf", but I didn't really understand it.

so it's good to keep it for reference. Also it's very important to make generating the certificate easy, or people might blindly reuse the example certificate.

anurag-kar: > it's good to keep it for reference. Also it's very important to make generating the certificate easy, or people might blindly reuse the example certificate.

I agree. This is a minor thing. What matters is that it should be convenient for the user.

Then you can keep the file with executable flag removed. If every one is okay with this during the internal review then it should be alright.

anurag-kar:
One more thing, please squash your commits into two. One commit for changes to esp_http_server. Second commit adds new esp_https_server component along with examples. This way it looks clean during merge.

components/esp_http_server/include/esp_http_server.h
MightyPork: initializers added

anurag-kar:
@MightyPork Thank you for cooperating. Will update you when it gets merged internally

X-Ryl669:
Just a question, how if this supposed to working in practice ? 
I mean, now the industry is moving to short lived certificates, you'll have a lot of trouble getting a certificate that's valid for, let say, 5 years (even 2 years will be hard in the future). If you make your own certificate, the browser will never accept it and it'll be a pain for user to use such server (they'll have a "security warning" most don't understand, and it's hard to explain to them to add an exception for your certificate). In the end, I guess the only workable way is to have the system somehow, download its certificate from some authority, or have a certbot on the ESP32, but this means always being connected to Internet.

MightyPork:
The main use case for me is to make the node securely accessible over the
network, for example to adjust it's configuration, check it works correctly
etc. I never meant this to be used to run a public server, the esp doesn't
have the horsepower for that anyway. Hence there is no problem for me with
a self signed certificate or one lasting e.g. 10 years. You could, I guess,
set up a dummy CA and preload the root into your browser, if you're
obsessed with having a "green lock". for the record, you could update the
certificate via OTA or in other way, all it takes is to restart the server.

MightyPork:
I don't want to sound impatient, but what takes so long about accepting an approved change request?

mahavirj:
@MightyPork 
> I don't want to sound impatient, but what takes so long about accepting an approved merge request?

Apologies regarding timelines. Approved PR goes through our internal review and testing (functional/integration), so its taking bit longer, in this specific case we also observed few areas for code cleanup. This will reflect in public repository in next few days.

mahavirj:
Thanks for contribution. Merged with https://github.com/espressif/esp-idf/commit/61ee1bd31ff9ad9f3fb62e01e11d5402c728430e

MightyPork:
Thanks for merging and the update!

one mistake i found - https://github.com/espressif/esp-idf/blob/master/docs/en/api-reference/protocols/esp_https_server.rst
this file must be updated to reflect that you renamed all the functions from xxx_sess to sess_xxx (and some of the functions were removed)

(the code changes were done in https://github.com/espressif/esp-idf/commit/9a9d18e466c0221dc048970efeef6c68fa5983a9)

mahavirj:
@MightyPork You are right, we will fix in followup commit.

PR #2645, opened 2018-10-31
Fix broken image tags

todduk:
Three image references corrected


Isl2017:
@todduk , Thanks very much. There is already an MR in Gitlab now which has much modified about the doc and includes the images issues. We will merge it soon.

PR #2650, opened 2018-10-31
Keep socket creation/bind/listen outside the loop (IDFGH-1790)

LoloVTS:


Alvin1Zhang:
@LoloVTS Thanks for the contribution and sorry for the very slow turnaround. We will look into this PR. Thanks.

igrr:
It seems that this improvement has already been implemented in b7bfcf3c936ac8eeb3f130abd1dc69436dfeae53.

PR #2651, opened 2018-10-31
Fixed error in return type of xTaskCreateStatic* (IDFGH-2185)

chanibal:
The `xTaskCreateStatic` and `xTaskCreateStaticPinnedToCore` claimed that they return `pdPASS` or an error value depending on their success, but what they really return is an `TaskHandle_t`.

This error is also in official FreeRTOS.


Dazza0:
@chanibal closed in c7dfb435f8ba3594848f130fe3d4eea20946f3b4. Thank you for your contribution.

PR #2654, opened 2018-11-01
Add esp_tls_init_global_ca_store function to esp-tls

paulreimer:
called from `esp_tls_set_global_ca_store`. (if `global_cacert` has already been initialized, re-initialize it -- same behaviour as before).

Having `esp_tls_init_global_ca_store` function allows allocating the mbedTLS certificate chain, but to manage it directly with mbedTLS APIs (in my use case, I only have der format certs, but I still need the cert chain initialized first, before adding them).

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -141,6 +141,19 @@ err_freeaddr:
     return ret;
 }
 
+esp_err_t esp_tls_init_global_ca_store()
+{
+    if (global_cacert == NULL) {
+        global_cacert = (mbedtls_x509_crt *)calloc(1, sizeof(mbedtls_x509_crt));
+        if (global_cacert == NULL) {
+            ESP_LOGE(TAG, "global_cacert not allocated");
+            return ESP_ERR_NO_MEM;
+        }
+        mbedtls_x509_crt_init(global_cacert);
+    }
+    return ESP_OK;
+}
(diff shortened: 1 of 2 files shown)


components/esp-tls/esp_tls.c
```diff
@@ -150,13 +163,11 @@ esp_err_t esp_tls_set_global_ca_store(const unsigned char *cacert_pem_buf, const
     if (global_cacert != NULL) {
         mbedtls_x509_crt_free(global_cacert);
     }
```
ChiragAtal: @paulreimer Could you move this `global_cacert` check and the corresponding free inside the new `esp_tls_init_global_ca_store()` API at the beginning? And the remove the initial NULL check from there.

paulreimer: Hi @ChiragAtal -- I can do this, but then repeated calls to `esp_tls_init_global_ca_store()` will wipe the store if it was already configured, e.g. you can't run it multiple times. Right now repeated calls to `esp_tls_set_global_ca_store` will wipe the store, but init is no-op if it is already allocated. Should I change it to the new behaviour?

mahavirj:
@paulreimer Thanks for your contribution. Change is merged with https://github.com/espressif/esp-idf/commit/a1204f8b161c0f451f130ba6d22e3f9a2e18323f

Issue #2342, opened 2018-08-28
FreeRTOS component, possible xTimerDelete / prvProcessReceivedCommands flaw? (IDFGH-352)


paulreimer:
Hopefully I have missed something or am using FreeRTOS incorrectly, but I have the following issue:

Preconditions:
- Using FreeRTOS Timer API
- Multiple "one-shot" timers or recurring timers which are stopped + deleted from other tasks.
- (Optionally) Sometimes deleting these timers from within the callback, in the timer svc task itself.
- (Possibly) Unprocessed timer events for the same `Timer_t` in the timer svc's queue at time of deletion
- (Possibly) Multiple events for the same `Timer_t` in the timer svc's queue e.g. in the case of a fast timer that occasionally does enough work to delay the next event.

As I understand it, the timer svc is constantly servicing the queue in its own task here:
https://github.com/espressif/esp-idf/blob/cc8ad721f98ffbc7953ece70616c07422b58e06b/components/freertos/timers.c#L701

And if a timer is deleted by a different task, then taking this pxTimer pointer would be invalid:
https://github.com/espressif/esp-idf/blob/cc8ad721f98ffbc7953ece70616c07422b58e06b/components/freertos/timers.c#L731

So this does seem to be the case for me -- if I omit `xTimerDelete` completely (leak), then everything is fine. However if I set a fast-ish timer (e.g. 50ms) and attempt to use `xTimerDelete` immediately after `xTimerStop`, then I get the following panic w/stacktrace:
```
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.
Core 0 register dump:
PC      : 0x400904a2  PS      : 0x00060330  A0      : 0x8008e980  A1      : 0x3ffc4350
0x400904a2: vListInsert at esp-idf/components/freertos/list.c:188 (discriminator 1)

A2      : 0x3ffb63ec  A3      : 0x3ffd8a9c  A4      : 0x1c378132  A5      : 0x00000000
A6      : 0x00000000  A7      : 0x3ffc0480  A8      : 0x00000004  A9      : 0x3ffe1eb4
A10     : 0x00000006  A11     : 0x00007203  A12     : 0x00060320  A13     : 0x3ffc43ac
A14     : 0x3ffd8a98  A15     : 0x3ffd8a98  SAR     : 0x0000000a  EXCCAUSE: 0x0000001c
[...]

PR #2655, opened 2018-11-01
Add dynamic deleted timer list to freertos timer loop (IDFGH-2497)

paulreimer:
To avoid double free on timers.

It resolves my issue here:
https://github.com/espressif/esp-idf/issues/2342

Perhaps seeing the solution (that worked for me) will illustrate the issue? I haven't been able to replicate it, only to verify that this PR fixes it for me (and perhaps protects against the double free issue if it did somehow happen anyway).

### DIFF
--- a/components/freertos/timers.c
+++ b/components/freertos/timers.c
@@ -698,6 +698,9 @@ Timer_t *pxTimer;
 BaseType_t xTimerListsWereSwitched, xResult;
 TickType_t xTimeNow;
 
+List_t* xDeletedTimerList = NULL;
+
+int deleted_timer_count = 0;
 	while( xQueueReceive( xTimerQueue, &xMessage, tmrNO_DELAY ) != pdFAIL ) /*lint !e603 xMessage does not have to be initialised as it is passed out, not in, and it is not used unless xQueueReceive() returns pdTRUE. */
 	{
 		#if ( INCLUDE_xTimerPendFunctionCall == 1 )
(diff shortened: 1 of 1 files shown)


paulreimer:
@negativekelvin yes, good suggestions all, implemented now. Should I add some logging in the case the List or ListItems cannot be allocated? For now it would just continue silently without that feature and try to allocate again next time a timer is deleted -- same behaviour as before this PR which did cause issues for me, but if you are out of memory then probably there are bigger problems elsewhere.

negativekelvin:
@mahavirj esp-idf freertos is not based on Amazon's fork of freertos. Esp-idf already has significant changes from base freertos to support SMP.

mahavirj:
@negativekelvin 
> esp-idf freertos is not based on Amazon's fork of freertos. Esp-idf already has significant changes from base freertos to support SMP.

Yes, aware about that. But this is core feature/change that should benefit all users of FreeRTOS and we can always backport this, once it gets merged upstream.

X-Ryl669:
Instead of creating a new list and having to O(N) search it for each timer to delete, it would be more efficient to have a bit/boolean/flag to say if the timer is deleted. Flag the bit when it's about to be deleted, then, finally, iterate all timers for deleted bit. This is O(1) and does not consume any memory (if you can't find a free bit anywhere in the xTimer structure, you can set the name to a sentinel for example when it's deleted). You can, for example, set `pcTimerName = DeletedSentinel` with `const char * DeletedSentinel = (const char*)1`. Then, in the end, you'll simply need to iterate all the timers and delete only those where `pcTimerName == DeletedSentinel`.

PR #2683, opened 2018-11-08
driver/uart: round baudrate to the closest integer (IDFGH-2873)

bearsh:


components/driver/uart.c
```diff
@@ -204,7 +204,7 @@ esp_err_t uart_set_baudrate(uart_port_t uart_num, uint32_t baud_rate)
     } else {
         uart_clk_freq = esp_clk_apb_freq();
     }
-    uint32_t clk_div = (((uart_clk_freq) << 4) / baud_rate);
+    uint32_t clk_div = (((uart_clk_freq << 4) + (baud_rate / 2)) / baud_rate);
```
wingunder: The following will use a right shift, instead of a divide, and will be faster (and more pure).
`
uint32_t clk_div = ((uart_clk_freq << 4) + (baud_rate >> 1)) / baud_rate;
`

bearsh: agreed and done!
(although I expect that a sane compile as of today should do this optimization by itself...)

koobest:
Thanks for your submission :), The divider contains integer and decimal parts. In most cases, the accuracy is sufficient,  but this calculation method `((uart_clk_freq << 4) + (baud_rate >> 1)) / baud_rate;`  can reduce accuracy in some cases. e.g,: 
    when the `APB % baudrate == 0 `

Alvin1Zhang:
@bearsh Thanks for contribution.

PR #2735, opened 2018-11-23
BT: Check if CCB's fixed channel info is NULL (IDFGH-2874)

pvachon:
A BT device can disconnect while other actions (such as requesting BLE GATT attributes) can be underway. This sets various bits of state to NULL under our feet. When these two states raced, it was possible an ESP32 would crash with a load exception.

When checking for congestion, return TRUE, indicating the channel is congested, so that the caller's error paths can take over.


Alvin1Zhang:
@pvachon Thanks for contribution and sorry for slow turnaround, please help see commit 51a0fcea5. Thanks.

PR #2743, opened 2018-11-25
Feature/pthread enhancement

PerMalmberg:
I had some time over while waiting for hardware to arrive so I implemented the ability to set the name and core affinity on pthreads and thus _std::thread_ as I originally [originally asked for](https://www.esp32.com/viewtopic.php?f=13&t=7919) on the forums.

I'm sure @projectgus or @igrr will have something to say about default values etc.

Re. `thread_name` member in `esp_pthread_cfg_t`: it had to be an array and thus taking up extra space. I see no other way to implement it without introducing a real risk of dangling pointers when `inherit_cfg` is used. I'm also not sure if `configMAX_TASK_NAME_LEN` includes the terminating null, i.e. if the +1 in `thread_name` declaration is needed or not.

Perhaps the name shouldn't be part of the `esp_pthread_cfg_t` at all and another way of naming phreads should be used? I'm open to suggestions.

### DIFF
--- a/components/pthread/Kconfig
+++ b/components/pthread/Kconfig
@@ -19,4 +19,31 @@ config PTHREAD_STACK_MIN
     help
         Minimum allowed pthread stack size set in attributes passed to pthread_create
 
+choice ESP32_PTHREAD_TASK_CORE_DEFAULT
+    bool "Default pthread core affinity"
+    default ESP32_DEFAULT_PTHREAD_CORE_NO_AFFINITY
+    depends on !FREERTOS_UNICORE
+    help
+        The default core to which pthreads are pinned.
+        
+config ESP32_DEFAULT_PTHREAD_CORE_NO_AFFINITY
+    bool "No affinity"
+config ESP32_DEFAULT_PTHREAD_CORE_0
+    bool "Core 0"
+config ESP32_DEFAULT_PTHREAD_CORE_1
+    bool "Core 1"
+endchoice
+
+config ESP32_PTHREAD_TASK_CORE_DEFAULT
+    int
+    default -1 if ESP32_DEFAULT_PTHREAD_CORE_NO_AFFINITY || FREERTOS_UNICORE
+    default 0 if ESP32_DEFAULT_PTHREAD_CORE_0
+    default 1 if ESP32_DEFAULT_PTHREAD_CORE_1
+
+config ESP32_PTHREAD_TASK_NAME_DEFAULT
+    string "Default name of pthreads"
+    default "pthread"
+    help
+        The default name of pthreads.
+
 endmenu
--- a/components/pthread/include/esp_pthread.h
+++ b/components/pthread/include/esp_pthread.h
@@ -14,6 +14,9 @@
 
 #pragma once
 
+#include "esp_err.h"
+#include <freertos/FreeRTOSConfig.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -24,10 +27,21 @@ extern "C" {
 
 /** pthread configuration structure that influences pthread creation */
 typedef struct {
-    size_t stack_size;    ///< the stack size of the pthread
-    size_t prio;          ///< the thread's priority
-    bool inherit_cfg;     ///< inherit this configuration further
+    size_t stack_size;  ///< The stack size of the pthread
+    size_t prio;        ///< The thread's priority
+    bool inherit_cfg;   ///< Inherit this configuration further
+    const char* thread_name;  ///< The thread name.
+    int pin_to_core;    ///< The core id to pin the thread to. Has the same value range as xCoreId argument of xTaskCreatePinnedToCore.
 } esp_pthread_cfg_t;
+
+/**
+ * @brief Creates a default pthread configuration based
+ * on the values set via menuconfig.
+ * 
+ * @return
+ *      A default configuration structure.
+ */
+esp_pthread_cfg_t esp_pthread_get_default_config();
 
 /**
  * @brief Configure parameters for creating pthread
--- a/components/pthread/pthread.c
+++ b/components/pthread/pthread.c
@@ -136,7 +136,6 @@ static void pthread_delete(esp_pthread_t *pthread)
     free(pthread);
 }
 
-
 /* Call this function to configure pthread stacks in Pthreads */
 esp_err_t esp_pthread_set_cfg(const esp_pthread_cfg_t *cfg)
 {
@@ -168,6 +167,24 @@ esp_err_t esp_pthread_get_cfg(esp_pthread_cfg_t *p)
     return ESP_ERR_NOT_FOUND;
 }
 
+static int get_default_pthread_core()
+{
+    return CONFIG_ESP32_PTHREAD_TASK_CORE_DEFAULT == -1 ? tskNO_AFFINITY : CONFIG_ESP32_PTHREAD_TASK_CORE_DEFAULT;
+}
+
+esp_pthread_cfg_t esp_pthread_get_default_config()
+{
+    esp_pthread_cfg_t cfg = {
+        .stack_size = CONFIG_ESP32_PTHREAD_TASK_STACK_SIZE_DEFAULT,
+        .prio = CONFIG_ESP32_PTHREAD_TASK_PRIO_DEFAULT,
+        .inherit_cfg = false,
+        .thread_name = NULL,
+        .pin_to_core = get_default_pthread_core()
+    };
+
+    return cfg;
+}
+
 static void pthread_task_func(void *arg)
 {
     void *rval = NULL;
@@ -179,8 +196,13 @@ static void pthread_task_func(void *arg)
     xTaskNotifyWait(0, 0, NULL, portMAX_DELAY);
 
     if (task_arg->cfg.inherit_cfg) {
-        /* If inherit option is set, then do a set_cfg() ourselves for future forks */
-        esp_pthread_set_cfg(&task_arg->cfg);
+        /* If inherit option is set, then do a set_cfg() ourselves for future forks,
+        but first set thread_name to NULL to enable inheritance of the name too.
+        (This also to prevents dangling pointers to name of tasks that might
+        possibly have been deleted when we use the configuration).*/
+        esp_pthread_cfg_t *cfg = &task_arg->cfg;
+        cfg->thread_name = NULL;
+        esp_pthread_set_cfg(cfg);
     }
     ESP_LOGV(TAG, "%s START %p", __FUNCTION__, task_arg->func);
     rval = task_arg->func(task_arg->arg);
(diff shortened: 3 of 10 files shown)


components/pthread/include/esp_pthread.h
```diff
@@ -24,11 +27,33 @@ extern "C" {
...
 typedef struct {
-    size_t stack_size;    ///< the stack size of the pthread
-    size_t prio;          ///< the thread's priority
-    bool inherit_cfg;     ///< inherit this configuration further
+    size_t stack_size;  ///< The stack size of the pthread
+    size_t prio;        ///< The thread's priority
+    bool inherit_cfg;   ///< Inherit this configuration further
+    // This member has to be an array and not a pointer due to the fact that the entire
+    // struct is copied by the thread when inherit_cfg is true which in turn likely
+    // will result in dangling pointers, especially if threads are created/destroyed
```
igrr: Could we possibly use some FreeRTOS API to get the name from the parent task, instead of getting it from this structure? I.e. if inherit_cfg is true then make a copy of this structure and set thread_name pointer to NULL. When inheriting, if thread_name is NULL, get the name of the current task

components/pthread/Kconfig
```diff
@@ -19,4 +19,21 @@ config PTHREAD_STACK_MIN
     help
         Minimum allowed pthread stack size set in attributes passed to pthread_create
 
+config ESP32_PTHREAD_TASK_CORE_DEFAULT
+    int "Default pthread core affinity"
+    default -1
+    range -1 1
```
igrr: Has to be conditional on FREERTOS_UNICORE, don't allow setting 1 in single core mode.

components/pthread/include/esp_pthread.h
```diff
@@ -24,11 +27,33 @@ extern "C" {
...
+    // dynamically at run time with different names etc.
+    char thread_name[configMAX_TASK_NAME_LEN + 1];  ///< The thread name.
+    int pin_to_core;    ///< The core id to pin the thread to. See components/pthread/KConfig for valid ranges.
```
igrr: Suggest something like 'has the same meaning as xCoreId argument of xTaskCreatePinnedToCore' instead of pointing to Kconfig.

components/pthread/include/esp_pthread.h
```diff
@@ -24,11 +27,33 @@ extern "C" {
...
 } esp_pthread_cfg_t;
 
+/**
+ * @brief Creates a default pthread configuration based
+ * on the values set via menuconfig.
+ * 
+ * @return
+ *      A default configuration structure.
+ */
+esp_pthread_cfg_t esp_create_default_pthread_config();
```
igrr: Can you make it esp_pthread_get_default_config (so that API name is prefixed with esp_pthread)?

Same with esp_pthread_set_name below.

components/pthread/pthread.c
```diff
@@ -168,6 +167,41 @@ esp_err_t esp_pthread_get_cfg(esp_pthread_cfg_t *p)
     return ESP_ERR_NOT_FOUND;
 }
 
+static int min(int a, int b)
```
igrr: You can use MIN macro from sys/param.h

components/pthread/pthread.c
```diff
@@ -168,6 +167,41 @@ esp_err_t esp_pthread_get_cfg(esp_pthread_cfg_t *p)
...
+{
+    return a < b ? a : b;
+}
+
+void esp_set_pthread_name(esp_pthread_cfg_t* cfg, const char* name)
+{
+    if(cfg != NULL)
+    {
+        const char* s = (name == NULL || *name == 0) ? CONFIG_ESP32_PTHREAD_TASK_NAME_DEFAULT : name;
+        const int length = min(sizeof(cfg->thread_name)/sizeof(cfg->thread_name[0]) - 1, strlen(s));
```
igrr: or just use strlcpy :)

examples/system/cpp_pthread/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# The following lines of boilerplate have to be in your project's CMakeLists
+# in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.5)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
```
igrr: Are these three line necessary?

examples/system/cpp_pthread/main/cpp_pthread.cpp
```diff
@@ -0,0 +1,109 @@
+#include <iostream>
```
igrr: Please add here the public domain/CC0 header found in other examples.

examples/system/cpp_pthread/main/cpp_pthread.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <esp_log.h>
+
+using namespace std::chrono;
+
+const auto sleep_time = seconds{5};
+
+void print_thread_info(const char* extra = nullptr)
+{
+        std::stringstream ss;
+        if(extra)
```
igrr: Tabs, space between if and paren, opening brace should be on the same line with the condition.

(Or may just run the file through the formatting script)

examples/system/cpp_pthread/main/cpp_pthread.cpp
```diff
@@ -0,0 +1,109 @@
...
+void thread_func_inherited()
+{
+    while(true)
+    {
+        print_thread_info("This is the INHERITING thread with the same parameters as our parent, including name. ");
+        std::this_thread::sleep_for(sleep_time);
+    }
+}
+
+void spwan_another_thread()
```
igrr: Typo, spawn

examples/system/cpp_pthread/main/cpp_pthread.cpp
```diff
@@ -0,0 +1,109 @@
...
+{
+        std::stringstream ss;
+        if(extra)
+        {
+            ss << extra;
+        }
+        ss << "Core id: " << xPortGetCoreID()
+            << ", prio: " << uxTaskPriorityGet(nullptr) 
+            << ", minumim free stack: " << uxTaskGetStackHighWaterMark(nullptr) << " bytes.";
+        ESP_LOGI(pcTaskGetTaskName(nullptr), "%s", ss.str().c_str());
```
igrr: Wouldn't it be nice if we could keep using Pthread APIs here... Oh well, maybe some other time.

examples/system/cpp_pthread/main/cpp_pthread.cpp
```diff
@@ -0,0 +1,109 @@
...
+    esp_set_pthread_name(&cfg, name);
+    cfg.pin_to_core = core_id;
+    cfg.stack_size = stack;
+    cfg.prio = prio;
+    return cfg;
+}
+
+extern "C"
+{
+    void app_main()
```
igrr: Can you please make it `extern "C" void app_main()` to save one indentation level?

components/pthread/include/esp_pthread.h
PerMalmberg: I see no API that can get the name of the parent API, no.

components/pthread/Kconfig
PerMalmberg: I'll do something even better - menu selection when unicore is active.

components/pthread/include/esp_pthread.h
PerMalmberg: Sure.

components/pthread/include/esp_pthread.h
PerMalmberg: Yes.

components/pthread/pthread.c
PerMalmberg: I'll go with strlcpy

examples/system/cpp_pthread/CMakeLists.txt
PerMalmberg: Not strictly, no. I'll update the example code to make them unnecessary.

examples/system/cpp_pthread/main/cpp_pthread.cpp
PerMalmberg: Ofc.

examples/system/cpp_pthread/main/cpp_pthread.cpp
PerMalmberg: I looked at the contributing.rst file at github, but links in those files are broken when viewed there. Didn't realize that the docs.espressif.com shows them correctly. I'll format the file with the script.

examples/system/cpp_pthread/main/cpp_pthread.cpp
PerMalmberg: Ops.

examples/system/cpp_pthread/main/cpp_pthread.cpp
PerMalmberg: It would, but I guess it's more to it than just wrapping the functions?

examples/system/cpp_pthread/main/cpp_pthread.cpp
PerMalmberg: Yes, silly me.

PerMalmberg:
Code updated according to review comments by @igrr

components/pthread/include/esp_pthread.h
igrr: Consider a case when task A creates task B which creates task C.
A sets inherit=true and provides a name (pointer), then creates B. At this point, B gets it's name from the pointer. Pthread config associated with B gets updated to keep inherit=true and name=NULL. When B creates C, we see that inherit=true but name=NULL, so we take the name of the current task (B). Thus C gets created with the name of B.

PerMalmberg: Oh, I thought you meant there were an actual API to get the name of the creating task. I'll make a new commit with the proposed logic.

PerMalmberg:
@igrr New implementation pushed.

components/pthread/pthread.c
```diff
@@ -221,6 +245,28 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
         if (pthread_cfg->prio && pthread_cfg->prio < configMAX_PRIORITIES) {
             prio = pthread_cfg->prio;
         }
+
+        if(pthread_cfg->inherit_cfg) {
+            if(pthread_cfg->thread_name == NULL) {
+                // Inherit task name from current task.
+                task_name = pcTaskGetTaskName(NULL);
+            }
+            else {
```
igrr: Nitpick, `else` should be on the same line with the closing brace; also there is a missing space between `if` and opening paren. Same in a few other places below.

examples/system/cpp_pthread/main/cpp_pthread.cpp
igrr: Yes, let's keep this for now.

igrr:
just one minor note, looks good to me otherwise! Please combine the changes into a single commit.

components/pthread/pthread.c
PerMalmberg: Hm, guess I did I forget to run the formatter. Will fix.

PerMalmberg:
Commits squashed into a single commit as requested by @igrr

PR #2770, opened 2018-12-01
fixed bug where esp_vfs_unregister would unregister root file system (IDFGH-17810)

ThomasRogg:
when it should unregister sub-filesystem (such as /sdcard)
path comparison must be exact here

### DIFF
--- a/components/vfs/vfs.c
+++ b/components/vfs/vfs.c
@@ -171,12 +171,14 @@ esp_err_t esp_vfs_register_with_id(const esp_vfs_t *vfs, void *ctx, esp_vfs_id_t
 
 esp_err_t esp_vfs_unregister(const char* base_path)
 {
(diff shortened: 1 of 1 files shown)


neoniousTR:
Won't sign license agreement, not worth the hassle for me.
This is a clear bug. I would suggest you fix this.

neoniousTR:
You're welcome.

fivdi:
> Won't sign license agreement, not worth the hassle for me.

@neoniousTR I'm not sure if you are aware of this but signing the license agreement should actually be easy and if I'm not mistaken it can be done with your GitHub account. Just click on the "CLA | not signed yet" badge in the [above post](https://github.com/espressif/esp-idf/pull/2770#issuecomment-443391435) and follow the instructions.

PR #2847, opened 2018-12-18
fix(nvs) improve nvs_flash_init_partition() speed

GautierAtWork:
using  nvs_flash_init_partition() on a 2MBytes partition took 6sec.
This patch reduces this time

### DIFF
--- a/components/nvs_flash/src/nvs_page.cpp
+++ b/components/nvs_flash/src/nvs_page.cpp
@@ -48,16 +48,16 @@ esp_err_t Page::load(uint32_t sectorNumber)
         mState = header.mState;
         // check if the whole page is really empty
         // reading the whole page takes ~40 times less than erasing it
-        uint32_t line[8];
+        uint32_t line[64];
         for (uint32_t i = 0; i < SPI_FLASH_SEC_SIZE; i += sizeof(line)) {
             rc = spi_flash_read(mBaseAddress + i, line, sizeof(line));
             if (rc != ESP_OK) {
                 mState = PageState::INVALID;
                 return rc;
             }
-            if (std::any_of(line, line + 4, [](uint32_t val) -> bool { return val != 0xffffffff; })) {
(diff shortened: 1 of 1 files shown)


GautierAtWork:
@projectgus FYI

components/nvs_flash/src/nvs_page.cpp
```diff
@@ -48,16 +48,16 @@ esp_err_t Page::load(uint32_t sectorNumber)
         mState = header.mState;
         // check if the whole page is really empty
         // reading the whole page takes ~40 times less than erasing it
-        uint32_t line[8];
+        uint32_t line[64];
```
projectgus: Hi Gautier,

The only concern I have with this is that allocating 512 bytes on the stack (rather than 32) is quite a lot of stack, and some tasks may not have this spare.

A few possible ideas:

* Do you have any performance numbers on smaller size reads (bigger than 64 bytes, smaller than 512)?
* I suspect it might be possible to use spi_flash_mmap() the whole sector instead, provided flash encryption is disabled. This would probably be the fastest, although it needs two code paths (if flash encryption is enabled, spi_flash_read() will still be required)
* Could simply allocate the `line` buffer from heap, I guess.

I've asked @igrr  and @mahavirj to review as well, as they're more familiar with NVS than I am and might have different thoughts on this.

GautierAtWork: Hi Angus,

 - I didn't spend benchmark various values so far.
 - Yes, I was thinking about using spi_flash_mmap() as well, as it would be a single action, and no stack usage. Performance would be related to the size of a cache line.
 - correct, heap could be used here.

I agree, NVS maintainers could likely provide the best fix.

mahavirj:
Thanks for your contribution. Fixed with https://github.com/espressif/esp-idf/commit/b4c4f8514b24c0270e8419b9b4d1da2b5a77f32b

PR #2861, opened 2018-12-21
Sped the gpio_intr_service ISR up by 1.5 uSeconds (+-50% faster).

wingunder:
Removed as much branching (if statements) from the
gpio_intr_service ISR, as possible and split the while loop into
two. Also forced writing the two status*_w1tc variables only once,
instead of every time after calling the external function hooks.

The measurements below, was done using the following tools:

Toolchain version: crosstool-ng-1.22.0-80-g6c4433a
Compiler version: 5.2.0

Here follows a comparison of the gpio_intr_service ISR's
execution time, using a DS1054 oscilloscope. All the time spent
calling external functions, via the function pointers
gpio_isr_func[gpio_num].fn, were disregarded.

With OPTIMIZATION_FLAGS = -Og, 1.34 uSeconds faster:

3.22 uSec (with this patch)
4.56 uSec (with commit 71c90ac4)

100 - (100 * 4.56 / 3.22) = 42% faster

With OPTIMIZATION_FLAGS = -Os, 1.65 uSeconds faster:

2.89 uSec (with this patch)
4.54 uSec (with commit 71c90ac4)

100 - (100 * 4.54 / 2.89) = 57% faster

### DIFF
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -366,13 +366,17 @@ void IRAM_ATTR gpio_intr_service(void* arg)
 
     //read status to get interrupt status for GPIO0-31
     const uint32_t gpio_intr_status = GPIO.status;
-    gpio_isr_loop(gpio_intr_status, 0);
-    GPIO.status_w1tc = gpio_intr_status;
+    if (gpio_intr_status) {
+        gpio_isr_loop(gpio_intr_status, 0);
+        GPIO.status_w1tc = gpio_intr_status;
+    }
 
     //read status1 to get interrupt status for GPIO32-39
-    uint32_t gpio_intr_status_h = GPIO.status1.intr_st;
-    gpio_isr_loop(gpio_intr_status_h, 32);
-    GPIO.status1_w1tc.intr_st = gpio_intr_status_h;
+    const uint32_t gpio_intr_status_h = GPIO.status1.intr_st;
+    if (gpio_intr_status_h & ((1<<(GPIO_NUM_MAX-32))-1)) {
+        gpio_isr_loop(gpio_intr_status_h, 32);
+        GPIO.status1_w1tc.intr_st = gpio_intr_status_h;
+    }
 }
 
 esp_err_t gpio_isr_handler_add(gpio_num_t gpio_num, gpio_isr_t isr_handler, void* args)
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -330,19 +330,19 @@ esp_err_t gpio_reset_pin(gpio_num_t gpio_num)
 static inline int IRAM_ATTR num_lsb_zero_bits(const uint32_t v) {
     // From: https://en.wikipedia.org/wiki/De_Bruijn_sequence
     // This function returns the amount of unset (zero) bits starting from the LSB.
+    // Note: A value of 0 is returned when v == 0.
     static const int MultiplyDeBruijnBitPosition[32] = {
         0, 1, 28, 2, 29, 14, 24, 3, 30, 22, 20, 15, 25, 17, 4, 8,
         31, 27, 13, 23, 21, 19, 16, 7, 26, 12, 18, 6, 11, 5, 10, 9
     };
-    const int r = MultiplyDeBruijnBitPosition[((uint32_t)((v & -v) * 0x077CB531U)) >> 27];
-    return v ? r : 32;
+    return MultiplyDeBruijnBitPosition[((uint32_t)((v & -v) * 0x077CB531U)) >> 27];
 }
 
 static inline void IRAM_ATTR gpio_isr_loop(const uint32_t s, const uint32_t gpio_num_start) {
     uint32_t status = s;
     uint32_t gpio_num = gpio_num_start;
     while (status) {
-        const int n = num_lsb_zero_bits(s);
+        const int n = num_lsb_zero_bits(status);
         gpio_num += n; // Offset the gpio_num by the amount of unset (zero) bits found, starting from the LSB.
         if (gpio_isr_func[gpio_num].fn != NULL) {
             gpio_isr_func[gpio_num].fn(gpio_isr_func[gpio_num].args);
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -327,6 +327,36 @@ esp_err_t gpio_reset_pin(gpio_num_t gpio_num)
     return ESP_OK;
 }
 
+static inline int IRAM_ATTR num_lsb_zero_bits(const uint32_t v) {
+    // From: https://en.wikipedia.org/wiki/De_Bruijn_sequence
+    // This function returns the amount of unset (zero) bits starting from the LSB.
+    static const int MultiplyDeBruijnBitPosition[32] = {
+        0, 1, 28, 2, 29, 14, 24, 3, 30, 22, 20, 15, 25, 17, 4, 8,
+        31, 27, 13, 23, 21, 19, 16, 7, 26, 12, 18, 6, 11, 5, 10, 9
+    };
+    const int r = MultiplyDeBruijnBitPosition[((uint32_t)((v & -v) * 0x077CB531U)) >> 27];
+    return v ? r : 32;
+}
+
+static inline void IRAM_ATTR gpio_isr_loop(const uint32_t s, const uint32_t gpio_num_start) {
+    uint32_t status = s;
+    uint32_t gpio_num = gpio_num_start;
+    while (status) {
+        const int n = num_lsb_zero_bits(s);
+        gpio_num += n; // Offset the gpio_num by the amount of unset (zero) bits found, starting from the LSB.
+        if (gpio_isr_func[gpio_num].fn != NULL) {
+            gpio_isr_func[gpio_num].fn(gpio_isr_func[gpio_num].args);
+        }
+        if (n) {
+            status >>= n; // Advance to the next status bit, skipping all unset (zero) bits.
+        }
+        else { // LSB is not zero.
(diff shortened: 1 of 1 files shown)


wingunder:
> Is it worth it to blank check the status registers before looping over the bits?

Yes, you have a point, but we should probably not land in gpio_intr_service, if no status bits are set. I might be wrong...

Anyhow, I had a good look at the whole thing again, and submitted b0c1ea6, which will skip all, if no status bits were set. The patch also makes it much faster than before. For the gory details, please see b0c1ea6 and its commit message.

negativekelvin:
That's pretty cool.  What I meant was there is a good chance either gpio_intr_status or gpio_intr_status_h would be zero.

wingunder:
> That's pretty cool.

Thanks! :)

> What I meant was there is a good chance either gpio_intr_status or gpio_intr_status_h would be zero.

Ah, OK, I finally understand: We're mostly just handling only one status bit change inside gpio_intr_service(), but we have 2 separate 2 status words, so the bit can only be in one of the two: gpio_intr_status or gpio_intr_status_h. If one of these status words have a bit that is set, chances are really good that the other one has _no_ bits set, and is zero. Tada.... Sorry for not getting to this idea earlier, even with your hint.

OK, I can work on something to exploit this optimization as well, but the following makes me hesitate slightly: Currently, even with my patches, the timing inside  gpio_intr_service() is the same for all stati. It could be that the implementation of an optimization, as described in the previous paragraph, will break some true real time apps. I actually don't think it will, but one can never be sure.

I could implement the '1st/2nd status == 0' optimization check and add a configuration option to disable it, leaving it enabled by default. The possibility to disable it, will give us a backdoor, just in case it breaks some serious real time apps.

Some final questions:
   - Should I implement the '1st/2nd status == 0' optimization?
   - Should I add a configuration option for disabling it, and document it?
   - Is there anything else that I should document for this PR?

negativekelvin:
Your while loop does essentially do the zero check so moving it a few instructions sooner may not have much effect but you can test it if you have time.

negativekelvin:
Also do we know the upper 24 bits of gpio_intr_status_h will never be set or should they be masked out?

wingunder:
> Also do we know the upper 24 bits of gpio_intr_status_h will never be set or should they be masked out?

This is covered, as we read the GPIO_STATUS1_REG status register and write this value into the GPIO_STATUS1_W1TC_REG register after servicing all the registered ISRs. The upper 24 bits of both these registers are reserved. See page 65 of:
https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf

Furthermore, this functionality didn't really change, as we're now just writing the full status to GPIO_STATUS1_W1TC_REG after servicing the registered ISRs, instead of doing it bit by bit, after servicing or skipping every ISR.

negativekelvin:
Yes they are reserved but are they guaranteed to be 0?  Because now you are passing the raw value to gpio_isr_loop which does nothing to avoid those upper 24 bits and if one of them happens to be set it is going to do out of bounds reads on gpio_isr_func array.  So if you just mask it when you set gpio_intr_status_h then you don't have to mask it when doing the zero check.

wingunder:
> Yes they are reserved but are they guaranteed to be 0?

Thanks for pointing this out.
Commit b9f4ef6 sets the top 24 bits of status1 to zero, before using it.

projectgus:
Nice optimisation!

I think there may actually be some gcc builtins that are even more effective than this, for example `__builtin_ffs()` (which compiles to a single NSAU - Normalization Shift Amount Unsigned - instruction on ESP32) https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

wingunder:
> Nice optimisation!

Thanks!

> I think there may actually be some gcc builtins that are even more effective than this, for example `__builtin_ffs()` (which compiles to a single NSAU - Normalization Shift Amount Unsigned - instruction on ESP32) https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html . Did you look at these at all?

No, I didn't. I knew this existed for other CPUs, but I did not think of trying this as I didn't think that these would have been implemented for the esp32. A pleasant surprise :)

Here is the test code:
```
gpio_set_level(GPIO_NUM_19, 0);
gpio_set_level(GPIO_NUM_19, 1);
const volatile int n = num_lsb_zero_bits(tmp);
gpio_set_level(GPIO_NUM_19, 0);
gpio_set_level(GPIO_NUM_19, 1);
const volatile int n1 = __builtin_ffs(tmp) - 1;
gpio_set_level(GPIO_NUM_19, 0);
gpio_set_level(GPIO_NUM_19, 1);
const volatile int n2 = __builtin_ctz(tmp);
gpio_set_level(GPIO_NUM_19, 0);
```
Here are the (+- expected) benchmarks from my 50MHz scope:
   - num_lsb_zero_bits - 0.374 uSec
   - __builtin_ffs - 0.300 uSec
   - __builtin_ctz - 0.300 uSec

The num_lsb_zero_bits function is inline and is a simple indexed read (in the DeBruijn lookup table), and that's why it really holds up very well to the 'equivalent' built-ins.

As you can see, I also benchmarked the ctz builtin, which is a drop-in replacement for num_lsb_zero_bits, except for the case where the value is zero. This gets filtered in the while(status) condition, so this is no problem.

The first paragraph of https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html reads:
> GCC provides a large number of built-in functions other than the ones mentioned above. Some of these are for internal use in the processing of exceptions or variable-length argument lists and are not documented here because they may change from time to time; we do not recommend general use of these functions.

Disadvantages of using ctz:
   - The compiler might not support __builtin_ctz in the future.

Advantages of using ctz:
   - Save another 0.074 uSec.
   - Marginally less code. (The DeBruin lookup will not be needed.)

Would you say that I should submit a patch for using __builtin_ctz?

wingunder:
@projectgus, I added the \_\_bultin\_ patch in any case. If needed, it can be skipped by cherry-picking, once this PR gets merged.

Thanks again for the \_\_builtin\_* hint.

projectgus:
@wingunder Thanks for being patient while we worked our way around to this.

This patch is in our merge queue now. We made some small changes, using `__builtin_ffs()` rather than `__builtin_ctz()` as it allows the code to be structured in a slightly simpler way.

Also, there was a small trap if interrupt value is exactly 0x80000000, as the result of `status >>= 32` is undefined behaviour in C (can't shift by a number of bits equal to or greater than the size of the promoted lvalue). (Kudos to @KonstantinKondrashov for noticing this.)

The change should be merge to master soon.

wingunder:
>@wingunder Thanks for being patient while we worked our way around to this.
@projectgus Thanks, no problem.

> Also, there was a small trap if interrupt value is exactly 0x80000000, as the result of status >>= 32 is undefined behaviour in C (can't shift by a number of bits equal to or greater than the size of the promoted lvalue). (Kudos to @KonstantinKondrashov for noticing this.)

Excellent! :1st_place_medal: Thanks a lot, for pointing this out to me. I could however find _no_ reference on the entire github.com of this suggestion...

>The change should be merged to master soon.
Thanks, I appreciate this.

One thing that is however hard for me to understand, is that it seems as if a lot of stuff, concerning this PR has happened somewhere else than here. The parties concerned probably have their reasons for this, but I find this really strange, given the fact that this is an open source project. I have spent a hell of a lot of my free time on this PR, but being excluded from the progress on it, doesn't motivate me at all.

Best regards
@wingunder

projectgus:
Hi @wingunder,

> it seems as if a lot of stuff, concerning this PR has happened somewhere else than here. The parties concerned probably have their reasons for this, but I find this really strange, given the fact that this is an open source project. I have spent a hell of a lot of my free time on this PR, but being excluded from the progress on it, doesn't motivate me at all.

That's a fair criticism, thanks for pointing it out. I apologise for excluding you from these final changes. I'll explain a bit what happened here, to hopefully fill you in a bit more:

Normally, we review a Pull Request here and then push the branch into our internal gitlab server for automated testing and the final merge. [The process is fully documented here](https://docs.espressif.com/projects/esp-idf/en/latest/contribute/). Sometimes there's some minor cleanup that happens at this stage, not normally big changes like this.

I asked Konstantin to merge this PR and also look to see if the same approach could be optimization to any of our other drivers. I also mentioned to him the `__builtin_ffs` macro that I mentioned to you, as I still had a feeling the code could be simplified a little (but I wasn't sure).

Konstantin pushed the changes into our internal gitlab server. He also noted the "undefined behaviour" problem with right shifting a full word and made the changes for __builtin_ffs() at the same time. I suggested he make the changes as I thought it wasn't very polite to come back to the PR after six weeks and ask for more changes from you. In retrospect this was probably the wrong decision, I apologise for that.

Normally the PR process doesn't go like this one did, this was something of an outlier where we decided to do some additional internal work based on your work. I hope it doesn't discourage you from sending future PRs, but would understand if it did.

wingunder:
Hi @projectgus,

>In retrospect this was probably the wrong decision, I apologise for that.

Thank you very much for coming back to me on this. I truly except your apology and would like to thank you for explaining this in such a honest and polite way. 

>Normally the PR process doesn't go like this one did, this was something of an outlier where we decided to do some additional internal work based on your work.

It's heartening to know that this was an outlier case. I'm also glad that the community can eventually profit from your internal enhancements.

> I hope it doesn't discourage you from sending future PRs, but would understand if it did.

No problem :) I still have have #2873 open and I'm looking forward to some feedback on this. Once that's out of the way, there will surely be more to come.

Best regards
@wingunder

igrr:
Thank you for the PR @wingunder, has been cherry-picked as b8dc48a.

Issue #2845, opened 2018-12-17
GPIO interrupts service and disabling interrupts (IDFGH-38)


nameofuser1:
Hi there!

I have spent last few days looking for bug in my code and found out an interesting thing: GPIO interrupt service checks only for interrupt status. At the same time clearing `GPIO.pin[gpio_num].int_ena` flag, which is done by `gpio_intr_disable()` function, does not seem to mask gpio interrupt as a consequence interrupt service will still call disabled interrupt handler. 

https://github.com/espressif/esp-idf/blob/da2116f55c555ab6a4aa9d3a982e9bee990dd7b4/components/driver/gpio.c#L120-L126

https://github.com/espressif/esp-idf/blob/da2116f55c555ab6a4aa9d3a982e9bee990dd7b4/components/driver/gpio.c#L330-L361

Certainly calling `gpio_isr_handler_remove` or clearing `GPIO.pin[gpio_num].int_type` field can be used in order to stop gpio interrupt service to call disabled interrupts, but, naturally, if I use function named `whatever_disable_interrupt()`, I expect that corresponding irq handler will never be executed, which is not true in this case and requires me to look inside function and compare it with reference manual.

negativekelvin:
If the interrupt is disabled and the flag is cleared by gpio_intr_disable then it should be impossible for the flag to be set.  So the flag check in gpio_intr_service would never succeed and the handler for that gpio would not be called.

nameofuser1:
> If the interrupt is disabled and the flag is cleared by gpio_intr_disable then it should be impossible for the flag to be set. So the flag check in gpio_intr_service would never succeed and the handler for that gpio would not be called.

That's strange because until I clear `GPIO.pin[MY_GPIO_1].int_type`, interrupt handler for MY_GPIO_1 is executed whenever edge on MY_GPIO_1 line was found and any other GPIO interrupt is triggered. I can not attach my code and screenshots from logic analyzer right now but certainly will do it tomorrow.

X-Ryl669:
Is it called once only ? 
It's possible the interrupt is currently running (on the other core) and the test for interrupt status saying it's triggered already passed when you are disabling the interrupt. In that case, it'll be triggered once for all.

nameofuser1:
Updates :)

[...]

PR #2873, opened 2018-12-24
Consistently enable gpio interrupts on the core, stored in gpio_isr_handle.

wingunder:
This PR aims to do the following:                                                                                                                                                              
                                                                                                                                                                                               
- consistently enable gpio interrupts on the correct core.                                                                                                                                     
- enable and disable interrupts according to the value of the intr_type.                                                                                                                       
- fix all gpio tests, touched by the above changes.                                                                                                                                            
- added a gpio test for enabling interrupts on a core other than the one on                                                                                                                          
  which the isr_sevice was started.

It is now possible to:
   - call  gpio_intr_disable from anywhere and expect the interrupt _not_ to occur again.
   - rely on gpio_intr_disable to disable an interrupt consistently.
   - call gpio_intr_enable from any task, regardless on which core it runs.

For more information, please read the elaborate commit messages below.

Here follows the original text, with which this PR was started:

> At the moment (commit db8bc3ee), gpio_intr_enable and
gpio_intr_disable do not have critical sections for guarding status,
enable and disable flags, which are also manipulated in the ISR,
gpio_intr_service. It was probably OK up to now, as these functions
were only called from critical sections within the gpio.c code. One
exception is of course the call from the gpio_config function, but the
chances that this would have caused a problem, would have been
minimal, as no-one would probably have thought of calling gpio_config
more than once, for a given port.

> A major problem is however, that both gpio_intr_enable and
gpio_intr_disable are published in the gpio.h file, opening the door
to anyone to call these, at any time. This could certainly cause
havoc, as both these functions manipulate the interrupt status
flags. The ISR, gpio_intr_service, relies on the status flags to stay
constant during the interrupt, so these _must_ be protected.

> This patch protects gpio_intr_enable and gpio_intr_disable with a
critical section. Note that FreeRTOS allows and supports nested
critical sections, so no fancy checking is done inside the functions,
in order to see if we're already inside a critical section or
not, as it doesn't matter. This patch makes it safe to call
gpio_intr_enable and gpio_intr_disable from anywhere, except from
within an ISR.

### DIFF
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -114,14 +114,19 @@ static esp_err_t gpio_intr_enable_on_core (gpio_num_t gpio_num, uint32_t core_id
 
 esp_err_t gpio_intr_enable(gpio_num_t gpio_num)
 {
-    return gpio_intr_enable_on_core (gpio_num, xPortGetCoreID());
+    portENTER_CRITICAL(&gpio_spinlock);
+    const esp_err_t ret = gpio_intr_enable_on_core (gpio_num, xPortGetCoreID());
+    portEXIT_CRITICAL(&gpio_spinlock);
+    return ret;
 }
 
 esp_err_t gpio_intr_disable(gpio_num_t gpio_num)
 {
     GPIO_CHECK(GPIO_IS_VALID_GPIO(gpio_num), "GPIO number error", ESP_ERR_INVALID_ARG);
+    portENTER_CRITICAL(&gpio_spinlock);
     GPIO.pin[gpio_num].int_ena = 0;                             //disable GPIO intr
     gpio_intr_status_clr(gpio_num);
+    portEXIT_CRITICAL(&gpio_spinlock);
     return ESP_OK;
 }
 
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -100,24 +100,19 @@ static void gpio_intr_status_clr(gpio_num_t gpio_num)
     }
 }
 
-static esp_err_t gpio_intr_enable_on_core (gpio_num_t gpio_num, uint32_t core_id)
+esp_err_t gpio_intr_enable(gpio_num_t gpio_num)
 {
+    GPIO_CHECK(gpio_isr_func != NULL, "GPIO isr service is not installed, call gpio_install_isr_service() first", ESP_ERR_INVALID_STATE);
     GPIO_CHECK(GPIO_IS_VALID_GPIO(gpio_num), "GPIO number error", ESP_ERR_INVALID_ARG);
+    portENTER_CRITICAL(&gpio_spinlock);
     gpio_intr_status_clr(gpio_num);
-    if (core_id == 0) {
+    if (esp_intr_get_cpu(gpio_isr_handle) == 0) {
         GPIO.pin[gpio_num].int_ena = GPIO_PRO_CPU_INTR_ENA;     //enable pro cpu intr
     } else {
         GPIO.pin[gpio_num].int_ena = GPIO_APP_CPU_INTR_ENA;     //enable pro cpu intr
     }
-    return ESP_OK;
-}
-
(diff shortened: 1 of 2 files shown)


components/driver/gpio.c
```diff
@@ -88,6 +88,11 @@ esp_err_t gpio_set_intr_type(gpio_num_t gpio_num, gpio_int_type_t intr_type)
     GPIO_CHECK(GPIO_IS_VALID_GPIO(gpio_num), "GPIO number error", ESP_ERR_INVALID_ARG);
     GPIO_CHECK(intr_type < GPIO_INTR_MAX, "GPIO interrupt type error", ESP_ERR_INVALID_ARG);
     GPIO.pin[gpio_num].int_type = intr_type;
+    if (intr_type == GPIO_INTR_DISABLE) {
+        gpio_intr_disable(gpio_num);
+    } else {
+        gpio_intr_enable(gpio_num);
```
negativekelvin: I don't think we want to automatically enable the interrupt when setting the type.

wingunder: @negativekelvin Thanks for your review. It's almost been 7 months, since I had my head into this and I unfortunately do not have time to help out on this project at the moment. I however remembered that what you spotted here, bothered me as well. It was however the solution that I found with the bare minimum changes to the "structure", at that time. I wanted to rewrite a lot more, but I have had no response on my patch, from anyone at all, so it now is what it is.

BTW, just 2 days ago, @koobest wrote the following [here](https://github.com/espressif/esp-idf/issues/2845#issuecomment-513447994):

> We are use the GPIO interrupt raw register in the ISR. So, When the interrupt is disabled, the interrupt raw status register will still be set when the interrupt trigger signal arrives.By the way, when the int_type register is set to 0, the interrupt raw status register will not be set. It is recommended to use gpio_set_intr_type to enable or disable interrupts.
thanks !!

I'm not sure what to make of his last sentence, but if @koobest is right, then my patch should probably stay as it is.

I would suggest to rework the whole interrupt enabling and disabling thing, and name things according to what they do, once #2845 is solved and stable.

negativekelvin: If you look at the follow-up to that you'll see it will be changing but I see your point about forcing the enable since the current code will read the flag anyway.  It would be the least bad option if there wasnt a fix.

koobest: Hi, @wingunder 
Thanks for your contribution.  Can you help post this PR based on the latest master branch? so we can help review and merge into the master branch.

> We are use the GPIO interrupt raw register in the ISR. So, When the interrupt is disabled, the interrupt raw status register will still be set when the interrupt trigger signal arrives.By the way, when the int_type register is set to 0, the interrupt raw status register will not be set. It is recommended to use gpio_set_intr_type to enable or disable interrupts.
thanks !!

Here I mentioned a bug in GPIO driver.  If we want to disable a GPIO's interrupt , we can use `gpio_set_intr_type(gpio_num, GPIO_INTR_DISABLE)`, and do not use `gpio_intr_disable(gpio_num)`.  But  this is just a work around way, the bug still exists.

Now, we want to fix it. we should polling the interrupt status register instead of the raw status register.

wingunder: @koobest At the moment I really cannot help you out, due to my heavy work load and private life. I'm sorry, but if I am really optimistic, I could eventually look at it by the end of September 2019. This is however _not_ a promise.

I'm sure that there are some other capable contributors around that could help you out.

koobest: Hi, @wingunder 
It doesn't matter :-),  I just want to try to merge your PR into the master, because you spent your time on it. we can fix this bug by ourselves. 

Thank you.

wingunder: >I just want to try to merge your PR into the master

@koobest, No problem, just 'git format-patch' and 'git am' my commits onto your bugfix branch and continue. It shouldn't be too difficult to fix the conflicts.

wingunder:
Closing.

PR #2937, opened 2019-01-11
Pr/strict prototypes (IDFGH-757)

M1cha:
While in C++ `()` is the same as `(void)`, in C it's not and can be quite dangerous.


M1cha:
It's fine if you rebase it and solve the conflicts, anything else would be a catch-up game with new commits for a big change like this.

Alvin1Zhang:
@M1cha Thanks for the contribution and sorry for the slow turnaround. The PR has been merged, thanks.

PR #2940, opened 2019-01-11
docs(en): "Adding user to dialout on Linux" added specific command for Arch linux

ShellAddicted:
Improvement of English docs, added specific command for arch linux where `dialout` group doesn't exist (by default) and `/dev/ttyUSB0` is managed by `uucp` group.


krzychb:
Merged in aa36496, thank you @ShellAddicted!

PR #2944, opened 2019-01-13
corrected '$PATH' export order in docs (en)

lenhart:
fixed order of '$PATH' export. xtensa-esp32-elf/bin was added at beginning of '$PATH' instead of end (which the docs later suggested to check w/ printenv

### DIFF
--- a/docs/en/get-started-cmake/add-idf_path-to-profile.rst
+++ b/docs/en/get-started-cmake/add-idf_path-to-profile.rst
@@ -43,7 +43,7 @@ Linux and MacOS
 Set up ``IDF_PATH`` and add ``idf.py`` to the PATH by adding the following two lines to your ``~/.profile`` file::
 
     export IDF_PATH=~/esp/esp-idf
-    export PATH="$IDF_PATH/tools:$PATH"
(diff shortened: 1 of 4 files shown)


lenhart:
Also on these pages: You don't have to log off and on to apply PATH changes. 
It suffices to do 'source ~/.profile'.
Would this be useful to change in the docs? 
(Only downside: if you accidentally did multiple "export PATH=... " it won't reset the path variable.)

projectgus:
Thanks for correcting this, @lenhart.

> Also on these pages: You don't have to log off and on to apply PATH changes.
> It suffices to do 'source ~/.profile'.

Yes. If I remember correctly, we had some discussion about this and it was a conscious decision to simply the guide. If you know how ~/.profile works then you'll do this anyhow regardless of whether it's in the docs. If you don't know how this works, it can easily become confusing when you run "source" in one terminal and then later you open a new terminal and ESP-IDF suddenly doesn't work in it. Rather than explain all the various implications, we chose to explain the one straightforward method.

krzychb:
Hi @lenhart after some discussion the order of items in PATH have been standardized across documents in a separate MR. Thank you for initiating this update!

PR #2987, opened 2019-01-24
FIX: Updating the missing python packages section

kolipakakondal:
Adding Install the ``Required Python Packages section`` to the CMake get started guide


kolipakakondal:
I'll create MR in gitlab

PR #2996, opened 2019-01-28
Fix log tag comparison when using variable tags (such as non-literals). (IDFGH-897)

DavidAntliff:
Setting a log level by tag does not work when the tag is provided as a variable (rather than a literal string). This is because there is an incorrect equality performed in `esp_log_level_set()`. A copy of the tag is made but the contents of this copy is not compared - instead the tag _pointers_ are compared, which does not work for non-literal tags.

Refer to this forum post: https://www.esp32.com/viewtopic.php?f=13&t=9004

### DIFF
--- a/components/log/log.c
+++ b/components/log/log.c
@@ -164,7 +164,7 @@ void esp_log_level_set(const char* tag, esp_log_level_t level)
 #ifdef LOG_BUILTIN_CHECKS
         assert(i == 0 || s_log_cache[(i - 1) / 2].generation < s_log_cache[i].generation);
 #endif
-        if (s_log_cache[i].tag == tag) {
+        if (strcmp(s_log_cache[i].tag, tag) == 0) {
             s_log_cache[i].level = level;
             break;
         }


DavidAntliff:
Looks like this was fixed on master in 750c8c00c8 - can we have this fix backported to 3.0.7 please?

projectgus:
@DavidAntliff Thanks for being patient while someone got back to you.

Our policy is only to backport high severity bugs, to reduce the amount of code churn and reduce the chance of regressions.

Is updating to ESP-IDF v3.1.x a possibility for you?

DavidAntliff:
@projectgus looks like @igrr  may have already backported it to v3.0 (thank you) however I'm happy to start trialing v3.1 anyway. Thanks for getting back to me. I think this can be closed now.

projectgus:
@DavidAntliff It seems like @igrr is more generous than I am. ;).

That's great, I'm happy that this is a good outcome. Thanks again for being patient.

PR #3023, opened 2019-02-02
tcp_transport: remove mention of mqtt and minor fix (IDFGH-847)

aitjcize:
The websocket transport shouldn't have any mention of mqtt.

### DIFF
--- a/components/tcp_transport/transport_ws.c
+++ b/components/tcp_transport/transport_ws.c
@@ -97,7 +97,6 @@ static int ws_connect(esp_transport_handle_t t, const char *host, int port, int 
                          "Host: %s:%d\r\n"
                          "Upgrade: websocket\r\n"
                          "Sec-WebSocket-Version: 13\r\n"
-                         "Sec-WebSocket-Protocol: mqtt\r\n"
                          "Sec-WebSocket-Key: %s\r\n"
                          "User-Agent: ESP32 Websocket Client\r\n\r\n",
                          ws->path,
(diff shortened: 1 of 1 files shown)


bverhoeven:
Hi there, while I agree that the `Sec-WebSocket-Protocol: mqtt` header shouldn't be used in a "generic" web socket transport like this, it seems that this change will break MQTT over web sockets for the ESP-IDF if it's applied without changes.

The MQTT implementation uses transport_ws when connecting over websockets:

https://github.com/espressif/esp-mqtt/blob/f08f3b678717865234637164a29ed3a63e756ca7/mqtt_client.c#L320-L329

As per the [MQTT specification](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718127) MQTT over web sockets would require us to send "mqtt" when using mqtt:

> The client MUST include “mqtt” in the list of WebSocket Sub Protocols it offers [MQTT-6.0.0-3].  

And it seems that this change will break that. I do think the MQTT implementation needs to be fixed too before applying this change, or we will be breaking protocol.

david-cermak:
Hi @aitjcize 
I agree with bverhoeven, removing mqtt as subprotocol would break current mqtt over ws (generally it'd probably work in most cases but would not comply with rfc).

I'm thinking of two options to go forward with this PR:

1) Leave the subprotocol as is, which I think is alright as mqtt client is the only place in idf which uses it, plus we're planning on new ws-client (to come in 4.0, already internally reviewed) which will have to tackle this issue.

2) Parametrize subprotocol and modify esp-mqtt to use the correct protocol. These changes need to come (=be merged) at the same time, as one would break the other.

...feel free to suggest any other approach

Thank you for your contribution, even if you choose the minimal update (1st option), these small changes are useful as they help making the code cleaner and more maintainable!

components/tcp_transport/transport_ws.c
```diff
@@ -97,7 +97,6 @@ static int ws_connect(esp_transport_handle_t t, const char *host, int port, int
                          "Host: %s:%d\r\n"
                          "Upgrade: websocket\r\n"
                          "Sec-WebSocket-Version: 13\r\n"
-                         "Sec-WebSocket-Protocol: mqtt\r\n"
```
david-cermak: Please either leave the line or sync with mqtt for parametric configuration (as mentioned in generic comment)

Alvin1Zhang:
@aitjcize Thanks for the contribution and sorry for the slow turnaround. Would you please help share if any updates for this PR? Thanks.

aitjcize:
I think I will send a separate PR instead of this one, to create a separate web socket module instead of modifying the existing transport used for Mqtt.

PR #3061, opened 2019-02-17
Added check for over-sized heap allocation

gregjesl:
## Problem
The HTTPS OTA updater will fail if there is insufficient memory for the OTA buffer.  

## Solution
I've added a check that compares the largest available heap size with the buffer size being requested.  If the buffer size being requested is larger than the largest available heap size, the buffer is resized appropriately.


mahavirj:
@gregjesl 

Can you please help with scenario which required this fix?

If statistics of heap are such that it can not satisfy allocation for OTA buffer, is that good point for system OTA? I wonder if explicit failure is rather better way for early notification. Even for default WiFi configuration, it needs significant amount of free heap (in order of 8-10K, with each buffer size exceeding 1600 bytes each), so there is high probability of OTA failure even if this fix is applied.

gregjesl:
@mahavirj I'm actually going to close this pull request because the pull request accidentally contained some tweaks specific to our application.  If I get some free time I'll resubmit using a proper branch.  

> Can you please help with scenario which required this fix?

Absolutely.  Our binaries are ~1.3 MB.  Using the default buffer size resulted in downloads that took more than 5 minutes.  We are using Azure Functions as the file server; the default timeout for an HTTP request using Azure Functions is 5 minutes meaning the OTA update was being interrupted.  Increasing the buffer size increased the download speed ~20%; however, we're running an HTTP server and an MQTT client that uses SSL among other tasks, so our memory is tight.  We saw on some occasion that our "large" buffer failed to allocate. 

> If statistics of heap are such that it can not satisfy allocation for OTA buffer, is that good point for system OTA? I wonder if explicit failure is rather better way for early notification. 

I can't speak for other applications but in our application OTA failures are very bad.  If we can't update the unit, then we run the risk of "bricking" units.  A good testing regimen will reduce the risk but we want to make an OTA failure risk as low as possible.

PR #3073, opened 2019-02-19
Add another auto-generated file to .gitignore.

sarenameas:
* Update tools/kconfig/.gitignore for auto-generated file *.o.


igrr:
Thanks for the PR, @sarenameas, and sorry for the slow turn-around. Has been cherry-picked as e24ac7a!

PR #3096, opened 2019-02-24
Compilation error fix: invalid conversion

itay-grudev:
```
base64.cpp:42:6: error: invalid conversion from 'void*' to 'unsigned char*' [-fpermissive]
base64.cpp:119:12: error: invalid conversion from 'void*' to 'unsigned char*' [-fpermissive]
```

### DIFF
--- a/components/wpa_supplicant/src/wpa2/utils/base64.c
+++ b/components/wpa_supplicant/src/wpa2/utils/base64.c
@@ -39,7 +39,7 @@ unsigned char * base64_encode(const unsigned char *src, size_t len,
 	olen++; /* nul termination */
 	if (olen < len)
 		return NULL; /* integer overflow */
-	out = os_malloc(olen);
+	out = (unsigned char*)os_malloc(olen);
 	if (out == NULL)
 		return NULL;
 
@@ -116,7 +116,7 @@ unsigned char * base64_decode(const unsigned char *src, size_t len,
 		return NULL;
 
 	olen = count / 4 * 3;
-	pos = out = os_malloc(olen);
+	pos = out = (unsigned char*)os_malloc(olen);
 	if (out == NULL)
 		return NULL;
 


projectgus:
Hi @itay-grudev ,

Thanks for submitting this! I'm curious, did you find this problem when compiling base64.c as C++? I noticed the warning lines in the log you posted are for `base64.cpp`.

Explicit casting from `void *` is not required in C, but it is required in C++. Explicit casts of this kind are sometimes discouraged in C, because they can mask other problems (here, if the return type of `os_malloc()` was not `void *` then the cast would mask the warning that this is an invalid conversion. 

Whereas in C++ the type system is stronger, so you can have nicer type checks (and some C++ codebases make all C-style casts into warnings). 

BTW, if you're looking for a good base64 algorithm to use from C++ code, I strongly recommend the mbedTLS base64 API: https://github.com/espressif/mbedtls/blob/19eb57f7f7b2f4312d497ddbcb5f104fc6877c70/include/mbedtls/base64.h#L36

(The mbedTLS API can be called directly from C++ and produces correct results. The base64 algorithm in wpa_supplicant has some quirks around newline encodings. We actually plan to make all of the wpa_supplicant "utils" into private APIs in the next major ESP-IDF release, as there are better alternatives for almost all of them.)

itay-grudev:
The inconsistency you noticed is because I hand-crafted the compiler log excerpt when I posted this as I had already discarded my original log. Apparently I didn't do a very good job at it.

The reason why I probably got the error in the first place is because I was compiling my project with: `-Wall -pedantic`.

And thanks for the tip. The mbedTLS API is probably a much better choice. Thanks.

itay-grudev:
I can't reproduce the error any more, but I changed so much of my code and build system tweaks that I'm not surprised. Feel free to drop this if you don't want it.

projectgus:
> The reason why I probably got the error in the first place is because I was compiling my project with: -Wall -pedantic. It was a while ago and I can't recall the exact circumstances.

That's not it. As mentioned above, this is only ever a warning in C++ not C (as it's often an anti-pattern in C): https://gcc.godbolt.org/z/VhCYLA

(If you change the language to C++ in compiler explorer, you'll see it fail immediately even without any additional flags.)

Maybe at some point the tweaked build system was compiling some C files with g++?

> The mbedTLS API is probably a much better choice. Thanks.

Awesome. :)

> I can't reproduce the error any more, but I changed so much of my code and build system tweaks that I'm not surprised. Feel free to drop this if you don't want it.

Thanks.

Issue #11, opened 2016-09-15
Installation step 2: not a valid identifierline 89: export: `sha1


krzychb:
Hi Team,

Following [Step 2: Getting the esp-idf repository from github](https://github.com/espressif/esp-idf/blob/master/docs/windows-setup.rst#step-2-getting-the-esp-idf-repository-from-github) of the installation process on Windows 7 x64,  I got the following two messages regarding sha1 (at the end of the full log below):

```
$ git clone --recursive https://github.com/espressif/esp-idf.git
Cloning into 'esp-idf'...
remote: Counting objects: 3162, done.
remote: Total 3162 (delta 0), reused 0 (delta 0), pack-reused 3162
Receiving objects: 100% (3162/3162), 5.05 MiB | 1.03 MiB/s, done.
Resolving deltas: 100% (1695/1695), done.
Checking connectivity... done.
Submodule 'components/esp32/lib' (https://github.com/espressif/esp32-wifi-lib.git) registered for path 'components/esp32/lib'
Submodule 'components/esptool_py/esptool' (https://github.com/themadinventor/esptool.git) registered for path 'components/esptool_py/esptool'
Cloning into '/esp-idf/components/esp32/lib'...
Cloning into '/esp-idf/components/esptool_py/esptool'...
': not a valid identifierline 89: export: `sha1
': not a valid identifierline 89: export: `displaypath
Submodule path '': checked out '1303c92c1056b7f59b95360b58e70a21cb4a93e1'
': not a valid identifierline 89: export: `sha1
': not a valid identifierline 89: export: `displaypath
Submodule path '': checked out '7c84dd433512bac80e4c01c569e42b4fe76646a7'
```
I have tried this step twice with the same result.

The `/esp-idf/components/esptool_py/esptool` directory contains some files and sub directories.
I did not find any leads to this error on `https://github.com/themadinventor/esptool` repository.

Should I be concerned?

hardikharpal:
Hi @krzychb,
Until u got solution from esp32 team,try cloning git by desktop git app.I had some issue to but with desktop git app it worked.

hardikharpal:
Hi @krzychb,
Until u got solution from esp32 team,try cloning git by desktop git app.I had some issue too but with desktop git app it worked.

igrr:
Seen this message on colleague's windows computer as well. It looked like everything was actually checked out correctly.
[...]

Issue #10, opened 2016-09-13
project build error


hardikharpal:
![image](https://cloud.githubusercontent.com/assets/12824670/18468169/5c966a94-79c0-11e6-8814-c49ef7ed7335.png)

igrr:
Can you please do `cat /c/ESP/esp-idf/tools/kconfig/zconf.gperf` and post the output?

hardikharpal:
here it is,
`Hardik Harpal@Hardik MSYS /c/esp-idf-template
$ cd C:/esp-idf-template

Hardik Harpal@Hardik MSYS /c/esp-idf-template
$ cat /c/ESP/esp-idf/tools/kconfig/zconf.gperf
%language=ANSI-C
%define hash-function-name kconf_id_hash
%define lookup-function-name kconf_id_lookup
%define string-pool-name kconf_id_strings
%compare-strncmp
%enum
%pic
%struct-type

struct kconf_id;

static const struct kconf_id *kconf_id_lookup(register const char *str, register unsigned int len);

%%
mainmenu,       T_MAINMENU,     TF_COMMAND
menu,           T_MENU,         TF_COMMAND
endmenu,        T_ENDMENU,      TF_COMMAND
source,         T_SOURCE,       TF_COMMAND
choice,         T_CHOICE,       TF_COMMAND
endchoice,      T_ENDCHOICE,    TF_COMMAND
comment,        T_COMMENT,      TF_COMMAND
config,         T_CONFIG,       TF_COMMAND
menuconfig,     T_MENUCONFIG,   TF_COMMAND
help,           T_HELP,         TF_COMMAND
---help---,     T_HELP,         TF_COMMAND
if,             T_IF,           TF_COMMAND|TF_PARAM
endif,          T_ENDIF,        TF_COMMAND
depends,        T_DEPENDS,      TF_COMMAND
optional,       T_OPTIONAL,     TF_COMMAND
default,        T_DEFAULT,      TF_COMMAND, S_UNKNOWN
prompt,         T_PROMPT,       TF_COMMAND
tristate,       T_TYPE,         TF_COMMAND, S_TRISTATE
def_tristate,   T_DEFAULT,      TF_COMMAND, S_TRISTATE
bool,           T_TYPE,         TF_COMMAND, S_BOOLEAN
boolean,        T_TYPE,         TF_COMMAND, S_BOOLEAN
def_bool,       T_DEFAULT,      TF_COMMAND, S_BOOLEAN
[...]

Issue #3057, opened 2019-02-16
pm_esp32.c:426 IntegerDivideByZero (IDFGH-631)


coollofty:
occasionally encountered once, then no longer encountered:

Guru Meditation Error: Core  0 panic'ed (IntegerDivideByZero). Exception was unhandled.
Core 0 register dump:
PC      : 0x400821e8  PS      : 0x00060031  A0      : 0x800824df  A1      : 0x3ffb0640
0x400821e8: update_ccompare at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:426

A2      : 0x00000003  A3      : 0x00060023  A4      : 0x00000001  A5      : 0x00000001
A6      : 0x00060021  A7      : 0x00000000  A8      : 0x00000000  A9      : 0x02741217
A10     : 0x00000028  A11     : 0x7cf898f2  A12     : 0x00060021  A13     : 0x3ffd0d60
A14     : 0x00000003  A15     : 0x00060b23  SAR     : 0x00000013  EXCCAUSE: 0x00000006
EXCVADDR: 0x00000000  LBEG    : 0x00000000  LEND    : 0x00000000  LCOUNT  : 0x00000000
Core 0 was running in ISR context:
EPC1    : 0x400821e8  EPC2    : 0x400821dc  EPC3    : 0x00000000  EPC4    : 0x00000000
0x400821e8: update_ccompare at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:426

0x400821dc: update_ccompare at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:426

Backtrace: 0x400821e8:0x3ffb0640 0x400824dc:0x3ffb0660 0x40085934:0x3ffb0680 0x4019a6f7:0x00000000
0x400821e8: update_ccompare at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:426

0x400824dc: esp_pm_impl_isr_hook at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:462

0x40085934: _xt_lowint1 at E:/esp32/esp-adf/esp-idf/components/freertos/xtensa_vectors.S:1105

0x4019a6f7: esp_pm_impl_waiti at E:/esp32/esp-adf/esp-idf/components/esp32/pm_esp32.c:487

igrr:
Could you please attach your sdkconfig and mention the IDF commit ID?

[...]

PR #3110, opened 2019-02-28
pm: Fix atomicity of rtos idle lock acquiring (IDFGH-876)

tim-nordell-nimbelink:
It's possible for esp_pm_impl_isr_hook(...) to be nested due to the fact
that interrupts are nested on the ESP32.  To fix this we need to place the
acquiring of the lock into a critical section to ensure it does not get
nested on the system, otherwise the system will never release the idle
lock when this occurs and will not go into lower power states.

A sample backtrace encountering this (the code was instrumented to go into
a while(1) loop when the condition was hit to get this backtrace) from
commit d7a7a683:

    #0  leave_idle () at esp-idf/components/esp32/pm_esp32.c:444
    #1  0x4008143a in esp_pm_impl_isr_hook () at esp-idf/components/esp32/pm_esp32.c:473
    #2  0x40082750 in _xt_medint2 () at esp-idf/components/freertos/xtensa_vectors.S:1243
    #3  0x4000bff0 in ?? ()
    #4  0x40090bb0 in vTaskExitCritical (mux=0x3ffbd230) at esp-idf/components/freertos/tasks.c:4304
    #5  0x40081758 in esp_pm_lock_acquire (handle=0x3ffbd218) at esp-idf/components/esp32/pm_locks.c:126
    #6  0x40081399 in leave_idle () at esp-idf/components/esp32/pm_esp32.c:440
    #7  0x4008143a in esp_pm_impl_isr_hook () at esp-idf/components/esp32/pm_esp32.c:473
    #8  0x400826b8 in _xt_lowint1 () at esp-idf/components/freertos/xtensa_vectors.S:1154
    #9  0x400d14b0 in esp_pm_impl_waiti () at esp-idf/components/esp32/pm_esp32.c:483
    #10 0x400d2c77 in esp_vApplicationIdleHook () at esp-idf/components/esp32/freertos_hooks.c:63
    #11 0x40091008 in prvIdleTask (pvParameters=0x0) at esp-idf/components/freertos/tasks.c:3412
    #12 0x40090344 in vPortTaskWrapper (pxCode=0x40090ffc <prvIdleTask>, pvParameters=0x0) at esp-idf/components/freertos/port.c:143

### DIFF
--- a/components/esp32/pm_esp32.c
+++ b/components/esp32/pm_esp32.c
@@ -434,11 +434,13 @@ static void IRAM_ATTR update_ccompare()
 static void IRAM_ATTR leave_idle()
 {
     int core_id = xPortGetCoreID();
+    uint32_t state = portENTER_CRITICAL_NESTED();
     if (s_core_idle[core_id]) {
         // TODO: possible optimization: raise frequency here first
         esp_pm_lock_acquire(s_rtos_lock_handle[core_id]);
         s_core_idle[core_id] = false;
     }
+    portEXIT_CRITICAL_NESTED(state);
 }
 
 void esp_pm_impl_idle_hook()


tim-nordell-nimbelink:
ping.  (Mainly to get an internal Jira ticket assigned per the workflow added after this pull request was created.)

igrr:
Thanks for submitting this @tim-nordell-nimbelink! As we were running some stress tests with this change applied, we ran into a rare crash which is still being investigated. The fix itself looks correct, but we want to make sure we understand the cause of this new crash before we merge this. Will update when we have some progress.

tim-nordell-nimbelink:
Hi @igrr -

Thanks for looking it over.  Hopefully you can find the root cause of the new crash that you're observing!  I can understand that a crash that happens rarely is hard to debug, and _testing is always a good thing._

Good luck!
Tim

igrr:
@tim-nordell-nimbelink The issues I have previously observed were due to the fact that preemption may happen not only inside `leave_idle` function, but also inside the `update_ccompare` function (which is called from `esp_pm_impl_isr_hook`). This resulted in issues such as #3057, when the following sequence of events occurs:

1. Level 1 interrupt happens, `esp_pm_impl_isr_hook` is called
2. Hook runs up to `update_ccompare` function
3. Level 2 interrupt happens on the same CPU, `esp_pm_impl_isr_hook` is called again.
4. `update_ccompare` returns, `s_need_update_ccompare[core_id]` is set to false.
5. Other CPU (which was waiting for s_need_update_ccompare[core_id] to become `false`) sets `s_ccount_div = 0` (this happens in `on_freq_update` function).
6. Level 2 interrupt completes and level 1 interrupt resumes in `update_ccompare` function
7. Integer division by 0 happens in `update_ccompare`, because `s_ccount_div` is now zero.

Please check the attached patch, it moves the interrupt lock into `esp_pm_impl_isr_hook`. It solves the issues we have observed. Please let me know if you have any concerns, if not we will merge that instead of your PR.
[4775.patch.zip](https://github.com/espressif/esp-idf/files/3079450/4775.patch.zip)

tim-nordell-nimbelink:
Hi @igrr -

It should still maintain the fix for the issue I saw; I'm not running the update_ccompare() code here since I'm running on a single core for power consumption reasons.

I do think I see a potential problem within on_freq_update(...) though - is it possible for both cores to run this code simultaneously?  If so, then:

1. Core 1 and 2 are past setting s_ccount_div to old_ticks_per_us
3. Core 2 gets an an interrupt so it doesn't get as far as notifying core 1
4. Core 1 notifies core 2
5. Core 2 handles update of ccompare from interrupt
6. Core 2 gets another interrupt before notifying core 1
7. Core 1 finishes and sets s_ccount_div to 0
8. Core 2 notifies core 1 of update for ccompare[...]
9. Core 1 goes into interrupt for handling ccompare(...), and has a divide by 0

Another scenario:

1. Core 1 and 2 are past setting s_ccount_div to old_ticks_per_us
2. Core 2 gets an interrupt before reaching update_ccompare()
3. Core 1 notifies core 2 for updating
4. Core 2 processes interrupt for updating ccompare()
5. Core 1 sets s_ccount_div to 0
6. Core 2 exits interrupt service routines and gets into update_ccompare(...) routine, but s_ccount_div is set to 0 at this point so it faults.

It'd be a race condition for it to occur with enough interrupt activity at just the right time, but I bet it's possible.

- Tim

igrr:
@tim-nordell-nimbelink Sorry that I haven't noticed your comment before.

> potential problem within on_freq_update(...) though - is it possible for both cores to run this code simultaneously?

I think this situation should not happen due the check at the top of `do_switch` function. If a frequency switch is already in progress, then the other CPU will either cancel its switch or wait until the first one is completed.

The patch attached above has been merged in d31ee80. Please let me know if you see any related issues. And thanks for contributing!

PR #3118, opened 2019-03-02
lwip: enable ipv6 dns server retrieval from ndp (IDFGH-1769)

mrrosen:
Combined with https://github.com/espressif/esp-lwip/pull/5, enables IPv6 DNS server address retrieval from Router Advertisements in NDP. Macro was missing from settings in port header file to enable this feature, so networks only supporting IPv6 (without DHCPv6 which doesnt seem supported yet in lwip used by ESP IDF) or only with IPv6 DNS servers can now be used as the DNS servers for lookups. Previously, IPv4 address was needed (and dhcp returns good DNS servers) to actually do a DNS lookup while all other network traffic was using IPv6.

### DIFF
--- a/components/lwip/port/esp32/include/lwipopts.h
+++ b/components/lwip/port/esp32/include/lwipopts.h
@@ -640,6 +640,13 @@
  */
 #define LWIP_IPV6                       1
 
+/**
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@mrrosen Thanks for the contribution and sorry for the slow turnaround. We will look into PR. Thanks.

mrrosen:
@Alvin1Zhang any word/thoughts on this change?

PR #3129, opened 2019-03-05
Easier use of -Wsign-conversion warnings needs changes in ESP-IDF (IDFGH-849)

PerMalmberg:
I've enabled -Wsign-conversion and -Wconversion when compiling and stumbled upon these two required changes.

GPIO_NUM_NC is, as the comment indicates, to specify a not connected pin. I'm sure I've seen -1 used for this in other IDF-code too so lets give it a proper enum value, also removing the warning that -1 isn't a valid gpio_num_t value.

The cast in xPortGetCoreID() can perhaps be improved by defining `id` as uint32_t directly - I can't read the ASM so I can't tell if doing that would change anything.

I'm creating this as a PR-draft for now, might be more things that need changing too.

### DIFF
--- a/components/driver/include/driver/gpio.h
+++ b/components/driver/include/driver/gpio.h
@@ -128,6 +128,7 @@ extern "C" {
 #define GPIO_IS_VALID_OUTPUT_GPIO(gpio_num)      ((GPIO_IS_VALID_GPIO(gpio_num)) && (gpio_num < 34))         /*!< Check whether it can be a valid GPIO number of output mode */
 
 typedef enum {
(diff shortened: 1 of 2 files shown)


PerMalmberg:
Turns out no more changes were needed, opening for review.

PerMalmberg:
@igrr thanks for accepting the change, though I'm wondering why this PR is till open? Do PRs no longer auto-close when merged?

projectgus:
@PerMalmberg The PR was cherry-picked with a small change (xPortGetCoreID() parameter type changed rather than using the cast). PRs only auto-close if they're 100% equivalent to merging.

Thanks for sending these fixes. :)

PR #3138, opened 2019-03-06
Http server default config

gregjesl:
## Problem

Brace-enclosed initializer lists don't always work with C++ compilers:
```
error: no match for 'operator=' (operand types are 'httpd_config_t {aka httpd_config}' and '<brace-enclosed initializer list>')
```

## Solution
Default initializer is now a static inline function


requirements.txt
```diff
@@ -7,5 +7,5 @@ setuptools
 #
 pyserial>=3.0
 future>=0.15.2
-cryptography>=2.1.4
```
dobairoland: I'm afraid that this change cannot be accepted because the `nvs_flash` component relies on this version. Older versions produce the following error: `ValueError: Invalid key size (512) for AES.`

gregjesl:
@dobairoland It looks like I made a pull request from the wrong branch.  I'll close this and open a new pull request from the right branch.

igrr:
An alternative (smaller and less invasive) change would be to add a C-cast to `httpd_config_t` in macro definition, before the opening curly brace. This should work both forbC and C++, I think.

gregjesl:
> An alternative (smaller and less invasive) change would be to add a C-cast to httpd_config_t in macro definition, before the opening curly brace. This should work both for C and C++, I think.

@igrr I don't use brace-enclosed initializer lists often so I'm not knowledgeable enough to know if that will work; I'll certainly give a try.  Thank you for the suggestion.

PR #3148, opened 2019-03-08
Update component/coap to libcoap version release-4.2.0 (IDFGH-739)

mrdeep1:
This takes the code up to the latest released version of libcoap.

As there have been API changes, coap_client and coap_server in
examples/protocols have been updated to use the new APIs.

Further information on the new libcoap APIs can be found at
https://libcoap.net/doc/reference/4.2.0/

coap_client has been updated to handle BLOCK2 responses from
"coap://californium.eclipse.org"
coap_client has been modified to only send out one request (and wait for
the response)

coap_server has been updated to support Observe subscriptions, and well as
adding in PUT and DELETE handlers to work on the Espressif resource

coap_server and coap_client have had their stack sizes increased.

port/coap_io.c has been added, a copy of libcoap/src/coap_io.c with support
added for systems that do not have RFC 3542 section 20 support.

port/coap_io_socket.c has been removed as a lot of the code is now
replicated in different libcoap files.

Once this PR is place, then adding in DTLS will be a lot simpler (as a
separate PR)


jitin17:
@mrdeep1 Thanks for the update. We will review this on priority. On a side note, I would like to know whether you are working on adding support for DTLS for libcoap.

mrdeep1:
@jitin17 I am one of the libcoap developers and did this update to help someone who raised an issue on the libcoap github and they needed the later code.  I do not have an ESP environment for testing, but the user reports the example applications crashing on receipt of a packet.  Needless to say that the example code works fine when run within a standard libcoap environment in Linux (with esp stuff commented out).

It transpires that esp-idf lwip does not have RFC 3542 section 20 support which libcoap uses for its I/O - consequently the use of un-initialized data in struct mshdr's msg_control causes memory to get corrupted is my current thinking for the crashes being seen in a ESP environment.

I will update the libcoap I/O to not make use of the Ancillary Data (how it was done in the previous code may work, the use of Ancillary data was added before 2015, before I started work on libcoap), get the user needing this functionality to test it and then I will let you know it is good to go.  It may be easiest to push updates to this PR for the user to test.

In terms of DTLS, libcoap 4.2.0 supports OpenSSL, GnuTLS and TinyDTLS.  OpenSSL requires a minimum version of 1.1.0 which is not supported by the esp-idf openssl.  There are 2 ways forward here - make TinyDTLS a part of esp-idf, or add in support for mbedlts in libcoap.  The first may be the easiest, the second may be a better long term solution from libcoap's perspective.

mrdeep1:
@jitin17 This looks good to go now.  I have rebased it off the latest master  d34ca0b8977a8ddecd6c2b571c7cf9d8dc01aa90 
However, I see that there have been changes to other submodules in the interim for master
M       components/bt/lib
M       components/lwip/lwip
which are not part of this PR - I have ignored them and they are not a part of this patch.

jitin17:
@mrdeep1 I have raised a PR in our internal repo to accept this PR(that's how we accept PRs). It's up for review and will keep you posted about the updates.

mahavirj:
@mrdeep1 Change is now merged with master, we ended up adding small change to fine tune stack size in example code. Thanks.

Issue #2600, opened 2018-10-21
fatal error: 'libintl.h' file not found trying to install/run idf on OSX


bootrino:
I'm running OSX 10.13.6

I've followed the latest instructions for installing idf.

I've tried installing xcode, tried installing ports and then gettext and ncurses.

```
andrew-stuarts-macbook-pro:hello_world andrewstuartsupercoders$ make menuconfig
make[1]: /Users/andrewstuartsupercoders/esp/esp-idf/tools/kconfig/lxdialog/check-lxdialog.sh: Command not found
cc -c   -DLOCALE -MMD  /Users/andrewstuartsupercoders/esp/esp-idf/tools/kconfig/mconf.c -o mconf.o
In file included from /Users/andrewstuartsupercoders/esp/esp-idf/tools/kconfig/mconf.c:22:
/Users/andrewstuartsupercoders/esp/esp-idf/tools/kconfig/lkc.h:12:11: fatal error: 'libintl.h' file not found
# include <libintl.h>
          ^~~~~~~~~~~
1 error generated.
make[1]: *** [mconf.o] Error 1
andrew-stuarts-macbook-pro:hello_world andrewstuartsupercoders$
```

my env looks like:

```
andrew-stuarts-macbook-pro:hello_world andrewstuartsupercoders$ env
TERM_PROGRAM=iTerm.app
TERM=xterm-256color
SHELL=/bin/bash
TMPDIR=/var/folders/dw/60dbyvqd7v92pkm_515486x00000gw/T/
Apple_PubSub_Socket_Render=/private/tmp/com.apple.launchd.oxJh3e4Njr/Render
TERM_PROGRAM_VERSION=3.2.0
COMMAND_MODE=unix2003
__CF_USER_TEXT_ENCODING=0x0:0:0
[...]

PR #3166, opened 2019-03-12
Use /usr/bin/env bash for portability (IDFGH-708)

cemeyer:
bash is not always installed in /bin/bash.  This could be related to
issue #2600 (or at least, manifests with similar symptoms).

### DIFF
--- a/components/esp32/ld/elf_to_ld.sh
+++ b/components/esp32/ld/elf_to_ld.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 echo '/*'
 echo 'ESP32 ROM address table'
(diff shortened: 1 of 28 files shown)


cemeyer:
This change was performed mechanically, with:

```
$ find . -name \*.sh -exec sed -i "" -e 's|^#!.*bin/bash|#!/usr/bin/env bash|' {} +
```

(Note that BSD `sed -i ""` is basically identical to GNU `sed -i`.)

cemeyer:
Thanks, @igrr !

igrr:
@cemeyer No problem, sorry for the delay and sorry that I had to make a new commit rather than reusing the commit you have authored (it was a pain to rebase... again, due to the delay).

cemeyer:
No worries, this is definitely the kind of thing it makes sense to just re-apply mechanically instead of trying to rebase.

PR #3167, opened 2019-03-12
Added a name to typedef with anonymous struct (IDFGH-709)

huhasa:
To avoid warnings during compilation of gcc in cpp mode


projectgus:
@huhasa Can this be closed in favour of #3199?

PR #3175, opened 2019-03-13
A flag to control default watchdog feeding behavior (IDFGH-737)

digitalorder:
To configure watchdog task behavior there is only one flag at the moment: ```CONFIG_TASK_WDT_PANIC```. If it's set there will be a call for ```abort()``` function inside watchdog interrupt handler. This will result in device reboot with ```SW_CPU_RESET``` listed as a reason. This is not practical in every case, because user will not be able to distinguish between watchdog reboots and say OTA update related ones.

This PR introduces a flag that if set will not feed WDT inside IRQ handler. As a result, device will reboot after three ```CONFIG_TASK_WDT_TIMEOUT_S``` intervals with a reason ```TG0WDT_SYS_RESET```.

### DIFF
--- a/components/esp32/task_wdt.c
+++ b/components/esp32/task_wdt.c
@@ -64,6 +64,7 @@ struct twdt_config_t {
     twdt_task_t *list;      //Linked list of subscribed tasks
     uint32_t timeout;       //Timeout period of TWDT
     bool panic;             //Flag to trigger panic when TWDT times out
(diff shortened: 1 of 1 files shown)


igrr:
Please note that you can get the "correct" reason from esp_reset_reason function, in case Task Watchdog resets the chip.

digitalorder:
Haven't seen that one since I'm working in v3.1.3 branch at the moment. Existing feature solves the issue, so I'm closing this PR.

PR #3189, opened 2019-03-15
feat(heap) make heap size fixed and configurable (IDFGH-780)

GautierAtWork:
Currently the heap size is defined at link time as "the remaining ram", what is not deterministic.

Once the product is released, and a minor revision is created, it should be ensured that critical parameters of the system are not impacted.
With a non deterministic heap size, no one can guarantee that connectivity is not impacted by any code change, even a minimal innocent one.

This patch makes the heap size configurable from Kconfig. Any change of the heap size is fully under control.

### DIFF
--- a/components/esp32/ld/esp32.common.ld
+++ b/components/esp32/ld/esp32.common.ld
@@ -175,8 +175,6 @@ SECTIONS
     *(COMMON)
     . = ALIGN (8);
     _bss_end = ABSOLUTE(.);
-    /* The heap starts right after end of this section */
-    _heap_start = ABSOLUTE(.);
   } > dram0_0_seg
 
   .flash.rodata :
--- a/components/esp32/ld/esp32.ld
+++ b/components/esp32/ld/esp32.ld
@@ -21,6 +21,12 @@
 #define CONFIG_BT_RESERVE_DRAM 0
 #endif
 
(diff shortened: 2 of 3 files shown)


negativekelvin:
I don't think that will work given that the heap actually fills various regions around used and reserved memory blocks

https://github.com/espressif/esp-idf/blob/80c5547b971128fdecbbcf6d8178196d35a15b73/components/soc/src/memory_layout_utils.c
https://github.com/espressif/esp-idf/blob/f1b4c18c2a5bc4be83c260605424880e4093b9c5/components/soc/esp32/soc_memory_layout.c

projectgus:
Hi @GautierAtWork,

I think @negativekelvin is more or less right, at least that it's a complex concept to have a "fixed heap size" on ESP32 because some memory regions are only added to the heap at runtime..

I think might be possible to reframe this as a "Fixed static RAM size" option which reserves a fixed contiguous block of memory in the linker script for .data+.bss, and doesn't allow any of that reserved memory to be used for heap:

* Heap size & layout will be constant regardless of static memory usage
* If static memory use grows beyond the "Fixed static RAM size" size, a linker error will occur.

Would that work for you in the same way?

Angus

PS We don't accept PRs into release branches. PRs should always target the master branch. We are happy to backport bugfixes from master into release branches, but as this isn't a bugfix we probably would not backport it.

GautierAtWork:
Closed in favor of https://github.com/espressif/esp-idf/pull/3222

PR #3199, opened 2019-03-18
Feature/typedef (IDFGH-796)

huhasa:


projectgus:
Hi @huhasa ,

Thanks for sending this PR. There's a lot of new  struct names here, I'm guessing this was done with an automated tool of some kind? Does the tool happen to be published?

May I ask what the purpose of naming these structs is? Is there a case where referring to a type as `struct TYPENAME_s` is preferable to referring to it as `TYPENAME`?


Angus

huhasa:
Thanks for sending this PR. There's a lot of new struct names here, I'm guessing this was done with an automated tool of some kind? Does the tool happen to be published?
- I was using Eclipse and a (big) regular expressions to add these struct names. Then an example application was compiled and solved any problems or names conflicts.

May I ask what the purpose of naming these structs is?
- If you use these header files in a Cpp context the following warning is given: warning: '<class>' has a field '<field>' whose type uses the anonymous namespace. See also https://stackoverflow.com/questions/24948158/warning-anonymous-namespace-inside-struct-using-function-pointer#24948460 for a explanation
- Some debugger don't like anonymous structs

Is there a case where referring to a type as struct TYPENAME_s is preferable to referring to it as TYPENAME?
- Only TYPENAME gave sometimes conflicts.
- The _s for struct matches the _t for typedef

projectgus:
Hi @huhasa ,

Thanks for explaining.

> If you use these header files in a Cpp context the following warning is given: warning:

I understand that this is the case if the typedef modifies an anonymous struct, for example in the example on SO the typedef for RETVAL makes that type into a pointer to an anonymous struct. [An explanation here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53184#c4) explains that if the typedef attaches a volatile or const modifier to the struct, it's the same situation.

But in the case of a simple anonymous struct typedef-ed directly to a type name, the compiler should apply the typedef name to the struct for linkage/namespacing purposes (see explanation at the above link to gcc bug), so it seems like this warning should not occur.

Do you have an example of the actual typedef(s) that triggered errors, and an example of usage? I'd much rather we fix problematic cases than modify a large number of source files, if this is an option.

> Some debugger don't like anonymous structs

Do you have an example of where this trips up a debugger, please? Is it the simple case, or the more complex modified type case?

huhasa:
The attached zip file [uart_esp32.zip](https://github.com/espressif/esp-idf/files/2991981/uart_esp32.zip) contains a simple example. When this example is compiled the following warning is given:
```In file included from .../UartEsp32.cpp:1:0:
.../UartEsp32.hpp:8:11: warning: 'hal::UartEsp32' has a field 'hal::UartEsp32::regs' whose type uses the anonymous namespace
     class UartEsp32
```
If you change line 21 in file uart_struct.h to:
`typedef volatile struct uart_dev_ {`
then no warning is given

I have no example of a debugger which doesn't work anymore which anonymous namespace. This problem was occurring with a debugger from ten of fifteen year ago. May be this no issue any more.

Proposal: I use a spi and uart device, which gave both warnings, so I can make pull request with a change for  these two files only.

projectgus:
Ah, of course - I should have realised this. All of the register _struct.h files have a `typedef volatile struct {` in them, so will have this problem in C++ (because of the volatile + anonymous struct).

> Proposal: I use a spi and uart device, which gave both warnings, so I can make pull request with a change for these two files only.

ESP-IDF should fix this for all the `*_struct.h` registers that have typedef-d a volatile anonymous struct. Feel free to update the PR, or if you'd prefer then I can make the commit to fix this.

huhasa:
I have changed all  the *_struct.h files and push this change

huhasa:
@projectgus  I tried to sign the license/cla but it is still complaining that the license is not signed. What is the problem?

projectgus:
@huhasa I think the commit in Git has two email addresses, and only one of them is associated with your GitHub account so it thinks the other author has not signed off.

Don't worry about this too much, they're clearly both you so I think we can take the CLA as signed (especially for such a small change).

projectgus:
@huhasa Thanks again for the contribution. I've cherry-picked this into our review and merge queue.

projectgus:
Cherry-picked as 4aac441, thanks!

PR #3222, opened 2019-03-25
feat(memory-map) allow fixed static ram and heap size boundaries (IDFGH-858)

GautierAtWork:
Currently the heap size is defined at link time as "the remaining ram", what is not deterministic.

Once the product is released, and a minor revision is created, it should be ensured that critical parameters of the system are not impacted.
With a non deterministic heap size, no one can guarantee that connectivity is not impacted by any code change, even a minimal innocent one.

This patch makes the static ram size configurable from Kconfig, and indirectly fixed the heap size. Any change of the heap size is fully under control.

### DIFF
--- a/components/esp32/Kconfig
+++ b/components/esp32/Kconfig
@@ -929,24 +929,24 @@ menu "ESP32-specific"
             This option depends on the CONFIG_FREERTOS_UNICORE option because RTC fast memory
             can be accessed only by PRO_CPU core.
 
-config USE_FIXED_STATIC_RAM_SIZE
-	bool "Use fixed static RAM size"
-	default n
-    help
-	    Without this option, the Dram part of the heap starts right after the .bss sections,
-		within the dram0_0 region. As a result, adding or removing some static variables,
-		indirectly decreases or reduces the heap size.
-		With this option, the Dram part of the heap starts right after the dram0_0 region,
-		where its length is set with FIXED_STATIC_RAM_SIZE
+    config ESP32_USE_FIXED_STATIC_RAM_SIZE
+        bool "Use fixed static RAM size"
+        default n
+        help
+            Without this option, the Dram part of the heap starts right after the .bss sections,
+            within the dram0_0 region. As a result, adding or removing some static variables,
+            indirectly decreases or reduces the heap size.
+            With this option, the Dram part of the heap starts right after the dram0_0 region,
+            where its length is set with ESP32_FIXED_STATIC_RAM_SIZE
 
-config FIXED_STATIC_RAM_SIZE
-	hex "Fixed Static RAM size"
-	default 0x1E000
-	range 0 0x2c200
-    depends on USE_FIXED_STATIC_RAM_SIZE
-	help
-	    Ram size dedicated for static variables.
-		Please note that the actual length will be reduced by BT_RESERVE_DRAM if configured.
+    config ESP32_FIXED_STATIC_RAM_SIZE
+        hex "Fixed Static RAM size"
+        default 0x1E000
+        range 0 0x2c200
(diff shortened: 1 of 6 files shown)


negativekelvin:
The runtime heap allocator does not use _heap_start it only uses _bss_end  so this may have no effect on heap size and only restrict dram0_0_seg length

GautierAtWork:
@negativekelvin Thanks for that. Seems it changed since the ESP-Idf version we are using.
I will have a look.

GautierAtWork:
@negativekelvin: fixed

components/esp32/Kconfig
```diff
@@ -929,6 +929,24 @@ menu "ESP32-specific"
             This option depends on the CONFIG_FREERTOS_UNICORE option because RTC fast memory
             can be accessed only by PRO_CPU core.
 
+config USE_FIXED_STATIC_RAM_SIZE
+	bool "Use fixed static RAM size"
+	default false
```
projectgus: Minor, but Kconfig syntax should be `default n` (this will also work but our Kconfig linter will complain about it.)

components/esp32/Kconfig
```diff
@@ -929,6 +929,24 @@ menu "ESP32-specific"
...
+		indirectly decreases or reduces the heap size.
+		With this option, the Dram part of the heap starts right after the dram0_0 region,
+		where its length is set with FIXED_STATIC_RAM_SIZE
+
+config FIXED_STATIC_RAM_SIZE
+	hex "Fixed Static RAM size"
+	default 0x1E000
+    depends on USE_FIXED_STATIC_RAM_SIZE
+	help
+	    Ram size dedicated for static variables. Should be smaller than 0x2c200.
```
projectgus: This can be achieved with the `range` directive in Kconfig itself (see some other Kconfig files for examples of this).

components/esp32/Kconfig
GautierAtWork: fixed.

components/esp32/Kconfig
GautierAtWork: fixed.

components/esp32/Kconfig
```diff
@@ -929,6 +929,25 @@ menu "ESP32-specific"
             This option depends on the CONFIG_FREERTOS_UNICORE option because RTC fast memory
             can be accessed only by PRO_CPU core.
 
+config USE_FIXED_STATIC_RAM_SIZE
```
igrr: We are adding additional checks for Kconfig option naming soon. Please prefix these new option names with `ESP32_` so that there is no conflict with the upcoming changes.

components/esp32/Kconfig
```diff
@@ -929,6 +929,25 @@ menu "ESP32-specific"
...
+	bool "Use fixed static RAM size"
+	default n
```
igrr: There is a mix of tabs and spaces here (tab on this line and spaces below). To be sure that the format of this Kconfig file is correct, please run `tools/check_kconfigs.py --directory components/esp32`.

GautierAtWork: fixed.

components/esp32/Kconfig
GautierAtWork: fixed.

projectgus:
Thanks for contributing this, @GautierAtWork . Thanks also for being patient while we worked through the review process.

I've squashed these commits into one and pushed them into our internal review & merge queue. The PR will be updated once the changes are merged and pushed to GitHub.

igrr:
Thanks for the PR @GautierAtWork, it has been cherry-picked as https://github.com/espressif/esp-idf/commit/542e544faa5977408d08d49406ace1d6f6e488cb.

PR #3239, opened 2019-04-01
Added _t to a couple of the NVS typedefs (IDFGH-891)

Frosticles:
Just went through and added _t to some typedefs in nvs.h to match the convention. This mainly helps with syntax highlighting, vscode recognises _t as a typedef and highlights it

### DIFF
--- a/examples/storage/ext_flash_fatfs/CMakeLists.txt
+++ b/examples/storage/ext_flash_fatfs/CMakeLists.txt
@@ -1,0 +1,6 @@
+# The following lines of boilerplate have to be in your project's CMakeLists
+# in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.5)
+
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+project(ext_flash_fatfs)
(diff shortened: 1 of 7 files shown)


components/nvs_flash/include/nvs.h
```diff
@@ -26,7 +26,7 @@ extern "C" {
 /**
  * Opaque pointer type representing non-volatile storage handle
  */
-typedef uint32_t nvs_handle;
+typedef uint32_t nvs_handle_t;
```
projectgus: We probably don't want to break all existing projects' code which uses nvs_handle, so at minimum we would want another typedef here to maintain compatibility.

Frosticles: Yep that's a good point, added back the original typedefs I replaced

Frosticles:
Ah ok I didn't realise the `_t` was originally a POSIX thing, I had just seen it used as the convention when declaring typedefs, but as @igrr says there seems to be more code in the IDF that uses `_t` than doesn't.

projectgus:
Thanks for making those changes, @Frosticles . I've cherry-picked this into our internal review & merge queue, with one small additional small change (added a backwards-compatible nvs_openmode to go with nvs_openmode_t).

This PR will be updated when it's merged.

PR #3304, opened 2019-04-12
Feature/deep sleep fast wake (IDFGH-967)

tim-nordell-nimbelink:
This patchset adds an optional compile-time flag to the bootloader to reduce the time spent in image validation when waking from deep sleep.  This changeset assumes that the image hasn't changed
since the last time the bootloader ran which may not always be a valid assumption to take depending on board design.  This can reduce the boot time to 10% of normal when an otherwise normal boot would be doing full image validation (including validating the secure boot signature).

There are a couple of potential corner cases that this would be bad for secure boot in which are documented inside of the Kconfig menu option, and it can only be enabled if insecure options are enabled in the bootloader.

### DIFF
--- a/components/bootloader/Kconfig.projbuild
+++ b/components/bootloader/Kconfig.projbuild
@@ -400,6 +400,26 @@ menu "Security features"
 
             Refer to https://docs.espressif.com/projects/esp-idf/en/latest/security/secure-boot.html before enabling.
 
+    config BOOT_SKIP_VALIDATE_OUT_OF_DEEP_SLEEP
+        bool "Skip image validation when exiting deep sleep"
+        depends on (SECURE_BOOT_ENABLED && SECURE_BOOT_INSECURE) || !SECURE_BOOT_ENABLED
+        select BOOTLOADER_RESERVE_RTC
+        default N
+        help
+            This option disables the normal validation of an image coming out of
+            deep sleep (checksums, SHA256, and signature).  This is a trade-off
+            between wakeup performance from deep sleep, and image integrity checks.
+
+            Only enable this if you know what you are doing.  It should not be used
+            in conjunction with using deep_sleep() entry and changing the active OTA
+            partition as this would skip the validation upon first load of the new
+            OTA partition.
+
+    config BOOTLOADER_RESERVE_RTC
+        hex
+        default 0x10 if BOOT_SKIP_VALIDATE_OUT_OF_DEEP_SLEEP
+        default 0
+
     config FLASH_ENCRYPTION_ENABLED
         bool "Enable flash encryption on boot (READ DOCS FIRST)"
         default N
--- a/components/bootloader_support/src/esp_image_format.c
+++ b/components/bootloader_support/src/esp_image_format.c
@@ -186,9 +186,9 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
 
     data->image_len = end_addr - data->start_addr;
     ESP_LOGV(TAG, "image start 0x%08x end of last section 0x%08x", data->start_addr, end_addr);
-    if (NULL != checksum && !esp_cpu_in_ocd_debug_mode()) {
+    if (NULL != checksum) {
         err = verify_checksum(sha_handle, checksum_word, data);
-        if (err != ESP_OK) {
+        if (err != ESP_OK && !esp_cpu_in_ocd_debug_mode()) {
             goto err;
         }
     }
--- a/components/bootloader_support/src/esp_image_format.c
+++ b/components/bootloader_support/src/esp_image_format.c
@@ -97,7 +97,7 @@ static esp_err_t __attribute__((unused)) verify_simple_hash(bootloader_sha256_ha
 static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_t *part, esp_image_metadata_t *data)
 {
 #ifdef BOOTLOADER_BUILD
-    bool do_load = (mode == ESP_IMAGE_LOAD);
+    bool do_load = (mode == ESP_IMAGE_LOAD || mode == ESP_IMAGE_LOAD_NO_VALIDATE);
 #else
     bool do_load = false; // Can't load the image in app mode
 #endif
@@ -105,6 +105,7 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
     esp_err_t err = ESP_OK;
     // checksum the image a word at a time. This shaves 30-40ms per MB of image size
     uint32_t checksum_word = ESP_ROM_CHECKSUM_INITIAL;
+    uint32_t *checksum = NULL;
     bootloader_sha256_handle_t sha_handle = NULL;
 
     if (data == NULL || part == NULL) {
@@ -125,17 +126,26 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
         goto err;
     }
 
-    // Calculate SHA-256 of image if secure boot is on, or if image has a hash appended
-#ifdef SECURE_BOOT_CHECK_SIGNATURE
-    if (1) {
+#ifdef BOOTLOADER_BUILD
+    if (mode != ESP_IMAGE_LOAD_NO_VALIDATE)
 #else
-    if (data->image.hash_appended) {
+    if (1)
 #endif
-        sha_handle = bootloader_sha256_start();
-        if (sha_handle == NULL) {
-            return ESP_ERR_NO_MEM;
+    {
+        checksum = &checksum_word;
+
+        // Calculate SHA-256 of image if secure boot is on, or if image has a hash appended
+#ifdef SECURE_BOOT_CHECK_SIGNATURE
+        if (1) {
+#else
+        if (data->image.hash_appended) {
+#endif
+            sha_handle = bootloader_sha256_start();
+            if (sha_handle == NULL) {
+                return ESP_ERR_NO_MEM;
+            }
+            bootloader_sha256_data(sha_handle, &data->image, sizeof(esp_image_header_t));
         }
-        bootloader_sha256_data(sha_handle, &data->image, sizeof(esp_image_header_t));
     }
 
     ESP_LOGD(TAG, "image header: 0x%02x 0x%02x 0x%02x 0x%02x %08x",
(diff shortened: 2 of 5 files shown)


tim-nordell-nimbelink:
There is another pitfall I didn't realize initially, too.  The bootloader probably *does* need to know the last successfully booted partition since otherwise the current patchset could end up booting an image that failed validation the last time it went through and checked.  E.g. if the check order is:

1. Corrupted OTA Slot 1
2. OTA Slot 2

The current patchset as-is would boot into OTA Slot 2 during initial power-up, but then boot into OTA Slot 1 coming out of deep sleep.  This needs to be addressed before being merged.  Unfortunately, there isn't any carved out area in memory for the bootloader to persist some state information through deep sleep that I'm aware of, and this would need to be added into the linker scripts for the bootloader and the main application.

tim-nordell-nimbelink:
I added a check that the partition matches the last known booted partition by the bootloader.  This places some information into RTC FAST RAM for this purpose, and reserves it from the normal linkage of the main applications.  This only works for deep sleeps which retain RTC FAST RAM.

tim-nordell-nimbelink:
I see now that someone else had proposed a similar change a while ago in #585, but they didn't go as far as removing all of the checks I removed in this patchset, nor was it an optional add-on via KConfig.  Some sample numbers from my current codebase:

1. ~750ms load time with public key validation
2. ~300ms load time with normal validation of image
3. ~50ms load time with this patch applied coming out of deep sleep

-- Tim

components/bootloader/Kconfig.projbuild
```diff
@@ -400,6 +400,26 @@ menu "Security features"
 
             Refer to https://docs.espressif.com/projects/esp-idf/en/latest/security/secure-boot.html before enabling.
 
+    config BOOT_SKIP_VALIDATE_OUT_OF_DEEP_SLEEP
+        bool "Skip image validation when exiting deep sleep"
+        depends on (SECURE_BOOT_ENABLED && SECURE_BOOT_INSECURE) || !SECURE_BOOT_ENABLED
+        select RTC_BOOTLOADER_RESERVED
```
igrr: Seems to be a non-existent option?

tim-nordell-nimbelink: @igrr -

Good catch.  This should have been BOOTLOADER_RESERVE_RTC, not RTC_BOOTLOADER_RESERVED.  I changed the naming order to more closely match the other option similar within the linker script - CONFIG_BT_RESERVE_DRAM - but missed one spot.  Thanks.

- Tim

tim-nordell-nimbelink:
> The feature looks good to me. Have asked a couple more colleagues to review, since this touches the image verification code.

No problem.

I just updated the pull request to point to the correct KConfig option.

 -- Tim

tim-nordell-nimbelink:
Hi @igrr -

I pushed another fix to this branch that isn't directly related, but touches some of the same lines as this patch.  E.g. that means either this patch needs to be applied first (and this branch based on that patch) or vice-versa.  This fix helps for when OpenOCD is attached to the system and you have flash validation turned on.  See the [new commit](https://github.com/espressif/esp-idf/pull/3304/commits/86aba4e91d869c0d431dd1016e683209c4401a1d) for details.  It's independent from the other commits in this series, technically, and I wasn't sure I could do a pull request in github based on another pull request as a base.  Otherwise it can be adapted to match the logic without the other patches in this series easily.

- Tim

components/bootloader_support/src/esp_image_format.c
```diff
@@ -225,7 +243,7 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
     }
 
 #ifdef BOOTLOADER_BUILD
-    if (do_load) { // Need to deobfuscate RAM
+    if (ram_obfs_value[0] != 0 && ram_obfs_value[1] != 0 && do_load) { // Need to deobfuscate RAM
```
projectgus: Minor, but can this be `(do_load && (ram_obfs_value[0] != 0 || ram_obfs_value[1] != 0))`?

One of the two ram_obfs_values will be zero once in every 2^31 boots, but both will only be zero once in every 2^64 boots. I like the second odds better. :)

components/bootloader_support/src/esp_image_format.c
```diff
@@ -125,17 +126,26 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
         goto err;
     }
 
-    // Calculate SHA-256 of image if secure boot is on, or if image has a hash appended
+#ifdef BOOTLOADER_BUILD
+    if (mode != ESP_IMAGE_LOAD_NO_VALIDATE)
+#else
+    if (1)
+#endif
```
projectgus: To simplify the code here and also for a more readable version of the `if(NULL != checksum)` check below, how about setting up a `bool do_verify` at the top of the function, which is true for non-bootloader builds and true for bootloader builds if `mode != ESP_IMAGE_LOAD_NO_VALIDATE`?

components/bootloader_support/src/esp_image_format.c
projectgus: (Or test the `do_verify` flag as suggested in the other comment, maybe? As ram_obfs_value being zero is being tested because it's a side effect of skipping verification, yes?)

components/bootloader_support/src/esp_image_format.c
```diff
@@ -193,15 +204,22 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
        rewritten the header - rely on esptool.py having verified the bootloader at flashing time, instead.
     */
     if (!is_bootloader) {
+#ifdef BOOTLOADER_BUILD
+        if (mode != ESP_IMAGE_LOAD_NO_VALIDATE)
+#else
+        if (1)
+#endif
```
projectgus: Same suggestion about `do_verify` applies here

components/bootloader_support/src/esp_image_format.c
```diff
@@ -398,13 +425,28 @@ static esp_err_t process_segment(int index, uint32_t flash_addr, esp_image_segme
 
 static esp_err_t process_segment_data(intptr_t load_addr, uint32_t data_addr, uint32_t data_len, bool do_load, bootloader_sha256_handle_t sha_handle, uint32_t *checksum)
 {
+    // If we are not loading, and the checksum is empty, skip processing this
+    // segment for data
+    if(!do_load && NULL == checksum)
```
projectgus: Same suggestion applies here. This function already has a lot of arguments but testing `if(!do_load && !do_verify)` is be more readable. Same applies in other places.

components/bootloader_support/src/esp_image_format.c
tim-nordell-nimbelink: > Minor, but can this be `(do_load && (ram_obfs_value[0] != 0 || ram_obfs_value[1] != 0))`?
> 
> One of the two ram_obfs_values will be zero once in every 2^31 boots, but both will only be zero once in every 2^64 boots. I like the second odds better. :)

It's guaranteed to be non-zero on a normal boot - take a look a the existing code that generates ram_obfs_value[x].  It has the following loop:

```
    while (ram_obfs_value[0] == 0 || ram_obfs_value[1] == 0) {
        ....
    }
```

tim-nordell-nimbelink:
> Hi @tim-nordell-nimbelink ,
> 
> This looks like a very helpful feature for boot times, thanks! I'm guessing the majority of the 90% boot time savings comes from skipping secure boot signature check? Did you happen to do any benchmarking with secure boot disabled?

The majority of time is spent validating the images, whether for the SHA hash, or for the public key signature validation.  This patch skips touching any of the segments that are purely mapped into memory (mmap only), so the time savings are larger for images that have more flash content than RAM content that needs to be copied.  As mentioned earlier in the comments, with this patch, the a validation skip boot takes ~20-30ms, a normal boot (without secure boot but with the SHA checksum matching) is ~250-300ms, and a secure boot validation is in the 700ms range for my ~950KiB image.

-- Tim

components/bootloader_support/src/esp_image_format.c
projectgus: Oh, my mistake! Thanks.

projectgus:
>  As mentioned earlier in the comments, with this patch, the a validation skip boot takes ~20-30ms, a normal boot (without secure boot but with the SHA checksum matching) is ~250-300ms

@tim-nordell-nimbelink Thanks for clarifying, sorry I missed the comment with the more detailed numbers.

components/bootloader_support/src/esp_image_format.c
tim-nordell-nimbelink: I'm just realizing this shortcut to skipping an unneeded mmap() invocation probably should also check that `sha_handle != NULL` as well.  (That way it's possible to use the function for a SHA check without checksum.)  The only point this check is currently hit is with the 

Yeah... there are a lot of arguments to this function, and to process_segment(), too.  Do_load/do_verify could be made into single parameter via an enum/bitmask along the lines of:

typedef enum {
    ESP_SEGMENT_DO_NOTHING = 0,
    ESP_SEGMENT_DO_LOAD         = 1,
    ESP_SEGMENT_DO_VERIFY      = 2,
    ESP_SEGMENT_DO_LOAD_DO_VERIFY = 3,
} esp_process_segment_opt_t;

Then the check could be `if (ESP_SEGMENT_DO_NOTHING == operation)`.  Although, having a pointer that is NULL also (at least to me) says that it's not intended to be used.  Granted, it's a little more obvious what it is doing with an intentional desired operation parameter passed along.

-- Tim

negativekelvin:
@projectgus @KonstantinKondrashov maybe you can also consider optimizations to esp_clk_init as part of your changes such as  https://github.com/espressif/esp-idf/pull/2398#issuecomment-420160856

tim-nordell-nimbelink:
@igrr @KonstantinKondrashov -

Looks like a partially reworked version of this pull request was introduced in a21ca2270a961798569f7d10e8cd4474546071e6.  Thanks for merging this in!  Should we close this PR?

I had pushed one additional commit on top of this series that was in this PR (86aba4e91d869c0d431dd1016e683209c4401a1d) that wasn't merged in, but wasn't directly related to this since it was only based on this PR.  Should I resubmit this as a separate pull request?

Thanks,
Tim

KonstantinKondrashov:
Hi @tim-nordell-nimbelink !
I'm very sorry about it. This part of the changes has been missed. I will add this, and when this commit is merged, PR will be closed.

KonstantinKondrashov:
Hi @tim-nordell-nimbelink !
Could you resubmit this (86aba4e) as a separate pull request? and close this PR.
Sorry for the inconvenience.
Thanks.

tim-nordell-nimbelink:
Hi @KonstantinKondrashov -

Opened new PR #4290.

 \- Tim

PR #3309, opened 2019-04-15
Fixed excess data coming from uninitialized variable in esp_efuse_get_pkg_ver (IDFGH-976)

Frosticles:
The call to `esp_efuse_read_field_blob` only seems to set the 3 last bits of `pkg_ver`. So the other uninitialised 29 bits in `pkg_ver` gets returned.

Before: 1073677826 (in binary) `111111111111110000011000000010`, note the last 3 bits are correct
After: 2


KonstantinKondrashov:
Thanks, @Frosticles. It has picked into our internal repo. It will be merged with master ASAP.

PR #3310, opened 2019-04-15
nvs: Check if an item is modified before writing out an identical copy (IDFGH-977)

tim-nordell-nimbelink:
This prevents wear and tear on the flash, and it also is faster in some
cases since the read-out of flash is a cheaper operation than the erasure
of flash.  Some library modules (such as the esp_wifi) write out to NVS
upon every initialization without checking first that the existing value
is the same, and this speeds up initialization of modules that make
these design choices and moves it into a centralized place.

The comparison functions are based on the read-out functions of the same
name, and changes out the memcpy(...) operations for memcmp(...)
operations.

### DIFF
--- a/components/nvs_flash/include/nvs.h
+++ b/components/nvs_flash/include/nvs.h
@@ -54,6 +54,7 @@ typedef uint32_t nvs_handle;
 #define ESP_ERR_NVS_KEYS_NOT_INITIALIZED    (ESP_ERR_NVS_BASE + 0x16)  /*!< NVS key partition is uninitialized */
 #define ESP_ERR_NVS_CORRUPT_KEY_PART        (ESP_ERR_NVS_BASE + 0x17)  /*!< NVS key partition is corrupt */
 
+#define ESP_ERR_NVS_CONTENT_DIFFERS         (ESP_ERR_NVS_BASE + 0x18)  /*!< Internal error; never returned by nvs API functions.  NVS key is different in comparison */
 
 #define NVS_DEFAULT_PART_NAME           "nvs"   /*!< Default partition name of the NVS partition in the partition table */
 /**
--- a/components/nvs_flash/src/nvs_page.cpp
+++ b/components/nvs_flash/src/nvs_page.cpp
@@ -308,6 +308,58 @@ esp_err_t Page::readItem(uint8_t nsIndex, ItemType datatype, const char* key, vo
     return ESP_OK;
 }
 
+esp_err_t Page::cmpItem(uint8_t nsIndex, ItemType datatype, const char* key, const void* data, size_t dataSize, uint8_t chunkIdx, VerOffset chunkStart)
+{
+    size_t index = 0;
+    Item item;
+
+    if (mState == PageState::INVALID) {
+        return ESP_ERR_NVS_INVALID_STATE;
+    }
+
+    esp_err_t rc = findItem(nsIndex, datatype, key, index, item, chunkIdx, chunkStart);
+    if (rc != ESP_OK) {
+        return rc;
+    }
+
+    if (!isVariableLengthType(datatype)) {
+        if (dataSize != getAlignmentForType(datatype)) {
+            return ESP_ERR_NVS_TYPE_MISMATCH;
+        }
+
+        if (memcmp(data, item.data, dataSize)) {
+            return ESP_ERR_NVS_CONTENT_DIFFERS;
+        }
+        return ESP_OK;
+    }
+
+    if (dataSize < static_cast<size_t>(item.varLength.dataSize)) {
+        return ESP_ERR_NVS_INVALID_LENGTH;
+    }
+
+    const uint8_t* dst = reinterpret_cast<const uint8_t*>(data);
+    size_t left = item.varLength.dataSize;
+    for (size_t i = index + 1; i < index + item.span; ++i) {
+        Item ditem;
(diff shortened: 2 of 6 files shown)


tim-nordell-nimbelink:
I updated the pull request to fix a couple of things with the changeset:
 - Remove potential erasures that would happen regardless in the callee functions.  The comparison functions don't need to clean up after the error case.
 - Replaced an assertion that the read size mismatched with a return condition that the contents were different for cmpMultiPageBlob(...).

components/nvs_flash/include/nvs.h
```diff
@@ -54,6 +54,7 @@ typedef uint32_t nvs_handle;
 #define ESP_ERR_NVS_KEYS_NOT_INITIALIZED    (ESP_ERR_NVS_BASE + 0x16)  /*!< NVS key partition is uninitialized */
 #define ESP_ERR_NVS_CORRUPT_KEY_PART        (ESP_ERR_NVS_BASE + 0x17)  /*!< NVS key partition is corrupt */
 
+#define ESP_ERR_NVS_COMPARE_IDENTICAL       (ESP_ERR_NVS_BASE + 0x18)  /*!< NVS key is identical in comparison */
```
igrr: Can you please add same comment as for `ESP_ERR_NVS_PAGE_FULL`, that this is never returned by the API?

components/nvs_flash/src/nvs_page.cpp
```diff
@@ -308,6 +308,63 @@ esp_err_t Page::readItem(uint8_t nsIndex, ItemType datatype, const char* key, vo
...
+        return rc;
+    }
+
+    if (!isVariableLengthType(datatype)) {
+        if (dataSize != getAlignmentForType(datatype)) {
+            return ESP_ERR_NVS_TYPE_MISMATCH;
+        }
+
+        if(!memcmp(data, item.data, dataSize))
+        {
```
igrr: (minor) Please move the opening brace on the same line as the if, and add a space between `if` and the opening paren.

igrr: (same in a few places below)

components/nvs_flash/src/nvs_page.cpp
```diff
@@ -308,6 +308,63 @@ esp_err_t Page::readItem(uint8_t nsIndex, ItemType datatype, const char* key, vo
...
+    }
+
+    if (dataSize < static_cast<size_t>(item.varLength.dataSize)) {
+        return ESP_ERR_NVS_INVALID_LENGTH;
+    }
+
+    bool isSame = true;
+    const uint8_t* dst = reinterpret_cast<const uint8_t*>(data);
+    size_t left = item.varLength.dataSize;
+    for (size_t i = index + 1; i < index + item.span; ++i) {
```
igrr: Since we only care if this function returns `ESP_ERR_NVS_COMPARE_IDENTICAL` or not, is it possible to exit this loop early if `!isSame`, and skip the CRC check?

In this case, it might be nicer to redefine the return codes of this function: `ESP_OK` — content is the same; `ESP_ERR_NVS_CONTENT_DIFFERS` or `ESP_ERR_NOT_FOUND` — other possible return values.

components/nvs_flash/src/nvs_storage.cpp
```diff
@@ -269,6 +269,14 @@ esp_err_t Storage::writeItem(uint8_t nsIndex, ItemType datatype, const char* key
         VerOffset prevStart,  nextStart;
         prevStart = nextStart = VerOffset::VER_0_OFFSET;
         if (findPage) {
+            // Do a sanity check that the item in question is actually being modified.
+            // If it isn't, it is cheaper to purposefully not write out new data.
+            // since it may invoke an erasure of flash.
+            if (ESP_ERR_NVS_COMPARE_IDENTICAL == cmpMultiPageBlob(nsIndex, key, data, dataSize))
```
igrr: Here and below, can you please write the conditions in "normal" order (`something_complex == constant`)? The compiler is capable of detecting accidental assignments in conditional expressions, so no need to reverse them.

tim-nordell-nimbelink:
> @tim-nordell-nimbelink Thank you for the PR, it looks very useful. 

No problem!  It helps save wear and tear on the flash on our board (and start-up time out of deep sleep due to extra erases of flash).

> I have left a few mostly cosmetic comments.

Sorry, it's easy for me to fall back to the coding style I usually use.  I'll go back and fix-up through to match espressif's coding style.

> Since the logic added in `cmpItem` and `cmpMultiPageBlob` is not exactly trivial, may I ask you to add a couple of unit tests for these functions

It shouldn't be too bad.  The other thought I had was that really Page::cmp... and Page::read... are virtually identical.  They could utilize a visitor pattern, but that would have a slightly deeper stack depth than the two separate routines.  The benefit would be code maintenance, but it would touch more code within the NVS routine.

> If not, I can add the tests on top of your commit.

I take it you prefer the tests to be added within the same commit?

 -- Tim

tim-nordell-nimbelink:
@igrr  -

I've gone ahead and switched the logic around to return ESP_OK instead when the comparison is the same.  I had to touch a couple of the tests for duplicates that assumed a new write occurred each time a value was written even if it was the same value.  (I didn't run the test suite previously, but I had now that I added in the tests for this change set.)  Hopefully I got all of the cosmetic changes, too.

-- Tim

PR #3328, opened 2019-04-18
partition: Fix "encrypted" read/write when encryption is disabled (IDFGH-1000)

tim-nordell-nimbelink:
According to the documentation[1][2] for partitions, setting the encrypted
flag for partitions should be a no-op when system level encryption isn't
enabled.  The current implementation, however, does not actually match
the documentation and it ends up with an unreadable partition via the
partition API if a partition flag is marked as encrypted without
system-level encryption enabled.  (This is because the writes go through
the encryption block, and reads do not go through the encryption block
when this situation occurs causing unreadable data to the application
running.) This fixes up the read-back of the partition table to match
whether or not the partition is currently encrypted under the hood.

This should not affect the bootloader's code for reading/writing encrypted
partitions as the bootloader directly invokes the spi_flash_write*(...)
APIs.

[1] https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/partition-tables.html#flags
[2] https://docs.espressif.com/projects/esp-idf/en/latest/security/flash-encryption.html#encrypted-partition-flag

### DIFF
--- a/components/spi_flash/partition.c
+++ b/components/spi_flash/partition.c
@@ -168,10 +168,13 @@ static esp_err_t load_partitions()
         item->info.type = it->type;
         item->info.subtype = it->subtype;
         item->info.encrypted = it->flags & PART_FLAG_ENCRYPTED;
-        if (esp_flash_encryption_enabled() && (
-                it->type == PART_TYPE_APP
+
+        if (!esp_flash_encryption_enabled()) {
+            /* If flash encryption is not turned on, no partitions should be encrypted */
+            item->info.encrypted = false;
+        } else if (it->type == PART_TYPE_APP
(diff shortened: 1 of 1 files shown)


mahavirj:
@tim-nordell-nimbelink Thanks for your contribution, I have put this up in our internal review process.

tim-nordell-nimbelink:
@mahavirj -

Thanks.  I just realized I didn't quite conform to espressif's C style, so I just updated the pull request to reflect that cosmetic change.  (The opening brace for the if statement should be on the same line of the if statement if possible.)

-- Tim

mahavirj:
@tim-nordell-nimbelink I did that stylistic cleanups while putting it in (coding style as well as minor rewording to comment), sorry forgot to update about that. Once merged this PR will get closed. Thanks.

Issue #3295, opened 2019-04-11
Build fail on v4.0-dev-311(linker script generation failed) when ESP32_WIFI_IRAM_OPT = n (IDFGH-958)


no1seman:
## Environment

- Development Kit: ESP32-Wrover-Kit
- Kit version (for WroverKit/PicoKit/DevKitC): v4.1
- Module or chip used:  ESP32-WROVER
- IDF version (run ``git describe --tags`` to find it): v4.0-dev-311-g70eda3d22
- Build System:         CMake
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): 1.22.0-80-g6c4433a
- Operating System:     Linux
- Power Supply:         USB|external 5V|Battery

## Problem Description
Project build fails if ESP32_WIFI_IRAM_OPT = n after update to v4.0-dev-311-g70eda3d22 esp-idf

### Expected Behavior
Project must be built

### Actual Behavior
Linker error (see log below)
### Steps to repropduce

1. idf.py build
2. ...

### Code to reproduce this issue
Any example that using wifi

## Debug Logs

Checking Python dependencies...
Python requirements from /home/user/esp/esp-mdf/esp-idf/requirements.txt are satisfied.
Running cmake in directory /home/user/esp/beacon/build
Executing "cmake -G Ninja -DPYTHON_DEPS_CHECKED=1 -DESP_PLATFORM=1 --warn-uninitialized /home/user/esp/beacon"...
Warn about uninitialized values.
-- Found Git: /usr/bin/git (found version "2.17.1") 
-- IDF_TARGET not set, using default target: esp32
-- Building for target esp32
-- The C compiler identification is GNU 5.2.0
-- The CXX compiler identification is GNU 5.2.0
-- The ASM compiler identification is GNU
-- Found assembler: /home/user/esp/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc
-- Check for working C compiler: /home/user/esp/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc
-- Check for working C compiler: /home/user/esp/xtensa-esp32-elf/bin/xtensa-esp32-elf-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /home/user/esp/xtensa-esp32-elf/bin/xtensa-esp32-elf-g++
[...]

PR #3335, opened 2019-04-19
Moving flag into section for esp_event linker file (IDFGH-1007)

tomseago:
If this flag is at the top level the linker generation script throws an error.

It appears that when this file was added an incorrect syntax was used. This pull request follows the syntax found for other components and allows the build of hello world to succeed. This fixes https://github.com/espressif/esp-idf/issues/3295

### DIFF
--- a/components/esp_event/linker.lf
+++ b/components/esp_event/linker.lf
@@ -1,6 +1,6 @@
-if POST_EVENTS_FROM_IRAM_ISR = y:
-    [mapping:esp_event]
-    archive: libesp_event.a
-    entries:
+[mapping:esp_event]
+archive: libesp_event.a
+entries:
+    if POST_EVENTS_FROM_IRAM_ISR = y:
         esp_event:esp_event_isr_post_to (noflash)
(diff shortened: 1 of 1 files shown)


sieren:
I've filed a PR for this earlier as well: https://github.com/espressif/esp-idf/pull/3322/commits/8c82c8f4506aa919bae6f5074f5cf71f8b928b32

Can you get away without the default section (as in mine)? I recall running into some issues without it.

renzbagaporo:
@tomseago @sieren 

The top level flag check syntax was added recently. Unfortunately the recent pyparsing 2.4.0 *might* have introduced regression which affects our grammar, we've submitted an issue to the PyParsing Github issue tracker and waiting on a response.

re. the abscence of default section, an update is coming to use default mappings when no conditions match; removing this requirement. But for now having the final `else` block is more correct.

igrr:
Closing as this got fixed in https://github.com/espressif/esp-idf/commit/2522268c209e580e34f26c8e06e31eeed25f41e2 (but it took another week for the fix to get synced to Github, sorry about that!)

PR #3400, opened 2019-05-01
MCPWM pulse duration measurement (IDFGH-1079)

cor3ntin:
These modifications allow to measure the duration of a pulse
(duration between a rising edge and a falling edge)

Original fix by @vgonet


Alvin1Zhang:
@cor3ntin Thanks for the contribution and sorry for the slow turnaround. We will look into this PR. Thanks.

Issue #3399, opened 2019-05-01
I2S output left/right channels swapped (IDFGH-1078)


redchenjs:
## Environment

- Development Kit: Widora AIR v6.0
- Module or chip used: ESP32 Rev.1
- IDF version: v3.2
- Build System: CMake
- Compiler version: 1.22.0-80-g6c4433a
- Operating System: Linux
- Power Supply: USB

## Problem Description

When ESP32 is used for A2DP streaming, the I2S output left/right channels are swapped.

### When playing left channel only audio, the LA output is:
![left](https://user-images.githubusercontent.com/9910809/57007773-ba7b3980-6c1d-11e9-8531-227226820318.png)

### When playing right channel only audio, the LA output is:
![right](https://user-images.githubusercontent.com/9910809/57007776-c0711a80-6c1d-11e9-81e3-1c26e03a9602.png)

### The expected left/right channel behavior according to TRM:
![trm](https://user-images.githubusercontent.com/9910809/57007779-c49d3800-6c1d-11e9-9500-aff4a55c4d64.png)

### Steps to repropduce

1. Get the `a2dp_sink` example from esp-idf v3.2.
2. Build the example and flash it to esp32.
3. Connect to esp32 with your phone and play the test files.
[test_files.zip](https://github.com/espressif/esp-idf/files/3133899/test_files.zip)

### Code to reproduce this issue

The `a2dp_sink` example from esp-idf v3.2

redchenjs:
Well, the `a2d_data_cb()` is just a wrapper for `i2s_write()`, so the problem might be SBC decoding output:
```
void bt_app_a2d_data_cb(const uint8_t *data, uint32_t len)
{
    size_t bytes_written;
    i2s_write(0, data, len, &bytes_written, portMAX_DELAY);
    if (++s_pkt_cnt % 100 == 0) {
        ESP_LOGI(BT_AV_TAG, "Audio packet count %u", s_pkt_cnt);
    }
}
```

blueMoodBHD:
Hi @redchenjs,  We have do some test, and the PCM decoded by the SBC actually has the correct left and right channel order. 
And the problem appear on the I2S side, our Drvier code defaults to the left channel in the high position, only need to set I2S_CONF_REG (0x0008) when I2S configuration the I2S_TX_MSB_RIGHT bit of the register, L/R, can be flipped over.

PR #3401, opened 2019-05-01
component/bt: fix SBC audio L/R channels swapped issue. #3399 (IDFGH-1080)

redchenjs:
Modify the SBC SynthFrame output pointer to swap the channel data output.

Tested with esp-idf v3.2, this commit fixed issue #3399


blueMoodBHD:
Hi @redchenjs, Thanks for your help. But there is some problem in this pull request.
    We have do some test, and the PCM decoded by the SBC actually has the correct left and right channel order.
    And the problem appear on the I2S side, our Drvier code defaults to the left channel in the high position, only need to set I2S_CONF_REG (0x0008) when I2S configuration the I2S_TX_MSB_RIGHT bit of the register, L/R, can be flipped over.

redchenjs:
@blueMoodBHD Thanks for your help!

PR #3404, opened 2019-05-01
make: Fix erase_otadata target to use sdkconfig UART (IDFGH-1083)

tim-nordell-nimbelink:
The erase_otadata target should utilize the same UART as specified in
the sdkconfig.

### DIFF
--- a/components/app_update/Makefile.projbuild
+++ b/components/app_update/Makefile.projbuild
@@ -2,6 +2,8 @@
 #
 .PHONY: blank_ota_data erase_otadata read_otadata
 
+ESPPORT ?= $(CONFIG_ESPTOOLPY_PORT)
+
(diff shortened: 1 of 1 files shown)


renzbagaporo:
Hi @tim-nordell-nimbelink, thanks for noticing this issue. I'm planning to submit a change internally that changes the interaction between `--port` argument and `--partition-table-file`, so they are not mutually exclusive (which is probably why you opted to use ESPTOOL_PORT instead). When that time comes the implementation in this MR would have to be changed to use the `--port` argument.

tim-nordell-nimbelink:
Hi @renzbagaporo -

Yep, your analysis is right as to why I used the environment variable.  I agree that changing the tool to be able to utilize a port would be a better solution in the long term.

-- Tim

renzbagaporo:
Hi @tim-nordell-nimbelink, this should already have been solved by 94e0569. Sorry for the delay.

PR #3426, opened 2019-05-05
project_config_mk: fix if condition (IDFGH-1105)

vetaldudnik:
Fixes failed build on Windows (Eclipse). Condition **ifdef MSYSTEM** not performs.
"make" uses /bin/sh as default shell interpreter where variable **MSYSTEM** is not defined. I suggest to use **OS** variable instead


renzbagaporo:
Thanks for the contribution @vetaldudnik. The changes here are under review in our internal system.

PR #3449, opened 2019-05-09
esp32: Fix rtos pm locks to use a no_light_sleep lock

tim-nordell-nimbelink:
Commit message:

The documentation[1] for the pm configuration states that when no locks
are taken, the minimum CPU frequency is in use.  However, since the
idle code takes a CPU_FREQ_MAX lock whenever an task is running, this
does not actually hold true.  If we shift the default lock to a no light
sleep lock, then unless the user is holding another power management lock,
the CPU will be at the minimum CPU frequency specified.

[1] https://github.com/espressif/esp-idf/blob/bec70ce2986b79a28ed1269760fa79da9603c90b/components/esp32/include/esp32/pm.h#L37

Additional notes:

I'm not sure what the intent of this lock was - I'm assuming it was to prevent the system from entering light sleep, so I think it makes sense to leave it as a light sleep lock.  If I have a task that simply spins (or just wakes up for a little bit), without acquiring the maximum CPU frequency lock, the system always enters the maximum CPU frequency while this task is doing something.  This is contrary to what one would expect due to the availability of a maximum CPU frequency lock.

### DIFF
--- a/components/esp32/pm_esp32.c
+++ b/components/esp32/pm_esp32.c
@@ -600,11 +600,11 @@ void esp_pm_impl_init()
 #ifdef CONFIG_PM_TRACE
     esp_pm_trace_init();
 #endif
-    ESP_ERROR_CHECK(esp_pm_lock_create(ESP_PM_CPU_FREQ_MAX, 0, "rtos0",
+    ESP_ERROR_CHECK(esp_pm_lock_create(ESP_PM_NO_LIGHT_SLEEP, 0, "rtos0",
             &s_rtos_lock_handle[0]));
     ESP_ERROR_CHECK(esp_pm_lock_acquire(s_rtos_lock_handle[0]));
 #if portNUM_PROCESSORS == 2
-    ESP_ERROR_CHECK(esp_pm_lock_create(ESP_PM_CPU_FREQ_MAX, 0, "rtos1",
+    ESP_ERROR_CHECK(esp_pm_lock_create(ESP_PM_NO_LIGHT_SLEEP, 0, "rtos1",
             &s_rtos_lock_handle[1]));
     ESP_ERROR_CHECK(esp_pm_lock_acquire(s_rtos_lock_handle[1]));
 #endif // portNUM_PROCESSORS == 2


igrr:
Hi @tim-nordell-nimbelink,

> documentation[1] for the pm configuration states that when no locks are taken, the minimum CPU frequency is in use

I think this is correct, taking these 2 RTOS locks into account. Perhaps the documentation needs to be updated to explicitly mention them.

> I'm not sure what the intent of this lock was - I'm assuming it was to prevent the system from entering light sleep

This lock is intentional, it guarantees that when any code is running, CPU frequency does not change. A lot of the code written for IDF before DFS functionality was introduced, relied on the fact that CPU frequency was constant.

Perhaps an alternative would be to add an extra parameter to esp_pm_config structure, defining the optional lock handle which will be used by the RTOS. If not supplied, then the default one (CPU_MAX) will be used. It's hard to say though, what side effects this might have for the existing code. Potentially problematic places are esp_random, ets_delay_us, esp_timer implementation, and some drivers for peripherals which depend on APB frequency being 80 MHz.

Also, if you have a task that spins, try replacing that with blocking on an RTOS queue.

tim-nordell-nimbelink:
Hi @igrr -

Thanks for the quick response.

> I think this is correct, taking these 2 RTOS locks into account. Perhaps the documentation needs to be updated to explicitly mention them.

I guess it is mentioned in the documentation here (https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/power_management.html#overview):

> Another example is that the RTOS will request the CPU to run at the highest configured frequency while there are tasks ready to run.

However, the way it is worded, makes it sound like an example and not the way the RTOS behaves.

> This lock is intentional, it guarantees that when any code is running, CPU frequency does not change. A lot of the code written for IDF before DFS functionality was introduced, relied on the fact that CPU frequency was constant.

Interesting.  So that means an asynchronous call to change the frequency while one of those routines are running is also not a good thing given the assumptions they make currently.  The places that make these assumptions (and possibly require these to be actuality) should probably hold a PM lock.  That, however, does not guarantee that a user did not call esp_pm_configure(...) in the meantime.

> Perhaps an alternative would be to add an extra parameter to esp_pm_config structure, defining the optional lock handle which will be used by the RTOS. If not supplied, then the default one (CPU_MAX) will be used. It's hard to say though, what side effects this might have for the existing code. Potentially problematic places are esp_random, ets_delay_us, esp_timer implementation, and some drivers for peripherals which depend on APB frequency being 80 MHz.

Fair enough.  An extra parameter for esp_pm_config could be problematic for a couple of reasons:

1. Since in C you can't specify a default value in a structure, if a user omitted specifying the new parameter it wouldn't work as expected.
2. The lock is currently created at startup, and there isn't a provision to change lock-types currently for a given handle, although, it probably wouldn't be too bad to implement a way to change lock-types on the fly.

Maybe for now it'd make sense to have a KConfig option that changes the default for now?  If I changed this patch to that behavior could I get it accepted?  I always prefer changes I make locally into the overall SDK to be "mainlined" if possible.  

> Also, if you have a task that spins, try replacing that with blocking on an RTOS queue.

This is a test function in the console that spins.  In the platform I don't have any normally running code that sits in a loop (event driven is always better).  I was noticing after re-enabling an actual range (I had been hard-coding one frequency range and switching it) in the power management configuration that my power usage was spikier than it had been and I was surprised since I wasn't holding any locks while idling.  That's why I was trying the spin test code and noticed the power consumption of our device was continuously akin to the highest power level.

I also have some local changes that checks if none of the CPU frequencies specified in the esp_pm_configure(...) call has the PLL enabled, and then disables the PLL afterwards.  Running the board at 40MHz from the XTAL with the PLL off is half the power of running the board at 40MHz with the unused PLL on.  (I'm not sure what the additional latency for turning the PLL off is though.)  For our application we don't always have a radio on and we don't need the performance of the platform with the PLL on.  However, on occasion we need to raise the CPU speed and I'd prefer to do that with just obtaining locks rather than re-calling esp_pm_configure(...) to change the overall CU speed.

-- Tim

igrr:
> However, the way it is worded, makes it sound like an example and not the way the RTOS behaves.

Agree, the way it is written, it may be misleading.

> Fair enough. An extra parameter for esp_pm_config could be problematic for a couple of reasons

I think this should still be doable: assuming C99 initializers are used, new parameters will be zero initialized by default. If the parameter is a lock handle, then the logic can be: if the handle is non-null, then use the given lock handle as the RTOS lock. Otherwise, use the default lock handles created at start-up. Don't delete the default lock handles, keep them around in case esp_pm_configure is called again with a NULL handle. Of course, esp_pm_configure will need to take care of acquiring the lock passed into the function and releasing the one used before.
Also possibly we need to pass in two handles, one for each CPU, to make the logic in esp_pm_configure simpler.

Regarding keeping the PLL on: this is on our list of possible optimizations. Despite the large latency of enabling the PLL, it is possible to wake up early enough and enable the PLL in the Wi-Fi connected light sleep use case (with a minor tweak to the Wi-Fi driver). Unfortunately this task got pushed back a bit due to other activities. The idea there is also to introduce a new flag to the esp_pm_config structure, to control whether the PLL is disabled or not.

tim-nordell-nimbelink:
> I think this should still be doable: assuming C99 initializers are used, new parameters will be zero initialized by default.

True.  I could also see someone creating the structure on a stack without zero initializing the structure and setting the variables one by one, although, it's certainly less convenient to do it this way than a C99 initializer.  I don't know if _everyone_ uses C99 initializers.  (I prefer to use them, but I don't like to make assumptions about others, especially in an API that is backwards compatible.)

> If the parameter is a lock handle, then the logic can be: if the handle is non-null, then use the given lock handle as the RTOS lock.

There's a lock per core, too.  The other part of this that would be interesting is that the lock will be in-use whenever esp_pm_configure(...) is invoked, and you'd need to ensure both cores are coordinated in switching locks since there is a unique lock per core.  It might be easier to swap out the lock type for the in-use handles.  (And even easier to switch out the default type into a KConfig option.)

> Despite the large latency of enabling the PLL, it is possible to wake up early enough and enable the PLL in the Wi-Fi connected light sleep use case (with a minor tweak to the Wi-Fi driver). Unfortunately this task got pushed back a bit due to other activities. 

It seems like there ought to be some way of abstracting this early wake-up out so that the user doesn't have to be directly aware of the latency for the frequency switch.  Something along the lines of "set a future constraint that the PLL is enabled by time X" into the PM subsystem.  The PM subsystem could then decide when it needs to take action based on the current system state (and future).

> The idea there is also to introduce a new flag to the esp_pm_config structure, to control whether the PLL is disabled or not.

Maybe a new lock type would be more appropriate so that drivers can enable/disable this constraint (being that the PLL can not be disabled) as needed?  Although a latency requirement aware PM subsystem wouldn't directly need this type of additional lock type.

PR #3456, opened 2019-05-10
Synchronize the mDNS values with LWIP values for protocol type (IDFGH-1141)

davidzuhn:
mDNS data contains an IP protocol type field, and these values are
sometimes assigned directly from the IP data.  Using the same
field values allows this to work with all combinations of data &
checks (if IPv4 else vs. if IPv6 else).   This prevents IPv6
addresses from sometimes being interpreted at IPv4 addresses.


davidzuhn:
Withdrawn.  This is a poor workaround to solve the wrong problem.

PR #3466, opened 2019-05-13
Fixed rare crash of DHCP Server during WIFI AP network reconfiguratio… (IDFGH-1151)

despierto:
…n and DHCP restart due to missed callback

There are few places of dhcps_start(), but at one of them doesn't have initialization of dhcps_cb by dhcps_set_new_lease_cb(). 

Unexpected condition while network reconfiguration with calling of dhcps_start() from another place happened a few times and caused to system crash by reading NULL ptr.


Alvin1Zhang:
@despierto Thanks for the contribution and sorry for the slow turnaround. We will look into this. Thanks.

freakyxue:
@despierto
this bug has been fixed in 3.3.
later i will backport to 3.2

AxelLin:
v3.3: the fix is merged by faf23c095cb94cc1179ba3ddd9d2affa8ae7e821
v3.2: the fix is included in 49202b7f03e1c8dd63ccbb8a9bce7f01949e2973
So this should be closed.

Alvin1Zhang:
@despierto Thanks for contribution, and thanks @AxelLin for the notes, changes have been merged with https://github.com/espressif/esp-idf/commit/faf23c095cb94cc1179ba3ddd9d2affa8ae7e821

PR #3486, opened 2019-05-16
Fix deps check on Arch Linux and msys2 (IDFGH-1173)

lf-:
This still needs testing on macOS: I think they have a python2.7 symlink. If they don't, this situation needs to be completely reconsidered so the python2 from the kconfig is used rather than whatever python is in the system path. Additionally, `menuconfig` needs to be allowed to run if the user doesn't have the right dependencies on their system `python` executable.

Anyway, this fixes menuconfig and probably other stuff on Arch Linux and any other system which has a `python2.7` symlink to a python 2.7.

### DIFF
--- a/tools/check_python_dependencies.py
+++ b/tools/check_python_dependencies.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 #
 # Copyright 2018 Espressif Systems (Shanghai) PTE LTD
 #


dobairoland:
Hi @lf- , thank you for your efforts in improving esp-idf.

May I ask you what was your issue with menuconfig and Python 3 on Archlinux?

I'm also using Archlinux and didn't have any problem with setting PYTHON=python2 back then when Python 3 wasn't supported. Now, however, it should work well with `python` (3) as well. If you experience any problem then please open a new issue.

To be honest, I don't think that hardcoding anything to `python2` is the right way to solve this problem. Python 2.7 will be deprecated in a couple of months. Python 3 is the future of Python and everything should work well with it.

lf-:
I 100% agree that python2 needs to die. This is actually a documentation issue: I was misled by the IDF installation docs which seemed to vaguely suggest that one should switch to python2 in some cases, but it wasn't at all clear what those cases were.

Furthermore, I had *no idea* that there was a `PYTHON` environment variable, as that is not specified in the new user guide.

As for the actual issue I was having, I was (incorrectly) assuming that the toolchain only worked on python 2, and was getting extremely confused when it still was calling python 3 and complaining about python 3 dependencies despite me specifying python 2 in menuconfig.

Unclear docs suggesting changing to python 2 in case of issues + going on autopilot having experienced this type of issue before = wasting an hour of time.

dobairoland:
I'm sorry for that misunderstanding. The recommendations were ment in the way that one can temporarily switch to Python 2 if experiencing problem with Python 3 (until we fix the bug). Currently, I have no knowledge about any Python 3 incompatibility issue.

If you specify `python2` in menuconfig (using `make`) then it should not complain about missing `python3` packages. However, the new `cmake` toolchain doesn't use that variable and if you use `idf.py` then it will use the default interpreter. Switching is possible by using `python2 $(which idf.py) ...`. But again, that shouldn't be necessary under normal circumstances.

If you run into truble then please feel to open an issue.

Issue #3285, opened 2019-04-09
Feature request: ability to specify recevie and transmit buffers' sizes for esp_http_client


olegantonyan:
At the moment, there's only one `buffer_size` parameter in `esp_http_client_config_t` which is used for both receive and transmit. This is not very good because I need a big receive buffer and small transmit buffer to download big files. I need as big a possible rx buffer and as small as possible tx buffer fro this task.

Please let me know if this is something you're willing to accept as a PR

jitin17:
We have accepted https://github.com/espressif/esp-idf/pull/3528, so closing this issue.

PR #3528, opened 2019-05-23
esp_http_client: separate buffer_size config option for transmit (IDFGH-1226)

olegantonyan:
This PR adds `buffer_size_tx` to `esp_http_client_config_t`.
It's important to have different buffer sizes when you need to download large files. To do so, you need RX buffer to be as big as you can afford. But you don't want to waste memory on TX buffer in this case. 

For example. I want 8K RX buffer (all I can afford). With single `buffer_size` this will allocate another 8K for TX buffer which I don't need. For TX buffer 2K is more than enough. 6K wasted. I'd rather allocate 14K for RX buffer in this case. 

Original issue https://github.com/espressif/esp-idf/issues/3285

Even if this PR will be rejected I really hope you'll implement this functionality b/c it's crucial for my project.

Thanks.

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -111,6 +111,7 @@ struct esp_http_client {
     http_event_handle_cb        event_handler;
     int                         timeout_ms;
     int                         buffer_size;
+    int                         buffer_size_tx;
     bool                        disable_auto_redirect;
     esp_http_client_event_t     event;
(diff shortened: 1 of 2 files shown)


jitin17:
@olegantonyan This feature looks good and has been added for review in our internal work queue. Thanks for your efforts!!

jitin17:
Thanks for your contribution( we did some cosmetic changes), merged to master now.

Issue #3260, opened 2019-04-05
a2dp_sink with internal DAC doesn't work (IDFGH-915)


keel-210:
## Environment

- Development Kit : ESP32-DevKitC
- Module or chip used : ESP32-WROOM-32D
- IDF version : v3.1.2-99-gcf5dbadf4
- Compiler version : 1.22.0-80-g6c4433a5
- Operating System :  Windows
- Power Supply :  USB

## Problem Description
Hi. I want to modify the sample of a2dp_sink a little and process and output the original audio signal. However, the original audio signal can not be output using the internal DAC.
The code used the original sample and the one corrected in the comment of the issue, but both did not output the correct waveform. And I changed the a2dp_sink config to use the internal DAC as follows:
![image](https://user-images.githubusercontent.com/21135326/55607305-cc9cc000-57b6-11e9-84f2-72584db03f91.png)

[a2dp_sink sample](https://github.com/espressif/esp-idf/tree/master/examples/bluetooth/a2dp_sink)

### Expected Behavior
I expected that analog waveforms using the internal DAC would be output from GPIO 25 and 26.

### Actual Behavior
I heard noise when I connected to a speaker.
When I looked at the output of the GPIO 25 pin with an oscilloscope, the following waveforms were output.
![sample](https://user-images.githubusercontent.com/21135326/55608267-0ff82e00-57b9-11e9-9150-baffe251d58c.png)

According to the following advice, after modifying the code, the GPIO 25 pin outputs the following.
[macaba's modified bt_app_a2d_data_cb](https://github.com/espressif/esp-idf/issues/1614#issuecomment-432858740)
![sampleplus](https://user-images.githubusercontent.com/21135326/55607902-408b9800-57b8-11e9-9c10-ec824523d451.png)

Console Output
> I (177) I2S: DMA Malloc info, datalen=blocksize=240, dma_buf_count=6
I (177) I2S: PLL_D2: Req RATE: 44100, real rate: 2777.000, BITS: 16, CLKM: 30, B
CK: 60, MCLK: 30.234, SCLK: 88864.000000, diva: 64, divb: 14

The above waveform looks more like a digital waveform than an analog waveform.Since these waveforms have a very long period compared to the I2S codec, it can be understood that they output waveforms other than the I2S codec output, but I don't know what waveform they are.

koobest:
Hi, @keel-210
Can you provide you configuration of I2S and where is your original audio signal came from? this info  can help us debug this issue.

thanks!!

[...]

Issue #1819, opened 2018-04-04
noise of a2dp_sink


mingshima:
Hello all,
I use A2DP source and sink on esp32 github. I just simply did a test, I use my smartphone to connect to the A2DP sink by bluetooth and play music, I chose internal dac. When I don't play music, there is no noise, when I play music, there is a lot of noise. I found that the sample frewuency is 2777Hz
```
I (12121) I2S: PLL_D2: Req RATE: 44100, real rate: 2777.000, BITS: 16, CLKM: 30,
 BCK: 60, MCLK: 30.234, SCLK: 88864.000000, diva: 64, divb: 14
```
Why will this happen and why there are so much noise? I have tried to add a low pass filter between ESP32 board and speaker, but can't get improve. Can someone explain to me and help me to decrease the noise? Many thanks.

koobest:
@mingshima The internal DAC has only 8 bit width, so there is noise.

macaba:
For anyone else who comes across this, please see my comment:
https://github.com/espressif/esp-idf/issues/1614#issuecomment-432858740

MrityunjaiKumar:
> For anyone else who comes across this, please see my comment:
> [#1614 (comment)](https://github.com/espressif/esp-idf/issues/1614#issuecomment-432858740)

@macaba this step do help, but there are still noises here and there, on oscilloscope it shows sudden peaks in waveform. Tried low pass filter but that doesn't help. Any clue?

ronyeapen:
Hello all, I am new to esp32 and I started off with A2DP_sink as my first code to experiment. @macaba Thank you very much for that comment. It helps a LOT! It improved the sound quality by 90%. But as @MrityunjaiKumar  mentioned, there is still noise and 'crack' like noise every so often when playing music. Any ideas why that is the case? any ideas on how it can be fixed?

Issue #1614, opened 2018-02-12
[TW#18629] I2S driver for internal 8 bit DAC is broken


pufstudio:
I can't get internal DAC to produce sound when using the a2dp_sink example.

This used to work when I first tested it around September 2017.
I have been using external I2S DAC ever since so I'm not sure when the issue with internal DAC started.

I sometimes get crackling noises but music can't be heard.

I noticed that calculated real rate is way off the target:
> I (192) I2S: PLL_D2: Req RATE: 44100, real rate: 2777.000, BITS: 16, CLKM: 30, BCK: 60, MCLK: 30.234, SCLK: 88864.000000, diva: 64, divb: 14

If I set .use_apll = 1 the output changes to:
> I (204) I2S: APLL: Req RATE: 44100, real rate: 44099.020, BITS: 16, CLKM: 1, BCK: 8, MCLK: 30.234, SCLK: 1411168.625000, diva: 1, divb: 0

But still it's not working.

koobest:
Hi, @pufstudio, we can not use apll clk in i2s adc mode.

pufstudio:
Thanks for your reply koobest, but it's also not working without apll.

I'm using a2dp_sink example without any modification (except setting to internal dac in menuconfig) with the esp-idf from master branch.

Looks to me like sample rate is calculated wrong but maybe that's how it's supposed to be...
When it was working back in the day I haven't actually been paying close attention to the "real rate".

koobest:
Hi, pufstudio, Where is a2dp_sink example, I didn't find it example and recently we have found that opening wifi affects I2s ADC.  Have you used wifi in your code?

negativekelvin:
https://github.com/espressif/esp-idf/tree/master/examples/bluetooth/a2dp_sink

pufstudio:
@koobest: No. I'm not using WiFi with the example. Only classic Bluetooth.

koobest:
@pufstudio 
I have a test with this example, but no sound can be heard. so, I change  `communication_format` to
`communication_format = I2S_COMM_FORMAT_I2S_MSB`. music can be heard but not so clear.
because the DAC is 8 bit out put.

pufstudio:
@koobest 
Thank you. That was certainly part of the issue.
[...]

PR #3542, opened 2019-05-25
examples/bluetooth: Fix output to internal DAC and add optional dithering (IDFGH-1245)

treitmayr:
This first commit fixes the output to the ESP32's internal DAC in the a2dp_sink example, which repeatedly created bugs (#1614, #1819, #3260). It implements the suggested solution of changing signed samples to unsigned samples before submitting them to the I2S controller.

The second commit adds optional support for dithering when using the internal DAC by adding some (triangular density) noise from the ESP32's hardware random number generator during the 16->8 bit conversion.

The changes were done and tested against the current master branch.

### DIFF
--- a/examples/bluetooth/a2dp_sink/main/Kconfig.projbuild
+++ b/examples/bluetooth/a2dp_sink/main/Kconfig.projbuild
@@ -18,6 +18,13 @@ menu "A2DP Example Configuration"
 
     endchoice
 
+    config EXAMPLE_A2DP_SINK_OUTPUT_INTERNAL_DAC_DITHER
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@treitmayr Thanks for the contribution and sorry for the slow turnaround. We will look into this PR. Thanks.

treitmayr:
On second thought, you may want to skip the dithering part because I have doubts it really implements a *triangular* probability density function (TPDF). However the part for the sample conversion is essential for the internal DAC and proven to be correct across different other projects.

PR #3577, opened 2019-06-03
Fix "%d' directive output may be truncated writing between 1 and 10 b… (IDFGH-1282)

BrianPugh:
…ytes into a region of size 7" error in bluedroid

The 5 value implies a `uint16_t` was used at some point. I'd recommend either changing the appropriate values to `uint16_t`, or to simply increase the buffer size to handle a UINT32_MAX value to prevent compiler errors.


BrianPugh:
bump

igrr:
Sorry @BrianPugh, it seems that the same has been fixed (in a slightly different way) in https://github.com/espressif/esp-idf/commit/f8c107fbe3ba1ab9825973b8b52941f4577db88e.

PR #3580, opened 2019-06-03
Expand ULP macro programming functionality (IDFGH-1285)

boarchuz:
This updates and expands ULP macro programming functionality, bringing it into line with other programming methods.

**Implement counter register functionality**
- The 8 bit counter register is now available. 
- Instructions for resetting (I_STAGE_RST), incrementing (I_STAGE_INC) and decrementing (I_STAGE_DEC) the counter are defined. 
- Counter register branch instruction is defined (I_JUMPS). 
- Additionally, convenient branch macros are defined in the existing style (as per M_BL and M_BGE) to match the functionality described here https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/ulp_instruction_set.html#jumps-jump-to-a-relative-address-condition-based-on-stage-count. These are M_BSLT (Macro: Branch if Stage <), M_BSGE (>=), M_BSLE (<=), M_BSEQ (==), and M_BSGT (>).

**Implement I2C functionality**
- While the instruction format already existed, the macro is now defined so that it is accessible: I_I2C_RW
- Additionally, convenience macros - I_I2C_READ and I_I2C_WRITE - have been defined for reading and writing a byte.

**Implement MOVE [label] functionality**
- A new macro sub-opcode (SUB_OPCODE_MACRO_LABELPC) has been defined to facilitate copying the program counter of a label into a register. This is equivalent to the "MOVE RX, label" instruction.
- This makes I_BXR, I_BXZR and I_BXFR useful without troublesome pre-calculation of instruction addresses. Subroutines may now be easily programmed.
- A macro (M_MOVL) has been defined for convenient use.
- Usage might look something like this, for example:

```
// ...
M_MOVL(R3,2),  //Copy the address of label 2 into R3
M_BX(7),             //Branch to the subroutine
M_LABEL(2),       //The return point of the subroutine
// ...
M_MOVL(R3,3),  //Copy the address of label 3 into R3
M_BX(7),             //Branch to the subroutine
M_LABEL(3),       //The return point of the subroutine
// ...
I_HALT(),

M_LABEL(7),       //Subroutine entry
//Do work
I_BXR(R3),          //Return to the address in R3
```

**Update comments to reflect current state of ULP macro support**
- Temperature sensor, ADC and others were incorrectly commented as "not implemented yet", causing confusion. I've verified that they are implemented and working. Such comments have been updated, along with any comments requiring updating due to the changes above. 
- (It seems a lot of the incorrect comments are downstream from the binutils-esp32ulp repo; there are many "not implemented yet" comments that require updating in there too.)

### DIFF
--- a/components/ulp/include/esp32/ulp.h
+++ b/components/ulp/include/esp32/ulp.h
@@ -1016,7 +1016,7 @@ static inline uint32_t SOC_REG_TO_ULP_PERIPH_SEL(uint32_t reg) {
  * 
  *  Slave address (in 7-bit format) has to be set in advance into SENS_I2C_SLAVE_ADDRx register field, where x == slave_sel. 
  *  For read operations, 8 bits of read result is stored into R0 register.
- *  For write operations, bits outside of val[high_bit:low_bit] are masked.
+ *  For write operations, val will be written to sub_addr at [high_bit:low_bit]. Bits outside of this range are masked.
  */
 #define I_I2C_RW(sub_addr, val, low_bit, high_bit, slave_sel, rw_bit) { .i2c = {\
         .i2c_addr = sub_addr, \
--- a/components/ulp/include/esp32/ulp.h
+++ b/components/ulp/include/esp32/ulp.h
@@ -263,8 +263,8 @@ typedef union {
     struct {
         uint32_t i2c_addr : 8;      /*!< I2C slave address */
         uint32_t data : 8;          /*!< 8 bits of data for write operation */
-        uint32_t low_bits : 3;      /*!< number of low bits to mask for write operation */
-        uint32_t high_bits : 3;     /*!< number of high bits to mask for write operation */
+        uint32_t low_bits : 3;      /*!< low bit of range for write operation (lower bits are masked) */
+        uint32_t high_bits : 3;     /*!< high bit of range for write operation (higher bits are masked) */
         uint32_t i2c_sel : 4;       /*!< index of slave address register [7:0] */
         uint32_t unused : 1;        /*!< Unused */
         uint32_t rw : 1;            /*!< Write (1) or read (0) */
@@ -1016,7 +1016,7 @@ static inline uint32_t SOC_REG_TO_ULP_PERIPH_SEL(uint32_t reg) {
  * 
  *  Slave address (in 7-bit format) has to be set in advance into SENS_I2C_SLAVE_ADDRx register field, where x == slave_sel. 
  *  For read operations, 8 bits of read result is stored into R0 register.
- *  For write operations, bits outside range [high_bit:low_bit] of val are masked.
+ *  For write operations, bits outside of val[high_bit:low_bit] are masked.
  */
 #define I_I2C_RW(sub_addr, val, low_bit, high_bit, slave_sel, rw_bit) { .i2c = {\
         .i2c_addr = sub_addr, \
--- a/components/ulp/include/esp32/ulp.h
+++ b/components/ulp/include/esp32/ulp.h
@@ -46,20 +46,22 @@ extern "C" {
  * @{
  */
 
-#define OPCODE_WR_REG 1         /*!< Instruction: write peripheral register (RTC_CNTL/RTC_IO/SARADC) (not implemented yet) */
+#define OPCODE_WR_REG 1         /*!< Instruction: write peripheral register (RTC_CNTL/RTC_IO/SARADC) */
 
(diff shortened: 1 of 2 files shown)


components/ulp/include/esp32/ulp.h
```diff
@@ -46,50 +46,62 @@ extern "C" {
...
+#define OPCODE_RD_REG 2         /*!< Instruction: read peripheral register (RTC_CNTL/RTC_IO/SARADC) */
 
 #define RD_REG_PERIPH_RTC_CNTL 0    /*!< Identifier of RTC_CNTL peripheral for RD_REG and WR_REG instructions */
 #define RD_REG_PERIPH_RTC_IO   1    /*!< Identifier of RTC_IO peripheral for RD_REG and WR_REG instructions */
 #define RD_REG_PERIPH_SENS     2    /*!< Identifier of SARADC peripheral for RD_REG and WR_REG instructions */
 #define RD_REG_PERIPH_RTC_I2C  3    /*!< Identifier of RTC_I2C peripheral for RD_REG and WR_REG instructions */
 
-#define OPCODE_I2C 3            /*!< Instruction: read/write I2C (not implemented yet) */
+#define OPCODE_I2C 3            /*!< Instruction: read/write I2C */
+#define OPCODE_I2C_RD 0         /*!< I2C read */
```
igrr: For similarity with other opcodes, suggest using `SUB_OPCODE_I2C_RD`/`_WR` here

components/ulp/include/esp32/ulp.h
```diff
@@ -46,50 +46,62 @@ extern "C" {
...
 #define BX_JUMP_TYPE_ZERO 1     /*!< Branch if last ALU result is zero */
 #define BX_JUMP_TYPE_OVF 2      /*!< Branch if last ALU operation caused and overflow */
 #define SUB_OPCODE_B  1         /*!< Branch to a relative offset */
 #define B_CMP_L 0               /*!< Branch if R0 is less than an immediate */
 #define B_CMP_GE 1              /*!< Branch if R0 is greater than or equal to an immediate */
+#define JUMPS_LT 0              /*!< Branch if counter register < */
+#define JUMPS_GE 1              /*!< Branch if counter register >= */
+#define JUMPS_LE 2              /*!< Branch if counter register <= */
+// #define JUMPS_EQ 3              /*!< Branch if counter register == */
+// #define JUMPS_GT 4              /*!< Branch if counter register > */
```
igrr: Do these commented out lines serve any purpose? If not, suggest removing them.

components/ulp/include/esp32/ulp.h
```diff
@@ -46,50 +46,62 @@ extern "C" {
...
 #define ALU_SEL_LSH 5           /*!< Shift left by given number of bits */
 #define ALU_SEL_RSH 6           /*!< Shift right by given number of bits */
+#define ALU_SEL_SINC  0         /*!< Increment counter register */
+#define ALU_SEL_SDEC  1         /*!< Decrement counter register */
+#define ALU_SEL_SRST  2         /*!< Reset counter register */
 
 #define OPCODE_BRANCH 8         /*!< Branch instructions */
 #define SUB_OPCODE_BX  0        /*!< Branch to absolute PC (immediate or in register) */
+#define SUB_OPCODE_BR  1        /*!< Branch to relative PC */
+#define SUB_OPCODE_BS  2        /*!< Branch to relative PC */
```
igrr: Could you please add ", conditional on R0" and ", conditional on the stage counter" to these two comments?

boarchuz:
Done! I also edited mentions of "counter register" so that it is now consistently referred to as "stage counter".

boarchuz:
A little issue I'm having is that the I2C read operation doesn't match the behaviour documented here: https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/ulp_instruction_set.html#i2c-rd-read-single-byte-from-i2c-slave
Specifically, the low_bit and high_bit have no effect; the full, 8-bit read value is always stored in R0 regardless of the provided range.
I wonder if this is normal (ie. the info at the link is incorrect) or if there's something wrong with the I2C read macros here? Since the write operation works fine with a given range, I suspect the former. Maybe someone with the binutils toolchain installed could test I2C_RD behaviour to compare.

igrr:
Cherry-picked in https://github.com/espressif/esp-idf/commit/88a69823cfb23359a0bbb116ea71355c59756739, thanks!

PR #3597, opened 2019-06-06
To avoid error 112 (Address already in use), set socket to Reuse Addr… (IDFGH-1306)

DavidAntliff:
…ess. Ensure listening socket is closed when client disconnects.

This is a quick fix to the example as it is written. However a better example would probably not close and reopen the listening socket each iteration. There is no need to continually close and rebind this socket when clients disconnect.

Therefore I suggest that this PR be considered for the time being, but that this example be revisited to behave in a more standard manner.

### DIFF
--- a/examples/protocols/sockets/tcp_server/main/tcp_server.c
+++ b/examples/protocols/sockets/tcp_server/main/tcp_server.c
@@ -64,7 +64,10 @@ static void tcp_server_task(void *pvParameters)
         ESP_LOGI(TAG, "Socket created");
 
         int flag = 1;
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@DavidAntliff Thanks for the contribution and sorry for the slow turnaround. We have been looking into PR, thanks.

blippy:
I have discussed this issue [here](https://www.esp32.com/viewtopic.php?f=2&t=7791&p=49688#p49688), along with others. I think that there is still a problem (qv).

Issue #1684, opened 2018-03-06
libstdc++ with RTTI (i.e. without -fno-rtti) (IDFGH-751)


jovermann:
Apparently the libstdc++ in the current toolchain (xtensa-esp32-elf-linux64-1.22.0-75-gbaf03c2-5.2.0.tar.gz) is compiled with -fno-rtti. 

I have a project which relies on RTTI and I modified project.mk to have -frtti instead of -fno-rtti. This compiled fine, but when linking I get lots of errors like:

undefined reference to `typeinfo for std::basic_istream<char, std::char_traits<char> >'

for various types of course. I think this is because libstdc++ is compiled with -fno-rtti.

Request: Please compile libstdc++ in the toolchain with -frtti (or with no rtti flag at all because RTTI is enabled by default). Please change the default flags for CXX to -frtti (or to no rtti flag at all because RTTI is enabled by defaul).

The linker will ignore the additional data when it is not needed, so I do not see a drawback.

I currently see non-deterministic crashes when doing dynamic_cast on user defined types which are derived from std:: types (like locale etc). This is perfectly explainable since the RTTI settings between my code and libstdc++ are inconsistent. But as an esp-idf user I have no way to fix libstdc++.

chengyuhui:
Any update on  this? 

Recently I have been trying to get https://github.com/ThePhD/sol2 running on EPS32, which depends on rtti to compile successfully.

igrr:
> The linker will ignore the additional data when it is not needed, so I do not see a drawback.

Unfortunately, this is not entirely the case. If RTTI is enabled when compiling libstdc++.a, then vtables of classes compiled into libstdc++.a will contain pointers to the typeinfo structures. These pointers will be in place even if the application does not enable RTTI, and will cause typeinfo structures to be linked into the application. In some tests we have seen increase of application binary size by 10kB or more.

We are looking into approaches to reduce this penalty for users who do not need to enable RTTI.

v1993:
@igrr Can't you build libstdc++ together with project and not provide pre-compiled library? That would introduce some time penalty for complete rebuilds, but they aren't required pretty often I think.

bpietsch:
Or provide two versions of the library and choose which to link to based on whether rtti was used in the project via the makefile?

v1993:
@bpietsch Yep, that would work as well I guess. Options from "menuconfig" can be read from CMake/Makefiles.

igrr:
[...]

PR #3649, opened 2019-06-18
option to configure runtime type information (IDFGH-2495)

itmm:
To allow dynamic_cast<> in C++ the runtime type information needs to be enabled. This can be done via menuconfig. It slightly increases the virtual table of all classes and structs that have virtual functions.

Maybe this change can be cherry-picked to master. I cannot get the the needed Toolchain for master compiling on my Raspberry Pi (assembler error in gmp).


igrr:
Sorry for not updating this earlier. RTTI is now supported on the master branch (Kconfig option, disabled by default). This will be part of release v4.1. Commits are linked from https://github.com/espressif/esp-idf/issues/1684.

PR #3659, opened 2019-06-19
Fix compiler warning/error when HFP_DYNAMIC_MEMORY is not TRUE. (IDFGH-1378)

nifomorph:
The ```error_exit``` label is not used in a goto statement unless dynamic memory allocation is enabled. Depending on the setting of compiler flags, this may generate an error during compilation.


nifomorph:
Has been fixed before espressif's bluetooth refactoring where bluedroid was bounced to host/bluedroid.

PR #3665, opened 2019-06-20
BT HFP: Add AT+NREC=0 command for disabling AG echo cancellation. (IDFGH-1385)

nifomorph:
For testing HFP Client in the case where audio is routed via HCI, it is practical to route the incoming audio from the AG to the outgoing audio via a buffer. However, if the AG has echo cancellation (EC) enabled, the EC will suppress the audio from the HF.

This PR adds an API to use the AT+NREC command for disabling echo cancellation in the AG.
This is probably also needed if the HF has it's own echo canceller.

### DIFF
--- a/components/bt/host/bluedroid/api/esp_hf_client_api.c
+++ b/components/bt/host/bluedroid/api/esp_hf_client_api.c
@@ -432,6 +432,22 @@ esp_err_t esp_hf_client_request_last_voice_tag_number(void)
     return (stat == BT_STATUS_SUCCESS) ? ESP_OK : ESP_FAIL;
 }
 
+esp_err_t esp_hf_client_send_nrec(void)
+{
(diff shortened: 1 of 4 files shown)


nifomorph:
Please ignore the .gitmodules change. It has already been submitted as a separate PR. 
It is a bit tedious to remove from the branch :-/

Alvin1Zhang:
@nifomorph Thanks for the contribution and sorry for the slow turnaround. We have looked into this PR. Thanks.

blueMoodBHD:
@nifomorph Thanks for the contribution. And the PR have been merged into master.
Refer to commit 75578300f1e11d79d496434f7c394a76b20bb22b & 9f145ff165099f32b92b7ee022a1debc5f0a583d(merged commit)

Alvin1Zhang:
Thanks for contribution.

Issue #3692, opened 2019-06-26
I2S reported sample rate is halved, real sample rate is erratic when using APLL and ADC built-in mode (IDFGH-1415)


detly:
## Environment

- Development Kit: Wemos Lolin D32 
- Module or chip used:  ESP32-WROOM-32
- IDF version: `master` (`58df1d93`)
- Build System: Cmake
- Compiler version: `1.22.0-80-g6c4433a`
- Operating System: Linux (Ubuntu 18.04)
- Power Supply: USB

## Problem Description

When I configure I2S to use a sample rate of 24kHz, it displays 12kHz.

### Expected Behavior

The I2S sample rate is confirmed to be 24kHz.

### Actual Behavior

The I2S info messages tell me it's 12kHz. ~~The actual sample rate seems to be correct though — if I run a 1kHz tone generator into a microphone connected to the ADC, I get peaks 24 samples apart, which is correct.~~

### Steps to repropduce

See the code in [this Gitlab project](https://gitlab.com/detly/i2s_test/tree/i2s_rate_bug). Compile and flash it with `idf.py build` and `idf.py flash monitor`. You will see the following I2S-related messages:

```text
I (284) I2S: DMA Malloc info, datalen=blocksize=4092, dma_buf_count=2
I (284) I2S: APLL: Req RATE: 24000, real rate: 11999.990, BITS: 16, CLKM: 1, BCK_M: 8, MCLK: 3071997.500, SCLK: 383999.687500, diva: 1, divb: 0
I (304) I2S: APLL: Req RATE: 24000, real rate: 11999.990, BITS: 16, CLKM: 1, BCK_M: 8, MCLK: 3071997.500, SCLK: 383999.687500, diva: 1, divb: 0
```

It's even worse if you have `use_apll = 0`:

```text
I (283) I2S: DMA Malloc info, datalen=blocksize=4092, dma_buf_count=2
I (283) I2S: PLL_D2: Req RATE: 24000, real rate: 1515.000, BITS: 16, CLKM: 55, BCK: 60, MCLK: 55.556, SCLK: 48480.000000, diva: 64, divb: 35
I (303) I2S: PLL_D2: Req RATE: 24000, real rate: 1515.000, BITS: 16, CLKM: 55, BCK: 60, MCLK: 55.556, SCLK: 48480.000000, diva: 64, divb: 35
```
### Code to reproduce this issue

```c
void app_main()
{
    int i2s_num = I2S_NUM_0;

[...]

PR #3702, opened 2019-06-28
Error if attempting to use APLL for I2S clock source in ADC/DAC mode (IDFGH-1426)

detly:
The technical manual (12.5) says:

> In the ADC/DAC mode, use PLL_D2_CLK as the clock source.

I added an error check (maybe it should be a warning?) and a note in the I2S header about this.


negativekelvin:
... but it was specifically allowed by this commit https://github.com/espressif/esp-idf/commit/4b34ae1258c03e1c67cf51348855ff6e74dd21fd#diff-27688c6b3c29373d2a2b142b8471981c and is used in the example https://github.com/espressif/esp-idf/blob/bbafd506624b2a3d1e58f2dd3154286b79a8632b/examples/peripherals/i2s_adc_dac/main/app_main.c#L71
So clarification needed

detly:
It causes a lot of issues, see #3692.

negativekelvin:
Yes but is that because it isn't properly supported or is it really unusable

detly:
I eventually bypassed the IDF libs and configured the relevant registers directly, and it was still unstable or resulted in the wrong frequency. So it looks like the warning in the TRM is serious.

koobest:
Hi, we support use apll in ADC mode, but if the sample rate is two low, we will use `I2S_D2_CLK`.

PR #3703, opened 2019-06-28
bugfix: mbedtls elliptic curve headers included (IDFGH-1427)

PeterHorvathIsento:
mbedtls elliptic curve cryptography headers are missing, that makes secure_boot_signatures.c uncompilable with the following error:

```
.../esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c: In function 'esp_secure_boot_verify_signature_block':
.../esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c:96:5: error: unknown type name 'mbedtls_ecdsa_context'; did you mean 'mbedtls_rsa_context'?
     mbedtls_ecdsa_context ecdsa_context;
     ^~~~~~~~~~~~~~~~~~~~~
     mbedtls_rsa_context

```

This trivial fix includes the required headers.


PeterHorvathIsento:
Unfortunately, cla-assistant.io timeouts, so I can't sign the CLA. But feel free to insert the headers independently from me.

In the future, please use lesser crap solutions for such important things like licensing.

ghemal:
Sorry for late response on this. 

Currently ecdsa.h is already included through [pk.h](https://github.com/espressif/mbedtls/blob/mbedtls-2.16.2-idf/include/mbedtls/pk.h#L44) which is included through [x509.h](https://github.com/espressif/mbedtls/blob/mbedtls-2.16.2-idf/include/mbedtls/x509.h#L34) in secure_boot_signatures.c

This requires elliptic curve DSA library to be enabled though (which is the case by default) in mbedtls component. Could you please confirm if you have this feature enabled (menuconfig -> component config -> mbedtls -> Elliptic Curve Ciphers -> Elliptic Curve DSA)?

projectgus:
@PeterHorvathIsento I wanted to follow up on my [colleague's earlier comment](https://github.com/espressif/esp-idf/pull/3703#issuecomment-510786832) as it's not clear to us why this header needed to be directly included (rather than via x509.h).

The change which introduced mbedTLS signature checks is 202f689b and it includes a config rule that `MBEDTLS_ECP_DP_SECP256R1_ENABLED` should be enabled whenever signature checks are enabled, however it doesn't also ensure that `MBEDTLS_ECDSA_C` is enabled, despite this also being required. We'll fix the IDF config rules so that this config combination is enforced by the configuration system.

If you're still getting compile errors without the explicit include and you have ECDSA enabled in the project config, could you please share your sdkconfig file?

> Unfortunately, cla-assistant.io timeouts, so I can't sign the CLA. But feel free to insert the header independently from me.

Thanks for letting us know it didn't work and sorry for the inconvenience caused by it not working correctly. 

Maybe this is not the original problem you encountered, but what I can see at the moment is that commit d474e86 has separate "author" and "committer" email addresses, corresponding to GitHub accounts @PeterHorvathIsento and @HorvathAkosPeter. The former has signed through CLA-Assistant but the latter hasn't. Maybe due to the timeout problem you mentioned?

This change is small enough to not attract copyright issues, and it's clear that both these accounts are you, so this doesn't pose any problem for us.

> In the future, please use better solutions for such important things like licensing. You might lose more important improvements, too.

Thanks, we'll keep an eye out for anyone else reporting technical problems when signing. To my knowledge you're the first in over 400 contributors, so for now we're treating this as an unfortunate glitch rather than a systematic problem.

make/project.mk
```diff
@@ -440,6 +440,7 @@ CFLAGS ?=
 EXTRA_CFLAGS ?=
 CFLAGS := $(strip \
 	-std=gnu99 \
+  -fplan9-extensions \
```
projectgus: FWIW, you can achieve this change without patching IDF by setting `EXTRA_CFLAGS := -fplan9-extensions` in your project Makefile, before the line that includes `project.mk`

projectgus:
Closing due to a lack of any updates. @PeterHorvathIsento please let us know if you're still experiencing any problems due to this.

PR #3704, opened 2019-06-28
components: ethernet: emac_main: support giving 50mhz rmii clock with… (IDFGH-1432)

tomer-shefler:
… both 40mhz and 26 mhz rtc xtal

Signed-off-by: Tomer Shefler <sheflertomer@gmail.com>


suda-morris:
Thank you @tomer-shefler we appreciate your effort in the Ethernet driver, the code looks GOOD to me. But the Ethernet driver has been refactored theses days, so would please move the changes to [here](https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/emac_hal.c#L97)? I'm so sorry for any inconvenience.

PR #3708, opened 2019-06-29
espcoredump.py: detect base for '--off' parameter

thierer:
Please consider enabling support for non base 10 integer literals for the `--off` parameter of espcoredump.py. 

It's a minor annoyance that I have to convert it to decimal even though it says *type "make partition_table" to see* - which displays the coredump partition's offset in hex.

Also: Fix a small typo.

Courtesy of https://stackoverflow.com/a/48950906


thierer:
Oops. Lost a comma and missed that the parameter is actualy used in two different locations... "Should" word now

PR #3717, opened 2019-07-01
i2s: fix i2s left/right channels swapped issue. close #3399 (IDFGH-1447)

### LINKED ISSUE #3399: I2S output left/right channels swapped (IDFGH-1078)

redchenjs:
The right channel data is in the high position, so the I2S_TX_MSB_RIGHT & I2S_RX_MSB_RIGHT bits should be set.

Tested with IDF v4.0-dev-1071-g70474ae84.

This commit fixes the #3399.

### DIFF
--- a/components/driver/i2s.c
+++ b/components/driver/i2s.c
@@ -944,7 +944,7 @@ static esp_err_t i2s_param_config(i2s_port_t i2s_num, const i2s_config_t *i2s_co
     I2S[i2s_num]->conf.rx_start = 0;
 
     if (i2s_config->mode & I2S_MODE_TX) {
-        I2S[i2s_num]->conf.tx_msb_right = 0;
+        I2S[i2s_num]->conf.tx_msb_right = 1;
         I2S[i2s_num]->conf.tx_right_first = 0;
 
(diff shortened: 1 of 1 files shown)


redchenjs:
Merged.

PR #3721, opened 2019-07-01
Fix: Make submodules work in forks (IDFGH-1451)

nifomorph:
Change URLs for Espressif modules to be relative to 
the parent of espressif/esp-idf rather than esp-idf itself.

This will make submodules work out of the box on GitHub forks.
See also:
https://esp32.com/viewtopic.php?f=13&t=11255&sid=196511452bd1a4faa0c9cf7b960db096


projectgus:
Hi @nifomorph ,

Thanks for this fix! As you can see it's been cherry-picked into IDF, sorry noone got back to you about it earlier.

Angus

nifomorph:
Hi Angus,
You're welcome. You've got a huge project in development, and I think it's nicely managed.
I saw that the fix was included in the master branch and subsequently closed the PR, as I assume that is the proper procedure.
ATB Niels

projectgus:
@nifomorph Thanks for all the kind words.

And yes, thanks for closing it. Right now we try to manually follow-up on PRs which are cherry picked like that and then close them. But we can be a bit slow, as in this case.

PR #3750, opened 2019-07-06
esp_partition_erase_range(): rename parameter "start_addr" to "offset" (IDFGH-1484)

thierer:
The name "start_addr" (which goes straight into the docs) implies
it's an absolute address while in fact it's an offset into the
partition like what's used in all the other esp_partition_*
functions.

So in order to avoid confusion make the name consistent with the
parameter names used for the other partition functions and call it
"offset".


projectgus:
@thierer Sorry I missed that this PR targets to release/v3.2 branch.

Our policy is to merge into master branch first, and then optionally we backport bug fixes to previous releases depending on severity. In this case this is not a very severe bug so I'm afraid we probably won't backport.

I'll rebase your branch on master and submit it internally, no need for any changes on your side.

projectgus:
Cherry-picked as d6c2dad, thanks!

PR #3757, opened 2019-07-10
malloc memory request size is updated (IDFGH-1490)

rmshub:
sizeof(CAP_SIG_NUM) will return only 4 bytes. However, we need 3*4 = 12 for three signals capture. So, CAP_SIG_NUM*sizeof(uint32_t) is valid for memory allocation. Otherwise, it will access out-of boundary memory.


koobest:
Thanks for your PR . We will fix this bug soon :-)

Alvin1Zhang:
@rmshub Thanks for the contribution. This PR has already been merged, thanks.

PR #3776, opened 2019-07-12
Wifi Powersave (IDFGH-1511)

arnold-b:
Hello, 

currently ESP is not realy useable on Wifimode. Its a waist of energy. It would be good if there would be a WiFi Powersave mode to reduce the current. 
:Zentri AMWx36 0,77mA in Powersave mode, Active Mode 5.7mA

A Solution wich is using HTTP for exp. is wasting a lot power. You can only do some tricks (sleep 2 or 3 seconds) to save power. Thats not good. 

Or there could be a mode, in which the MCU slows down the speed but still have full function. May on this we also could save power. 

Think about: everything in your home is going over Wifi. I guess then you have around 500 Wifi modules. We say around 200 will have response over Wifi (Relays, Dimmer, ...) 
Everyone of this, is using around 150mA.  => 200*(0,05 no active communication/ 0,15= = 10A/30A *3,3= (33/99) Watt
=> 33/99*24h = 792/2376Wh per Day => 289/867KWh per year => It cost a lot money per year even.

And now you add this to milions of home´s or factories.
ROI for better modules will be very fast. 
I did not even calculate the power waste to get the 3V3 or the passive Sensors (Temprature, Key,...) 

Sleep or DeepSleep is only useable for passive.

I like ESP, but this will be a huge Problem in the IoT market 

i hope someone have a solution for this.


negativekelvin:
Why is this a pull request??  Please learn the functions of GitHub and their purpose.

This is a good topic, but it belongs at esp32.com forum.  Esp32 does have modem sleep and DFS btw.

Alvin1Zhang:
@arnold-b Thanks for the contribution and sorry for the slow turnaround. We have looked into this, thanks.

mahavirj:
Please post questions on esp32.com forum. Closing as this is not valid PR.

PR #3783, opened 2019-07-15
tcp transport ssl DER-support (IDFGH-1518)

huelsenfruchtzwerg:
This pull-request is to add der-support to esp_transport_ssl. Or to be more precise, to clarify the already existing possibility to use der format for certificates and keys courtesy of mbedtls.
mbedtls supports der format just as well as pem format, assuming it is not disabled at compile time. The relevant functions autodetect the format, so all that is necessary is to pass the correct certificate-/key- length.
Unfortunately the existing esp_transport_ssl_set_client_cert etc. functions assume NULL-terminated pem, and (because mbedlts requires the NULL-byte to be included in the key-/cert-length parameter) add 1 to the length. This is wrong for DER, so I added _der variants that don't add 1 to the length back. (that's literally the only difference).
Ideally I'd also like to rename all *_pem_* variables, to reflect that they in fact don't need to be in PEM format, but since they are part of the public api of the libraries this would break existing code.

### DIFF
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -90,22 +90,28 @@ typedef struct esp_tls_cfg {
                                                  - where 'h2' is the protocol name */
  
     const unsigned char *cacert_pem_buf;    /*!< Certificate Authority's certificate in a buffer.
-                                                 This buffer should be NULL terminated */
+                                                 Format may be PEM or DER, depending on mbedtls-support
+                                                 This buffer should be NULL terminated in case of PEM */
  
     unsigned int cacert_pem_bytes;          /*!< Size of Certificate Authority certificate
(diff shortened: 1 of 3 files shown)


huelsenfruchtzwerg:
[https://github.com/espressif/esp-mqtt/pull/119](Related PR on esp_mqtt to add der-support)

david-cermak:
Thanks @huelsenfruchtzwerg for this helpful contribution!
@mahavirj @jitin17 PTAL.

david-cermak:
@huelsenfruchtzwerg Thank you for your patience and sorry for the slow turnaround. I've taken your changes to Espressif internal round of reviews and testing to be merged to master eventually.

Alvin1Zhang:
@huelsenfruchtzwerg Thanks for the contribution. The PR has been merged, thanks.

PR #3792, opened 2019-07-17
Low power management : Automatic light and deep sleep implementation (IDFGH-1528)

NegatuMichael:
The aim of this "pull request" is to contribute to the development of the low power management on esp. Automatic light sleep has already been implemented (pm_esp32.c) so these changes focus on the implementation of automatic deep sleep. 
You can find an example illustrating the working of the automatic light and deep sleep under 'esp-idf/examples/system/low_power_management`


negativekelvin:
That is interesting, but coming out of deep sleep you have lost all your application and hardware context so there is no way to automatically continue

hchaudhary1:
will ".light_sleep_enable = true", work with wifi?

NegatuMichael:
@negativekelvin : I agree with you : coming out of deep sleep you have lost all your application and hardware context. However the goal of this patch, as I said, was to easily switch between different power consumption modes using the locks. So when deep sleep is enabled and all the locks have been released, the chip goes automatically into deep sleep until a configured wakeup_call is called.

NegatuMichael:
@hchaudhary1 : Normally automatic light sleep works with WiFi. You just have to make sure that before entering any sleep mode you have properly stopped the WiFi. This is why I implemented the enter_hook and exit_hook functions. However, I'm having some trouble with the TCP/IP stack. Calling the  tcpip_adapter_init() function activates timers that force the system to quit light sleep mode even tough there are no tasks waiting to be executed. I have already posted an issue on the esp-idf forum.

shadowpho:
> @hchaudhary1 : Normally automatic light sleep works with WiFi. You just have to make sure that before entering any sleep mode you have properly stopped the WiFi. This is why I implemented the enter_hook and exit_hook functions. However, I'm having some trouble with the TCP/IP stack. Calling the tcpip_adapter_init() function activates timers that force the system to quit light sleep mode even tough there are no tasks waiting to be executed. I have already posted an issue on the esp-idf forum.

Do you have a link to the issue? I think I might be hitting that same issue -- I tear down esp32_wifi  but still does not go into light sleep mode.

ESP-Marius:
@NegatuMichael Thanks for the contribution and sorry for the (extremely) delayed response.

This MR is actually very well done: comments and an example, everything we want to see in a MR!

But from a use-case perspective I'm not sure how you would use it. I think it breaks a bit with the current expected behavior for esp-pm, i.e. calculate the time until something happens in our system, briefly sleep, and then continue handling the event we were waiting for. As mentioned above, with deep sleep everything is just gone, and there is no way of continuing execution.

So I think for now we are not going to merge this one and i'll mark it as closed.

As always, if anyone else feels like there is a good use-case for this feature then feel free to keep leaving comments in this thread and we will consider re-opening it.

PR #3799, opened 2019-07-19
Fix bug: Avoid sbc_plc_deinit(...) on unallocated memory. (IDFGH-1535)

nifomorph:
Once in a while, usually in situations when a "Peer connection failed (res: 16)" error is observed, it seems that ```sbc_plc_deinit(...)``` is called on unallocated memory. This causes a core panic due to an unhandled ```StoreProhibited``` exception. This happens when Bluedroid is using dynamic allocation.

```sbc_plc_deinit(...)``` just clears the ```plc_state``` instance memory and does not (currently) have any side-effects.

In any case, I think that ```bta_hf_client_sco_co_close()``` should be written so that it is safe to call it, even if initialisation/allocation has not been completed.

### DIFF
--- a/components/bt/host/bluedroid/btc/profile/std/hf_client/bta_hf_client_co.c
+++ b/components/bt/host/bluedroid/btc/profile/std/hf_client/bta_hf_client_co.c
@@ -254,12 +254,13 @@ void bta_hf_client_sco_co_close(void)
 
     if (hf_air_mode == BTM_SCO_AIR_MODE_TRANSPNT) {
 #if (PLC_INCLUDED == TRUE)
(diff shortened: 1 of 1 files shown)


blueMoodBHD:
Hi @nifomorph, Thanks very much for your issue and pull request. But there is some problems in your PR, `sbc_plc_deinit()` should be called even if dynamic memory is on. I have find anther way to fix it, and  it will be merged ASAP.

nifomorph:
Hi @blueMoodBHD, 
Thank you for addressing this. 
I am currently using the proposed fix in our own fork, so could you maybe clarify if the proposed fix itself  introduces problems in the current version of the code?

blueMoodBHD:
This pull request works OK, but not suggest to do as this.

PR #3839, opened 2019-07-26
Added menuconfig option for wpa_supplicant logging (IDFGH-1580)

badcf00d:
All the internal logs for wpa_supplicant seemed to be enabled by default so I added a menuconfig option to enable / disable them.

At the time of writing disabling this logging reduced the size of the binary file by about 60kb on LOG_LEVEL_DEBUG.

EDIT: On second thoughts the added size would depending what logging level you have set in your project, added that in the help message.

### DIFF
--- a/components/wpa_supplicant/Kconfig
+++ b/components/wpa_supplicant/Kconfig
@@ -6,4 +6,14 @@ menu "Supplicant"
         help
             Select this option to use MbedTLS crypto API's which utilize hardware acceleration.
 
+    config WPA_DEBUG_PRINT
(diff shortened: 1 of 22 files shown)


Alvin1Zhang:
Thanks for your contribution and very sorry for the slow reply, changes have been merged with https://github.com/espressif/esp-idf/commit/851658109c6d792deece46138333dcad54c77801#diff-0adb7edf1803ea7ca953fa09489415611126866fab8ceab77dcedc599b1ceaf3.

PR #3897, opened 2019-08-08
Accept relative path as env variable - CMake

domenbukovac:
When I tried to set up CMake and defined IDF_PATH relative to user's home `~/esp/esp-idf'` I got an error:  `No such file or directory: '~/esp/esp-idf/requirements.txt'` in check_python_dependencies.py as `with open("~/esp/esp-idf)` requires full file path. 

Fix expands IDF_PATH with user home if it starts with ~

### DIFF
--- a/tools/check_python_dependencies.py
+++ b/tools/check_python_dependencies.py
@@ -52,7 +52,9 @@ if __name__ == "__main__":
     args = parser.parse_args()
 
     not_satisfied = []
-    with open(args.requirements) as f:
+
+    absolute_path = os.path.expanduser(args.requirements)
+    with open(absolute_path) as f:
         for line in f:
             line = line.strip()
             try:


projectgus:
Hi @domenbukovac,

Thanks for sending this contribution.

I have some concern this may fix the immediate problem with having a `~` in IDF_PATH but move it further down into other parts of the system, as paths containing `IDF_PATH` may get passed into other Python scripts by CMake.

Did you find that everything else works OK after applying this change?

Alvin1Zhang:
@domenbukovac Thanks for the contribution. Would you please help share if any updates regarding @projectgus question "Did you find that everything else works OK after applying this change?" ? Thanks.

domenbukovac:
I didn't experience any problems so far on Mac OS.

PerMalmberg:
Adding my two cents to this. Expansion of ~ is really shell thing and as such I don't see it as a good thing being used in IDF_PATH. Having just done a lot of work re. CMake in my own project, I have no hard time imagining it becoming an issue.

I'm curious just how you set your environment variable, @domenbukovac as this yields the correct result:

```shell script
export IDF_PATH=~/esp/esp-idf
echo $IDF_PATH               
/home/permal/esp/esp-idf
```

Did you set IDF_PATH in your IDE? If that's the case then the underlying issue is that the IDE doesn't expand it when passing on the environment to the build.

projectgus:
Closing as @PerMalmberg's reasoning seems correct in this case, and there hasn't been any other update. (For the record: I also set IDF_PATH with `~` but I do it in a way that it's expanded by the shell and not pushed into CMake or idf.py).

PR #3940, opened 2019-08-19
Fixed syntax error in esp_http_server.rst (IDFGH-1701)

morsisko:
Changed the "const char[] name" into "const char name[]", so now the code doesn't produce the "expected identifier or '(' before '[' token" error.


mahavirj:
@morsisko Thanks for sending this across, I have put this in our internal review queue.

PR #3947, opened 2019-08-21
Improve bootloader partition table printing (IDFGH-1708)

Adam5Wu:
Fixed two bugs with bootloader partition printing:
1. Off-by-one error: counts the end marker as a normal partition,
last entry prints garbage;
2. Inconsistent handling of MD5 partition: esp_partition_table_basic_verify()
discounts the MD5 partition, but bootloader does not skip it when printing.

One improvement: Recognizes FAT and SPIFFS sub-type and prints them properly.


Alvin1Zhang:
@Adam5Wu Thanks for contributing. We will look into the PR. Thanks.

projectgus:
I'm really sorry that we never reviewed this PR when it was submitted back in 2019. The partition table implementation has changed substantially since then and this patch no longer applies.

In terms of the new feature to print additional partition types, we have a constant struggle to keep the bootloader small so for this reason we'd prefer not to add these.

PR #3958, opened 2019-08-23
Added the option to print system time rather than RTOS time in ESP_LOGx functions (IDFGH-1773)

badcf00d:
I mainly wrote this because I wanted to syncronize logs from an ESP32 with logs from a cloud platform that were timestamped in UTC, and I think other people would probably find this option useful too.

Obviously this time will just start from 0 unless you set the system time (with an SNTP sync for example), but assuming you do set the time it makes the logs much more useful when trying to match up events happening on multiple platforms.

Example from the hello world program:

```D (168) efuse: In EFUSE_BLK0__DATA5_REG is used 1 bits starting with 20 bit
I (168) cpu_start: Chip Revision: 1
I (170) cpu_start: Starting scheduler on PRO CPU.
D (0) intr_alloc: Connected src 25 to int 2 (cpu 1)
I (0) cpu_start: Starting scheduler on APP CPU.
D (171) heap_init: New heap initialised at 0x3ffe0440
D (171) heap_init: New heap initialised at 0x3ffe4350
D (171) intr_alloc: Connected src 16 to int 12 (cpu 0)
Hello world!
D (00:00:00.011) efuse: coding scheme 0
D (00:00:00.012) efuse: In EFUSE_BLK0__DATA3_REG is used 1 bits starting with 15 bit
D (00:00:00.013) efuse: coding scheme 0
D (00:00:00.013) efuse: In EFUSE_BLK0__DATA5_REG is used 1 bits starting with 20 bit
This is ESP32 chip with 2 CPU cores, WiFi/BT/BLE, silicon revision 1, 4MB external flash
Restarting in 10 seconds...

### DIFF
--- a/components/log/Kconfig
+++ b/components/log/Kconfig
@@ -44,17 +44,30 @@ menu "Log output"
 
             In order to view these, your terminal program must support ANSI color codes.
 
-    config LOG_SYSTEM_TIME
-        bool "Use system time in log output"
-        default "n"
+    choice LOG_TIMESTAMP_SOURCE
+        prompt "Log Timestamps"
+        default LOG_TIMESTAMP_SOURCE_RTOS
         help
-            Display the system time (HH:MM:SS.sss) in the log output rather than the 
-            RTOS time. Currently this will not get used in logging from binary blobs
-			(i.e WiFi & Bluetooth libraries), these will still print the RTOS tick time.
+            Choose what sort of timestamp is displayed in the log output:
 
-            The system time is initialized to 0 on startup, this can be set to 
-            the correct time with an SNTP sync, or manually with standard 
-            POSIX time functions.
+            - Milliseconds since boot is calulated from the RTOS tick count multiplied
+              by the tick period. This time will reset after a software reboot.
+              e.g. (90000)
+            
+            - System time is taken from POSIX time functions which use the ESP32's
+              RTC and FRC1 timers to maintain an accurate time. The system time is 
+              initialized to 0 on startup, it can be set with an SNTP sync, or with 
+              POSIX time functions. This time will not reset after a software reboot.
+              e.g. (00:01:30.000)
 
+            - NOTE: Currently this will not get used in logging from binary blobs
+              (i.e WiFi & Bluetooth libraries), these will always print 
+              milliseconds since boot.
+
+        config LOG_TIMESTAMP_SOURCE_RTOS
+            bool "Milliseconds Since Boot"
+        config LOG_TIMESTAMP_SOURCE_SYSTEM
+            bool "System Time"
+    endchoice
 
 endmenu
--- a/components/log/Kconfig
+++ b/components/log/Kconfig
@@ -44,5 +44,15 @@ menu "Log output"
 
             In order to view these, your terminal program must support ANSI color codes.
 
+    config LOG_SYSTEM_TIME
+        bool "Use system time in log output"
+        default "n"
+        help
+            Display the system time (HH:MM:SS.sss) in the log output rather than the 
+            RTOS time.
+
+            In order for this time to be correct, you must do an SNTP time sync,
+            otherwise this time will start from 0.
+
 
 endmenu
--- a/components/log/Kconfig
+++ b/components/log/Kconfig
@@ -49,10 +49,12 @@ menu "Log output"
         default "n"
         help
             Display the system time (HH:MM:SS.sss) in the log output rather than the 
-            RTOS time.
+            RTOS time. Currently this will not get used in logging from binary blobs
+			(i.e WiFi & Bluetooth libraries), these will still print the RTOS tick time.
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: components/log/Kconfig [hunk 1]
```diff
@@ -44,5 +44,17 @@ menu "Log output"
 
             In order to view these, your terminal program must support ANSI color codes.
 
+    config LOG_SYSTEM_TIME
```
projectgus: To make it easier for users to see the possibilities here, could you please make this a `choice` config item "Log timestamps", with two choices "System time" & "Milliseconds since boot"?

components/log/include/esp_log.h
```diff
@@ -86,6 +86,21 @@ vprintf_like_t esp_log_set_vprintf(vprintf_like_t func);
...
+/**
+ * @brief Function which returns system timestamp to be used in log output
+ *
+ * This function is used in expansion of ESP_LOGx macros to print
+ * the system time as "HH:MM:SS.sss". The system time is initialized to
+ * 0 on startup, this can be set to the correct time with an SNTP sync,
+ * or manually with standard POSIX time functions.
+ *
+ * Currently this will not get used in logging from binary blobs
+ * (i.e WiFi & Bluetooth libraries), these will still print the RTOS tick time.
```
projectgus: I'll check if we can resolve this atsome point in the future.

components/log/log.c
```diff
@@ -332,6 +334,52 @@ uint32_t ATTR esp_log_early_timestamp(void)
...
+    else
+    {
+        struct tm timeinfo;
+        struct timespec tv;
+        time_t now;
+
+        time(&now);
+        localtime_r(&now, &timeinfo);
+        clock_gettime(CLOCK_REALTIME, &tv);
+        sprintf(buffer,
```
projectgus: Logging can happen from multiple tasks simultaneously, so there is some potential here for the static `buffer` buffer to be simultaneously updated and logged in two different tasks.

I don't think this can result in a buffer overflow, but it can probably result in the timestamp showing a mixture of two time strings (because snprintf is updating buffer at the same time it's being copied to the output.) If this coincidences with a hour/minute/second overflow then it may give misleading results.

Suggest adding a lock in this function (newlib `_lock_t` should work) and holding it when updating `buffer`. This way any race condition will only result in the logging of the old timestamp or the new timestamp, never a partially updated buffer.

Also, for "belts and braces" safety please use snprintf() here and assert the result of the function call is in the valid range. Even though the format string should make it always safe to call, someone may update this in the future and not notice that the buffer is not overflowing.

components/log/include/esp_log.h
Frosticles: Yeh, essentially the problem was the print format for those log messages was still expecting an integer, not a string, so all the logging from the wifi binary was just printing the memory address of the buffer as an integer. If there is a way you could get the binaries to be compatible that would be great.

Frosticles:
I've added in the requests from @projectgus, given it a quick test my end and all seems well. Please do comment on any other changes you'd like but otherwise I'm happy to merge this for now.

components/log/log.c
```diff
@@ -332,6 +334,61 @@ uint32_t ATTR esp_log_early_timestamp(void)
...
+    }
+    else
+    {
+        struct tm timeinfo;
+        struct timespec tv;
+        time_t now;
+
+        if (bufferLock == 0)
+        {
+            _lock_init(&bufferLock);
```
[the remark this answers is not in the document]
Frosticles: Ah ok, didn't know that, I'll remove this bit

components/log/log.c
```diff
@@ -332,6 +334,61 @@ uint32_t ATTR esp_log_early_timestamp(void)
...
 #ifndef BOOTLOADER_BUILD
 
+char* IRAM_ATTR esp_log_system_timestamp(void)
+{
+    static char buffer[15] = {0};
+    static _lock_t bufferLock = 0;
+
+    if (xTaskGetSchedulerState() == taskSCHEDULER_NOT_STARTED) 
+    {
+        uint32_t timestamp = esp_log_early_timestamp();
```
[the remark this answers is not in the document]
Frosticles: Yeh I added this code because `snprintf()` isn't available while it's booting, so when I was using this function for early logging it was crashing because of something to do with newlib not initialised I think.

projectgus:
Thanks again @Frosticles . I've squashed the commits and pushed them into our internal review & merge queue. This PR will be updated again once that process is complete.

Frosticles:
Apologies for the late commit, not sure how I missed that but cppcheck spotted that I was accessing that buffer array out of bounds, just pushed a fix.

projectgus:
@Frosticles Thanks, can't believe we both missed that! Have updated our internal branch, should be merged soon (has been a national holiday in China so things got a little delayed).

projectgus:
Oops, sorry this wasn't closed earlier @badcf00d . Changes were cherry-picked as 
23e9224 .

PR #3972, opened 2019-08-27
kconfig: fix `unexpected operator` error on Linux (IDFGH-1742)

trombik:
and use portable backticks

introduced by me, d296aad causes `unexpected operator` warning during the build on Linux machine, which is harmless, but confuses users.

### DIFF
--- a/tools/kconfig/lxdialog/check-lxdialog.sh
+++ b/tools/kconfig/lxdialog/check-lxdialog.sh
@@ -4,10 +4,10 @@
 # What library to link
 ldflags()
 {
-	if [ $(uname -s) == "Darwin" ]; then 
+	if [ `uname -s` = "Darwin" ]; then
 		#OSX seems to need ncurses too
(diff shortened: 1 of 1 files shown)


dobairoland:
Hi @trombik. I'm not getting any warnings on Linux. However, we are reverting back the shebang from `#!/bin/sh` to `#!/bin/bash` because it caused some incompatibilities on MacOs as well.

Do you still getting those warnings if you revert it back locally on latest master?

trombik:
@dobairoland no, i do not get warnings on the latest master. i am fine with bash.

dobairoland:
Thank you @trombik and we also appreciate your intentions to fix this.

I will close this after commit f2a4131c241fa1959283e903bc884bc63b948af5 gets pushed to Github.

dobairoland:
Closed by f2a4131c241fa1959283e903bc884bc63b948af5. Please report if there is any subsequent issue.

Issue #3971, opened 2019-08-27
Problem with mbedtls component in IDF v4.1-dev-58-g02c7c3885 (IDFGH-1737)


ROYMURLI:
I was working with IDF v3.3-beta1-136-g97eecfa1b
 and developed my whole project in this version only. It is working good only I am facing a some small compatibility problem for which I downloaded new IDF v4.1-dev-58-g02c7c3885 and tried after removing my old  "build" folder from project. I am getting some mbedtls undefined error which seems to be some library and command line declaration problem. 
[sdkconfig.txt](https://github.com/espressif/esp-idf/files/3544652/sdkconfig.txt)

OS- Windows 10 (msys32)
xtensa-esp32-elf-gcc.exe (crosstool-NG crosstool-ng-1.22.0-80-g6c4433a5) 5.2.0

Things I did:- 
-Cloned the latest IDF and moved my project to example folder.
-Added "COMPONENT_SDKCONFIG_RENAMES := $(shell cygpath -w $(COMPONENT_SDKCONFIG_RENAMES))"
after line https://github.com/espressif/esp-idf/blob/master/make/project_config.mk#L11 
Because I was facing problem with "no such file and directory .............. sdkconfig_rename" which is solved using above solution.
- rm -r build did for build directory.
- sdkconfig are attached below.
Debug logs are given below:-

D:/msys32/home/MurlidharRoy/Newfolder/esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c: In function 'esp_secure_boot_verify_signature_block':
D:/msys32/home/MurlidharRoy/Newfolder/esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c:96:5: error: unknown type name 'mbedtls_ecdsa_context'
mbedtls_ecdsa_context ecdsa_context;
^
D:/msys32/home/MurlidharRoy/Newfolder/esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c:97:5: error: implicit declaration of function 'mbedtls_ecdsa_init' [-Werror=implicit-function-declaration]
mbedtls_ecdsa_init(&ecdsa_context);
^
D:/msys32/home/MurlidharRoy/Newfolder/esp-idf/components/bootloader_support/src/idf/secure_boot_signatures.c:99:5: error: implicit declaration of function 'mbedtls_ecp_group_load' [-Werror=implicit-function-declaration]
[...]

Issue #3961, opened 2019-08-24
mbedTLS: Client only does not compile


Reiner1210:
When I switch mbedTLS TLS role from Server & Client to Client the code will not compile

<code>
/home/reiner/data/programm/esp/esp-idf/components/coap/port/coap_mbedtls.c: In function 'setup_server_ssl_session':
/home/reiner/data/programm/esp/esp-idf/components/coap/port/coap_mbedtls.c:897:3: error: implicit declaration of function 'mbedtls_ssl_conf_dtls_cookies'; did you mean 'mbedtls_ssl_conf_sig_hashes'? [-Werror=implicit-function-declaration]
   mbedtls_ssl_conf_dtls_cookies(&m_env->conf, mbedtls_ssl_cookie_write,
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   mbedtls_ssl_conf_sig_hashes
CC build/bootloader/bootloader_support/src/bootloader_clock.o
/home/reiner/data/programm/esp/esp-idf/components/coap/port/coap_mbedtls.c: In function 'coap_dtls_hello':
/home/reiner/data/programm/esp/esp-idf/components/coap/port/coap_mbedtls.c:1646:15: error: implicit declaration of function 'mbedtls_ssl_set_client_transport_id'; did you mean 'mbedtls_ssl_conf_transport'? [-Werror=implicit-function-declaration]
     if((ret = mbedtls_ssl_set_client_transport_id(&m_env->ssl,
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
               mbedtls_ssl_conf_transport

</code

negativekelvin:
@mrdeep1

PR #3977, opened 2019-08-28
Fix libcoap building with various MbedTLS compile time options

mrdeep1:
See #3961 and #3971 

components/coap/port/coap_mbedtls.c:

Add in additional compile time check wrappers for MbedTLS configuration

components/coap/port/include/coap_config_posix.h:
components/lwip/Kconfig:
components/lwip/port/esp32/include/lwipopts.h:

Add in LwIP configuration support for IP_PKTINFO (default disabled) to
determine Server's destination IP address used for CoAP logging.

examples/protocols/coap_client/main/coap_client_example_main.c:

Inform user that MbedTLS Client Mode is required for DTLS if not emabled.
[Lower libcoap library will still error and report this condition]

examples/protocols/coap_server/main/coap_server_example_main.c:

Inform user that MbedTLS Server Mode is required for DTLS if not emabled.
[Lower libcoap library will still error and report this condition]


mrdeep1:
@mahavirj FYI

examples/protocols/coap_server/main/coap_server_example_main.c
```diff
@@ -252,12 +252,16 @@ static void coap_example_server(void *p)
         }
 #if defined(CONFIG_COAP_MBEDTLS_PSK) || defined(CONFIG_COAP_MBEDTLS_PKI)
         if (coap_dtls_is_supported()) {
+#ifndef CONFIG_MBEDTLS_TLS_SERVER
+            ESP_LOGI(TAG, "MbedTLS (D)TLS Server Mode not configured");
```
mahavirj: This can be changed to `ESP_LOGE` and add `goto clean_up`

components/lwip/Kconfig
```diff
@@ -90,6 +90,13 @@ menu "LWIP"
         help
             Enabling this option allows checking for available data on a netconn.
 
+    config LWIP_NETBUF_RECVINFO
```
mahavirj: I would request this to be separate PR, unless it has dependency on other fix (i.e. build issues with mbedtls client only mode configuration).

mahavirj:
@mrdeep1 I was working on similar change internally, but I will pull in your PR. Probably splitting this in 2 PRs would be helpful. Thanks.

components/lwip/Kconfig
mrdeep1: Ok - will split it later today when I get back to the office

examples/protocols/coap_server/main/coap_server_example_main.c
mrdeep1: If not configured, the the server needs to support just plain coap:// and not fail - hence what I did

mrdeep1:
I have realised that if mbedtls is completely disabled, things will not build - I have a clean fix in mind - will do this later today

mrdeep1:
@mahavirj PR split and fixed up to including rebuilding when MbedTLS is not configured for a TLS mode.

examples/protocols/coap_server/main/coap_server_example_main.c
mahavirj: I was thinking that we should not allow non-secure connections, but should be okay for this case as user would have explicitly disabled default configuration from mbedtls/coap.

components/coap/port/coap_notls.c
```diff
@@ -0,0 +1,190 @@
+/*
+* coap_notls.c -- Stub Datagram Transport Layer Support for libcoap
```
mahavirj: (One doubt) This file will eventually move to `libcoap` going ahead, right?

mrdeep1: Yes, once https://github.com/obgm/libcoap/pull/353 is merged and then libcoap is updated in esp-idf, there will be no need for port/coap_notls.c or port/coap_debug.c.

Once these MbedTLS changes to port/coap_mbedtls.c are merged back into libcoap (which I am working on), there will be no need for port/coap_mbedtls.c when libcoap is updated in esp-idf.

mahavirj: Sounds good, thanks.

Issue #3941, opened 2019-08-19
CONFIG_BTU_TASK_STACK_SIZE no longer works (IDFGH-1702)


zacwbond:
## Environment

- Development Kit:      [none]
- Module or chip used:  [ESP32]
- IDF version (run ``git describe --tags`` to find it): 
    // v4.0-dev-1443-g39f090a4f
- Build System:         [CMake]
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it):
    // 1.22.0-80-g6c4433a
- Operating System:     [Linux|]
- Power Supply:         [external 3.3V]

## Problem Description

In order to turn on Bluetooth logging, in the past I have had to increase the config option CONFIG_BTU_TASK_STACK_SIZE.   It looks like this config option is no longer used in recent versions of the code and the BTU stack size has been hard-coded to 4096.  I think this options should be added again; without it, I receive 

```***ERROR*** A stack overflow in task btuT has been detected.```

whenever my BLE peripheral connects.  Also, it seems like all the other tasks have configurable stack sizes, and this option is even mentioned in documentation. I think its removal was probably an accident.

### Expected Behavior

CONFIG_BTU_TASK_STACK_SIZE changes the btuT stack size.

### Actual Behavior

CONFIG_BTU_TASK_STACK_SIZE doesn't change the btuT stack size.

### Steps to reproduce

1. Change CONFIG_BTU_TASK_STACK_SIZE.
2. Rebuild.
3. Discover that the btuT stack size is unchanged.

negativekelvin:
Relevant commit
https://github.com/espressif/esp-idf/commit/faf23df19ac936710b12d36d56683d0bf0468189

blueMoodBHD:
Hi, This bug have been fixed, refer to commit 07bd6d4

Alvin1Zhang:
@zacwbond Hi, would you please help try the fix commit and check whether your issue has been fixed? Thanks.

zacwbond:
Looks like this has been resolved!  Thanks.

PR #3984, opened 2019-08-28
Use CONFIG_BTU_TASK_STACK_SIZE instead of hard-coded value in Bluedroid component (IDFGH-1757)

zacwbond:
I simply replaced a hard-coded stack size with the value pulled from the configuration, which is how it worked in the past.  Without this, I can't use Bluetooth at all without getting stack overflows.

I logged [IDFGH-1702](https://github.com/espressif/esp-idf/issues/3941) for this last week, and it seems to be a very simple fix.

### DIFF
--- a/components/bt/host/bluedroid/common/include/common/bluedroid_user_config.h
+++ b/components/bt/host/bluedroid/common/include/common/bluedroid_user_config.h
@@ -33,6 +33,12 @@
 #else
 #define UC_A2DP_SOURCE_TASK_STACK_SIZE      2048
 #endif
+#ifdef CONFIG_BT_BTU_TASK_STACK_SIZE
+#define UC_BT_BTU_TASK_STACK_SIZE           CONFIG_BT_BTU_TASK_STACK_SIZE
+#else
+#define UC_BT_BTU_TASK_STACK_SIZE           4096
+#endif
+
 
 /**********************************************************
  * Profile reference
--- a/components/bt/host/bluedroid/common/include/common/bt_target.h
+++ b/components/bt/host/bluedroid/common/include/common/bt_target.h
@@ -45,6 +45,7 @@
 /* OS Configuration from User config (eg: sdkconfig) */
 #define A2DP_SINK_TASK_STACK_SIZE   UC_A2DP_SINK_TASK_STACK_SIZE
 #define A2DP_SOURCE_TASK_STACK_SIZE UC_A2DP_SOURCE_TASK_STACK_SIZE
+#define BT_BTU_TASK_STACK_SIZE      UC_BT_BTU_TASK_STACK_SIZE
 
 /******************************************************************************
 **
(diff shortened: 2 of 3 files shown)


negativekelvin:
Should be BT_BTU_TASK_STACK_SIZE

zacwbond:
Thanks for the feedback.  I have updated `bluedroid_user_config.h` and `bt_target.h` as suggested.

blueMoodBHD:
Thanks, but it should be `CONFIG_BTU_TASK_STACK_SIZE`, as we rename it in `sdkconfig.rename`
I have create a MR, and add a commit behead yours, to change the name of macro.

negativekelvin:
@blueMoodBHD why would you want to use the deprecated name?

```
# sdkconfig replacement configurations for deprecated options formatted as
# CONFIG_DEPRECATED_OPTION CONFIG_NEW_OPTION
CONFIG_BTU_TASK_STACK_SIZE                                  CONFIG_BT_BTU_TASK_STACK_SIZE
```

blueMoodBHD:
It is just for the same with `BTC_TASK_STACK_SIZE`. 
```
#ifdef CONFIG_BTC_TASK_STACK_SIZE
#define UC_BTC_TASK_STACK_SIZE              CONFIG_BTC_TASK_STACK_SIZE
#else
#define UC_BTC_TASK_STACK_SIZE              3072
#endif
```
In bt_user_config.h, `UC_BTC_TASK_STACK_SIZE` is defined according to `CONFIG_BTC_TASK_STACK_SIZE`,  instead of the new one. So, I think it is better to keep them same.

negativekelvin:
If it is going to be deprecated it should be changed everywhere and only exist in sdkconfig.rename

TianHao-Espressif:
@negativekelvin , The `CONFIG_BT_BTU_TASK_STACK_SIZE` is to convert `CONFIG_BTU_TASK_STACK_SIZE` in sdkconfig.rename. Now, all the Bluedroid code is using `CONFIG_BTU_TASK_STACK_SIZE`, so although we add new code, we should use `CONFIG_BTU_TASK_STACK_SIZE` instead `CONFIG_BT_BTU_TASK_STACK_SIZE` to keep the usage uniq. We will consider to use `CONFIG_BT_xxx` to insead of `CONFIG_xxx` in bluedroid and remove sdkconfig.rename later.

zacwbond:
Is there a consensus on the naming?  To me it seems bizarre to have sdkconfig.rename define `CONFIG_BT_BTU_TASK_STACK_SIZE` but then never use it anywhere.

blueMoodBHD:
Hi, This PR have been merged, refer to commit f8212988c4c9c52849c96dd60e80f4376cfaafc0 & 07bd6d49ca4894dc5a19b2c4607ed5eb2d27b728(merged commit)

zacwbond:
Looks like it's working now.  I'll go ahead and close this PR.

PR #3999, opened 2019-09-02
gpio_isr_register(): Correct order of arguments in docs (IDFGH-1780)

thierer:
Make documentation order consistent with function signature.

### DIFF
--- a/components/driver/include/driver/gpio.h
+++ b/components/driver/include/driver/gpio.h
@@ -395,9 +395,9 @@ esp_err_t gpio_wakeup_disable(gpio_num_t gpio_num);
  * per-GPIO ISRs.
  *
  * @param  fn  Interrupt handler function.
+ * @param  arg  Parameter for handler function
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@thierer Thanks for the contribution. We will look into this PR, thanks.

X-Ryl669:
Is the order important ? 
I thought the documentation tool would find out the param name in the line (just after the `@param` token) and match it with the function signature. At least Doxygen does this.

thierer:
> I thought the documentation tool would find out the param name in the line (just after the @param token) and match it with the function signature. At least Doxygen does this.

[The generated documentation suggests that it does not](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/peripherals/gpio.html#_CPPv417gpio_isr_registerPFvPvEPviP17gpio_isr_handle_t):

Screenshot 2019-09-05 at 23 22 34

igrr:
Thanks for the PR @thierer, the fix has been cherry-picked as 75e0c6a.

PR #4025, opened 2019-09-05
fix(FreeRTOS): Initialize uxTaskNumber at task initialization (IDFGH-1815)

GautierAtWork:
```uxTaskNumber``` member of ```struct tskTaskControlBlock``` is not initialized at thread creation.
As the TCB can be dynamically allocated, ```uxTaskNumber``` might then get a value from the heap.
As a result, the user might get a seems-valid value when calling ```uxTaskGetTaskNumber()``` and then take wrong decisions.

This fix zeroes ```uxTaskNumber``` when the task is initialized.


Alvin1Zhang:
@GautierAtWork Thanks for the contribution, we will look into this PR. Thanks.

PR #4051, opened 2019-09-11
putchar is not thread safe and eventually halts the ESP32 under certa… (IDFGH-1842)

embeblue:
…in conditions. No WDT or task WDT work in that situation nor crash report is displayed. The device freezes for good.

### DIFF
--- a/examples/protocols/https_request/main/https_request_example_main.c
+++ b/examples/protocols/https_request/main/https_request_example_main.c
@@ -129,7 +129,7 @@ static void https_get_task(void *pvParameters)
             ESP_LOGD(TAG, "%d bytes read", len);
             /* Print response directly to stdout as it is read */
             for(int i = 0; i < len; i++) {
-                putchar(buf[i]);
+                printf("%c", buf[i]);
             }
         } while(1);
 


negativekelvin:
How about removing the for loop?

embeblue:
Hi @negativekelvin,
We thought about that but if the payload we receive is binary a printf("%s"...) wouldn't work, hence keeping the for loop.

negativekelvin:
It is HTTP so there shouldn't be any null in it and printf with specified max length should work or you could do fwrite to stdout

embeblue:
I agree, but that is not the point, nor it is the problem. The for loop was allready there on the first place. It can also be deleted entirely if that pleases you because it's only for demo purposes (this modification is withing an example project) Or it can be substituted by a printf("%s") or whatever other solution that does not include a putchar.

projectgus:
Hi @embeblue ,

> putchar is not thread safe and eventually halts the ESP32 under certain conditions. No WDT or task WDT work in that situation nor crash report is displayed. The device freezes for good.

Do you have any more details about this bug that you can share?

Does the bug happen in the unmodified example, or is there some additional change that causes it to freeze up in this way?

The reason I'm asking is that calling `putchar` should be thread safe, apart from the risk that another task will also write to stdout and the characters will become garbled. Also, as far as I understand newlib I believe the two code paths should end up in the same place:

* The `putchar()` macro expands to `putc(x, stdout)` which will result in an internal newlib call to `_putc_r`. `_putc_r` will lock stdout before writing into it.
* `printf()` will result in a call to `vprintf()` which outputs characters by expanding a macro `PRINT` which expands `__SPRINT` which expands to static function `__sfputs_r` which calls `_fputc_r` which calls `_putc_r`.

If changing `putchar` to `printf` fixes a hang, it may be because it changes the timing (the CPU spends more time processing the formatting string each time) rather than because it fixes whatever the root cause of the bug is.

One possible explanation is that the code in IDF that writes stdout to the UART is deadlocking, because there's a code path in there that calls `printf()` or similar (which will try to lock stdout again, causing the deadlock). This shouldn't happen, but something like this might cause the symptoms that you describe (the watchdogs can't detect a deadlock). Using `gdb` to halt the CPU at this point and dump a backtrace would confirm it.

embeblue:
Hi @projectgus,

First, thanks for your comments on this matter. I suppose this bug does not happen in the example given it's reduced number of tasks and complexity. We found it in our project (where we used the example code for https communication). Our project has several tasks for different purposes and it only freezes once a day (we communicate every 5 seconds)

What you say about the deadlock makes sense to me but then I do not understand why using putchar is any different from using printf appart from the different timing indeed. To be on the safe side I will completely get rid of the for loop and the printf. Maybe the right thing would be using the ESP_LOGx() function right?

projectgus:
> To be on the safe side I will completely get rid of the for loop and the printf. Maybe the right thing would be using the ESP_LOGx() function right?

ESP_LOGx() is a wrapper around printf, so they will end up calling the same code in newlib.

This kind of timing issue can be really hard to track down, it's even possible that it's not related to stdout locking but it just happens that making this print loop less efficient (due to lots of printf formatting) is changing the timing of something else and avoids a deadlock or some other problem.

Attaching JTAG to a "stuck" ESP32 and using the debugger to dump a backtrace for all the tasks is the only to see for sure where it's stuck.

Are you using PSRAM, by the way?

embeblue:
No, we are using only internal RAM.
If we manage to have the problem with the JTAG connected I will post the backtrace here.
Thanks again.

PR #4085, opened 2019-09-18
Fix websocket frame handling in components/tcp_transport (IDFGH-1882)

johnnyxwan:
The _components/esp_websocket_client_ sends PING frame to server every 10 seconds by default, this usually triggers the server replying with a PONG frame, which could trigger a read error. This pull request fixed this issue, and made the following minor changes in _components/tcp_transport_:
1. Reply with PONG frame when a PING frame from server is received. This prevent timeout if server has a keep-alive mechanism.
2. Handle websocket frame with size over buffer size without error.

While using _components/esp_websocket_client_ to test, also fixed the following issue:
1. Fix path in client config is not being applied.
2. Rephrasing debug message prevent misunderstanding.


david-cermak:
@johnnyxwan Thanks for your help in fixing the websocket client. I'm sorry we have to close this PR as the websocket client has been refactored and the issues fixed in your code are already addressed.

Alvin1Zhang:
@johnnyxwan Thanks for the contribution.

PR #4093, opened 2019-09-19
Expand console example to support ping (ICMP echo) (IDFGH-1890)

aknooh:
This pull request adds support for ping (ICMP) echo.
It piggybacks on the console example to add ping support and CLI.
Ping is usually the first tool that engineers use to diagnose network problems, so it's very useful to have.


suda-morris:
Merged in https://github.com/espressif/esp-idf/commit/a77e69c40820873654b288107b60cc685f93227b Thanks!

PR #4098, opened 2019-09-19
Typo fix (IDFGH-1899)

boarchuz:
rd_reg comment references incorrect OPCODE ("OPCODE_WR_REG"); amended to "OPCODE_RD_REG".


Alvin1Zhang:
Thanks for reporting. We will look into. @boarchuz

PR #4139, opened 2019-09-30
esp_event: fix crash when unregistering a handler instance in itself (IDFGH-1940)

xentec:
When a handler instance is the last one in the list und unregisters itself, the handler iterator will be invalidated by entering free'd memory. Same applies for event base and id, if they become empty.

### DIFF
--- a/components/esp_event/esp_event.c
+++ b/components/esp_event/esp_event.c
@@ -526,30 +526,30 @@ esp_err_t esp_event_loop_run(esp_event_loop_handle_t event_loop, TickType_t tick
 
         bool exec = false;
 
-        esp_event_handler_instance_t *handler;
+        esp_event_handler_instance_t *handler, *temp_handler;
         esp_event_loop_node_t *loop_node;
-        esp_event_base_node_t *base_node;
-        esp_event_id_node_t *id_node;
+        esp_event_base_node_t *base_node, *temp_base;
+        esp_event_id_node_t *id_node, *temp_id_node;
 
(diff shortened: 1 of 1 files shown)


components/esp_event/esp_event.c
```diff
@@ -526,30 +526,30 @@ esp_err_t esp_event_loop_run(esp_event_loop_handle_t event_loop, TickType_t tick
...
 
-        esp_event_handler_instance_t *handler;
+        esp_event_handler_instance_t *handler, *temp_handler;
         esp_event_loop_node_t *loop_node;
-        esp_event_base_node_t *base_node;
-        esp_event_id_node_t *id_node;
+        esp_event_base_node_t *base_node, *temp_base;
+        esp_event_id_node_t *id_node, *temp_id_node;
 
         SLIST_FOREACH(loop_node, &(loop->loop_nodes), next) {
```
renzbagaporo: This seems to also need the safe foreach variant.

xentec: Sorry for the long silence, but you're right. I've also seen your fixup.

xentec:
After further debugging, I've detected a heap corruption when using event profiling. When unregistering event handlers in themselves, the `handler` instance is free'd, during execution, but its fields are still written to from profiling code afterwards: 
https://github.com/espressif/esp-idf/blob/0a0f2caa1d30e0f8cbfe31475093492e09995763/components/esp_event/esp_event.c#L132-L140

renzbagaporo:
Hi @xentec, can you open a new issue to track this? I only noticed your reply now.

PR #4142, opened 2019-10-01
Add KConfig option to allow using other Bluetooth Host Stack (IDFGH-2062)

mringwal:
With the current KConfig for the Bluetooth component, it is not possible to select neither Bluedroid nor NimBLE. If disabling both, the default (Bluedroid) is activated. 

The pull request adds an "Other" option that allows use a different Bluetooth Stack, e.g. [BTstack](https://github.com/bluekitchen/btstack/tree/master/port/esp32).

### DIFF
--- a/components/bt/Kconfig
+++ b/components/bt/Kconfig
@@ -378,6 +378,11 @@ menu "Bluetooth"
             help
                 This option is recommended for BLE only usecases to save on memory
 
+        config BT_OTHER_ENABLED
+            bool "Other Host Stack"
+            help
+                This option is recommended to use with other Bluetooth Host stack
+
     endchoice
 
     menu "Bluedroid Options"


piotrkundu:
👍

gengyuchao:
Thank you for your contribution, but we currently have no plans to add this menuconfig option. It is not reasonable to just close Bluedroid or Nimble without modifying the controller's configuration menu. If you want to use your own Bluetooth Host protocol stack, you can turn off our Bluetooth option, call the controller's API function directly, or write a new menuconfig of your example to configure the Bluetooth Stack. Thank you.

GYC-Espressif:
Thank you for your contribution, this is a great suggestion. We will add this option later. Currently, if you want to use your own Bluetooth host protocol stack, you can turn off our Bluetooth option, call the controller's API function directly, or write a sample new menuconfig to configure the Bluetooth Stack. It is possible to use our controller without adding this menuconfig option. You can refer to the two examples of controller_hci_uart and controller_vhci_ble_adv to implement your function. Thank you.

negativekelvin:
This PR allows setting CONFIG_BT_ENABLED without defining CONFIG_BT_BLUEDROID_ENABLED.  Is another clean way to do it?  Because CONFIG_BT_ENABLED is used in makefiles to build controller sources and user does not want to compile bluedroid sources.

https://github.com/espressif/esp-idf/blob/fb3d648e07033b952578f187b4fcbf7fd8aced3b/components/bt/CMakeLists.txt#L1

GYC-Espressif:
Sorry, we still need to open the Bluetooth option. But if you don‘t call the code related to Bluedroid, the Bluedroid code will not be compiled into the bin file. 
I tried to compile the gatt_server_demos, the total image size is 676802 bytes,
and then remove the Bluedroid code called in main, the total image size is 313642 bytes,
and then remove the CONFIG_BT_BLUEDROID_ENABLED definition, the total image size is 313642 bytes too.
So whether to change the compilation option will not affect the final compilation result. But we can reduce the compilation process by adding this compilation option.

chinmaychhajed:
Hello @mringwal this PR has been handled internally along with some other changes which would be reflected soon. You can close this PR. Thanks for your contribution!

mringwal:
Thanks for reviewing this. This, or a similar change, brings back the behaviour before NimBLE was added (it was possible to disable Bluedroid in KConfig). Closing as suggested and hoping to see it it soon on master.

PR #4167, opened 2019-10-06
idf_monitor: Exit with CTRL+X in menu (IDFGH-1971)

possan:
Currently, the only way of exiting the idf_monitor program is to hit the CTRL+] button, if your keyboard doesn't have that key unless you hit another modifier key, it's not super trivial to exit.

This change adds the option to exit bt CTRL+T (for menu) then hitting X (or CTRL+X) for exiting.


dobairoland:
Thank you @possan for your contribution. This will be reviewed internally and I get back to you if any request arise. Otherwise, it will be merged in this form.

Thank you!

PR #4171, opened 2019-10-07
newlib: fix compile error of toupper/tolower (IDFGH-1975)

gozfree:
otherwise, error will occured as follows:

...
ctype.h:57:54: error: array subscript has type 'char' [-Werror=char-subscripts]

### DIFF
--- a/components/newlib/include/ctype.h
+++ b/components/newlib/include/ctype.h
@@ -54,7 +54,7 @@ extern	__IMPORT char	* _CONST __ctype_ptr__;
    Meanwhile, the real index to __ctype_ptr__+1 must be cast to int,
    since isalpha(0x100000001LL) must equal isalpha(1), rather than being
    an out-of-bounds reference on a 64-bit machine.  */
-#define __ctype_lookup(__c) ((__ctype_ptr__+sizeof(""[__c]))[(int)(__c)])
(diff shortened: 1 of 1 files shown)


igrr:
Hi @gozfree, can you please describe the circumstances when you get this error?

The file you are modifying, `components/newlib/include/ctype.h`, is not used when building with the latest GCC 8.2 based toolchain. Instead, `include/ctype.h` is used from the toolchain archive. The file in the toolchain archive is from newlib 3.0.0, which is the current version we use. The file included in IDF is from an older newlib 2.2.0. If this file somehow gets included instead of the file from the toolchain archive, this error may occur. However per our tests, the correct file is included both with Make and CMake build systems, with latest IDF master branch.

Please provide more information to diagnose this issue: IDF commit ID, which build system you use, any modifications you make in the project or component makefiles which affect CPPFLAGS or CFLAGS. If you can provide a self-sufficient example to reproduce this, it would be prefect.

gozfree:
I just use an older toolchain xtensa-esp32-elf-linux64-1.22.0-80-g6c4433a-5.2.0.tar.gz
to compile project. After updated to the latest, it seems not report error any more.
Thanks.

igrr:
Thanks for confirming, closing this PR as the issue is resolved.

PR #4182, opened 2019-10-09
Fix UART RS485 Failing to Enable Transmitter After Some Time (IDFGH-1983) (IDFGH-1987)

dasrue:
Added the code to clear the tx done interrupt before enabling the interrupt to fix the issue of the transceiver not being enabled after some time of running in RS485 mode. I suspect an interrupt is being missed for some reason, which causes the interrupt and fifo fill to get out of sync. By clearing the interrupt before enabling it and writing to the fifo, we make sure it will always stay in sync


negativekelvin:
Suggest you use format 
`UART[uart_num]->int_clr.tx_done = 1;`
for consistency.  Also there is one other place where it should be added.

alisitsyn:
The solution above was not confirmed as a fix for issue. See https://esp32.com/viewtopic.php?f=13&t=21835&p=80501#p80501

Alvin1Zhang:
Thanks for your contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/a1b5c382181fa26844b201684dcaa89556696044.

PR #3365, opened 2019-04-25
Add config option for LWIP TCP_TMR_INTERVAL (IDFGH-1040)

JosuGZ:

PR #4213, opened 2019-10-18
Add config option for LWIP TCP_TMR_INTERVAL (IDFGH-2041)

JosuGZ:
This is the same change from https://github.com/espressif/esp-idf/pull/3365

I haven't tested it though; since I'm still on the 3.2 branch, but it should do the same thing.


suda-morris:
Thanks for your contribution, we will move it into our internal review queue

Alvin1Zhang:
@JosuGZ Thanks for the contribution.

PR #4216, opened 2019-10-19
Add execute permission to export.sh script (IDFGH-2026)

SchnitzlerJ:
`install.sh` has the execute permission, `export.sh` hasn't.


SchnitzlerJ:
Seems like I should read the docs more thorougly :D

PR #4219, opened 2019-10-19
esp_http_client: Add SOURCE method for NTRIP (IDFGH-2029)

nebkat:
Now that 742d4689dd728cb3bd5f06747b04b7d8930b7d02 has been merged, the [NTRIP v1.0 protocol](http://www.wsrn3.org/CONTENT/Reference/Reference_NTRIP-V1-Tech-paper.pdf) is almost entirely supported using the http client, which greatly simplifies its implementation.

In order to allow data upload to an NTRIP caster, the custom SOURCE method is required. The existing SUBSCRIBE/UNSUBSCRIBE/etc are already non-standard methods.


shubhamkulkarni97:
@nebkat Is there any publicly available server instance to validate working of NTRIP with ESP-IDF

PR #4251, opened 2019-10-27
Fix build error semphr.h:74:3: error: #error "include FreeRTOS.h" must appear in source files before "include semphr.h" (IDFGH-2077)

motla:
FreeRTOS.h must always be included before semphr.h


Alvin1Zhang:
@motla Thanks for the contribution.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/5491365874528b1532d1a79a397944515a7371de

PR #4255, opened 2019-10-28
Update Flash_ops.c to enable direct communication with the SPI FLASH (IDFGH-2085)

filzek:
This code enable the flash to be read and return it internal registers, extreme usefull to know flash id, flash version, flash vendor, flash size, flash serial number, and so many others info available from the flash manufacturer.

Example usage to return a valid flash ID:

uint8_t flash_unique_id[9] = {0};
// Get flash unique id 0x4b
uint8_t command[13] = {0x4b, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
// Get JEDEC ID 0x9f
//uint8_t command[13] = {0x9f, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
// Get Manufacturer ID 0x90
//uint8_t command[13] = {0x90, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
uint8_t response[13];
spi_flash_send_cmd(sizeof(command), command, response);

Simple, and can send any command from the flash vendor.

### DIFF
--- a/components/spi_flash/flash_ops.c
+++ b/components/spi_flash/flash_ops.c
@@ -679,6 +679,132 @@ static esp_err_t IRAM_ATTR spi_flash_translate_rc(esp_rom_spiflash_result_t rc)
     }
 }
 
+// Those code below enable direct communication with the SPI FLASH, such as almost any command available from any flash vendor or manufacturer	
+#include "soc/spi_reg.h"		
+#include "soc/soc.h"		
+static void IRAM_ATTR spi_master_ops(int unit, unsigned int word_size, unsigned int len, unsigned char *out, unsigned char *in) {		
+unsigned int bytes = word_size * len; // Number of bytes to write / read		
+unsigned int idx = 0;		
+/*		
+ * SPI data buffers hardware registers are 32-bit size, so we use a		
+ * transfer buffer for adapt user buffers to buffers expected by hardware, this		
+ * buffer is 16-word size (64 bytes)		
+ *		
+ */		
+uint32_t buffer[16]; // Transfer buffer		
+uint32_t wd;         // Current word		
+unsigned int wdb; 	 // Current byte into current word		
+// This is the number of bits to transfer for current chunk		
+unsigned int bits;		
+bytes = word_size * len;		
+while (bytes) {		
+  // Populate transfer buffer in chunks of 64 bytes		
+  idx = 0;		
+  bits = 0;		
+  while (bytes && (idx < 16)) {		
+    wd = 0;		
+    wdb = 4;		
+    while (bytes && wdb) {		
+      wd = (wd >> 8);		
+      if (out) {		
+        wd |= *out << 24;		
+        out++;		
+      } else {		
+        wd |= 0xff << 24;		
+      }		
+      wdb--;		
+      bytes--;		
+      bits += 8;		
+    }		
+    while (wdb) {		
+      wd = (wd >> 8);		
+      wdb--;		
+    }		
+    buffer[idx] = wd;		
+    idx++;		
+  }		
+  // Wait for SPI bus ready		
+  while (READ_PERI_REG(SPI_CMD_REG(unit))&SPI_USR);		
+  // Load send buffer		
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@filzek Thanks for the contribution.

filzek:
I will provide more codes but I need to port it to latest if first, so it need some corrections, as I finish I will request some new pushes.

igrr:
@filzek Please note that the flash_ops.c contains the legacy SPI flash code; there is a new driver, with APIs in esp_flash.h. It relies on a chip-specific driver in spi_flash_chip_driver.h.

If you are implementing support for a new SPI flash chip, you can define a driver for it, see spi_flash_chip_gd.c, spi_flash_chip_issi.c, etc.

filzek:
I will check tomorrow, as legacy it does work for any native flash (app) on esp32.

I will try to use the new driver code to read the I'd from flash, and cross check the results and post it tomorrow.

I did not find any example on read flash Id for the current spi flash on esp32, but will try to run it.

filzek:
@igrr  I try to run the code:

esp_flash_t * internnal_flash_chip;
uint32_t id, size;
esp_flash_read_id(&internnal_flash_chip, &id);
esp_flash_get_size(&internnal_flash_chip, &size);
printf("Flash ID %08x detected size %d bytes\n", id, size);

but it only returns 0000000..

I did not understant quite how to use the code to read the internal flash id and its size, so, no luck at all.

Can you clarify steps?

filzek:
> @igrr I try to run the code:
> 
> esp_flash_t * internnal_flash_chip;
> uint32_t id, size;
> esp_flash_read_id(&internnal_flash_chip, &id);
> esp_flash_get_size(&internnal_flash_chip, &size);
> printf("Flash ID %08x detected size %d bytes\n", id, size);
> 
> but it only returns 0000000..
> 
> I did not understant quite how to use the code to read the internal flash id and its size, so, no luck at all.
> 
> Can you clarify steps?

we need to read the internal ESP32 flash id.

all attemp we tried based on esp_flash_read_id fail.

Our legacy code runs smooth and without any problem, and ID match to direct reading flash.

mythbuster5:
@filzek if you initialize as `esp_flash_t * internnal_flash_chip;`, the function should be `esp_flash_read_id(internnal_flash_chip, &id);`. It will return the ture value.

mythbuster5:
@filzek Thanks for your suggestion with flash_ops, we are still maintaining it. What you 've done is meaningful, but if you use our new api, you will get better experience.

filzek:
I have tried to run the code but error is deployed:

```
esp_flash_t * internnal_flash_chip;
uint32_t id, size;
esp_flash_read_id(internnal_flash_chip, &id);
esp_flash_get_size(internnal_flash_chip, &size);
```


 error: 'internnal_flash_chip' is used uninitialized in this function [-Werror=uninitialized]
 esp_flash_read_id(internnal_flash_chip, &id);
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

so, still cant move to the new api to read the true ID of internnal flash, please advise the working code.

filzek:
> @filzek Thanks for your suggestion with flash_ops, we are still maintaining it. What you 've done is meaningful, but if you use our new api, you will get better experience.

esp_flash_read_id  returns a WRONG Flash ID. The function need to READ THE REAL FLASH ID from the IC!

filzek:
For example on a true board:
using  **esp_flash_read_id**
Initialized internnal_flash_chip, size=8192 KB, ID=0xc84017
flash_id: [13123607]

using the API we have develop
**REAL FLASH ID: 3139313237701032**

when remove the flash and external read it, the flash is same as our api read: **3139313237701032**

so, why the esp_flash_read_id doesnt work as intent?

PR #4259, opened 2019-10-29
for Wi-Fi initialization without esp_wifi_init but enabled power management (IDFGH-2092)

Rojer-x86:
for Wi-Fi initialization without esp_wifi_init but enabled power management

### DIFF
--- a/components/esp32/include/esp_wifi.h
+++ b/components/esp32/include/esp_wifi.h
@@ -212,6 +212,16 @@ extern const wpa_crypto_funcs_t g_wifi_default_wpa_crypto_funcs;
 };
 
 /**
+  * @brief  Init power management for WiFi
+  *         Initialize a lock handle for certain power management parameter.
+  *
+  * @return 
+  *    - ESP_OK: succeed
+  *    - ESP_ERR_NO_MEM: if the lock structure can not be allocated
+  */
+esp_err_t esp_wifi_pm_init(void);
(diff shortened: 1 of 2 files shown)


igrr:
Hi @Rojer-x86, can you add a bit more context, when is this new function required?

Rojer-x86:
> Hi @Rojer-x86, can you add a bit more context, when is this new function required?
if i want wifi initialization without default event handlers. function esp_wifi_init set default event handlers, but if initialization over esp_wifi_init_internal with enabled power management then pm lock don't create and any try acquire/releas leads to an error

igrr:
Thanks, understood.

There is a change to the tcpip_adapter coming soon, which will make it possible to bypass the event handlers registration in a nice way. Will update you when this change is merged.

igrr:
Another option is to remove the event handlers, as shown in this example: https://github.com/espressif/esp-idf/blob/758db1e008d4a524b4f386367760b6259e6d6737/examples/ethernet/eth2ap/main/ethernet_example_main.c#L195

Rojer-x86:
this function exists only in version 4.0 in 3.3 lts I didn't find

zhangyanjiaoesp:
@Rojer-x86 Does you problem still exist?

jack0c:
I will close this issue since release/v3.3 got EOL.

PR #4273, opened 2019-10-31
fix modbus error messages (IDFGH-2111)

alegen:
I believe due to a quick copy/paster, the error message displayed in function `mbc_master_send_request` was identical to the one in `mbc_master_get_parameter`. This led to confusion when calling the `mbc_master_send_request` function.

I also updated the error message from the `mbc_master_get_parameter` function to be in line with the other error messages.


Alvin1Zhang:
@alegen Thanks for the contribution.

alisitsyn:
@alegen, 
Thank you for your contribution. Sorry for the long delay with answer. This change (send_request) was implemented in master as part of separate MR and merged. However there are still some messages like "SERIAL master send request failure" ... in each API and it is better to update them as you suggested. This is true for common master and slave interface. Could you update the code in order to merge your PR?

Thanks.

PR #4274, opened 2019-11-01
Fix macOS install error "NotImplementedError: Unsupported archive type" (IDFGH-2112)

motla:
Just removed an extra trailing space.


igrr:
Thanks for the correction! We have merged the same fix internally, it should appear on GitHub soon.

dobairoland:
The fix (bdb08de87555bd070905d416840d46405ead1ee0) is now available in the master branch.

PR #4290, opened 2019-11-05
bootloader: Fix image signature validation when OpenOCD is attached (IDFGH-2130)

tim-nordell-nimbelink:
The code as-is invoked verify_checksum() only if OpenOCD was not attached,
but if SECURE_BOOT_CHECK_SIGNATURE is defined, the invocation of
verify_secure_boot_signature(...) depends on verify_checksum() being run.

Instead, always invoke verify_checksum(), but discard the result if
OpenOCD is attached.  This still allows the software breakpoints per
the original use case of having this check (see commit 439140d26),
but allows the signature validation to still work if there are no
software breakpoints set.  (The signature validation check is always
run despite OpenOCD being attached.)

### DIFF
--- a/components/bootloader_support/src/esp_image_format.c
+++ b/components/bootloader_support/src/esp_image_format.c
@@ -120,6 +120,21 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
         return ESP_ERR_INVALID_ARG;
     }
 
+#if defined(BOOTLOADER_BUILD) && !defined(SECURE_BOOT_CHECK_SIGNATURE)
+    // If we're in the bootloader, the debugger is attached, and we're not
+    // doing secure boot signature checks, disable verification since we may
+    // have a software breakpoint set within the image we're validating which
+    // will throw off validation.
+    if (do_verify && esp_cpu_in_ocd_debug_mode()) {
+        ESP_LOGW(TAG, "skipping image validation due to debugger");
+        do_verify = false;
+    }
+#else
+    if (do_verify && esp_cpu_in_ocd_debug_mode()) {
+        ESP_LOGW(TAG, "debugger is attached; image validation may fail due to software breakpoints");
+    }
+#endif
+
     if (part->size > SIXTEEN_MB) {
         err = ESP_ERR_INVALID_ARG;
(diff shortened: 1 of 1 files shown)


tim-nordell-nimbelink:
Note: I ran into this while having a debugger attached while working with FOTA - it would never pass image validation.  In this case, one is trying to run the image validation code without setting breakpoints in what is being validated at all (but might in the actually running image).

components/bootloader_support/src/esp_image_format.c
```diff
@@ -190,9 +190,9 @@ static esp_err_t image_load(esp_image_load_mode_t mode, const esp_partition_pos_
...
         data->image_len = end_addr - data->start_addr;
         ESP_LOGV(TAG, "image start 0x%08x end of last section 0x%08x", data->start_addr, end_addr);
-        if (NULL != checksum && !esp_cpu_in_ocd_debug_mode()) {
+        if (NULL != checksum) {
             err = verify_checksum(sha_handle, checksum_word, data);
-            if (err != ESP_OK) {
+            if (err != ESP_OK && !esp_cpu_in_ocd_debug_mode()) {
                 goto err;
             }
         }
```
tim-nordell-nimbelink: Just realizing it might be better to change it into:

```
#ifdef BOOTLOADER_BUILD
        if (NULL != checksum && !esp_cpu_in_ocd_debug_mode()) {
#else
        if (NULL != checksum) {
#endif
            err = verify_checksum(sha_handle, checksum_word, data);
            if (err != ESP_OK) {
                goto err;
            }
        }
```

instead, and change the other occurrence below into the same format:

```
#ifdef BOOTLOADER_BUILD
            if (sha_handle != NULL && !esp_cpu_in_ocd_debug_mode()) {
#else
            if (sha_handle != NULL) {
#endif
                err = verify_simple_hash(sha_handle, data);
            }
```

The case below I wasn't hitting because I had SECURE_BOOT_CHECK_SIGNATURE set.

 \- Tim

Alvin1Zhang:
@tim-nordell-nimbelink Thanks for reporting.

mahavirj:
@tim-nordell-nimbelink Sorry about not providing any update here. This was fixed in https://github.com/espressif/esp-idf/commit/07bf1e23db718f4c5cfd800b5dc1e4ecaf88ecb5. Closing the PR.

PR #4306, opened 2019-11-08
Adding WebSocket server support for esp_http_server (IDFGH-2151)

huming2207:
Hi all,

This Pull Request adds the WebSocket support for esp_http_server library. It mainly does:

- Handling WebSocket handshake
  - Parsing HTTP upgrade request
  - Reply the upgrade request
- Receive WebSocket packets
  - Parse header, decode to a struct
  - Unmask payload (if required)
- Send WebSocket packets

I've tested my implementation with `websocat` (ref: https://github.com/vi/websocat) and it works just fine for now. 

Here's a quick and dirty example for testing:

```c
static esp_err_t ws_ping_handler(httpd_req_t *req)
{
    uint8_t buf[128] = { 0 };
    httpd_ws_pkt_t ws_pkt;
    memset(&ws_pkt, 0, sizeof(httpd_ws_pkt_t));
    ws_pkt.payload = buf;

    esp_err_t ret = httpd_ws_recv_pkt(req, &ws_pkt, 128, 60000);
    ESP_LOGI(TAG, "Got packet with message: %s", ws_pkt.payload);

    ret = ret ? : httpd_ws_send_pkt(req, &ws_pkt);
    return ret;
}

static const httpd_uri_t ws = {
    .uri        = "/ws",
    .method     = HTTP_GET,
    .handler    = ws_ping_handler,
    .user_ctx   = NULL,
    .is_websocket = true
};

static httpd_handle_t start_webserver(void)
{
    httpd_handle_t server = NULL;
    httpd_config_t config = HTTPD_DEFAULT_CONFIG();

    // Start the httpd server
    ESP_LOGI(TAG, "Starting server on port: '%d'", config.server_port);
    if (httpd_start(&server, &config) == ESP_OK) {
        // Set URI handlers
        ESP_LOGI(TAG, "Registering URI handlers");
        // ...some other HTTP endpoint handlers also add at here
        httpd_register_uri_handler(server, &ws);
        return server;
    }

    ESP_LOGI(TAG, "Error starting server!");
    return NULL;
}
```

Then run `websocat ws://{ESP32_IP_ADDR}/ws`, type in something and press Enter.

Meanwhile, I will submit some other examples soon on another Pull Request later.

Please review this Pull Request and provide some advice if possible. Thanks! 

 Regards,
Jackson Hu


huming2207:
Hi guys, any updates?

Alvin1Zhang:
@huming2207 Thanks for the contribution, we will look into. Thanks.

huming2207:
@Alvin1Zhang Thanks. If you have any suggestions for improvements, please tell me soon.

huming2207:
By the way, here are some of my suggestions for `esp-http-server`. I may consider submitting some other PRs if you guys agree. @Alvin1Zhang @anurag-kar @chinmaychhajed @shahpiyushv 
  
1. Do we really need control socket just for passing messages internally? If not, then I think we should use something FreeRTOS event queue instead. Or at least leave it as an option in `menuconfig`.
2. Some variable namings are very confusing. For example, `r` for `httpd_req_t` and `ra` for `httpd_req_aux`. We should consider refactoring this as well.

lucastcox:
Definitely agreed on the control socket. It seems completely necessary.

huming2207:
@lucastcox you mean necessary or unnecessary?

huming2207:
In regards to the control socket,  I will try removing it but leave an option in `menuconfig` in case someone needs it. I will do it once this PR is merged.

By the way, @Alvin1Zhang any further updates?

lucastcox:
> @lucastcox you mean necessary or unnecessary?

Haha, yes. I meant unnecessary.

components/esp_http_server/src/esp_httpd_priv.h
```diff
@@ -471,6 +474,31 @@ int httpd_default_recv(httpd_handle_t hd, int sockfd, char *buf, size_t buf_len,
...
+ * @{
+ */
+
+
+/**
+ * @brief   This function is for responding the client when in a WebSocket handshake phrase
+ *
+ * @param[in] hd    Server instance data
+ * @return
+ *  - ESP_OK                        : if sucessful
```
shahpiyushv: Remove the "ifs" to be consistent with rest of the documentation?

components/esp_http_server/src/esp_httpd_priv.h
```diff
@@ -471,6 +474,31 @@ int httpd_default_recv(httpd_handle_t hd, int sockfd, char *buf, size_t buf_len,
...
+/** @name WebSocket
+ * Functions for WebSocket header parsing
+ * @{
+ */
+
+
+/**
+ * @brief   This function is for responding the client when in a WebSocket handshake phrase
+ *
+ * @param[in] hd    Server instance data
```
shahpiyushv: Probably incorrect param name and description

components/esp_http_server/src/httpd_main.c
```diff
@@ -39,7 +39,7 @@ static esp_err_t httpd_accept_conn(struct httpd_data *hd, int listen_fd)
              * therefore httpd_accept_conn() will be called again, but this time
              * with space available for one session
              */
-       }
+        }
```
shahpiyushv: Some stray change?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,302 @@
...
+#include "esp_httpd_priv.h"
+
+#define TAG "httpd_ws"
+
+#define HTTPD_WS_FIN_BIT        0x80U
+#define HTTPD_WS_OPCODE_BITS    0x0fU
+#define HTTPD_WS_MASK_BIT       0x80U
+#define HTTPD_WS_LENGTH_BITS    0x7fU
+
+static const char ws_magic_uuid[] = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
```
shahpiyushv: Is this some randomly generated UUID?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,302 @@
...
+
+#include <esp_http_server.h>
+#include "esp_httpd_priv.h"
+
+#define TAG "httpd_ws"
+
+#define HTTPD_WS_FIN_BIT        0x80U
+#define HTTPD_WS_OPCODE_BITS    0x0fU
+#define HTTPD_WS_MASK_BIT       0x80U
+#define HTTPD_WS_LENGTH_BITS    0x7fU
```
shahpiyushv: If these are from some standard, can you provide a reference?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,302 @@
...
+    if (!req || !req->aux) {
+        ESP_LOGW(TAG, LOG_FMT("Argument is invalid"));
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    /* Detect handshake - reject if handshake was ALREADY performed */
+    struct httpd_req_aux *req_aux = req->aux;
+    if (req_aux->sd->has_ws_handshake_done) {
+        ESP_LOGW(TAG, LOG_FMT("State is invalid - Handshake has been performed"));
+        return ESP_ERR_INVALID_ARG;
```
shahpiyushv: Return ESP_ERR_INVALID_STATE instead?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,302 @@
...
+
+    /* Detect WS version existence */
+    char version_val[3] = { '\0' };
+    if (httpd_req_get_hdr_value_str(req, "Sec-WebSocket-Version", version_val, sizeof(version_val)) != ESP_OK) {
+        ESP_LOGW(TAG, LOG_FMT("\"Sec-WebSocket-Version\" is not found"));
+        return ESP_ERR_NOT_FOUND;
+    }
+
+    /* Detect if WS version is "13" or not */
+    if (strcasecmp(version_val, "13") != 0) {
```
shahpiyushv: Some reference for this versioning?

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv This is a WS magic UUID, not random. Ref: https://stackoverflow.com/questions/13456017/what-does-258eafa5-e914-47da-95ca-c5ab0dc85b11-means-in-websocket-protocol

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv This is according to RFC6455 standard. I will add some comments here later.

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv This is also from RFC6455. The request should be Version 13 or otherwise, it should be rejected.

components/esp_http_server/src/httpd_parse.c
```diff
@@ -374,13 +375,29 @@ static esp_err_t cb_headers_complete(http_parser *parser)
...
-        ESP_LOGW(TAG, LOG_FMT("upgrade from HTTP not supported"));
-        /* There is no specific HTTP error code to notify the client that
-         * upgrade is not supported, thus sending 400 Bad Request */
-        parser_data->error = HTTPD_400_BAD_REQUEST;
-        parser_data->status = PARSING_FAILED;
-        return ESP_FAIL;
+        ESP_LOGD(TAG, LOG_FMT("Got a upgrade request"));
+
+        /* Now start to have a look if it's WebSocket - prepare header buffers */
+        size_t upgrade_val_len = httpd_req_get_hdr_value_len(r, "Upgrade");
```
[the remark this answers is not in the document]
huming2207: I think you read on a wrong commit. I have removed it. Ref: https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L382

components/esp_http_server/src/httpd_main.c
huming2207: @shahpiyushv This is formatted by CLion and the `astyle` script provided by ESP-IDF.

components/esp_http_server/include/esp_http_server.h
```diff
@@ -1452,6 +1458,65 @@ esp_err_t httpd_queue_work(httpd_handle_t handle, httpd_work_fn_t work, void *ar
...
+/** @name WebSocket
+ * Functions and structs for WebSocket server
+ * @{
+ */
+
+/**
+ * @brief Enum for WebSocket packet types (Opcode in the header)
+ */
+typedef enum {
+    HTTPD_WS_TYPE_CONT   = 0x0,
```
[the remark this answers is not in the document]
huming2207: @shahpiyushv Thanks, I will rename this.

components/esp_http_server/include/esp_http_server.h
```diff
@@ -1452,6 +1458,65 @@ esp_err_t httpd_queue_work(httpd_handle_t handle, httpd_work_fn_t work, void *ar
...
+
+/**
+ * @brief Receive and parse a WebSocket packet
+ * @param[in]   req         Current request
+ * @param[out]  pkt         WebSocket packet
+ * @param[in]   max_len     Maximum length for receive
+ * @param[in]   timeout_sec Timeout in second(s), leave it to 0 for forever
+ * @return
+ *  - ESP_OK                    : if successful
+ *  - ESP_FAIL                  : if socket errors occurs
```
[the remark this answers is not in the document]
huming2207: @shahpiyushv I've updated the docs here. Please check again if it's appropriate or not.

components/esp_http_server/src/esp_httpd_priv.h
huming2207: @shahpiyushv I've updated the docs here. Please check again if it's appropriate or not.

huming2207:
Hello @shahpiyushv ,

Great thanks for your comments.

I have fixed some grammar issues and typos that you've pointed out. However, one or two of your comments are not based on the latest commit. Please make sure you are reading my latest commit.

Please also have a read about these documents. It is very helpful for understanding WebSocket protocol:

- RFC6455 document: https://tools.ietf.org/html/rfc6455
- Mozilla's tutorial: https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers
- Wikipedia: https://en.wikipedia.org/wiki/WebSocket#Web_Server_implementation
- Node.js `ws` library: https://github.com/websockets/ws

Meanwhile...

> Control socket is useful for asynchronously triggering operations like stopping the webserver, closing any sessions or triggering some response in an already opened session. Since the webserver is always blocked on a select call, a socket descriptor is more convenient to wake up the server, than using a separate message queue. We will see if this can be put under a menuconfig option, or by adding some other run time logic (like control port set to 0). However, even a simple call like httpd_stop() would not work in that case.

In regards to the control socket issue, I think it's appropriate only for some beefy microcontroller. For example, Kendryte K210 or even some ARM Cortex-A platforms. On ESP32, it consumes too much RAM and flash. Please consider leaving an option in `menuconfig` instead.

> We can change these to req and req_aux probably? Will have to check that these names aren't already used in the same functions and also if it is indeed causing any confusion.

Please do so, thanks. I will also submit a draft PR for this later after this PR is merged.

Regards,
Jackson

harveyconnor:
great work guys!

huming2207:
Hello @Alvin1Zhang @shahpiyushv ,

I think we should also consider having a look at this library: https://github.com/cesanta/mongoose

It's similar to `esp_http_server` but it has more features. 

Jackson

components/esp_http_server/src/httpd_uri.c
```diff
@@ -292,21 +293,33 @@ esp_err_t httpd_uri(struct httpd_data *hd)
...
+        if (ret != ESP_OK) {
+            return ret;
+        }
+
+        aux->sd->has_ws_handshake_done = true;
+        aux->sd->ws_handler = uri->handler;
+    }
+
     /* Invoke handler */
     if (uri->handler(req) != ESP_OK) {
```
shahpiyushv: Is this supposed to be called even during the websocket handshake request? If yes, will it also get some data, or just the headers? If the handler is expected to parse headers, should it also get the facility of adding own headers ?

components/esp_http_server/include/esp_http_server.h
```diff
@@ -1452,6 +1458,66 @@ esp_err_t httpd_queue_work(httpd_handle_t handle, httpd_work_fn_t work, void *ar
...
+ * @note Please refer to RFC6455 Section 5.4 for more details
+ */
+typedef enum {
+    HTTPD_WS_TYPE_CONTINUE   = 0x0,
+    HTTPD_WS_TYPE_TEXT       = 0x1,
+    HTTPD_WS_TYPE_BINARY     = 0x2,
+    HTTPD_WS_TYPE_CLOSE      = 0x8,
+    HTTPD_WS_TYPE_PING       = 0x9,
+    HTTPD_WS_TYPE_PONG       = 0xA
+} httpd_ws_type_t;
```
shahpiyushv: Would it help to have some inbuilt support or helper functions for handling Close, Ping and Pong Control packets?

components/esp_http_server/src/esp_httpd_priv.h
```diff
@@ -63,14 +63,16 @@ struct sock_db {
...
-    httpd_free_ctx_fn_t free_ctx;      /*!< Function for freeing the context */
+    httpd_free_ctx_fn_t free_ctx;           /*!< Function for freeing the context */
     httpd_free_ctx_fn_t free_transport_ctx; /*!< Function for freeing the 'transport' context */
     httpd_send_func_t send_fn;              /*!< Send function for this socket */
     httpd_recv_func_t recv_fn;              /*!< Receive function for this socket */
     httpd_pending_func_t pending_fn;        /*!< Pending function for this socket */
     uint64_t lru_counter;                   /*!< LRU Counter indicating when the socket was last used */
     char pending_data[PARSER_BLOCK_SIZE];   /*!< Buffer for pending data to be received */
     size_t pending_len;                     /*!< Length of pending data to be received */
+    bool has_ws_handshake_done;             /*!< True if it has done WebSocket handshake (if this socket is a valid WS) */
```
shahpiyushv: Probably personal preference, but removing the word "has" i.e. changing to ws_handshake_done would be easier to read.

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,311 @@
...
+        payload[idx] = (payload[idx] ^ mask_key[idx % 4]);
+    }
+
+    return ESP_OK;
+}
+
+esp_err_t httpd_ws_recv_pkt(httpd_req_t *req, httpd_ws_pkt_t *pkt, size_t max_len, uint32_t timeout_sec)
+{
+    esp_err_t ret = httpd_ws_check_req(req);
+    struct httpd_req_aux *aux = req->aux;
```
shahpiyushv: Return value of httpd_ws_check_req() should be checked before reading aux, as req may be NULL.

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,311 @@
...
+        /* Case 2: If length byte is 126, then this packet's length bit is 16 bits */
+        uint8_t length_bytes[2] = { 0 };
+        if (httpd_recv_with_opt(req, (char *)length_bytes, sizeof(length_bytes), false) <= 0) {
+            ESP_LOGW(TAG, LOG_FMT("Failed to receive 2 bytes length"));
+            return ESP_FAIL;
+        }
+
+        pkt->len = ((uint32_t)(length_bytes[0] << 8U) | (length_bytes[1]));
+    } else if (init_len == 127) {
+        /* Case 3: If length is byte 126, then this packet's length bit is 64 bits */
```
shahpiyushv: Typo in comment. 126 -> 127

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,311 @@
...
+
+    /* We only accept SIZE_MAX for now (in ESP-IDF, it's 2^32 - 1) */
+    if (pkt->len > SIZE_MAX) {
+        ESP_LOGW(TAG, LOG_FMT("WS Message too long"));
+        return ESP_ERR_INVALID_SIZE;
+    }
+
+    /* If this packet is masked, dump the mask as well */
+    uint8_t mask_key[4] = { 0 };
+    if (masked) {
```
shahpiyushv: As per the RFC 6455 framing requirements (https://tools.ietf.org/html/rfc6455#section-5.2) all data from client to server should be masked. So, should an error be returned if masked bit is 0?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,311 @@
...
+    struct httpd_req_aux *req_aux = req->aux;
+    if (!req_aux->sd->has_ws_handshake_done) {
+        ESP_LOGW(TAG, LOG_FMT("State is invalid - No handshake performed"));
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    return ESP_OK;
+}
+
+static esp_err_t httpd_ws_mask_payload(uint8_t *payload, size_t len, const uint8_t *mask_key)
```
shahpiyushv: Should this be called httpd_ws_unmask_payload() instead?

components/esp_http_server/src/httpd_ws.c
```diff
@@ -0,0 +1,311 @@
...
+    uint8_t mask_key[4] = { 0 };
+    if (masked) {
+        if (httpd_recv_with_opt(req, (char *)mask_key, sizeof(mask_key), false) <= 0) {
+            ESP_LOGW(TAG, LOG_FMT("Failed to receive mask key"));
+            return ESP_FAIL;
+        }
+    }
+
+    /* Receive buffer */
+    size_t recv_len = size_min(pkt->len, max_len);
```
shahpiyushv: What happens to the rest of the bytes if pkt->len is indeed more than max_len. How are rest of the bytes read?

components/esp_http_server/include/esp_http_server.h
```diff
@@ -1452,6 +1458,66 @@ esp_err_t httpd_queue_work(httpd_handle_t handle, httpd_work_fn_t work, void *ar
...
+
+/**
+ * @brief WebSocket packet format
+ */
+typedef struct httpd_ws_pkt {
+    bool final;                 /*!< Final packet */
+    httpd_ws_type_t type;       /*!< WebSocket packet type */
+    uint8_t *payload;           /*!< Pre-allocated data buffer */
+    size_t len;                 /*!< Length of the WebSocket data */
+} httpd_ws_pkt_t;
```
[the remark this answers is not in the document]
huming2207: @shahpiyushv ok, I will rename it.

components/esp_http_server/include/esp_http_server.h
huming2207: @shahpiyushv I think we can, but leave it as optional in `menuconfig`.

components/esp_http_server/src/esp_httpd_priv.h
huming2207: @shahpiyushv Yep, will do.

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv Oh sorry, got a logic issue here

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv hmmm...maybe yes.

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv The mask and unmask is the same process. In my opinion, whichever is fine. I will change it later anyway.

components/esp_http_server/src/httpd_ws.c
huming2207: @shahpiyushv Hmm...yes, here's another logic error. Sorry...

components/esp_http_server/include/esp_http_server.h
```diff
@@ -1452,6 +1458,66 @@ esp_err_t httpd_queue_work(httpd_handle_t handle, httpd_work_fn_t work, void *ar
...
+ * @brief Construct and send a WebSocket packet
+ * @param[in]   req     Current request
+ * @param[in]   pkt     WebSocket packet
+ * @return
+ *  - ESP_OK                    : On successful
+ *  - ESP_FAIL                  : When socket errors occurs
+ *  - ESP_ERR_INVALID_STATE     : Handshake was already done beforehand
+ *  - ESP_ERR_INVALID_ARG       : Argument is invalid (null or non-WebSocket)
+ */
+esp_err_t httpd_ws_send_pkt(httpd_req_t *req, httpd_ws_pkt_t *pkt);
```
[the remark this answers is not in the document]
huming2207: @shahpiyushv I think I can, but it might be a bit too complex. I think this PR is just for basic WebSocket server support. It should be considered in the later PRs.

However, if you really think I should include this async support anyway, I will do it in this PR.

huming2207:
Hello @shahpiyushv ,

Great thanks for your reply. 

> I would also like to know what are the anticipated use cases for having websocket server running on an IoT device on a local network.

I know the use case is very limited. However, WebSocket is somewhat more "compliant" with some corporate networks with restricted firewall rules (e.g. port 80/443/995/993/22 only). It will be less likely to be blocked compared to some other protocols. Prior to 2016, my University's internal network only has TCP port 80, 22 and 443 access for students, until later the IT Service upgraded the whole network infrastructure and somehow removed this constraint at the same time. It's awful but this scenario does exist.

Furthermore, it can be used for MQTT over WS/WSS. Someone later can implement a MQTT server on ESP platforms.

Besides, it provides an "intermediate" protocol compared to raw TCP sockets and HTTP. Raw TCP socket is more lightweight but harder for users to implement, while HTTP is too heavy for ESP32.

Meanwhile, for this PR, my opinion is to provide a basic support for WebSocket server. Some of the advanced features and improvements like asynchronous support should be considered later. However, if you really think I should include it right now in this PR, then I will take a look and try working on it. I will take an exam next Tuesday. I may not be available by then. Sorry about the delay.

Regards,
Jackson

components/esp_http_server/src/httpd_uri.c
huming2207: @shahpiyushv All the WS requests that have been done handshaking before, will be handled separately before going in to the HTTP parser (or otherwise the parser will fail as it can't speak WebSockets anyway). Please refer to this function: https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L739

components/esp_http_server/include/esp_http_server.h
huming2207: @shahpiyushv Same as the async send support as you've mentioned. I think this automatic Ping/Pong support should be implemented but not included in this PR. 

However, if you think we do need this shipped with this PR, I will do it later on next Tuesday after I finish my exam.

shahpiyushv:
> Hello @shahpiyushv ,
> 
> Great thanks for your reply.
> 
> > I would also like to know what are the anticipated use cases for having WebSocket server running on an IoT device on a local network.
> 
> I know the use case is very limited. However, WebSocket is somewhat more "compliant" with some corporate networks with restricted firewall rules (e.g. port 80/443/995/993/22 only). It will be less likely to be blocked compared to some other protocols. Prior to 2016, my University's internal network only has TCP port 80, 22 and 443 access for students, until later the IT Service upgraded the whole network infrastructure and somehow removed this constraint at the same time. It's awful but this scenario does exist.

Port 80 and 443 are the default ports for HTTP and HTTPS anyways, even without web-sockets. So, I am not sure I understand this point here.

> 
> Furthermore, it can be used for MQTT over WS/WSS. Someone later can implement an MQTT server on ESP platforms.

MQTT Server (a broker actually) could be quite heavy for ESP32 (even without websocket support), especially when combined with this webserver. I am not sure if we would want to add it in this component itself or keep it separate.

> 
> Besides, it provides an "intermediate" protocol compared to raw TCP sockets and HTTP. Raw TCP socket is more lightweight but harder for users to implement, while HTTP is too heavy for ESP32.

We aren't bypassing the HTTP Server in this case anyways, so even for websocket, entire HTTP server is getting pulled. So, this is not making the server lightweight as such. Can you elaborate on this?

> 
> Meanwhile, for this PR, my opinion is to provide basic support for WebSocket server. Some of the advanced features and improvements like asynchronous support should be considered later. However, if you really think I should include it right now in this PR, then I will take a look and try working on it. I will take an exam next Tuesday. I may not be available until then. Sorry about the delay.

We had some discussions over this. Without rest of the features, this cannot really be called "feature completion". So, please consider adding rest of the features as and when you get time. Please also remember to squash all commits into a single one for the purpose of this PR and also put the feature under a config option so that people can selectively enable it.

Regards,
Piyush

> 
> Regards,
> Jackson

shahpiyushv:
@huming2207 some comment in the context of control socket.

> > Control socket is useful for asynchronously triggering operations like stopping the webserver, closing any sessions or triggering some response in an already opened session. Since the webserver is always blocked on a select call, a socket descriptor is more convenient to wake up the server, than using a separate message queue. We will see if this can be put under a menuconfig option, or by adding some other run time logic (like control port set to 0). However, even a simple call like httpd_stop() would not work in that case.
> 
> In regards to the control socket issue, I think it's appropriate only for some beefy microcontroller. For example, Kendryte K210 or even some ARM Cortex-A platforms. On ESP32, it consumes too much RAM and flash. Please consider leaving an option in `menuconfig` instead.
> 

I did some quick memory analysis on this by using the esp-idf/examples/protocols/http_server/simple app. Removing control socket support (and so, also the httpd_stop() call from the example) reduced the flash usage of libesp_http_server.a by about 1K (checked using idf_size.py)

I also dumped the RAM usage info in the handler for /hello. The run time RAM usage difference was about 532 bytes, whereas difference in size of "Min. Ever Free Size" was about 2K. I do not find this this a convincing enough reason to make the support configurable, at the cost of losing features like clean httpd_stop, LRU closures, asynchronous communication, etc.

I made the code changes in a hurry and so, my numbers may not be correct. Have you done such memory analysis at your end?

Regards,
Piyush

huming2207:
Hello @shahpiyushv 

Thanks for your reply.

> Port 80 and 443 are the default ports for HTTP and HTTPS anyways, even without web-sockets. So, I am not sure I understand this point here.

No, some of the proxies/firewalls may require the user to install MITM certifications so that they can filter out the HTTPS content as well. Some plain text HTTP content is the same, as the firewalls/proxies may not only detect the ports but also filter the content. For WebSocket, since it has a HTTP handshake and it's a famous modern web protocol, it will be less likely to be in trouble with those firewalls.

> We aren't bypassing the HTTP Server in this case anyways, so even for websocket, entire HTTP server is getting pulled. So, this is not making the server lightweight as such. Can you elaborate on this?

Yes, but the HTTP handshake is done once only per connection. The rest of the WebSocket frames only have 2 bytes overhead (for frames smaller than 126 bytes).

> We had some discussions over this. Without rest of the features, this cannot really be called "feature completion". So, please consider adding rest of the features as and when you get time. Please also remember to squash all commits into a single one for the purpose of this PR and also put the feature under a config option so that people can selectively enable it.

Okay, will do.

> I also dumped the RAM usage info in the handler for /hello. The run time RAM usage difference was about 532 bytes, whereas difference in size of "Min. Ever Free Size" was about 2K. I do not find this this a convincing enough reason to make the support configurable, at the cost of losing features like clean httpd_stop, LRU closures, asynchronous communication, etc.

In that case, then we probably should leave control socket alone, as it is very useful behind the scene anyway. Sorry for that I misunderstood its purposes.

Regards,
Jackson

huming2207:
Hello @shahpiyushv 

I have a concern about your ideas of using `httpd_queue_work()` to achieve async sending. Maybe I was wrong, but how can I guarantee that the frame payload pointer is still valid?

In other words, since the sending function will be async, then the Tx is done outside of a handler. If the user allocated an `uint8_t` array locally in that handler function, then when the Tx is performed, the handler has done and out of scope.

Have I misunderstood something?

Regards,
Jackson

huming2207:
Hi @shahpiyushv ,

I've done a draft version of the control frame auto-reply. The timeout will be set separately. I will do that later.

In regards to the async sending via `httpd_queue_work()`, I will not include it in this pull request anyway. Unless if you allocate the buffer outside (e.g. in the heap with `malloc`, or globally) of the handler's scope, I think there's no way to do it.  For your concerns about the performance, it won't really affect much, as WebSocket is often used for handling short-ish data at once.

Regards,
Jackson

huming2207:
Hello @shahpiyushv ,

Now automatic control frame treatment is done. You can try this on your computer:

```
websocat ws://{YOUR_ESP_BOARD_IP}/ws --ping-interval=5
``` 

It will send a PING frame for every 5 seconds. You will see the ESP will reply a PONG later.

For the async send functions, I will not do it anyway, as it is implicit and out of scope as I mentioned above.

Regards,
Jackson

huming2207:
Hello @shahpiyushv ,

I've squashed all the previous commits. Please review it again. If it's okay, please merge it. Thanks.

Regards,
Jackson

huming2207:
Hi @shahpiyushv ,

I've added a WebSocket option in `menuconfig`. It should be all good now.

Regards,
Jackson

huming2207:
@shahpiyushv @Alvin1Zhang Any updates?

shahpiyushv:
@huming2207 , this may get delayed a bit as I am caught up in some other work.

huming2207:
@shahpiyushv Understood. I may also consider preparing another new PR for updating Asio library.

githubforsank:
Hi @huming2207  
I had a dire need of WebSocket server for my esp32 project. 
I pulled in your commit to test the implementation but I see the handshake is failing.
I did not use websocat but a chrome/browser ws client to test. This is the output  I am getting. 
Thanks for providing the implementation. Can you shed some insights over this?

image

huming2207:
Hello @githubforsank ,

Could you please show me your code? thanks.

githubforsank:
> Hello @githubforsank ,
> 
> Could you please show me your code? thanks.

@huming2207  I have used the dirty test example you have posted in one of the conversation. 

> Hi all,
> 
> This Pull Request adds the WebSocket support for esp_http_server library. It mainly does:
> 
> * Handling WebSocket handshake
>   
>   * Parsing HTTP upgrade request
>   * Reply the upgrade request
> * Receive WebSocket packets
>   
>   * Parse header, decode to a struct
>   * Unmask payload (if required)
> * Send WebSocket packets
> 
> I've tested my implementation with `websocat` (ref: https://github.com/vi/websocat) and it works just fine for now.
> 
> Here's a quick and dirty example for testing:
> 
> ```c
> static esp_err_t ws_ping_handler(httpd_req_t *req)
> {
>     uint8_t buf[128] = { 0 };
>     httpd_ws_pkt_t ws_pkt;
>     memset(&ws_pkt, 0, sizeof(httpd_ws_pkt_t));
>     ws_pkt.payload = buf;
> 
>     esp_err_t ret = httpd_ws_recv_pkt(req, &ws_pkt, 128, 60000);
>     ESP_LOGI(TAG, "Got packet with message: %s", ws_pkt.payload);
> 
>     ret = ret ? : httpd_ws_send_pkt(req, &ws_pkt);
>     return ret;
> }
> 
> static const httpd_uri_t ws = {
>     .uri        = "/ws",
>     .method     = HTTP_GET,
>     .handler    = ws_ping_handler,
>     .user_ctx   = NULL,
>     .is_websocket = true
> };
> 
> static httpd_handle_t start_ws_webserver(void)
> {
>     httpd_handle_t server = NULL;
>     httpd_config_t config = HTTPD_DEFAULT_CONFIG();
> 
>     // Start the httpd server
>     ESP_LOGI(TAG, "Starting server on port: '%d'", config.server_port);
>     if (httpd_start(&server, &config) == ESP_OK) {
>         // Set URI handlers
>         ESP_LOGI(TAG, "Registering URI handlers");
>         // ...some other HTTP endpoint handlers also add at here
>         httpd_register_uri_handler(server, &ws);
>         return server;
>     }
> 
>     ESP_LOGI(TAG, "Error starting server!");
>     return NULL;
> }
```
void app_main(void)
{
    start_ws_webserver();
}
```

huming2207:
Hi @githubforsank ,

That's weird. I will have a look later.

huming2207:
Hello @githubforsank ,

It works for me without any issues.

image

image

Please try this demo: https://github.com/huming2207/esp-httpd-ws-demo.git

Regards,
Jackson

huming2207:
@githubforsank I forgot to mention this yesterday. Please try enabling the logging level to Debug.

huming2207:
@githubforsank Any updates?

Btw @shahpiyushv , any updates for merging?

githubforsank:
> @githubforsank Any updates?
> 
> Btw @shahpiyushv , any updates for merging?

I got the connection working.
It seems the uri handler was conflicting. But there is a new observation
The esp32 seems to dump and restart whenever we close the WebSocket connection.

image
Closing connection.
image

Core dump : 
image

huming2207:
@githubforsank Thanks for your update. I guess it might be affected by the header buffer length. Try setting the header buffer length to a longer one. Also, please set the logging level to `Debug` so that I can gather more info.

githubforsank:
> 
> 
> @githubforsank Thanks for your update. I guess it might be affected by the header buffer length. Try setting the header buffer length to a longer one. Also, please set the logging level to `Debug` so that I can gather more info.

Hi @huming2207  
Was stuck in some office work. 
I tried verifying the code according to the log level but can't understand why the new handshake request is delegated to the user uri handler most of the times and not handled by the handshake_response_handler for which handshake is failing most times.
Description and logs : 
Websocket handshake fails most of the time.( It works sometime at erratic intervals and reboots. )
image

In the rare case, the connection is successful then packets are received but closing the connection causes exception. 
Description and logs :
Websocket close connection is always failing. 
image

huming2207:
Hi @githubforsank 

Thanks for your reply. Happy new year.

Are you using my latest commit?

Also please try increasing the header buffer. I guess the HTTP header buffer (512 bytes I remembered) is too small for you, so it works occasionally but not always.

I'm not at home at the moment so I can't review my code for now. Please try increasing the buffer and give it another go. 


Jackson Hu

githubforsank:
> 
> 
> Hi @githubforsank
> 
> Thanks for your reply. Happy new year.
> 
> Are you using my latest commit?
> 
> Also please try increasing the header buffer. I guess the HTTP header buffer (512 bytes I remembered) is too small for you, so it works occasionally but not always.
> 
> I'm not at home at the moment so I can't review my code for now. Please try increasing the buffer and give it another go.
> 
> Jackson Hu

Thank you and Happy new year !
I am using your latest commit. 
I increased the buffer to 1024. Still the same.
FYI the header sent is 480 bytes which is also the same in the esp. 
`W (37544) httpd_ws: httpd_ws_check_req: State is invalid - No handshake performed`
The control doesn't reaches to handle_handshake function.

While disconnecting : 
```
esp_err_t httpd_req_delete(struct httpd_data *hd)
{
    httpd_req_t *r = &hd->hd_req;
    struct httpd_req_aux *ra = r->aux;

    /* Finish off reading any pending/leftover data */
    while (ra->remaining_len) { / * The dump comes in this line (Guru Meditation) */ 
        /* Any length small enough not to overload the stack, but large
         * enough to finish off the buffers fast */
        char dummy[CONFIG_HTTPD_PURGE_BUF_LEN];
        int recv_len = MIN(sizeof(dummy), ra->remaining_len);
        recv_len = httpd_req_recv(r, dummy, recv_len);
        if (recv_len < 0) {
            httpd_req_cleanup(r);
            return ESP_FAIL;
        }
```

Source : httpd_parse.c: httpd_req_delete (791)

githubforsank:
Hi @huming2207 

Final observations in trying to make web socket server work : 
httpd_uri.c : httpd_uri : 294
```
#ifdef CONFIG_HTTPD_WS_SUPPORT
    if (uri->is_websocket && aux->ws_handshake_detect && uri->method == HTTP_GET) {
        ESP_LOGD(TAG, LOG_FMT("Responding WS handshake to sock %d"), aux->sd->fd);
        esp_err_t ret = httpd_ws_respond_server_handshake(&hd->hd_req);
        if (ret != ESP_OK) {
            return ret;
        }

        aux->sd->ws_handshake_done = true;
        aux->sd->ws_handler = uri->handler;

        /* Return immediately after handshake, no need to call handler here */
        return ESP_OK;
    }
#endif
```
This block is not executed certain times. For the same the upgrade HTTP_GET websocket request is not handled by :
`httpd_ws_respond_server_handshake(&hd->hd_req);`
For which the result occurs as Invalid state : handshake not done. 
Few times the request goes through the handshake after which the connection is established.

On disconnecting websocket from client the core dumps at 
while (ra->remaining_len)
httpd_parse.c : httpd_req_delete : 791
```
/* Function that resets the http request data
 */
esp_err_t httpd_req_delete(struct httpd_data *hd)
{
    httpd_req_t *r = &hd->hd_req;
    struct httpd_req_aux *ra = r->aux;

    /* Finish off reading any pending/leftover data */
    **while (ra->remaining_len)** {
        /* Any length small enough not to overload the stack, but large
         * enough to finish off the buffers fast */
        char dummy[CONFIG_HTTPD_PURGE_BUF_LEN];
        int recv_len = MIN(sizeof(dummy), ra->remaining_len);
        recv_len = httpd_req_recv(r, dummy, recv_len);
        if (recv_len < 0) {
            httpd_req_cleanup(r);
            return ESP_FAIL;
        }

        ESP_LOGD(TAG, LOG_FMT("purging data size : %d bytes"), recv_len);

#ifdef CONFIG_HTTPD_LOG_PURGE_DATA
        /* Enabling this will log discarded binary HTTP content data at
         * Debug level. For large content data this may not be desirable
         * as it will clutter the log */
        ESP_LOGD(TAG, "================= PURGED DATA =================");
        ESP_LOG_BUFFER_HEX_LEVEL(TAG, dummy, recv_len, ESP_LOG_DEBUG);
        ESP_LOGD(TAG, "===============================================");
#endif
    }
```

Refer above screenshots for the details of the log .

Thank you.

huming2207:
Hi @githubforsank 

I still can't reproduce your issue. Here's my quick analysis:

On line 313: https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_uri.c#L313

- `uri->is_websocket` is set by user (i.e. by yourself) from `httpd_req_t` 
- `aux->ws_handshake_detect` is set on here: https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L400
    - According to your screenshots, this flag has been set to true, as it has "Got an upgrade request" log in it without any more complaints.
- ...and finally, the `uri->method == HTTP_GET` is also set by user.


Btw, have you tried setting your `httpd_req_t` to be static? 

Jackson

JonathanLourens:
> Hi @huming2207
> 
> Final observations in trying to make web socket server work :
> httpd_uri.c : httpd_uri : 294
> 
> ```
> #ifdef CONFIG_HTTPD_WS_SUPPORT
>     if (uri->is_websocket && aux->ws_handshake_detect && uri->method == HTTP_GET) {
>         ESP_LOGD(TAG, LOG_FMT("Responding WS handshake to sock %d"), aux->sd->fd);
>         esp_err_t ret = httpd_ws_respond_server_handshake(&hd->hd_req);
>         if (ret != ESP_OK) {
>             return ret;
>         }
> 
>         aux->sd->ws_handshake_done = true;
>         aux->sd->ws_handler = uri->handler;
> 
>         /* Return immediately after handshake, no need to call handler here */
>         return ESP_OK;
>     }
> #endif
> ```
> 
> This block is not executed certain times. For the same the upgrade HTTP_GET websocket request is not handled by :
> `httpd_ws_respond_server_handshake(&hd->hd_req);`
> For which the result occurs as Invalid state : handshake not done.
> Few times the request goes through the handshake after which the connection is established.
> 
> On disconnecting websocket from client the core dumps at
> while (ra->remaining_len)
> httpd_parse.c : httpd_req_delete : 791
> 
> ```
> /* Function that resets the http request data
>  */
> esp_err_t httpd_req_delete(struct httpd_data *hd)
> {
>     httpd_req_t *r = &hd->hd_req;
>     struct httpd_req_aux *ra = r->aux;
> 
>     /* Finish off reading any pending/leftover data */
>     **while (ra->remaining_len)** {
>         /* Any length small enough not to overload the stack, but large
>          * enough to finish off the buffers fast */
>         char dummy[CONFIG_HTTPD_PURGE_BUF_LEN];
>         int recv_len = MIN(sizeof(dummy), ra->remaining_len);
>         recv_len = httpd_req_recv(r, dummy, recv_len);
>         if (recv_len < 0) {
>             httpd_req_cleanup(r);
>             return ESP_FAIL;
>         }
> 
>         ESP_LOGD(TAG, LOG_FMT("purging data size : %d bytes"), recv_len);
> 
> #ifdef CONFIG_HTTPD_LOG_PURGE_DATA
>         /* Enabling this will log discarded binary HTTP content data at
>          * Debug level. For large content data this may not be desirable
>          * as it will clutter the log */
>         ESP_LOGD(TAG, "================= PURGED DATA =================");
>         ESP_LOG_BUFFER_HEX_LEVEL(TAG, dummy, recv_len, ESP_LOG_DEBUG);
>         ESP_LOGD(TAG, "===============================================");
> #endif
>     }
> ```
> 
> Refer above screenshots for the details of the log .
> 
> Thank you.
> Sasank Panda

Hi @huming2207,

Thank you for your time and effort on this, I was looking for a simple websocket implementation without the arduino core for a while now. From the little info I could find online there were a few options. but your version seems to be to only one working for me.

I'm no real hero on using git, so I did not know how to pull in your updated branch, so I backed up the old esp_http_server and replaced it with your version I downloaded from the git page.

After increasing the header buffer to 1024 I was able to connect and send/receive messages,
but when disconnecting I got the same error as @githubforsank did.

while (ra->remaining_len)
httpd_parse.c : httpd_req_delete : 791

`
/* Function that resets the http request data
 */
esp_err_t httpd_req_delete(struct httpd_data *hd)
{
    httpd_req_t *r = &hd->hd_req;
    struct httpd_req_aux *ra = r->aux;

    /* Finish off reading any pending/leftover data */
    **while (ra->remaining_len)** {
        /* Any length small enough not to overload the stack, but large
`

Looking into this I got a bit overwhelmed and was not able to find the cause nor solution for this,
so the only thing I could come up with to get it working was to modify it like:

`
/* Function that resets the http request data
 */
esp_err_t httpd_req_delete(struct httpd_data *hd)
{
    httpd_req_t *r = &hd->hd_req;
    struct httpd_req_aux *ra = r->aux;

    if (!ra) {
        return ESP_FAIL;
    }
`

Can you please comment on that ?

Thanks,
J. Lourens

huming2207:
Hello @JonathanLourens 

Thanks for your reply. I'm currently left for holiday and busy dealing with some of my other personal projects. Yes there are some problems in `httpd_req_delete()`. I will take a look at it tomorrow.

Regards,
Jackson

huming2207:
Hello @githubforsank and @JonathanLourens ,

Firstly sorry for the late reply. I've left for a trip with my family in December 2019 and I forgot this issue. Then when I returned back, I was busy working on my graduate projects.

I think I've got the answer. In `httpd_req_new()`, I've called `httpd_req_cleanup()` [when handling WS Close request](https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L754). Then `req_aux` got wiped up (and set to NULL). Later when  `httpd_req_new()` returns, `httpd_req_delete()` uses the `req_aux` again, where it has been set to `NULL`.

I will fix up the handling logic ASAP. You guys can comment out this line: https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L754 for now as a workaround

Regards,
Jackson

huming2207:
Hello @Alvin1Zhang  and @shahpiyushv ,

I've got a question need to confirm first. 

On https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L703

I would like to know why there is no `return;` after `httpd_sess_free_ctx()` is called? Did I miss something or it's intended to behave like that?

Regards,
Jackson

huming2207:
Hello again @Alvin1Zhang and @shahpiyushv 

> On https://github.com/huming2207/esp-idf/blob/feature/httpd-ws/components/esp_http_server/src/httpd_parse.c#L703
> 
> Or, on: https://github.com/espressif/esp-idf/blob/master/components/esp_http_server/src/httpd_parse.c#L679
> 
> I would like to know why there is no return; after httpd_sess_free_ctx() is called? Did I miss something or it's intended to behave like that?
> 

Sorry I misread the code. I've also removed the duplicated `httpd_req_cleanup()`. Please review it. Thanks.

Jackson

huming2207:
Hello @JonathanLourens and @shahpiyushv ,

Please have a try on my latest force-pushed commit. Now all the invalid connections (broken FD) and Close requests should be handled properly.

Regards,
Jackson

shahpiyushv:
@huming2207 , thanks for the updates. I have been a bit busy lately. I will check this soon and if it seems fine, I will send it for internal review.

huming2207:
@shahpiyushv Ok. By the way, is that for ESP32s2?

githubforsank:
Hi, 
@huming2207 thanks for the correction on the closing handle request. I have been scratching my head because I was able to connect to WS but only unpredictably. Upon looking closer to the code and debugging I found the line : 
https://github.com/espressif/esp-idf/blob/4cf2bfc1eed5f35d8abf2f25159c741011e49cfb/components/esp_http_server/src/httpd_uri.c#L313

uri->is_websocket  was returning false and true at random times. 

https://github.com/espressif/esp-idf/blob/4cf2bfc1eed5f35d8abf2f25159c741011e49cfb/components/esp_http_server/src/httpd_uri.c#L174

forgets to copy the is_websocket data to the corresponding handler. 
```
#ifdef CONFIG_HTTPD_WS_SUPPORT
            hd->hd_calls[i]->is_websocket = uri_handler->is_websocket;
#endif
```
This line corrects the handling of the connection. 
You have been lucky all this time to the classy garbage memory error ;). 
You may consider adding this correction. 
Thanks,
Sasank Panda

huming2207:
Hello @githubforsank 

Sorry for this late reply. I am very busy right now. I think that's the problem is. I will fix it this weekend.

Jackson

jaghatei:
First of all: great extension!
I have it running in a developement project for 2 weeks now.
Up to now I am missing only one point:
As there is no other id for identifying the client / session than the file descriptor, you should consider to shift httpd_req_aux structure from private include file esp_httpd_priv.h to public esp_http_server.h to enable handler to know from which particular origin a request is comming.

static esp_err_t ws_ping_handler(httpd_req_t *req)
{
	uint8_t buf[128] = { 0 };
	httpd_ws_frame_t ws_frame;
	memset(&ws_frame, 0, sizeof(httpd_ws_type_t));
	ws_frame.payload = buf;
	esp_err_t ret = httpd_ws_recv_frame(req, &ws_frame, 128);
	ESP_LOGI(TAG, "Got packet from fd: %d with message: %s", ((httpd_req_aux*) req->aux)->sd->fd , ws_frame.payload);

	ret = ret ? : httpd_ws_send_frame(req, &ws_frame);
	return ret;
}

huming2207:
> First of all: great extension!
> I have it running in a developement project for 2 weeks now.
> Up to now I am missing only one point:
> As there is no other id for identifying the client / session than the file descriptor, you should consider to shift httpd_req_aux structure from private include file esp_httpd_priv.h to public esp_http_server.h to enable handler to know from which particular origin a request is comming.
> 


Hello @jaghatei ,

You can separate them from getting their FDs. If that's not enough, try adding your own identifier (something like Socket.IO or MQTT over WS).

Jackson

huming2207:
Hello @githubforsank 

Sorry again for the delay, I was busy preparing my graduate project these days.

I've fixed the issue you've mentioned. Please test again if necessary. Great thanks again!

By the way, I'm just wondering why I didn't encounter this issue before. Are you using the latest toolchain (8.2, 2019r2) or the older 5.x toolchain? I know that was a bug, but I'm just curious why you encountered it but I haven't...😂

Regards,
Jackson

jaghatei:
Ups, I missed the existing function httpd_req_to_sockfd. So ignore my remark.

huming2207:
> Ups, I missed the existing function httpd_req_to_sockfd. So ignore my remark.

Yep, that's what I mean. 

WS is just a wrapper of raw TCP socket that design for co-existing with some HTTP services. You need some upper-layer stuff for some stateful message transfer.

huming2207:
Btw @Alvin1Zhang @shahpiyushv , any updates on your side?

Alvin1Zhang:
@huming2207 Thanks for the contribution and updates, and sorry for the slow reply. We have taken the PR for an internal review. Will keep you posted about the updates. Thanks.

huming2207:
Thanks @Alvin1Zhang ! Please keep me informed if possible,

fri-sch:
@huming2207 First of all thanks for your efforts on working on the WebSocket server implementation.

> For the async send functions, I will not do it anyway, as it is implicit and out of scope. I have mentioned the detailed reasons in the previous replies.

In my mind, the main benefit for using WebSockets is async send and receive on server and client side. An implementation that does not support async send is pretty useless in many cases because one can only use the same request/reply flow that would also be possible with simple HTTP. That's why I think we should work on a proper solution for this before merging anything.

I also wonder why httpd_ws_recv_frame() doesn't allocate the payload buffer for the frame dynamically on the heap instead of relying on a static buffer.

Furthermore the PR doesn't look like being ready for merging. It seems the fixes commits should be squashed into the first commit and the unrelated merge commits should probably be dropped.

huming2207:
Hello @fri-sch 

> I also wonder why httpd_ws_recv_frame() doesn't allocate the payload buffer for the frame dynamically on the heap instead of relying on a static buffer.

Because this is a library that works on ESP32 with a few hundred KB RAM, not a server that comes with a few hundred GB RAM.

> In my mind, the main benefit for using WebSockets is async send and receive on server and client side. An implementation that does not support async send is pretty useless in many cases because one can only use the same request/reply flow that would also be possible with simple HTTP. That's why I think we should work on a proper solution for this before merging anything.

Same reason as the first one. I don't think this is really needed on a microcontroller. You need to somehow keep the payload in a queue (copy to somewhere, or take the buffer pointer and ask the user to keep the buffer by themselves). This increases code complexity and RAM usage.

> Furthermore the PR doesn't look like being ready for merging. It seems the fixes commits should be squashed into the first commit and the unrelated merge commits should probably be dropped.

Thanks for the suggestion. I will do it tomorrow.

Regards,
Jackson

fri-sch:
> Because this is a library that works on ESP32 with a few hundred KB RAM, not a server that comes with a few hundred GB RAM. Keep doing heap allocation will cause heap fragmentation.

Yes, but in my experience this is not really an issue for most applications. The dynamic buffer would be allocated when the frame is received and live only until the frame is handled. When using a static buffer you have to reserve memory the size of the maximum frame size, which could be even more critical in applications with memory constraints. Anyway I think there should be two versions of the receive function at least, so one can choose between static and dynamic buffering.

> Same reason as the first one. I don't think this is really needed on a microcontroller. You need to somehow keep the payload in a queue (copy to somewhere, or take the buffer pointer and ask the user to keep the buffer by themselves). This increases code complexity and RAM usage, and heap fragmentation as well.

Yes, there is some effort needed to implement this, but I think it's feasible. And yes this is needed, even on a microcontroller, because as I explained above it's the basic principle and the main reason for using WebSockets. Without it I can just use simple HTTP requests from the client that are answered by the HTTP server.

> Again, just a reminder, this is a microcontroller. Yes, it's a big one compared to some Cortex-M0+ chips, but we still can't keep relying on too many "common senses" that designing a library for the beefy x86 server.

I think you somehow have a false impression of the capabilities of the hardware. We have a FreeRTOS system running in the background and many protocol stacks and drivers that just do what you claim is too complex, without any hassle.

huming2207:
Hi @fri-sch ,

> Yes, there is some effort needed to implement this, but I think it's feasible. And yes this is needed, even on a microcontroller, because as I explained above it's the basic principle and the main reason for using WebSockets. Without it I can just use simple HTTP requests from the client that are answered by the HTTP server.

I understand but unfortunately, I can't fully agree with your proposal. I will consider moving my forked library as a 3rd party library instead.

Regards,
Jacskon

fri-sch:
> I'm sorry I have to close this, as I want this library to be simple and suitable for a microcontroller. However, you've delayed this PR for nearly 6 months and asking for a lot of extra functions that are not really useful, so I would choose to re-consider my decision.

Please note that what I said above is just my personal opinion as a developer. I'm not related to Espressif and I'm not maintaining esp-idf, so it's not up to me to decide if your proposal is suitable or not. I just joined the discussion today and I haven't been delaying anything.

You have to understand that the maintainers of esp-idf require that your contributions are not only suitable for yourself, but for the majority of users and that certain things need to be discussed before accepting.

When I have some time I will take a closer look at the code and maybe I can help with some of the parts, that I consider to be missing.

david-cermak:
@huming2207 
Hi Jackson,

As @Alvin1Zhang mentioned [above](https://github.com/espressif/esp-idf/pull/4306#issuecomment-595576474) your PR has already been taken in for internal review (and adjustments). No further work is expected from your side. Thanks again for your effort and mainly for the patience with us.

Please note that anybody could comment, open a discussion, raise concerns. And there are indeed some open points which we've decided to address internally as well as squashing commits, formatting, adding tests, because it is just easier and faster for us. Your implementation is a good starting point which can be extended latter.

If you are okay with this approach we could continue and keep you updated. Just to be clear I *am* related to Espressif.

General note: All comments, questions, concerns are welcome (thanks everyone involved)

huming2207:
Hi @fri-sch 

Oops sorry I thought you were Espressif staffs or someone who has access to the

huming2207:
@fri-sch Ah so sorry... I thought you were the Espressif staff or ESP-IDF maintainer. I was a bit busy and I didn't have a detailed check last night.

huming2207:
@fri-sch However I will not consider adding async r/w support in this PR, as I really don't want to make this library too complex and RAM-hungry on a uC, even if it's powerful. But I'm definitely happy to see if someone will add optional async functions (and can be selected in menuconfig) on later PRs.

For @Alvin1Zhang and @shahpiyushv , Please proceed to review and very sorry for the trouble I've made...

Alvin1Zhang:
@huming2207 Thanks for contribution, the changes have been merged, will close this PR. Thanks.

PR #4326, opened 2019-11-12
Allow processing of ULP macro array in RTC slow mem (IDFGH-2172)

boarchuz:
This adds **ulp_process_macros_in_place()**, a variant of ulp_process_macros_and_load(). It takes a ULP macro array that is already in RTC slow memory and _replaces_ it with the output program.

```
RTC_DATA_ATTR const ulp_insn_t program[] = { M_LABEL(1), M_BX(1), };
size_t programSize = sizeof(program) / sizeof(ulp_insn_t);
ulp_process_macros_in_place(program, &programSize);
ulp_run((uint32_t*)program - RTC_SLOW_MEM);
```

The underlying processing functions are already safe (read_ptr>=write_ptr) so there isn't much to it.

By initialising directly in RTC memory, flash and/or dram usage is reduced.

Incidentally, this will also permit bypassing the restrictive pre-compiled ULP reserved memory size in the Arduino environment. (For greater flexibility, this value may even be reduced to 0 so that the full RTC slow memory is available for use, with only a minor inconvenience to those needing ULP memory.)

Of course it does mean more RTC memory is used, proportional to the number of macros in the program. A motivated programmer could repurpose this memory, if necessary, after processing is complete, by using the program size returned from ulp_process_macros_in_place() to determine the offset and size of unused memory.

### DIFF
--- a/components/ulp/ulp_macro.c
+++ b/components/ulp/ulp_macro.c
@@ -313,16 +313,15 @@ esp_err_t ulp_process_macros_and_load(uint32_t load_addr, const ulp_insn_t* prog
 esp_err_t ulp_process_macros_in_place(const ulp_insn_t* program, size_t* psize)
 {
     size_t macro_count = count_ulp_macros(program, *psize);
-    size_t real_program_size = *psize - macro_count;
     const size_t ulp_mem_end = 0x1000 / sizeof(ulp_insn_t);
     uint32_t load_addr = program - (ulp_insn_t*)RTC_SLOW_MEM;
     if (program < RTC_SLOW_MEM || load_addr >= ulp_mem_end) {
         ESP_LOGW(TAG, "program in invalid location in memory, check attributes");
         return ESP_ERR_ULP_INVALID_LOAD_ADDR;
     }
-    if (real_program_size + load_addr > ulp_mem_end) {
+    if (*psize + load_addr > ulp_mem_end) {
         ESP_LOGE(TAG, "program too big: %d words, max is %d words",
-                real_program_size, ulp_mem_end - load_addr);
+                *psize, ulp_mem_end - load_addr);
         return ESP_ERR_ULP_SIZE_TOO_BIG;
     }
     // If no macros found, return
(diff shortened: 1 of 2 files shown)


components/ulp/ulp_macro.c
```diff
@@ -313,16 +313,15 @@ esp_err_t ulp_process_macros_and_load(uint32_t load_addr, const ulp_insn_t* prog
...
         ESP_LOGW(TAG, "program in invalid location in memory, check attributes");
         return ESP_ERR_ULP_INVALID_LOAD_ADDR;
     }
-    if (real_program_size + load_addr > ulp_mem_end) {
+    if (*psize + load_addr > ulp_mem_end) {
         ESP_LOGE(TAG, "program too big: %d words, max is %d words",
-                real_program_size, ulp_mem_end - load_addr);
+                *psize, ulp_mem_end - load_addr);
         return ESP_ERR_ULP_SIZE_TOO_BIG;
     }
```
boarchuz: - Should an alignment check be added here or can it be assumed for ulp_insn_t?

- Is the 4K reserved slow memory strict, or should the program be allowed to 'overflow' into this region? If it's not enforced, should the ESP_ERR_ULP_SIZE_TOO_BIG check be removed (as there should be an error at compile time for rtc slow seg overflow)?

Alvin1Zhang:
@boarchuz Thanks for the contribution.

boarchuz:
Actually, it would be better to reserve destination memory (eg. using RTC_DATA_ATTR) and pass this destination pointer to the process_macros function along with the source pointer. That would be more flexible (it doesn't matter in processing whether or not src==dest so there's no need for separate functions).
It would arguably make the config option to reserve ULP memory obsolete, as well as being zero-initialised by default in case some is used for ulp data, etc.
eg. 

```
//Process and copy
RTC_DATA_ATTR ulp_insn_t programDest[123];
void init_ulp() {
  const ulp_insn_t programSrc[] = { /* ... */ };
  size_t programSize = sizeof(programSrc);
  ulp_process_macros_and_load(programDest, sizeof(programDest), programSrc, &programSize,  );
  ulp_run((uint32_t*)programDest - RTC_SLOW_MEM);
}
//or process and overwrite
RTC_DATA_ATTR ulp_insn_t program[123]  = { /* ... */ };
void init_ulp() {
  size_t programSize = sizeof(program);
  ulp_process_macros_and_load(program, sizeof(program), program, &programSize);
  ulp_run((uint32_t*)program - RTC_SLOW_MEM);
}
```

I realise it's fairly unimportant but I think it is definitely an improvement so I'll submit a PR for consideration whenever I get around to it.

PR #4352, opened 2019-11-17
Provide optional callback for wakeup from automatic light sleep (IDFGH-2200)

boarchuz:
Unless I'm missing something, it seems often impossible to be notified of a wakeup from automatic light sleep.

Consider this typical scenario:
An ISR (that gives a semaphore) is configured to trigger on a GPIO falling edge. PM config is set. When tasks are idle, the ESP32 automatically enters light sleep.
Later, a falling edge occurs -> the ESP32 begins to wake up from light sleep -> the falling edge ends -> the ESP32 is ready -> <???> -> No tasks are ready to run -> the ESP32 re-enters light sleep. 
The ISR never runs, the semaphore is never given, the app is blissfully unaware that the ESP32 even woke up.

This PR allows inserting an optional callback where <???> is so that the app is notified when a wakeup occurs, and can handle it accordingly. Consider it a light sleep 'wakeup stub'.

eg.
```
void myLightSleepWakeupHandler()
{
    if(gpio_get_level(GPIO_NUM_0) == 0) {
        xSemaphoreGive(someSemaphore);
    }
    xTaskNotifyGive(someTaskHandle);
    ++lightSleepWakeupCount;
}

esp_pm_set_light_sleep_wakeup_cb(myLightSleepWakeupHandler);
```

If there's an existing way to do this, please let me know!

Additionally, when WiFi is enabled, this may be used to synchronise tasks with WiFi-related wakeups in order to maximise the potential power savings of automatic light sleep.

### DIFF
--- a/components/esp32/pm_esp32.c
+++ b/components/esp32/pm_esp32.c
@@ -88,6 +88,9 @@ static uint32_t s_ccount_mul;
  */
 static bool s_skipped_light_sleep[portNUM_PROCESSORS];
 
+/* Optional user-provided callback function for wakeup from automatic light sleep */
+static esp_pm_light_sleep_wakeup_callback s_light_sleep_wakeup_cb = NULL;
+
 #if portNUM_PROCESSORS == 2
 /* When light sleep is finished on one CPU, it is possible that the other CPU
  * will enter light sleep again very soon, before interrupts on the first CPU
@@ -487,6 +490,11 @@ void esp_pm_impl_waiti(void)
 
 #if CONFIG_FREERTOS_USE_TICKLESS_IDLE
 
+void esp_pm_set_light_sleep_wakeup_cb(esp_pm_light_sleep_wakeup_callback cb)
+{
+    s_light_sleep_wakeup_cb = cb;
+}
+
 static inline bool IRAM_ATTR should_skip_light_sleep(int core_id)
 {
 #if portNUM_PROCESSORS == 2
@@ -513,6 +521,7 @@ static inline void IRAM_ATTR other_core_should_skip_light_sleep(int core_id)
 
 void IRAM_ATTR vApplicationSleep( TickType_t xExpectedIdleTime )
 {
+    bool went_to_sleep = false;
     portENTER_CRITICAL(&s_switch_lock);
     int core_id = xPortGetCoreID();
     if (!should_skip_light_sleep(core_id)) {
@@ -535,6 +544,8 @@ void IRAM_ATTR vApplicationSleep( TickType_t xExpectedIdleTime )
             int64_t slept_us = esp_timer_get_time() - sleep_start;
             ESP_PM_TRACE_EXIT(SLEEP, core_id);
 
+            went_to_sleep = true;
+
             uint32_t slept_ticks = slept_us / (portTICK_PERIOD_MS * 1000LL);
             if (slept_ticks > 0) {
                 /* Adjust RTOS tick count based on the amount of time spent in sleep */
@@ -554,6 +565,10 @@ void IRAM_ATTR vApplicationSleep( TickType_t xExpectedIdleTime )
         }
     }
     portEXIT_CRITICAL(&s_switch_lock);
+    if(went_to_sleep && s_light_sleep_wakeup_cb != NULL)
+    {
+        s_light_sleep_wakeup_cb();
+    }
 }
 #endif //CONFIG_FREERTOS_USE_TICKLESS_IDLE
 
--- a/components/esp32/pm_esp32.c
+++ b/components/esp32/pm_esp32.c
@@ -565,8 +565,7 @@ void IRAM_ATTR vApplicationSleep( TickType_t xExpectedIdleTime )
         }
     }
     portEXIT_CRITICAL(&s_switch_lock);
-    if(went_to_sleep && s_light_sleep_wakeup_cb != NULL)
-    {
+    if (went_to_sleep && s_light_sleep_wakeup_cb != NULL) {
         s_light_sleep_wakeup_cb();
     }
 }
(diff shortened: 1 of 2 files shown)


components/esp32/pm_esp32.c
```diff
@@ -554,6 +565,9 @@ void IRAM_ATTR vApplicationSleep( TickType_t xExpectedIdleTime )
         }
     }
     portEXIT_CRITICAL(&s_switch_lock);
+    if (went_to_sleep && s_light_sleep_wakeup_cb != NULL) {
+        s_light_sleep_wakeup_cb();
```
igrr: Note that in practice this will be called not after the wakeup, but before entering the sleep again. This is probably okay for many use cases, but it should be noted in the description of the new API.

components/esp_common/include/esp_pm.h
```diff
@@ -175,6 +180,13 @@ esp_err_t esp_pm_lock_delete(esp_pm_lock_handle_t handle);
  */
 esp_err_t esp_pm_dump_locks(FILE* stream);
 
+#if CONFIG_FREERTOS_USE_TICKLESS_IDLE
+/**
+ * Set an optional callback function to be invoked on wakeup from automatic light sleep.
+ */
+void esp_pm_set_light_sleep_wakeup_cb(esp_pm_light_sleep_wakeup_callback cb);
```
igrr: Have you considered adding the callback as a new member to the `esp_pm_config_t` structure? This would eliminate the need for adding `esp_pm_set_light_sleep_wakeup_cb`.

components/esp_common/include/esp_pm.h
```diff
@@ -65,6 +65,11 @@ esp_err_t esp_pm_configure(const void* config);
  */
 typedef struct esp_pm_lock* esp_pm_lock_handle_t;
 
+/**
+ * @brief Automatic light sleep wakeup callback function type
+ */
+typedef void  (*esp_pm_light_sleep_wakeup_callback)();
```
igrr: it is often useful to pass an optional user-provided `void* ctx` argument to such callbacks, to transfer some state, could you add that?

components/esp_common/include/esp_pm.h
igrr: could you please also add a sentence to the description of this callback, explaining that no blocking functions can be called from the callback, including printf/ESP_LOG.

components/esp32/pm_esp32.c
boarchuz: Oops, thanks for pointing that out. I think it would be best to notify the app via the callback as soon as possible.
I've had another look though and I'm not sure how to make that change - can you point me in the right direction?

components/esp_common/include/esp_pm.h
boarchuz: I'm wary about fiddling with the struct as it may be breaking (in my understanding, at least).
eg. If currently using a local esp_pm_config_esp32_t variable, when the struct definition is updated with a new pointer this field will be uninitialised, causing esp_pm_configure to set the callback to an undefined address (if not null). I'll leave it to someone else to decide if that's acceptable.

boarchuz:
@igrr I think this could be easily and cleanly implemented via the new v4 events API.

Base = SYSTEM_EVENT / POWER_EVENT, id = x_EVENT_LIGHT_SLEEP_WAKEUP

Thoughts?

igrr:
It's probably (?) okay if we only raise this event on wake-ups not caused by timer: EXT0/EXT1/GPIO/ULP/Touch, so that there is no overhead for most of the wake-ups.

One practical issue is how to send this event to the event loop task. `esp_event_post` calls xQueueSend, so it can't be called from the critical section enclosing esp_light_sleep_start. And calling this function outside of the critical section will in most cases result an extra context switch: idle task -> tick interrupt -> idle task -> event loop task, instead of idle task -> tick interrupt -> event loop task.

Another aspect to consider is whether the same event should be raised on wake-ups from forced light sleep as well. On one hand this is good for consistency, on the other hand this may lead to weird issues when the light sleep is initiated on the core other than the one where the event loop task runs. In this case the code may enter light sleep again before the event callback has a chance to execute, leading to missed wakeup events.

Overall I think that your initial approach is less error prone. Although need to think how to ensure that the callback is called after wake-up, rather than upon return to the idle task.

esp-lis:
> Unless I'm missing something, it seems often impossible to be notified of a wakeup from automatic light sleep.
> 
> Consider this typical scenario: An ISR (that gives a semaphore) is configured to trigger on a GPIO falling edge. PM config is set. When tasks are idle, the ESP32 automatically enters light sleep. Later, a falling edge occurs -> the ESP32 begins to wake up from light sleep -> the falling edge ends -> the ESP32 is ready -> <???> -> No tasks are ready to run -> the ESP32 re-enters light sleep. The ISR never runs, the semaphore is never given, the app is blissfully unaware that the ESP32 even woke up.
> 
> This PR allows inserting an optional callback where <???> is so that the app is notified when a wakeup occurs, and can handle it accordingly. Consider it a light sleep 'wakeup stub'.
> 
> eg.
> 
> ```
> void myLightSleepWakeupHandler()
> {
>     if(gpio_get_level(GPIO_NUM_0) == 0) {
>         xSemaphoreGive(someSemaphore);
>     }
>     xTaskNotifyGive(someTaskHandle);
>     ++lightSleepWakeupCount;
> }
> 
> esp_pm_set_light_sleep_wakeup_cb(myLightSleepWakeupHandler);
> ```
> 
> If there's an existing way to do this, please let me know!
> 
> Additionally, when WiFi is enabled, this may be used to synchronise tasks with WiFi-related wakeups in order to maximise the potential power savings of automatic light sleep.

I think a better implementation is to register higher level interrupts (level2 or level3) and implement the corresponding logic in the context of the interrupt service routine. For example:
```
void myLightSleepWakeup_isr_handler()
{
    if(gpio_get_level(GPIO_NUM_0) == 0) {
        xSemaphoreGiveFromISR(someSemaphore);
    }
    xTaskNotifyGiveFromISR(someTaskHandle);
    ++lightSleepWakeupCount;
}
```

Alvin1Zhang:
Thanks for your contribution again, will close due to short of feedback, feel free to reopen with more updates. Thanks.

PR #4374, opened 2019-11-20
espcoredump: Add option to prepend user-specified data to coredump (IDFGH-2225)

chrismerck:
I am looking for comment on this **draft** PR. This is a modification we needed for a commercial product, but I believe the community would benefit from this capability being baked into the IDF.

----

## Background

The coredump feature can provide crucial information from IoT devices in the field back to a manufacturer's cloud. However, interpretation of those coredumps requires some metadata which is not present in the dump itself:

- hash of corresponding ELF file (essential)
- app version (useful)
- project name (useful)
- app build datetime (useful, can even replace hash of ELF file)

This data *could* be added to the coredump at time of upload, however this introduces the possibility of a mismatch between the FW version which crashed and the FW version doing the uploading.

Furthermore, it is useful to have additional context around the crash, such as:

- last console messages
- list of errors which occurred on that boot

## Current PR

This PR proposes a user-driven solution where an sdkconfig option enables the pre-pending of a block of "userdata" prior to the regular coredump. It is then the responsibility of the manufacturer's cloud to strip the userdata prior to passing the remainder of the coredump into espcoredump.py.

One awkward point here is, I found it necessary to prepend the `data_len` as the first word of coredump---prior to userdata---so that `esp_core_dump_image_get` can delimit the dump and successfully mmap and CRC check.

I have implemented this solution in a commercial product, using AWS S3 for uploads, AWS Lambda for decoding the coredumps, and Sentry for reporting the crashes. It is immensely valuable.

## Suggestion

The proposed solution gives maximum flexibility to the app developer, but perhaps a different approach would be more in the spirit of ESP-IDF:

1. We could instead extend the format of the coredump to (either always or optionally) add the `esp_app_desc_t` structure, which has all the metadata needed. Then the espcoredump.py program could continue to be run against the whole coredump partition. Espcoredump.py could even check the hash of the ELF provided and warn if it does not match. Best of all, the app developer would not need to write *any* code to take advantage of this feature.

2. It is still desirable to allow a userdata block to be provided (say, as one of the records in the standard coredump format), so that the app developer can provide a list of errors or console messages, or perhaps other data that is useful for that specific application developer to debug. --- But I see this as more of a roadmap issue, let's make sure we don't preclude this being eventually possible, no need for the current PR to allow arbitrary userdata if the version metadata is embedded.

---

NOTE: I originally posted about this on the forum, but I believe that a draft PR is a more productive place to discuss. (https://esp32.com/viewtopic.php?f=13&t=13047)


chrismerck:
[Purpose of force-push was to update committer email to appease @CLAassistant]

Alvin1Zhang:
@chrismerck Thanks for the contribution.

chrismerck:
heads up @igrr the CI seems to be stuck in a loop. I've gotten ~100 emails over the past few days about some error. Pic attached.

Screen Shot 2019-11-25 at 14 07 27

projectgus:
@chrismerck Really sorry about this email spam. I think I understand what's happened:

We have an hourly scheduled GitHub Action that runs in the repo to sync new PRs into our JIRA tracker. (This is because otherwise PR events run in the context of the originating repo, not our repo.)

The scheduled Action is also running in your fork, but of course it fails there because your fork doesn't have any of the credentials configured that our repo has: https://github.com/chrismerck/esp-idf/actions?query=workflow%3A%22Sync+remain+PRs+to+Jira%22

This was an unexpected side effect of adding this schedule, sorry!

You should be able to disable Actions from running at all in your repo here: https://github.com/chrismerck/esp-idf/settings/actions - which will stop the spam.

I'll add a check in our internally-oriented actions that causes them to pass silently if they're not running as part of the "espressif" organization. Still a bit wasteful that all our forks will be spinning up a docker container once an hour to do nothing, but I can't see any fix for that!

chrismerck:
> You should be able to disable Actions from running at all in your repo here:

That was easy! Thanks. I was more concerned that we were consuming inordinate amounts of espressif CPU cycles :)

---

Any thoughts on the PR? Or how others solve this issue without having the metadata?

gerekon:
@chrismerck In upcoming IDF v4.1 we are changing coredump format from bin to ELF, but keeping backward compatibility with old binary format (updated script can parse it).
Regarding you suggestions:
- Coredump ELF format will include SHA256 for app's ELF it was generated for, using this you can clearly determine app's ELF file.
 - App's version and project name, from my point of view should be embedded into app ELF file. It seems to be more convenient place. In any case coredump cannot be decoded w/o program ELF and you will need to keep it for dumps analysis. So I do not see the reason to duplicate this info in coredump.

chrismerck:
@gerekon Thank you for suggesting a solution based on v4.1. I'm now undertaking to upgrade our products to v4.1-beta1 (no small task it seems), and will try as you suggest. I intend to share my findings here and close the issue within 3 weeks.

igrr:
I'm going to close the PR since the ELF file hash has been included into the core dumps since IDF v4.1.

PR #4416, opened 2019-11-28
jsmn configuration added (IDFGH-2274)

Lapshin:
It is convenient to have this conf for my project. I think it may be helpful for others


suda-morris:
Thanks for your nice contribution, we will merge this ASAP.

PR #4414, opened 2019-11-27
Added driver for ENC28J60 Ethernet Controller (IDFGH-2272)

hendog82:
I wrote a driver for the ENC28J60 Ethernet Controller.

The ENC28J60 uses SPI (similar to the DM9051) to communicate with the ESP32. The format of driver I wrote is based off of the DM9051 driver. 

I also re-wrote the "basic" ethernet example so that this driver can be used. 

I have also tested this driver with the MQTT Publish and HTTP request examples without any problems.

PR #4435, opened 2019-12-03
Added ENC28J60 Driver and added ENC28J60 to eth basic example (IDFGH-2295)

hendog82:
I wrote a driver for the ENC28J60 Ethernet Controller. (Use this instead of pull request #4414 )

The ENC28J60 uses SPI (similar to the DM9051) to communicate with the ESP32. The format of driver I wrote is based off of the DM9051 driver.

I also re-wrote the "basic" ethernet example so that this driver can be used.

I have also tested this driver with the MQTT Publish and HTTP request examples without any problems.

### DIFF
--- a/components/esp_eth/Kconfig
+++ b/components/esp_eth/Kconfig
@@ -143,14 +143,24 @@ menu "Ethernet"
             ESP-IDF can also support some SPI-Ethernet modules.
 
     if ETH_USE_SPI_ETHERNET
-        menuconfig ETH_SPI_ETHERNET_DM9051
+        choice ETH_SPI_INTERFACE
+            prompt "SPI PHY interface"
+            default ETH_SPI_ETHERNET_DM9051
+            help
+                select which SPI-Ethernet module is being used
+            config ETH_SPI_ETHERNET_DM9051
             bool "Use DM9051"
-            default y
             help
                 DM9051 is a fast Ethernet controller with an SPI interface.
                 It's also integrated with a 10/100M PHY and MAC.
                 Set true to enable DM9051 driver.
-
+            config ETH_SPI_ETHERNET_ENC28J60
+            bool "Use ENC28J60"
+            help
+                ENC28J60 is a Stand-Alone Ethernet Controller with SPI Interface.
+                It has an integrated MAC and 10Base-T PHY.
+                Select to enable ENC28J60 driver.
+        endchoice
         if ETH_SPI_ETHERNET_DM9051
             config ETH_DM9051_INT_GPIO
                 int "DM9051 Interrupt GPIO number"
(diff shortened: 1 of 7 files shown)


hendog82:
Hi @suda-morris,

What is the expected timeline of this review?

Thanks,
Henry

suda-morris:
> Hi @suda-morris,
> 
> What is the expected timeline of this review?
> 
> Thanks,
> Henry

Hi @hendog82 Thank you for your nice code! What I can tell is it's already in our internal review queue, we need some time to test and clean up. At least it should appear as a 4.2 feature :wink:

lucastcox:
@hendog82 , might not be a bad idea to have a config option for using either software reset or the hardware reset pin.

termtek:
> I wrote a driver for the ENC28J60 Ethernet Controller. (Use this instead of pull request #4414 )
> 
> The ENC28J60 uses SPI (similar to the DM9051) to communicate with the ESP32. The format of driver I wrote is based off of the DM9051 driver.
> 
> I also re-wrote the "basic" ethernet example so that this driver can be used.
> 
> I have also tested this driver with the MQTT Publish and HTTP request examples without any problems.

Hi hendog82,
Firstly, thanks for your effort.
I am trying to your example and ENC28J60 driver.
But I have a problem.
I can read the Vendor ID. 
Vendor ID is 6.
But I cant read the PHY Chip Id.
Always returns 0.

Do have any idea?

suda-morris:
Have merged into ESP-IDF and we now have a new example! Thanks @hendog82 for your contributions.

@termtek Suggest to try the example on master and create a new issue when you encounter any problems.

lucastcox:
@suda-morris Do you know why the interrupt pin in sdkconfig is limited to gpio 0-33? Can't 34-39 still be used, since the interrupt pin is push-pull, so no pullup resistor is needed?

lucastcox:
Also, @suda-morris , Thank you for merging this! Just starting a project with the ENC28J60, and I'm glad I can use the unmodified idf for it! :)

phatpaul:
I'm glad to have a lower pin-count Ethernet option on ESP32!  It has been such a hassle to add new features to my product with the LAN8720A using up 9 of the available 22 GPIO's on the WROVER.

1. Will the network latency and throughput be significantly worse with the ENC28J60 vs. LAN8720A, given it is only 10Base-T?  (I'm serving a <1MB web application with libesphttpd.)

2.  I want to support my older devices still using LAN8720A and I don't want to have to distribute more than 1 OTA firmware update file.  Is it possible to change the Ethernet driver at runtime to either LAN8720 or ENC28J60?  (I have the hardware version of my PCB encoded in a special flash partition, so no need to auto-detect the PHY type.)

Thanks!

PR #4470, opened 2019-12-10
add missing gatt descriptors (IDFGH-2339)

bfriedkin:
Hello -

This PR adds definitions for GATT descriptors 0x2909 (Number of Digitals) through 0x290E (Time Trigger Setting) as defined by the current Bluetooth GATT Descriptors specification:

https://www.bluetooth.com/specifications/gatt/descriptors/

In addition, the new descriptors are added to the btc_gatts_act_create_attr_tab() function, so that server applications can deploy services that include these descriptors, e.g. AIOS service. Without this change, it is not possible to deploy an AIOS service on ESP32/Bluedroid, because some of the required descriptors are missing and hence not added to the attribute table.

Regards,
Brian


Alvin1Zhang:
@bfriedkin Thanks for the contribution.

GYC-Espressif:
Thank you for your contribution. We have merged your submissions in the latest IDF with the commit of efc514fa0ad9b33d46d17960107e23125d3923e3. Can you close this? Thank you.

PR #4512, opened 2019-12-18
Created a new unmap to speed up when cache saving is not needed (IDFGH-2395)

Francois-Belanger:
I'm using the Himem to save time when reading multiple time the same files.  But the esp_himem_unmap function runs esp_spiram_writeback_cache() that kills it for me.  It take almost 5ms to write back the cache !  I don't know why.  So I simply added a new function to simply discard the content of the cache when just needing to reading back the Himem.


Alvin1Zhang:
@Francois-Belanger Thanks for the contribution.

PR #4518, opened 2019-12-19
Major performance improvements to idf_size.py (IDFGH-2404)

ajcasagrande:
I really like the information available from `idf_size.py`, however my `firmware.map` file is `15 MB` and it takes a whopping **20 seconds** to run!

After investigating I discovered that a majority of the time is spent running the regexes, especially `RE_SOURCE_LINE`. I tweaked that one and combined it with the cmake archive-less alternative. I also added a bunch of other small tweaks such as pre-filtering the input lines and exiting when we reach the next section in the file.

If you have ever run `idf.py size` or `idf.py size-components` on a fairly large project you will most likely not even believe how fast my version runs.

**Input file size**
```shell
anthony@linux:~/esp/esp-idf/tools$ du -h /tmp/firmware.map 
15M	/tmp/firmware.map
```
**Current code in master branch**
```shell
anthony@linux:~/esp/esp-idf/tools$ time ~/esp/esp-idf-master/tools/idf_size.py /tmp/firmware.map
Total sizes:
 DRAM .data size:   14132 bytes
 DRAM .bss  size:   35128 bytes
Used static DRAM:   49260 bytes (  75320 available, 39.5% used)
Used static IRAM:  108605 bytes (  22467 available, 82.9% used)
      Flash code:  923103 bytes
    Flash rodata:  300984 bytes
Total image size:~1346824 bytes (.bin may be padded larger)

real	0m19.440s
user	0m19.395s
sys	0m0.044s
```

**My version**
```shell
anthony@linux:~/esp/esp-idf/tools$ time ./idf_size.py /tmp/firmware.map
Total sizes:
 DRAM .data size:   14132 bytes
 DRAM .bss  size:   35128 bytes
Used static DRAM:   49260 bytes (  75320 available, 39.5% used)
Used static IRAM:  108605 bytes (  22467 available, 82.9% used)
      Flash code:  923103 bytes
    Flash rodata:  300984 bytes
Total image size:~1346824 bytes (.bin may be padded larger)

real	0m0.385s
user	0m0.361s
sys	0m0.024s
```

**Results**
> From **19.4 seconds** down to **0.4 seconds**

### DIFF
--- a/tools/idf_size.py
+++ b/tools/idf_size.py
@@ -101,10 +101,10 @@ def load_sections(map_file):
     # 0x0000000040080400       0xa4 /home/gus/esp/32/idf/examples/get-started/hello_world/build/esp32/libesp32.a(cpu_start.o)
     # cmake build system links some object files directly, not part of any archive, so make that part optional
     #  .xtensa.info   0x0000000000000000       0x38 CMakeFiles/hello-world.elf.dir/project_elf_src.c.obj
-    RE_SOURCE_LINE = re.compile(r"\s*(?P<sym_name>\S*) +0x(?P<address>[\da-f]+) +0x(?P<size>[\da-f]+) (?P<archive>.+\.a)?\(?(?P<object_file>.+\.(o|obj))\)?", re.M)
+    RE_SOURCE_LINE = re.compile(r"\s*(?P<sym_name>\S*) +0x(?P<address>[\da-f]+) +0x(?P<size>[\da-f]+) (?P<archive>.+\.a)?\(?(?P<object_file>.+\.(o|obj))\)?")
 
     # Fast check to see if line is a potential source line before running the slower full regex against it
-    RE_PRE_FILTER = re.compile(r".*\.(o|obj)\)?", re.M)
+    RE_PRE_FILTER = re.compile(r".*\.(o|obj)\)?")
 
     # Check for lines which only contain the sym name (and rest is on following lines)
     RE_SYMBOL_ONLY_LINE = re.compile(r"^ (?P<sym_name>\S*)$")
--- a/tools/idf_size.py
+++ b/tools/idf_size.py
@@ -66,9 +66,10 @@ def load_memory_config(map_file):
     """ Memory Configuration section is the total size of each output section """
     result = {}
     scan_to_header(map_file, "Memory Configuration")
-    RE_MEMORY_SECTION = r"(?P<name>[^ ]+) +0x(?P<origin>[\da-f]+) +0x(?P<length>[\da-f]+)"
+    RE_MEMORY_SECTION = re.compile(r"(?P<name>[^ ]+) +0x(?P<origin>[\da-f]+) +0x(?P<length>[\da-f]+)")
+
     for line in map_file:
-        m = re.match(RE_MEMORY_SECTION, line)
+        m = RE_MEMORY_SECTION.match(line)
         if m is None:
             if len(result) == 0:
                 continue  # whitespace or a header, before the content we want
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@ajcasagrande Thanks for the contribution.

tools/idf_size.py
```diff
@@ -109,37 +129,34 @@ def load_sections(map_file):
...
-            source["file"] = "%s:%s" % (source["archive"], source["object_file"])
-            section["sources"] += [source]
-
-        # In some cases the section name appears on the previous line, back it up in here
-        RE_SYMBOL_ONLY_LINE = r"^ (?P<sym_name>\S*)$"
-        m = re.match(RE_SYMBOL_ONLY_LINE, line)
-        if section is not None and m is not None:
-            sym_backup = m.group("sym_name")
+            if not RE_PRE_FILTER.match(line):
+                # line does not match our quick check, so skip to next line
```
projectgus: Just out of total curiosity, what's the performance difference from leaving out the pre-filter step?

0xjakob:
@ajcasagrande Thanks a lot for this PR. I think it even improves readability of the code. As @projectgus said, please test with test.sh in tools/test_idf_size/ to verify that it works with the test .map file.

ajcasagrande:
Thanks for your review, guys. I appreciate it.

`flake8` had 1 warning, the `RE_SOURCE_FILE` line was 163 chars (max 160). I was able to fix this by removing the `re.M` flag which I believe to be superfluous since we are reading the file one line at a time.

`flake8` is now passing.

`test.sh` passes:
```
Name                                                       Stmts   Miss  Cover
------------------------------------------------------------------------------
/home/anthony/esp/esp-idf/tools/idf_size.py                  193      2    99%
test_idf_size.py                                              17      1    94%
------------------------------------------------------------------------------
TOTAL                                                        210      3    99%
```

tools/idf_size.py
ajcasagrande: > tl;dr  without the pre-filter it goes from 375 ms to 561 ms

So i started looking into optimizing it with various small things here and there before I narrowed down the issue was with that specific regex (`RE_SOURCE_FILE`). I ended up getting fairly good gains, but not in the ballpark of what I have now. It wasn't until I discovered one very small thing that gave very high gains: there was an extra un-needed wildcard after the symbol name.

```python                                                      
                                       |
                                       V
RE_SOURCE_LINE = r"\s*(?P<sym_name>\S*).* +0x(?P<address> ....
```

So by combining the cmake and source file regexes into 1 it cut the time down in almost half, but I just checked and with this small regex tweak and nothing else changing, even leaving the cmake as a separate regex (with the fix), brings it down to ~1.4 seconds. 

Of course by the time i discovered this I had already made a lot of the other small tweaks whose performance gains became a lot less significant.

I played around with it a bit, commenting out fixes, and wrote a bash script to run it 10 times and print the average run. The results are kinda interesting to see. The first row is the code as it is now, the following rows an `⨯` represents a fix/tweak that I undid for bench-marking. As you can see without the regex fix the pre-filter and early exit made a much larger difference than they do once you include the regex fix.

Pre-Filter | Early Exit | Fixed Regex | Avg Millis
-- | -- | -- | --
✓ | ✓ | ✓ | 375
⨯ | ✓ | ✓ | 561
✓ | ⨯ | ✓ | 505
⨯ | ⨯ | ✓ | 675
✓ | ✓ | ⨯ | 1291
⨯ | ✓ | ⨯ | 4672
✓ | ⨯ | ⨯ | 6023
⨯ | ⨯ | ⨯ | 9294

projectgus: Thanks for the very comprehensive writeup!

projectgus:
@ajcasagrande Thanks for that. I've pushed this into our internal review & merge queue, PR will be updated once it's merged to master.

igrr:
Merged in https://github.com/espressif/esp-idf/commit/874cfda54c24ea43d1b045c5d7c472eac8b5f5bc, thanks @ajcasagrande!

PR #4494, opened 2019-12-16
esp_event_loop_deinit() should declare in esp_event_legacy.h file. (IDFGH-2367)

ankur2854:
I am getting this error when using esp_event_loop_deinit() function. Once I declare in esp_event_legacy.h, the error get resolved. Technically it should be declare in esp_event_legacy.h because the definition of this function is in event_loop_legacy.c

PR #4525, opened 2019-12-19
esp_event_loop_deinit should define in esp_event_legacy.h (IDFGH-2411)

ankur2854:
I am getting this error when using esp_event_loop_deinit() function. Once I declare in esp_event_legacy.h, the error get resolved. Technically it should be declare in esp_event_legacy.h because the definition of this function is in event_loop_legacy.c


igrr:
Per our contribution policy we can only accept PRs targeted at the `master` branch. On request we can also backport the bug fix to the release branches. I will close this PR and keep https://github.com/espressif/esp-idf/pull/4494 open. When that MR is merged, we will backport the fix to v4.0.

PR #4536, opened 2019-12-22
Wifi event sta disconnected reason type (IDFGH-2423)

NidasioAlberto:
Shouldn't the reason for wifi sta disconnection be the enum wifi_err_reason_t?


Alvin1Zhang:
@NidasioAlberto Thanks for the contribution.

esp-HarveyRong:
Hi @NidasioAlberto,
Thank you very much for your contribution. But because the internal code mesh does not get the error reason from the enum `wifi_err_reason_t`, it gets the cause of the error reason from another enum. 
Therefore, the variable `reason` cannot currently be changed to type `wifi_err_reason_t`.

PR #4573, opened 2020-01-02
Fix esp_wifi for C++ linking. (IDFGH-2461)

shaeberling:
Without this, functions such as "esp_netif_create_default_wifi_sta"
will not be found by the C++ linker.


david-cermak:
@shaeberling Thanks for your help in addressing the IDF issues and posting fixes!
This has already been fixed in an internal [commit](https://github.com/espressif/esp-idf/commit/ddf59520c65093ea8ccfeb15e848f8224ead5514) before and we have to close this PR. Thanks again for contributing and sorry for the slow feedback.

PR #4579, opened 2020-01-04
export: put old path to front of new path export (IDFGH-2467)

schmidtandreas:
In case of using of ccache, the entry of ccache have to be on front of
path exports. At least before the toolchain paths. This patch change the
order of new and old paths, so that the old paths are at front. This
avoid disabling of ccache functionality.

Signed-off-by: Andreas Schmidt <mail@schmidt-andreas.de>


igrr:
> In case of using of ccache, the entry of ccache have to be on front of
path exports.

Could you please explain why? This certainly is the case when ccache is used by creating a symlink from xtensa-esp32-elf-gcc to ccache.

However the CMake build system can use ccache even without symlinks, if it detects that ccache is present in the path. You may need to pass --enable-ccache flag to idf.py, or set IDF_CCACHE_ENABLE environment variable.

PR #4597, opened 2020-01-07
Avoid compiler error (IDFGH-2485)

revk:
Compiler errors saying always true in some uses of this macro.

### DIFF
--- a/components/heap/multi_heap_platform.h
+++ b/components/heap/multi_heap_platform.h
@@ -29,14 +29,14 @@ typedef portMUX_TYPE multi_heap_lock_t;
 /* Because malloc/free can happen inside an ISR context,
    we need to use portmux spinlocks here not RTOS mutexes */
 #define MULTI_HEAP_LOCK(PLOCK) do {                         \
-        if((PLOCK) != NULL) {                               \
+        if((PLOCK) != (void*)NULL) {                               \
             portENTER_CRITICAL((PLOCK));                    \
         }                                                   \
     } while(0)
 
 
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @revk,

Thanks for letting us know about this problem and sending the patch.

Can you please give us more context or information about the compiler warning/error you're seeing? 

The C99 standard (GNU99 is the default for IDF builds) provides that when applying an Equality operator (section 6.5.9):

> If one operand is a pointer and the other is a
> null pointer constant, the null pointer constant is converted to the type of the pointer

And the type of the `PLOCK` macro argument should always be `void *`. So it looks like this conversion should always be implicitly applied, and not need the cast.

(I'm not saying we won't merge this patch, I just don't quite understand why it's needed... the warning may be masking some other problem, and casting may not be addressing the root cause.)

Angus

revk:
I was getting this, which stopped build:
components/heap/multi_heap_platform.h:32:20: error: the comparison will always evaluate as 'true' for the address of 'registered_heaps_write_lock' will never be NULL [-Werror=address]
         if((PLOCK) != NULL) {

dobairoland:
The same issue seems to be captured here: https://github.com/espressif/esp-idf/issues/4601

@revk The issue will probably go away after you re-run `install.sh` (or the other install script depending on your system).

revk:
Just tried that and the issue is the same, but I'll see if I can reinstall the tool chain from scratch.

revk:
OK, agreed, that fixed it.

projectgus:
@revk Thanks for letting us know, glad it's working now.

PR #4623, opened 2020-01-12
Change Ubuntu/Debian command to `apt` (IDFGH-2536)

AndrewQuinn2020:
`apt` is the preferred installation method, I gather, from all the recent Ubuntu tutorials I've been reading. Moreover `apt-get` threw a weird error about some `libssh` package that didn't occur for me under `apt`.

### DIFF
--- a/docs/en/get-started/linux-setup.rst
+++ b/docs/en/get-started/linux-setup.rst
@@ -15,7 +15,7 @@ To compile with ESP-IDF you need to get the following packages:
 
 - Ubuntu and Debian::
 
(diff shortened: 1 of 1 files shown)


dobairoland:
Hi @AndrewQuinn2020 thank you for your contribution.

`apt` is high level interface to `apt-get`. If you take a look at the man page of `apt` then you can see that the `install` subcommand is further explained in the man page of `apt-get`. 

`apt` and `apt-get` both should work.  And they should work the same way since we don't have in or documentation any special argument for `install`.

Your `libssh` issue relates probably to your broken `apt-get` package. You could try to resolve that issue by yourself or report it to Ubuntu. 

I don't think that changing `apt-get` to `apt` is necessary for our documentation.

PR #4642, opened 2020-01-17
esp_modem.c: post PPP stop event when requested (IDFGH-2557)

fgiancane8:
When library users call `esp_stop_ppp()`, it is expected that the PPP link
is closed and the DCE is moved into command mode.

If the STOP event is not posted into the event loop, then the PPP stack
would not be turned off, resulting in the impossibility to switch into
command mode, as the modem is still sending PPP binary data.

Signed-off-by: Francesco Giancane <francesco.giancane@accenture.com>

### DIFF
--- a/examples/protocols/pppos_client/components/modem/src/esp_modem.c
+++ b/examples/protocols/pppos_client/components/modem/src/esp_modem.c
@@ -478,6 +478,10 @@ esp_err_t esp_modem_stop_ppp(modem_dte_t *dte)
 {
     modem_dce_t *dce = dte->dce;
     MODEM_CHECK(dce, "DTE has not yet bind with DCE", err);
+    esp_modem_dte_t *esp_dte = __containerof(dte, esp_modem_dte_t, parent);
+
+    /* post PPP mode stopped event */
+    esp_event_post_to(esp_dte->event_loop_hdl, ESP_MODEM_EVENT, ESP_MODEM_EVENT_PPP_STOP, NULL, 0, 0);
     /* Enter command mode */
     MODEM_CHECK(dte->change_mode(dte, MODEM_COMMAND_MODE) == ESP_OK, "enter command mode failed", err);
     /* Hang up */


AxelLin:
@fgiancane8 
Thanks for the fix.
I just tested your fix and it looks fine for me.

Without your fix, the pppos_example has below issue.
I (35807) pppos_example: Calling esp_modem_stop_ppp
E (36047) esp-modem: esp_dte_handle_line(90): handle line failed
W (36047) pppos_example: Unknow line received: ~!E
E (37797) esp-modem: esp_dte_handle_line(90): handle line failed
W (37797) pppos_example: Unknow line received: ~!E
E (38807) esp-modem: esp_modem_dte_send_cmd(221): process command timeout
E (38807) bg96: bg96_set_working_mode(298): send command failed
E (38807) esp-modem: esp_modem_dte_change_mode(311): set new working mode:0 failed
E (38817) esp-modem: esp_modem_stop_ppp(482): enter command mode failed
ESP_ERROR_CHECK failed: esp_err_t 0xffffffff (ESP_FAIL) at 0x4008631c

With your fix, the pppos_example works.

@projectgus  @suda-morris 
Any chance to review the fix?

AxelLin:
@david-cermak 
I think it's commit 25913af2cc12 ("pppos_client: udated example code to use esp-netif in PPP configuration") causes the regression:

-esp_err_t esp_modem_exit_ppp(modem_dte_t *dte)
+esp_err_t esp_modem_stop_ppp(modem_dte_t *dte)
 {
     modem_dce_t *dce = dte->dce;
     MODEM_CHECK(dce, "DTE has not yet bind with DCE", err);
-    esp_modem_dte_t *esp_dte = __containerof(dte, esp_modem_dte_t, parent);
-    /* Shutdown of PPP protocols */
-    MODEM_CHECK(pppapi_close(esp_dte->ppp, 0) == ERR_OK, "close ppp connection failed", err);
     /* Enter command mode */
     MODEM_CHECK(dte->change_mode(dte, MODEM_COMMAND_MODE) == ESP_OK, "enter command mode failed", err);
     /* Hang up */

And on the second though, it looks like calling pppapi_close is better than post PPP stop event.
Post stop event may have racing problem because you don't know if the ppp is  closed *before* calling change_mode.

fgiancane8:
> @david-cermak
> I think it's commit [25913af](https://github.com/espressif/esp-idf/commit/25913af2cc12cfa4667e23d931f45915a22532fe) ("pppos_client: udated example code to use esp-netif in PPP configuration") causes the regression:
> ``
> -esp_err_t esp_modem_exit_ppp(modem_dte_t *dte)
> +esp_err_t esp_modem_stop_ppp(modem_dte_t *dte)
> {
> modem_dce_t *dce = dte->dce;
> MODEM_CHECK(dce, "DTE has not yet bind with DCE", err);
> 
> * esp_modem_dte_t *esp_dte = __containerof(dte, esp_modem_dte_t, parent);
> * /* Shutdown of PPP protocols */
> * MODEM_CHECK(pppapi_close(esp_dte->ppp, 0) == ERR_OK, "close ppp connection failed", err);
>   /* Enter command mode _/
>   MODEM_CHECK(dte->change_mode(dte, MODEM_COMMAND_MODE) == ESP_OK, "enter command mode failed", err);
>   /_ Hang up */
>   ``
>   And on the second though, it looks like calling pppapi_close is better than post PPP stop event.
>   Post stop event may have racing problem because you don't know if the ppp is  closed _before_ calling change_mode.

Hi @AxelLin,

That commit is indeed correct. What has happened is that the previous `pppos-client` example was implemented as a standalone module, while now it is using the the `esp-netif` which required those modification.

If you look at the example, there is a https://github.com/espressif/esp-idf/blob/master/examples/protocols/pppos_client/components/modem/src/esp_modem_compat.c compatibility interface reimplementing the previous methods but bridged to the new interface.

My suggestion would be not to revert the commit you are indicating, but just adding my fix.

david-cermak:
@fgiancane8 Thanks for the fix, LGTM! Taking it for the internal review.

@AxelLin Thanks for noting, yes, I missed this part in the `esp_modem_stop_ppp()`.

fgiancane8:
@david-cermak @AxelLin 

Hi guys,
just to open the discussion. I was thinking that may be interesting moving the `esp-modem` inside the main `esp-idf` `components` folder. It should be more maintainable

What do you think?

AxelLin:
> @david-cermak @AxelLin
> 
> Hi guys,
> just to open the discussion. I was thinking that may be interesting moving the `esp-modem` inside the main `esp-idf` `components` folder. It should be more maintainable
> 
> What do you think?

Hi @fgiancane8 
That's a good suggestion.
However current pppos quality is just good enough as example since it's working, it is not good enough for commercial use. 
I would suggest to fix the *issues* first, especially in taking care of signal lost and reconnect issues.
https://github.com/espressif/esp-idf/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+pppos

fgiancane8:
> > @david-cermak @AxelLin
> > Hi guys,
> > just to open the discussion. I was thinking that may be interesting moving the `esp-modem` inside the main `esp-idf` `components` folder. It should be more maintainable
> > What do you think?
> 
> Hi @fgiancane8
> That's a good suggestion.
> However current pppos quality is just good enough as example since it's working, it is not good enough for commercial use.
> I would suggest to fix the _issues_ first, especially in taking care of signal lost and reconnect issues.
> https://github.com/espressif/esp-idf/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+pppos

Hi @AxelLin ,

I think it is a good point. I am using the PPPoS protocol in a development project and apart from adapting the AT driver for the modem we are using, I can confirm that results are very good.

However, as you are pointing out, code still needs to be cleaned (and in particular, what I needed to manually add is the negotiation with the modem at boot time, since now it is assumed that by the time you are running the code, modem is already initialized and network attached, which is not always the case).

I am willing to contribute, if you need. Maybe we can share a roadmap and work together.

Thanks,
Francesco

igrr:
Thanks again @fgiancane8, this was cherry-picked as https://github.com/espressif/esp-idf/commit/d8755fff3fddd1f05df56173ad6f9549bde985c5.

PR #4653, opened 2020-01-21
Fix modem stuck during ppp exit function and exception if starting ppp then (IDFGH-2568)

d3vil-st:
Before calling pppapi_close function need to check if connection is already dead (e.g. due to poor signal), cause it may cause MCU to stuck in that state. Reproducible if after successful PPP connection pull-out SIM card, wait for disconnect and try to call this function.

Also there is need to free up resources after stopping, cause esp_modem_setup_ppp function will face the exception somewhere in dhcp module that cause MCU reboot.

### DIFF
--- a/examples/protocols/pppos_client/components/modem/src/esp_modem.c
+++ b/examples/protocols/pppos_client/components/modem/src/esp_modem.c
@@ -616,7 +616,10 @@ esp_err_t esp_modem_exit_ppp(modem_dte_t *dte)
     MODEM_CHECK(dce, "DTE has not yet bind with DCE", err);
     esp_modem_dte_t *esp_dte = __containerof(dte, esp_modem_dte_t, parent);
     /* Shutdown of PPP protocols */
-    MODEM_CHECK(pppapi_close(esp_dte->ppp, 0) == ERR_OK, "close ppp connection failed", err);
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@d3vil-st Thanks for the contribution.

AxelLin:
@d3vil-st 
FYI: This does not apply to current master tree.

david-cermak:
Hi @d3vil-st 

Could you please explain a bit, how this issue could be reproduced? I've been playing with this and it seems to work as expected, even in the dead phase. And indeed it should, since the `ppp_close()` simply checks the phase in exits if in the `PPP_PHASE_DEAD`

https://github.com/espressif/esp-lwip/blob/2.1.2-esp/src/netif/ppp/ppp.c#L344-L347

Of course this, is called from the tcpip thread, so might be another phase (a network recovery perhaps) enters after the dead phase check in the main thread, but still, it might longer a bit in the close function and exit and free the ppp all the same. (we may still call this with `nocarrier=1` to skip the shutdown procedure). Any thought here?

Alvin1Zhang:
Thanks for your contribution, @d3vil-st  do you have any updates regarding @david-cermak 's comments? Thanks.

Alvin1Zhang:
Thanks for your contribution again, feel free to reopen.

PR #4667, opened 2020-01-25
bluedroid: Replace U+0092 with regular apostrophes in avrc_defs.h (IDFGH-2583)

ulfalizer:
avrc_defs.h has some scattered U+0092 characters in it for apostrophes.
Replace them with standard U+0027 apostrophes.

The U+0092 apostrophes somehow caused invalid UTF-8 in
https://github.com/zephyrproject-rtos/esp-idf, though it seems to be
fine here.

### DIFF
--- a/components/bt/host/bluedroid/stack/include/stack/avrc_defs.h
+++ b/components/bt/host/bluedroid/stack/include/stack/avrc_defs.h
@@ -47,21 +47,21 @@
 
 /* command type codes */
 #define AVRC_CMD_CTRL       0   /* Instruct a target to perform an operation */
-#define AVRC_CMD_STATUS     1   /* Check a devices current status */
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for reporting and sorry for the slow turnaround. Sorry that we are now during the holiday of Chinese Lunar New Year. Will update to you when we are back, thanks.

AbnerFederer:
Hi, @ulfalizer 

Thanks for your code.

But unfortunately, we have fixed the same annotation error in one MR in our internal repo. 

Due to the Chinese Spring Festival, the sync. into GitHub is slower than usual.

Thanks again for contributing to this PR, and we will close the PR and notify you when the fix is on GitHub.

Thanks
Abner

Alvin1Zhang:
@ulfalizer Thanks for your contribution.

Issue #4095, opened 2019-09-19
Missing libusb dependency on ubuntu:latest (IDFGH-1893)


mirronelli:
- IDF version: v4.1-dev-281-g96b96ae24
- Operating System:     Linux

## Problem Description

The toolchain setup on a ubuntu:latest container doesn't produce a working build environment

### Expected Behavior

- export.sh in esp-idf directory does not display errors and warnings
- OpenOcd tool works

### Actual Behavior
export.sh throws this error
```WARNING: tool openocd-esp32 version v0.10.0-esp32-20190708 is installed, but the tool failed to run```

trying to start openocd directly displays:
```root@d483f5c351af:~/.espressif/tools/openocd-esp32/v0.10.0-esp32-20190708/openocd-esp32/bin# ./openocd
./openocd: error while loading shared libraries: libusb-1.0.so.0: cannot open shared object file: No such file or directory
```

### Steps to repropduce

```
docker run -it ubuntu:latest
apt-get update
apt-get -y install git wget libncurses-dev flex bison gperf python python-pip python-setuptools python-serial python-click python-cryptography python-future python-pyparsing python-pyelftools cmake ninja-build ccache
mkdir /root/esp
cd /root/esp
git clone --recursive https://github.com/espressif/esp-idf.git
cd /root/esp/esp-idf
./install.sh
. ./export.sh 
```

### Steps to fix
apt-get install libusb-1.0

### Suggested change
add the libusb dependency to the documentation
```
apt-get -y install git wget libncurses-dev flex bison gperf python python-pip python-setuptools python-serial python-click python-cryptography python-future python-pyparsing python-pyelftools cmake ninja-build ccache  libusb-1.0
```

PerMalmberg:
I had this issue too so it's not local to @mirronelli's system.

KaeLL:
I'm having this exact problem too running the `v4.1-beta1` snapshot.
```
WARNING: tool openocd-esp32 version v0.10.0-esp32-20191114 is installed, but the tool failed to run
ERROR: tool openocd-esp32 has no installed versions. Please run 'install.sh' to install it.
```

cjds:
I couldn't find the package but I did

```
apt-get install libusb-1.0-0-dev
```

on Ubuntu 20.04

projectgus:
It should be enough to do `apt install libusb-1.0-0` (no `-dev` needed) on Debian/Ubuntu.

A fix to the docs is in review now, this issue will be updated when it merges.

[...]

PR #4669, opened 2020-01-26
Add libusb dependency for Ubuntu (IDFGH-2585)

mjdr:
I'm building development environment in Docker.
And I have same issue like here https://github.com/espressif/esp-idf/issues/4095
So I fixed it.

### DIFF
--- a/docs/en/get-started/linux-setup-scratch.rst
+++ b/docs/en/get-started/linux-setup-scratch.rst
@@ -19,7 +19,7 @@ To compile with ESP-IDF you need to get the following packages:
 
 - Ubuntu and Debian::
 
-    sudo apt-get install git wget libncurses-dev flex bison gperf python python-pip python-setuptools python-serial python-click python-cryptography python-future python-pyparsing python-pyelftools cmake ninja-build ccache libffi-dev libssl-dev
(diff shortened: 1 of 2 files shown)


dobairoland:
Hi @mjdr. The tricky part about these recommended packages is that they might be very different for various supported Ubuntu/Debian releases. I personally don't like adding version-specific packages to the list because it may cause failure for other developers. 

Another argument is that openocd is not a strict requirement for ESP-IDF. One can develop and flash chips without using openocd.

Also there is an ESP-IDF specific Docker image (https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/tools/idf-docker-image.html) where the right libusb package is already included: https://github.com/espressif/esp-idf/blob/master/tools/docker/Dockerfile#L18.

mjdr:
Yes, that is make sense.
Thanks. you

projectgus:
@mjdr We've reconsidered this, as a number of users are still finding the error mesage from `export.sh` confusing. Have cherry-picked your commit from this pull request into a new internal branch, this PR will be updated once it merges.

Thanks again for sending us this fix, even though it's taken a while to be picked up. :)

PR #4674, opened 2020-01-27
workaround for usbser.sys in idf_monitor (IDFGH-2590)

rousveiga:
Implemented workaround in idf_monitor for usbser.sys known bug which causes changes in RTS signal to only be sent if DTR signal changes. Works by forcing a dummy change in DTR signal (re-assigning its current value, as it has been previously done for the flashing tool).
As the ESP-IDF programming guide says that contributors who are unsure about the viability of a pull request should open it anyway and ask for feedback, I'd like to request some.


rousveiga:
I am currently having email issues with the account with which I made the commit; I will close the pull request so that I can repeat the process solving such issues.

PR #4676, opened 2020-01-27
workaround in idf_monitor for usbser.sys bug (IDFGH-2592)

rousveiga:
Implemented workaround in idf_monitor for known usbser.sys bug which causes changes in RTS signal to only be sent if DTR signal also changes. Works by forcing a dummy change in DTR signal (re-assigning its current value, as it has previously been done in the flashing tool).
As the ESP-IDF programming guide says that contributors who are unsure about the viability of a pull request should open it anyway and ask for feedback, I'd like to request some.

### DIFF
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -355,6 +355,7 @@ class SerialReader(StoppableThread):
             self.serial.rts = True  # Force an RTS reset on open
             self.serial.open()
             self.serial.rts = False
+            self.serial.dtr = self.serial.dtr   # usbser.sys workaround
         try:
             while self.alive:
                 data = self.serial.read(self.serial.in_waiting or 1)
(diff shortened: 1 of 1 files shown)


dobairoland:
Thank you @rousveiga for your contribution.

@pacucha42 PTAL

dobairoland:
@rousveiga Could you please describe the issue you were facing and how does this fix solved it (use case + logs). We would like to reproduce the issue if it possible.

rousveiga:
The Windows driver usbser.sys for USB CDC devices has a bug which causes changes in the RTS signal to not be sent until changes in the DTR signal occur, so it is not possible to reset or put into bootloader mode boards connected through an adapter that uses this driver. I did not draw information from any logs; idf_monitor.py would simply not cause a reset when launched. Checking the status of the DTR and RTS lines with an oscilloscope was what helped me diagnose the issue. 

esptool.py features a similar workaround for the flashing case.

pacucha42:
as the code change follows the same logic as in https://github.com/espressif/esptool/pull/290, I would consider it safe.
@rousveiga thank you.

Martin

rousveiga:
Thank you.

PR #4680, opened 2020-01-27
Build fix ESP_BLE_MESH. (IDFGH-2596)

Niram7777:
Revert partially commit f58d7d14c7e264375b6824dceb742f553b51269c to
define ADV_TASK_CORE if CONFIG_BT_{BLUEDROID,NIMBLE}_ENABLED are not
set.


Alvin1Zhang:
Thanks for reporting and sorry for the slow turnaround. Sorry that we are now during the holiday of Chinese Lunar New Year. Will update to you when we are back, thanks.

PR #4684, opened 2020-01-28
Fix overriding of PYTHON variable (IDFGH-2602)

anthony-smartnora:
Making this variable existentially conditional allows for overriding `PYTHON` and `CONFIG_SDK_PYTHON` without side effects

### DIFF
--- a/make/common.mk
+++ b/make/common.mk
@@ -2,7 +2,7 @@
 # and component makefiles (component_wrapper.mk)
 #
 
-PYTHON=$(call dequote,$(CONFIG_SDK_PYTHON))
+PYTHON ?= $(call dequote,$(CONFIG_SDK_PYTHON))
 
 # Include project config makefile, if it exists.
 #


dobairoland:
Thank you @anthony-smartnora for your contribution!

dobairoland:
Hello @anthony-smartnora. I'm sorry to tell you that we have to reject this change because it causes significant behavior change and the `make` build system is already a legacy build system since v4.0 and we are not planning to make big changes in it. The `PYTHON` variable is intended to be internal from the beginning which can be overwritten from menuconfig if it is necessary.  

Note that you can set `CONFIG_SDK_PYTHON` permanently in `sdkconfig.defaults` of your project if this helps.

Another good option is to create a Python environment for your builds (for example https://pypi.org/project/virtualenv/).

If you describe your use case and why do you need this change then maybe we can recommend you some better option.

PR #4723, opened 2020-02-06
Fix subtype in PartitionTable.find_by_type (IDFGH-2646)

boarchuz:
I was getting odd "Partition not found" results from parttool.py. Tracked it down to this line in gen_esp32part.py where 'ptype' has obviously not been updated to 'subtype'. Results are now as expected with the fix.


dobairoland:
Thank you @boarchuz for your contribution. We will review this internally.

Issue #4700, opened 2020-02-01
Support for using noinit variables in SPIRAM (IDFGH-2621)


devanlai:
**Is your feature request related to a problem? Please describe.**

I have an ESP-WROVER module with 8MB PSRAM, which I use to store large logs and debug traces that would not fit in the internal RAM with `CONFIG_SPIRAM_USE_MALLOC` enabled so that I can allocate buffers from SPIRAM. This works great, except that unlike regular internal memory, there's no way for me to statically allocate them and flag them as `noinit` so that they can be recovered after a software reset.

**Describe the solution you'd like**

It would useful to have an attribute macro like `__NOINIT_ATTR` that would work to place the noinit variable in the SPIRAM. I've done some testing to prove out the idea and I ended up making the following changes:

- Add a new linker script to the `esp32` component that defines a new linker section that is placed in the `extern_ram_seg` segment and provides start and end symbols.
- Add a new macro to `esp_attr.h` to place variables in the above linker section.
- Update `esp_spiram_test()` to skip over addresses that fall within the new SPIRAM noinit section.
- Update `esp_spiram_add_to_heapalloc()` to take into account the end of the new linker section when determining where available memory for the heap starts.

If the Espressif team is open to these kind of changes, I could make a pull request with the specific code.

Alvin1Zhang:
Thanks for reporting and sorry for the very slow turnaround, fix is available https://github.com/espressif/esp-idf/commit/0dad76329f9d8e98c02f478026bbf68f25bf0838, feel free to reopen.

PR #4728, opened 2020-02-07
Add support for using noinit variables in SPIRAM (IDFGH-2654)

devanlai:
As per #4700, this pull request add support for placing non-initialized variables in SPIRAM.

This includes the following changes:

- There's a new Kconfig option `SPIRAM_ALLOW_NOINIT_EXTERNAL_MEMORY` which enables the feature.
- When the feature is enabled, `EXT_RAM_NOINIT_ATTR` can be used to place variables into a new non-initialized section in SPIRAM. If `SPIRAM_ALLOW_NOINIT_EXTERNAL_MEMORY` is disabled, it attempts to place it in regular internal RAM instead.
- At startup, the SPIRAM memory test, if enabled, skips over any noinit sections.
- The SPIRAM noinit section is excluded from the SPIRAM memory made available to the heap allocator.

I've tested this internally and I've found it works pretty well (with the caveat that it's important to flush the SPIRAM write-back cache before restarting to ensure everything has been written back) and I've added a simple unit test for it.

Caveats:

- No ESP32S2 support, since I don't have any way to test it.
- No documentation yet

### DIFF
--- a/components/esp32/CMakeLists.txt
+++ b/components/esp32/CMakeLists.txt
@@ -60,6 +60,11 @@ else()
         target_linker_script(${COMPONENT_LIB} INTERFACE "ld/esp32.extram.bss.ld")
     endif()
 
+    if(CONFIG_SPIRAM_ALLOW_NOINIT_EXTERNAL_MEMORY)
+        # This has to be linked before esp32.project.ld
+        target_linker_script(${COMPONENT_LIB} INTERFACE "ld/esp32.extram.noinit.ld")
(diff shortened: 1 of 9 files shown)


Alvin1Zhang:
Thanks for the contribution.

Alvin1Zhang:
Thanks for your contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/0dad76329f9d8e98c02f478026bbf68f25bf0838.

PR #4750, opened 2020-02-11
Fix standard NEC IR transmission (IDFGH-2676)

Dee-Fuse:
There are some oversights in the RMT IR example that might break the functionallity of it.
One problem were the blocking RMT writes which elongated the delay between message and repeat frame to increase.
The other issue had to do with integer promotion which prevented the use of the utility in non-extended mode.


Alvin1Zhang:
Thanks for the contribution.

Alvin1Zhang:
Thanks for the contribution.

suda-morris:
@Dee-Fuse Thanks for your contributions, we will merge it ASAP.

suda-morris:
Thanks for your contribution! we have merged this PR. See in https://github.com/espressif/esp-idf/commit/c788351c946caad3a8c106782f9bf4b60a2b2fbd

PR #4772, opened 2020-02-16
lwip: fix bug for crash when socket lock not initialized (IDFGH-2701)

nebkat:
Update lwip to include fix for https://github.com/espressif/esp-lwip/pull/11


nebkat:
Already merged.

PR #4782, opened 2020-02-17
lwip: make IPV6 link-local support over PPP configurable (IDFGH-2712)

fgiancane8:
Make the link local negotiation for IPV6 in PPP optional and
configurable.
This is because some modems do not support the IPV6 negotiation and
sending IPV6CP frames would in some cases break the network
configuration phase, resulting in a timeout during the Phase Network.

Please note that this does not disable the IPV6 support for the outgoing
communication (IPV6 is still enabled even if this option is not
selected) but just for the local link between lwIP and modem.

I found this issue at least on the nB-IoT series of modem from SIMCOM and probably also with other products sharing the same platform.


Alvin1Zhang:
@fgiancane8 Thanks for the contribution.

fgiancane8:
@Alvin1Zhang 
thank you for the fast comment :)

Let me know if after review, there is something to be fixed or rephrased.

Francesco

fgiancane8:
@Alvin1Zhang 
Is there any update about the status of this PR?

Alvin1Zhang:
@fgiancane8 Thanks for the contribution and sorry for the late reply. We are reviewing the PR and will share updates once available. Thanks.

david-cermak:
@fgiancane8 Thanks again for posting this code -- very useful indeed! In addition it seems that the current pppos example doesn't correctly handle IPv6 addresses.

The change LGTM, the only additional suggestion I might have would be to add a line to the `sdkconfig.defaults` disabling this option for the pppos_client example in IDF. Not directly related to this PR, so feel free to ignore it, I can then add a separate commit updating the example. 

Another minor point is formatting of Kconfig, please run the `tools/check_kconfigs.py` and fix the warnings (long lines).

fgiancane8:
Hi @david-cermak ,

Wonderful! It is not a problem for me to fix both the sdkconfig and the Kconfig formatting.

> @fgiancane8 Thanks again for posting this code -- very useful indeed! In addition it seems that the current pppos example doesn't correctly handle IPv6 addresses.
I tested the pppos_client with SIMCOM 800 (2G ) and 7020 (nB-IoT). I did not have the need to disable IPV6 and IPV6 link-local negotiation with the 800, but only with the 7020. Unfortunately, the modem always negotiated an IPv4 address so I could not verify whether the IPV6 was working or not.

On the other hand, PPP on SIMCOM 7020 failed if requested for IPV6 addressing (IPV6CP). So please let me help understand better the problem so we can also fix the example.
> 
> The change LGTM, the only additional suggestion I might have would be to add a line to the `sdkconfig.defaults` disabling this option for the pppos_client example in IDF. Not directly related to this PR, so feel free to ignore it, I can then add a separate commit updating the example.
> 
> Another minor point is formatting of Kconfig, please run the `tools/check_kconfigs.py` and fix the warnings (long lines).

Will update the PR as soon as possible.

Thanks!

fgiancane8:
@david-cermak ,
> 
> Wonderful! It is not a problem for me to fix both the sdkconfig and the Kconfig formatting.
> 

PR updated.

david-cermak:
@fgiancane8 Thanks for the quick update! LGTM! This PR is then ready for merging (I'still might squash the two lwip related commits though).

> So please let me help understand better the problem so we can also fix the example.

There is yet another issue. Even if the IPv6 gets assigned correctly, the pppos example wouldn't properly receive the IP event and print the address. which means that the example would work only with modems using IPv4 only and it makes little sense to enable IPV6 config for that case. The only sensible solution is in my opinion is what you've pushed as https://github.com/espressif/esp-idf/pull/4782/commits/d6c316e3bb11d7436f7b1052a83c7088d5635966.
On top of that I've added one extra commit that prevents posting IP_EVENT when IPv6 address received.

fgiancane8:
@david-cermak ,

Many thanks for the feedback. I think that for the time being, we can be happy like that and continue testing and investigating.

Unfortunately, all the modem I have here do not mention supporting the IPV6 in the link-local PPP serial port, so I do not know how to help investigating better.

We need to understand if it is an lwIP-related problem or just the glue-code between lwIP and PPP and esp-modem.

iosix:
Just FYI, I backported this to 3.3.1 as we had a similar issue with SIM7000G (M1+NB1+2G) - the M1 and NB1 work just fine with IPV6CP - it locks up trying to connect most times on 2G only, with latest modem firmware only (used to work fine) - but even if I can get an IP I still can not communicate through the modem. I think this issue is a modem-side bug therefore. 


Robert Vogt IV
CEO
IOSiX, LLC
1300 Tefft Ct #1
Saline, MI  48176
P: +1-855-OBD-1939
C: +1-734-730-9690
robert@iosix.com

fgiancane8:
> Just FYI, I backported this to 3.3.1 as we had a similar issue with SIM7000G (M1+NB1+2G) - the M1 and NB1 work just fine with IPV6CP - it locks up trying to connect most times on 2G only, with latest modem firmware only (used to work fine) - but even if I can get an IP I still can not communicate through the modem. I think this issue is a modem-side bug therefore. Robert Vogt IV CEO IOSiX, LLC 1300 Tefft Ct #1 Saline, MI 48176 P: +1-855-OBD-1939 C: +1-734-730-9690 robert@iosix.com
> […](#)
> On Feb 25, 2020, at 8:16 AM, Francesco Giancane ***@***.***> wrote: ﻿ @david-cermak , Many thanks for the feedback. I think that for the time being, we can be happy like that and continue testing and investigating. Unfortunately, all the modem I have here do not mention supporting the IPV6 in the link-local PPP serial port, so I do not know how to help investigating better. We need to understand if it is an lwIP-related problem or just the glue-code between lwIP and PPP and esp-modem. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

Hello,
Thanks for your feedback. I also had no problem with the SIM 800C (which is a module 2g only, but different platform - it is Qualcomm based) and no issue at all with IPV6CP. I too think it is a modem-side bug.

fgiancane8:
Hi,
I am closing this pull requested since it has been merged.

PR #4789, opened 2020-02-17
install.sh: use python3 (IDFGH-2718) (IDFGH-2719)

szutenberg:
In some cases python3 is not default python library so probably it's better to use python3.


Alvin1Zhang:
@szutenberg Thanks for the contribution.

szutenberg:
That wasn't a problem with python2. The problem was fixed a few days ago but I used the stable version which wasn't patched. Sorry for the confusion.

PR #4796, opened 2020-02-18
Fixed rights for entrypoint.sh (IDFGH-2727)

aureliokessler:
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"/opt/esp/entrypoint.sh\": permission denied": unknown.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -61,7 +61,7 @@ RUN $IDF_PATH/tools/idf_tools.py --non-interactive install required \
 # Ccache is installed, enable it by default
 ENV IDF_CCACHE_ENABLE=1
 
-COPY entrypoint.sh /opt/esp/entrypoint.sh
+COPY entrypoint.sh $IDF_TOOLS_PATH/entrypoint.sh
 RUN chmod u+x $IDF_TOOLS_PATH/entrypoint.sh
 
 ENTRYPOINT [ "/opt/esp/entrypoint.sh" ]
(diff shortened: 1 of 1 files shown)


igrr:
You are building the docker image on Windows, right?

tools/docker/Dockerfile
```diff
@@ -62,6 +62,7 @@ RUN $IDF_PATH/tools/idf_tools.py --non-interactive install required \
 ENV IDF_CCACHE_ENABLE=1
 
 COPY entrypoint.sh /opt/esp/entrypoint.sh
+RUN chmod u+x $IDF_TOOLS_PATH/entrypoint.sh
```
igrr: Please either change the line above to use $IDF_TOOLS_PATH instead of /opt/esp, or change this line to use /opt/esp instead of $IDF_TOOLS_PATH, to make them consistent.

aureliokessler: That makes sense and has been adjusted. Thank you @igrr

aureliokessler:
> You are building the docker image on Windows, right?

No, with macOS Catalina with the VSCode extension Docker from Microsoft.

igrr:
That's odd then... entrypoint.sh already has executable permissions in Git, and if you are working on macOS the permissions shouldn't be getting lost when you are checking out the project. Can you please do `ls -l tools/docker/entrypoint.sh` to confirm the permissions?

aureliokessler:
I don't have a local git repository from esp-idf. Only that in the Docker image.

I copied the code from the files in the repository and created the files for me locally. Since I just wanted the whole esp-idf in a Docker container and didn't distribute everything on my system so I have to clean it in several places, I can simply delete the Docker image and it's good.

PR #4823, opened 2020-02-23
http authentication in esp_http_server (IDFGH-2757)

jaghatei:
This pull provides http authentication to the esp_http_server.
Function for basic auth is included. Other auth functions can be provided by user.
Normal URI handler will only be called in case authentication is disabled or a valid user id is found and authorized.
Authentication is disabled if: AuthType HTTP_AUTH_NONE or UserPassFn == NULL.
AuthType HTTP_AUTH_BASIC uses internal auth function. 
AuthType HTTP_AUTH_USER uses user provided function if reference is provided in field UserAuthFn.

Configuration is handled via uri config:
static const httpd_uri_t favicon = {
		.uri       = "/favicon.ico",
		.method    = HTTP_GET,
		.handler   = favicon_get_handler,
		.user_ctx  = NULL,
		.UserPassFn = myPassFn,
		.UserAuthFn = NULL,
		.AuthType = HTTP_AUTH_BASIC,
		.auth_realm = "yourhttprealm"
};

handler function can check result of authentication in req->auth_user_id.
-1: auth disabled
values > = 0: id of authenticated user
Results -2 ( no auth provided by client) and -3 (no user authenticated) of internal basic auth will close request and not call handler function.
Your own UserAuthFn may react different here but needs to provide return values >=-1 then.


Alvin1Zhang:
@jaghatei Thanks for your contribution.

xkevin190:
is this ready? What is missing to implement?

jaghatei:
Frm my point this is ready for merge. But this needs to be done by "authorized" users...

MihirShahLNM:
I am using IDF version: v4.2-dev-1905-g625bd5eb1-dirty

As I reviewed the  **httpd_uri_t** structure, it only has four parameters ....

```
static const httpd_uri_t favicon = {
.uri = "/favicon.ico",
.method = HTTP_GET,
.handler = favicon_get_handler,
.user_ctx = NULL
};
```

The following parameters are missing ...
```
.UserPassFn = myPassFn,
.UserAuthFn = NULL,
.AuthType = HTTP_AUTH_BASIC,
.auth_realm = "yourhttprealm"
```

MihirShahLNM:
I get the error: **implicit declaration of function** 'mbedtls_base64_decode'

```
C:/esp-idf/components/esp_http_server/src/httpd_uri.c: In function 'httpd_basic_auth':
C:/esp-idf/components/esp_http_server/src/httpd_uri.c:305:3: error: implicit declaration of function 'mbedtls_base64_decode' [-Werror=implicit-function-declaration]
   mbedtls_base64_decode((unsigned char*)&userpass,sizeof(userpass),(size_t*)&r,(const unsigned char*)&hdr +6,strlen(hdr)-6);
   ^~~~~~~~~~~~~~~~~~~~~
```

MihirShahLNM:
Further, if you do not specify following in the configuration & mcu panics & restarts !
```
.UserPassFn = myPassFn,
.UserAuthFn = NULL,
.AuthType = HTTP_AUTH_BASIC,
.auth_realm = "yourhttprealm"
```

shahpiyushv:
@jaghatei thanks for the contribution and sorry for the late reply. Can you have a look at our [contribution guidelines](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/index.html) and make appropriate changes. A couple of key points are:

- The nomenclature should be inline with what is already used (Eg. user_pass_fn instead of UserPassFn).
- An example to show how this should be used will be useful.
- Better to keep this feature disabled by default.

I will provide other comments on the actual code separately.

jaghatei:
@shahpiyushv: will look into these points
@MihirShahLNM: 
will check for embedtls include.
But for the mcu panic: 
Does this happen when HTTPD_AUTH_SUPPORT is Y or N on Kconfig? 
It should not when set to N.

Lisa999:
@jaghatei : Any progress yet?
Would be nice to have this...

Lisa999:
/home/lisa/.espressif/tools/xtensa-esp32-elf/esp-2020r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/bin/ld: esp-idf/main/libmain.a(restful_server.cpp.obj):/home/lisa/esp/esp-idf/components/esp_http_server/include/esp_http_server.h:83: multiple definition of `UserAuthFn'; esp-idf/main/libmain.a(mdns_server.cpp.obj):/home/lisa/esp/esp-idf/components/esp_http_server/include/esp_http_server.h:83: first defined here
/home/lisa/.espressif/tools/xtensa-esp32-elf/esp-2020r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/bin/ld: esp-idf/main/libmain.a(restful_server.cpp.obj):/home/lisa/esp/esp-idf/components/esp_http_server/include/esp_http_server.h:82: multiple definition of `UserPassFn'; esp-idf/main/libmain.a(mdns_server.cpp.obj):/home/lisa/esp/esp-idf/components/esp_http_server/include/esp_http_server.h:82: first defined here

When using mdns server also...

lhespress:
@jaghatei 

I don't think we need modify both httpd_uri.c and  esp_http_server.h file for the function, it's better to add it by user.

Please reference the code of https://github.com/espressif/esp-idf/issues/5646.

AxelLin:
> @jaghatei
> 
> I don't think we need modify both httpd_uri.c and esp_http_server.h file for the function, it's better to add it by user.
> 
> Please reference the code of #5646.

@lhespress 

The http_auth_basic in components/esp_http_client/lib/include/http_auth.h
is in PRIV_INCLUDE_DIRS of esp_http_client component.
So the application code cannot re-use the http_auth_basic() function.
e.g. In https://github.com/espressif/esp-idf/issues/5646#issuecomment-734765068
it implements http_auth_basic again.

Suggest to move http_auth_basic out of PRIV_INCLUDE_DIRS.

lhespress:
@AxelLin You can configure the information as follows when call `esp_http_client_init`

```
const char                  *username;           /*!< Using for Http authentication */
const char                  *password;           /*!< Using for Http authentication */
esp_http_client_auth_type_t auth_type;           /*!< Http authentication type, see `esp_http_client_auth_type_t` */
```
or  change it by 
```
esp_http_client_set_username
esp_http_client_set_password
esp_http_client_set_authtype
```

AxelLin:
> @AxelLin You can configure the information as follows when call `esp_http_client_init`

The thing is to use http_auth_basic() for httpd application, not for http client.

AxelLin:
> @AxelLin I still suggest you reference the code [#5646 (comment)](https://github.com/espressif/esp-idf/issues/5646#issuecomment-734765068) for httpd application.

I have no problem to implement it by myself.
I just point out the code http_auth_basic() can be re-used by both http-client and http-server.

lhespress:
@AxelLin Thanks for your understanding, I suggest it because of  if we make HTTP client API public then "HTTP server application" will create additional dependency on "HTTP client" component, which does not look very good. Of course, we will do it if we have a component which share some common utilities between HTTP client and server applications.

baldhead69:
Hi,

The "suggestion code" from @jaghatei seems very interesting to me.

Why hasn't it been implemented yet ?
This seems to me a basic requirement for a server.

There are some problems too.
Register users dynamically, password reset(in a local network) if the user forgets the password for example, etc.

Don't forget to consider authentication for websocket(WS) and websocket secure(WSS) too.

I open the feature request here:
https://github.com/espressif/esp-idf/issues/7566
https://github.com/espressif/esp-idf/issues/7431

Thank's.

baldhead69:
Hi,

Could someone suggest the best way to store users and passwords in flash memory ?
I would like to save this data to flash memory encrypted if it's not too complicated.
Before encrypted this data:
user: plain text
password: password hash + salt, (i still don't know how i will do it).

Fat Fs ?
Fat Fs with wear leveling ?
SPIFFS ?
SPIFFS with wear leveling ?
Other ?

Thank's.

lhespress:
@jaghatei Thanks for your contribution, but I suggest handle it by application itself.

PR #4846, opened 2020-02-26
Use C++20 by default instead of C++11 (IDFGH-2781)

biodranik:
The toolchain uses the recent gcc 8.2 which supports C++17. There are many useful features for embedded development like string_view, constexpr lambda and if, binary and utf8 literals, std::optional, many template and syntax improvements.

### DIFF
--- a/make/project.mk
+++ b/make/project.mk
@@ -467,7 +467,7 @@ CFLAGS := $(strip \
 CXXFLAGS ?=
 EXTRA_CXXFLAGS ?=
 CXXFLAGS := $(strip \
-	-std=gnu++11 \
+	-std=gnu++17 \
 	$(OPTIMIZATION_FLAGS) $(DEBUG_FLAGS) \
 	$(COMMON_FLAGS) \
 	$(COMMON_WARNING_FLAGS) \
--- a/tools/cmake/build.cmake
+++ b/tools/cmake/build.cmake
@@ -113,7 +113,7 @@ function(__build_set_default_build_specifications)
     list(APPEND c_compile_options   "-std=gnu99"
                                     "-Wno-old-style-declaration")
 
-    list(APPEND cxx_compile_options "-std=gnu++11")
+    list(APPEND cxx_compile_options "-std=gnu++17")
 
     idf_build_set_property(COMPILE_DEFINITIONS "${compile_definitions}" APPEND)
     idf_build_set_property(COMPILE_OPTIONS "${compile_options}" APPEND)


igrr:
Hi @biodranik, unfortunately we can't change the default standard from C++11 to C++17 in a minor release of IDF, since we try not to make breaking changes between minor releases. As you know, in addition to all the useful new features, C++17 also removes a number of older C++ features.

If you would like to use C++17 features in your code, you can add -std=c++17 via target_compile_options command in your component makefiles.

It has also been proposed to add a Kconfig option for setting the default standard. This would be a non-breaking change, but this would place additional burden upon us, as we would need to ensure that IDF code builds/works correctly with all possible standard options. Perhaps this approach is worth revisiting.

We will definitely schedule an update of the default C++ and C standards in a future major release of IDF. It might be C++20 by that time.

biodranik:
@igrr 
1. What is the right way to test all your code for C++17 compatibility? If it compiles, when there is no reason to delay the upgrade. So far, I tried to build my project with several components enabled, and everything works.
2. Removed features are the minor ones, and are easily fixable.
3. Mixing different standards of the C++ compiler is a bad idea, at least size-wise.
4. I don't think that Kconfig option is a good idea; it will increase the complexity and make support harder.
5. What do you count as a major release, and when (approximately) it can happen?

igrr:
> What is the right way to test all your code for C++17 compatibility? If it compiles, when there is no reason to delay the upgrade.

For the nvs_flash and wear_levelling components (the only two at the moment where we use C++), we have a set of unit and integration tests which we could run. I doubt there will be any issues, as we were using a very restricted set of C++ features, and certainly none which were deprecated/removed.

> If it compiles, when there is no reason to delay the upgrade.
> Removed features are the minor ones, and are easily fixable.

I can understand your perspective. But what is an insignificant change or a minor inconvenience for one user may be a bigger issue for another. Especially if they are using some legacy or third-party C++ codebase in their project, which they don't understand well, and can not easily fix/upgrade to the new standard. So I maintain the opinion that breaking changes should be avoided.

> Mixing different standards of the C++ compiler is a bad idea, at least size-wise.

I know that sometimes there are ABI differences between versions of the STL, but I haven't seen any issues when the same version of STL (e.g. libstdc++v3 shipped with specific version of GCC) is used with different `-std` flags. What are the cases when this causes size increase, do you have some reference?

> What do you count as a major release, and when (approximately) it can happen?

Last major release was 4.0, and it just came out a couple of weeks ago. The next major release will be 5.0, and it will likely be around the same time next year, or later, depending on the whether we have enough breaking changes in the backlog to justify a new major release. In the meantime there will be minor releases: 4.1, 4.2 and so on.

> I don't think that Kconfig option is a good idea; it will increase the complexity and make support harder.

I definitely agree with the latter. However seeing that the change of the default C++ standard is still some time away, and the "C++17" requirement is coming up more often, it may be a viable solution in the meantime...

0xjakob:
I also agree that it shouldn't be changed with a major release. Even though we use a restricted feature set inside the IDF, some users might have used many more C++ features in their application code already. Even this can be painful to change and is not something you want to do in case you just planned to release a product. The customers have to be aware of that change.

We also have asio. A quick look tells that asio "should" work with versions younger than c++11 but as @igrr pointed out already, also this has to be tested.

I'm fine with Kconfig but that's a matter of taste, I guess.

chipweinberger:
> cmake_minimum_required(VERSION 3.5)
> 
> set(EXTRA_COMPONENT_DIRS "../common")
> set(CMAKE_CXX_STANDARD 17)
> 
> include($ENV{IDF_PATH}/tools/cmake/project.cmake)
> project(japp)

I have this cmakelists but I hit 

> /Volumes/.espressif/tools/xtensa-esp32s2-elf/esp-2021r1-8.4.0/xtensa-esp32s2-elf/xtensa-esp32s2-elf/include/c++/8.4.0/bits/fs_path.h:644: undefined reference to `std::filesystem::__cxx11::path::has_root_directory() const' among other std::filesystem functions.

b1ackviking:
> Is there some trick to using std::filesystem?

@chipweinberger Unfortunately, it is impossible to use `std::filesystem` with the current toolchain because GCC 8.4 implements it as a separate library `libstdc++fs`, which is not included in the distribution. The [`libstdc++fs` became a part of `libstdc++`](https://gcc.gnu.org/gcc-9/changes.html) only since GCC 9.

jdoubleu:
Does that mean C++20 is the default starting from v5?

igrr:
@jdoubleu That's right!

PR #4849, opened 2020-02-27
Esp modem fix race condition on exit ppp mode (IDFGH-2786)

fgiancane8:
This pull request is a rework of #4642  which I previously provided.

As per the discussion, we saw the possibility to have a race condition between the posting of PPP_CLOSE event and its effective completion, thus calling change_mode() and hangup() without proper synchronization could lead to errors.

The correct solution is to differentiate between the events: PPP_STOP and PPP_STOPPED. The former is to identify the request to stop the PPP session and the latter is to signal that the session has been closed properly.

The discussion was already carried out with @AxelLin and @david-cermak so it may be sound if we review this patch together.

Please let me know if it works properly for you.

I also think that the `esp_event_loop_create_default();` may not be necessary.

### DIFF
--- a/examples/protocols/pppos_client/components/modem/include/esp_modem.h
+++ b/examples/protocols/pppos_client/components/modem/include/esp_modem.h
@@ -36,7 +36,8 @@ ESP_EVENT_DECLARE_BASE(ESP_MODEM_EVENT);
 typedef enum {
     ESP_MODEM_EVENT_PPP_START = 0,       /*!< ESP Modem Start PPP Session */
     ESP_MODEM_EVENT_PPP_STOP  = 3,       /*!< ESP Modem Stop PPP Session*/
-    ESP_MODEM_EVENT_UNKNOWN   = 4        /*!< ESP Modem Unknown Response */
+    ESP_MODEM_EVENT_UNKNOWN   = 4,       /*!< ESP Modem Unknown Response */
+    ESP_MODEM_EVENT_PPP_STOPPED          /*!< ESP Modem PPP Session completely stopped */
 } esp_modem_event_t;
 
 /**
--- a/examples/protocols/pppos_client/components/modem/src/esp_modem.c
+++ b/examples/protocols/pppos_client/components/modem/src/esp_modem.c
@@ -18,6 +18,7 @@
 #include "freertos/task.h"
 #include "freertos/semphr.h"
 #include "esp_modem.h"
+#include "esp_netif_ppp.h"
 #include "esp_log.h"
 #include "sdkconfig.h"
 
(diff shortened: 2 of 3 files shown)


fgiancane8:
@AxelLin @david-cermak 
Any update about this?

david-cermak:
@fgiancane8 Thank for posting the fix! I am not convinced however if this was the right way of fixing this issue:
* There was an idea of isolating the `esp-modem.c` from the corresponding network interface (unlike the previous version and the compatibility layer) to have them wired together in pppos_example main (not documented anywhere, but the basic idea at least...)
* Wouldn't this race be fixed just closing the network interface [here](https://github.com/espressif/esp-idf/blob/master/examples/protocols/pppos_client/main/pppos_client_main.c#L276). I think there's no need for an extra event if we closing the network manually. (I mean calling the `esp_netif_action_stop()` directly or adding a new interface to esp_modem_netif.h)
* Some minor issues in your changeset, as you've mentioned the `esp_event_loop_create_default()` would probably fail, and would prefer registering events in init code.

fgiancane8:
> @fgiancane8 Thank for posting the fix! I am not convinced however if this was the right way of fixing this issue:
> 
> * There was an idea of isolating the `esp-modem.c` from the corresponding network interface (unlike the previous version and the compatibility layer) to have them wired together in pppos_example main (not documented anywhere, but the basic idea at least...)
Ok then I will check the code after your rework and see if everything works correctly.
> * Wouldn't this race be fixed just closing the network interface [here](https://github.com/espressif/esp-idf/blob/master/examples/protocols/pppos_client/main/pppos_client_main.c#L276). I think there's no need for an extra event if we closing the network manually. (I mean calling the `esp_netif_action_stop()` directly or adding a new interface to esp_modem_netif.h)

Yes, of course. I liked the idea of "doing everything in the modem module" so that the user did not have to recall to close also the interface manually. I worked so that the previous interface would not break behavior. I am very aware that "modem" and "ppp interface on modem" are two different entities but this should not be a user's concern. I mean now we are discussing about module interface here :)

> * Some minor issues in your changeset, as you've mentioned the `esp_event_loop_create_default()` would probably fail, and would prefer registering events in init code.
In fact I also tested a version without the `esp_event_loop_create_default()` and it properly works as by the time the stop ppp function is called, the main event loop has been already created in the interface PPPoS initialization.
If this causes issue, you can safely remove that line (or I can push an update version).

Hi @david-cermak ,

Thank you for your feedback. I am answering inline.

I am open to the discussion, if you are going to modify the modules I am very happy and satisfied to close the PR and wait for next version.

Thanks,
Francesco

david-cermak:
@fgiancane8 The issue of switching modem modes is fixed in the referenced commit, so this PR could be closed.

PS: Sorry for the delay and for neglecting this issue; although i wasn't 100% sure about your changes it did help some users addressing these issues. Thanks again for posting your fix.

fgiancane8:
@david-cermak ,
Happy that the issue is fixed! Ever happier if my patch served as a starting point for the referenced issue!

Thanks again

PR #4865, opened 2020-03-01
Make espcoredump more resilient when decoding erroneous thread stack (IDFGH-2802)

amotl:
We found that `espcoredump` would stop decoding all subsequent thread stacks when it hits an error like that on a specific thread stack.
```
  10   process 1073466144 /builds/idf/crosstool-NG/.build/src/gdb-7.10/gdb/inline-frame.c:171: internal-error: inline_frame_this_id: Assertion `!frame_id_eq (*this_id, outer_frame_id)' failed.
```

This change accounts for that by detecting such a condition and continuing the process appropriately. This helped significantly as `espcoredump` just decoded thread stacks 19-11 before while it will now also decode thread stacks 9-1, providing valuable additional information.

### DIFF
--- a/components/espcoredump/espcoredump.py
+++ b/components/espcoredump/espcoredump.py
@@ -1624,15 +1624,17 @@ def info_corefile(args):
     p = gdbmi_getinfo(p, handlers, "info threads")
     # THREADS STACKS
     p,threads,cur_thread = gdbmi_get_thread_ids(p)
+    print()
     for thr_id in threads:
         task_index = int(thr_id) - 1
-        if thr_id == cur_thread:
-            continue
         p = gdbmi_switch_thread(p, thr_id)
         p,thr_info_res = gdbmi_get_thread_info(p, thr_id)
+        if not thr_info_res.target_id:
+            print("WARNING: Unable to switch to thread %s\n" % thr_id)
+            continue
         tcb_addr = gdb2freertos_thread_id(thr_info_res.target_id)
         p,task_name = gdbmi_freertos_get_task_name(p, tcb_addr)
-        print("\n==================== THREAD %s (TCB: 0x%x, name: '%s') =====================" % (thr_id, tcb_addr, task_name))
+        print("==================== THREAD %s (TCB: 0x%x, name: '%s') =====================" % (thr_id, tcb_addr, task_name))
         p = gdbmi_getinfo(p, handlers, "bt")
         if task_info and task_info[task_index].task_flags != EspCoreDumpTaskStatus.TASK_STATUS_CORRECT:
             print("The task '%s' is corrupted." % thr_id)
@@ -1640,6 +1642,7 @@ def info_corefile(args):
                   task_info[task_index].task_flags,
                   task_info[task_index].task_tcb_addr,
                   task_info[task_index].task_stack_start))
+        print()
     print("\n======================= ALL MEMORY REGIONS ========================")
     print("Name   Address   Size   Attrs")
     for ms in merged_segs:


Alvin1Zhang:
@amotl Thanks for the contribution.

amotl:
Thanks, @Alvin1Zhang.

I also wanted to note that this change also removes the shortcut
```
if thr_id == cur_thread:
    continue
```
when iterating the threads as I found it convenient to have **all thread stacks** printed within the "THREADS INFO" section of the output. I didn't see a reason why to omit the particular "current thread" within that output.

gerekon:
> Thanks, @Alvin1Zhang.
> 
> I also wanted to note that this change also removes the shortcut
> 
> ```
> if thr_id == cur_thread:
>     continue
> ```
> 
> when iterating the threads as I found it convenient to have **all thread stacks** printed within the "THREADS INFO" section of the output. I didn't see a reason why to omit the particular "current thread" within that output.

Because it is printed at the beginning of the core dump, see the very first back trace https://github.com/espressif/esp-idf/blob/master/components/espcoredump/espcoredump.py#L1615-L1623

components/espcoredump/espcoredump.py
```diff
@@ -1624,22 +1624,25 @@ def gdb2freertos_thread_id(gdb_thread_id):
     p = gdbmi_getinfo(p, handlers, "info threads")
     # THREADS STACKS
     p,threads,cur_thread = gdbmi_get_thread_ids(p)
+    print()
     for thr_id in threads:
         task_index = int(thr_id) - 1
-        if thr_id == cur_thread:
-            continue
```
gerekon: I would leave this, since stack of the current thread is printed at the beginning.

amotl: Hi Alexey,

I recognized that information about the current thread stack was printed in the beginning. 

However, I found it convenient to have the information complete within the whole "THREADS INFO" section of the output. That way, users do not have to scroll back and forth when analyzing all the stack information.

Does this make sense to you?

With kind regards,
Andreas.

gerekon: Yes, it makes sense

alisitsyn:
Hi @amotl,
Thank you for your PR.
Unfortunately it was not possible to reproduce error similar to your with GNU gdb (crosstool-NG crosstool-ng-1.22.0-80-g6c4433a5) 7.10.

```10   process 1073466144 /builds/idf/crosstool-NG/.build/src/gdb-7.10/gdb/inline-frame.c:171: internal-error: inline_frame_this_id: Assertion `!frame_id_eq (*this_id, outer_frame_id)' failed.```

Your changes are under review and will be merged as soon as possible.

**Comment by amotl** [comment]
Hi @alisitsyn,

as we learned from #2892, we are using the `-98-` toolchain instead of `-80-`.

```
./xtensa-esp32-elf-gcc --version
xtensa-esp32-elf-gcc (crosstool-NG crosstool-ng-1.22.0-98-g4638c4f) 5.2.0
```

```
./xtensa-esp32-elf-gdb --version
GNU gdb (crosstool-NG crosstool-ng-1.22.0-98-g4638c4f) 7.10
```

Just in this second we see this toolchain might have been made of two different GNUs (5.2.0 vs. 7.10)?

Cheers,
Andreas.

amotl:
Dear Alexey,

so do you believe this will be fine? I just wanted to make sure that I didn't miss anything and you are waiting for some changes of mine.

With kind regards,
Andreas.

alisitsyn:
Dear Andreas,

Everything is correct your PR reviewed and should be merged to github very soon. I appologize for the delay with answer.

igrr:
Merged in https://github.com/espressif/esp-idf/commit/dde4d2e9ebfad3e88fc80e5179b816dab16351c2, thanks for the contribution!

amotl:
Thanks for merging!

PR #4891, opened 2020-03-05
Fix crash due to modbus responses after timeout (IDFGH-2830)

sprmn:
When a modbus response is received after the timeout period, they are processed and lead to second call of `vMBMasterRunResRelease`. This seems to crash the ESP, because `xSemaphoreGive` is called while the semaphore is not taken. 
I don't know if this fix is the right approach, however it seems to involve the least amount of code and makes the call to `vMBMasterRunResRelease` more robust.

### DIFF
--- a/components/freemodbus/port/portevent_m.c
+++ b/components/freemodbus/port/portevent_m.c
@@ -167,6 +167,8 @@ BOOL xMBMasterRunResTake( LONG lTimeOut )
  */
 void vMBMasterRunResRelease( void )
 {
+    // Check if the semaphore is not released already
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@sprmn Thanks for the contribution.

alisitsyn:
Hello @sprmn,

Thank you for your contribution and it will be used. This issue really happens when Kconfig::FMB_MASTER_TIMEOUT_MS_RESPOND is not set correctly or UART interrupt handler is in flash. Your proposale is under review and there is other approch which is under testing.

PR #4898, opened 2020-03-06
Update the rmt documentation (IDFGH-2837)

raldone01:
State which functions require the rmt driver to be installed.

### DIFF
--- a/components/driver/include/driver/rmt.h
+++ b/components/driver/include/driver/rmt.h
@@ -533,6 +533,8 @@ esp_err_t rmt_set_err_intr_en(rmt_channel_t channel, bool en);
 /**
 * @brief Set RMT TX interrupt enable
 *
+* This function requires the driver to be installed first!
+*
 * @param channel RMT channel
 * @param en enable or disable TX interrupt.
 *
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@raldone01 Thanks for the contribution.

suda-morris:
Thanks @raldone01 for your contribution to esp-idf. We will merge it ASAP.

raldone01:
Registering a custom interrupt seems broken. I will reopen this as an issue.

projectgus:
Hi @raldone01,

I don't see the issue, did you end up opening one?

Thanks for sending this PR in any case. After looking at this, we've determined that these were accidental breaking changes in the driver API. The IDF v4.0 API doesn't require pre-installation.

So our plan at the moment is to fix the behaviour to make IDF v4.0.

raldone01:
I forgot to open a issue. Should I still open one? (For the time being I copied the rmt.h and rmt.c from 4.0 over 4.1 because my application uses the netif abstraction.)

projectgus:
@raldone01 We're planning to fix the "install driver before using this function" breaking changes already. If you found another bug that you think might not be covered by this, an issue with the details would be great - thanks.

PR #4926, opened 2020-03-12
Linenoise don't return NULL on 0 length input (IDFGH-2869)

MadnessASAP:
Linenoise should only return NULL if there's been an error on input or if EOF is reached (user enters Ctrl-D)  0 length inputs are still valid and should be treated and returned as such.

### DIFF
--- a/components/console/Kconfig
+++ b/components/console/Kconfig
@@ -1,0 +1,10 @@
+menu "Linenoise"
+
+    config ESP_LINENOISE_RETURN_ZERO_STRING
+        bool "Return 0 length strings"
+        default n
+        help
+            If enabled linenoise will return 0 length strings if the user presses enter with out typing 
+            a command, if disabled NULL will be returned instead.
+
+endmenu
--- a/components/console/linenoise/linenoise.c
+++ b/components/console/linenoise/linenoise.c
@@ -979,11 +979,11 @@ char *linenoise(const char *prompt) {
     } else {
         count = linenoiseDumb(buf, LINENOISE_MAX_LINE, prompt);
     }
-    if (count > 0) {
+    if (count >= 0) {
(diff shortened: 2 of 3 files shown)


igrr:
Does this change mean that empty lines will be added to history? If yes, there probably needs to be an additional check to avoid doing that.

MadnessASAP:
> Does this change mean that empty lines will be added to history? If yes, there probably needs to be an additional check to avoid doing that.

It does not, adding to history is still controlled via linenoiseHistoryAdd().  It's possible for a downstream developer to inadvertently add 0 length strings to the history if they don't check the string returned by linenoise() against strlen().  Fortunately linenoiseHistoryAdd() already partially mitigates that by not adding history lines that are identical to the previous entry, that is repeatedly pressing enter won't fill the history buffer with more than 1 blank line.

I'll add that my proposed behaviour is in line with the upstream linenoise library behaviour.

Alvin1Zhang:
@MadnessASAP Thanks for the contribution.

igrr:
Thanks for explaining! In this case there probably needs to be an extra check in the example before calling linenoiseHistoryAdd, so that empty lines don't get added to history at all.

That said, there is a minor concern that this would change the behavior of existing apps which used the console example as their base. It is easily fixable, but could be a breaking change for some nevertheless. So we may need to add some Kconfig option to allow reverting the old behavior.

MadnessASAP:
I've touched up the example to filter blank lines and added a Kconfig to default to the old behavior

igrr:
@MadnessASAP The PR was merged with a follow-up change to make the behavior run-time configurable (instead of the Kconfig option), since the cost of run-time configuration in this is quite low. Sorry for the slow turnaround and thanks again for contributing!

Issue #4744, opened 2020-02-10
ESP-IDF Tools Installer v2.2 does not work due to two different issues (IDFGH-2670)


vmsh0:
## Environment

Windows 10 with latest updates applied. (Other environment items were not applicable.)

## Problem Description

I cannot install ESP-IDF Tools and ESP-IDF using the official installer. I would like to start developing for the ESP32 but I cannot install the tools. I would prefer not to install the environment manually.

I encounter two issues:

### Expected Behavior

ESP-IDF is copied to destination directory after extraction and setp continues.

### Actual Behavior

Error message "Failed to copy ESP-IDF to the destination directory: E:\espidf\3.3" appears and setup halts.

### Steps to reproduce

1. Download and run `esp-idf-tools-setup-2.2.exe`
2. Choose option to download ESP-IDF (tested with 3.3 release and 3.3 branch)

### Debug Logs

[log1.txt](https://github.com/espressif/esp-idf/files/4182406/log1.txt)

Relevant part:

> 2020-02-10 19:00:44.329   --------
> 2020-02-10 19:00:44.332   Moving C:\Users\r\AppData\Local\Temp\is-EG1TH.tmp\esp-idf-v3.3.1 to E:\espidf\3.3
> 2020-02-10 19:00:44.332   Message box (OK):
>                           Failed to copy ESP-IDF to the destination directory: E:\espidf\3.3
> 2020-02-10 19:01:53.031   User chose OK.
> 2020-02-10 19:01:53.031   Message box (Yes/No):
>                           Installation log has been created, it may contain more information about the problem.
>                           Display the installation log now?
> 2020-02-10 19:01:53.962   User chose Yes.
> 2020-02-10 19:01:53.989   CurStepChanged raised an exception.

### Workaround

Go to the `.espressif\releases` directory, manually extract and copy esp-idf-v3.3.1.zip to ESP-IDF destination directory, and re-run installer.

### Expected Behavior

ESP-IDF Tools are set-up.

### Actual Behavior

Error message "Installation has failed with exit code 1" appears and setup halts.

### Steps to reproduce

1. Download and run `esp-idf-tools-setup-2.2.exe`
2. Choose option to download ESP-IDF (tested with 3.3 release and 3.3 branch)
3. When installation fails, apply workaround from Issue 1
4. Re-run `esp-idf-tools-setup-2.2.exe`
5. Choose option to use the previously extracted ESP-IDF

### Debug Logs

[...]

Issue #3581, opened 2019-06-03
ESP MQTT Client always connects via STA IF even if STA is not default IF (IDFGH-1286)


no1seman:
## Environment

- Development Kit:      ESP32-Wrover-Kit
- Kit version (for WroverKit/PicoKit/DevKitC): v4.1
- Module or chip used:  ESP32-WROVER-B
- IDF version (run ``git describe --tags`` to find it): v4.0-dev-667-gda13efc17
- Build System:         CMake
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): 1.22.0-80-g6c4433a
- Operating System:     Linux
- Power Supply:         USB|external 5V&Battery

## Problem Description
When 2 interfaces is up: ETH and STA MQTT Client always connects to server via STA interface even if default IF is ETH. 
### Expected Behavior
MQTT Client should use default IF for connection.
### Actual Behavior
MQTT Client always connects to server via STA interface if both ETH and STA is up and ETH is set to default IF.

### Steps to repropduce

1. Establish ETH and STA IFs
2. Set ETH as default IF
3. Establish MQTT connection
4. netstat -nt | grep 1883 on MQTT server server host to see that STA ip is used for connection

### Code to reproduce this issue

code to set default IF:
esp_err_t setDefaultIf(tcpip_adapter_if_t tcpip_if)
{
    struct netif *_netif;
    esp_err_t ret = tcpip_adapter_get_netif(tcpip_if, (void **)&_netif);
    if (ret != ESP_OK)
    {
        return ret;
    }
    netif_set_default(_netif);
    return ESP_OK;
}

.....
setDefaultIf(TCPIP_ADAPTER_IF_ETH);
....
esp_mqtt_client_start(client);
......

david-cermak:
Hi @no1seman 

The issue is probably that both your interfaces route to the server. Setting the default interface in LwIP means that the netif is "used to output all packets for which no specific route is found", 
so it applies only to scenarios where routing could not be determined. Please note the implementation of [`ip4_route` here](https://github.com/espressif/esp-lwip/blob/fe9a58cb7abdc74e7a2196b7c4e0a29ff9042f88/src/core/ipv4/ip4.c#L210), `default_netif` is used only if no match in IP 
could be found in the first section.

no1seman:
@david-cermak,

yes, you are right, both interfaces in the same subnet:
eth: IP: 192.168.3.222/24, GW: 192.168.3.1
[...]

PR #4985, opened 2020-03-25
LWIP SLIP integration (IDFGH-2952)

ryankurte:
update: SLIP integration for esp/lwip now done and working, see later comments for updates

~Working on LWIP SLIP module, not sure yet what the endgame looks like.~

## Mods
- ~`.gitmodules` patching required for non-github VCS.~ (removed)
- ~LWIP must be built with `SLIP_RX_FROM_ISR=1` to enable ISR based feeding and with the static (whyyyy?) `slipif_output` function exposed.~ (resolved with kconfig)
 
## Questions
- ~Current approach uses an application task to feed the SLIP interface, a default impl of the task could be provided however parsing is likely to be application-specific due to common command extensions to SLIP (see: contiki and `!P`, `?P`).~
- ~What is the best way to patch LWIP to work with this?~

## References
- https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_netif.html
- https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_netif_driver.html
- https://github.com/espressif/esp-idf/blob/e599b79/examples/ethernet/basic/main/ethernet_example_main.c


projectgus:
Hi @ryankurte,

> .gitmodules patching required for non-github VCS.

Just FWIW, this `../..` is what we use for our internal non-github VCS (which has mirrors of all these submodule repos in it). We know it can be annoying for hosting forks of IDF in other places. :/

>  Current approach uses an application task to feed the SLIP interface, a default impl of the task could be provided however parsing is likely to be application-specific due to common command extensions to SLIP (see: contiki and !P, ?P).

@suda-morris @david-cermak Do you have any suggestion for this?

> What is the best way to patch LWIP to work with this?

If the only option is to expose a static function as non-static, then send a PR to the [esp-lwip repo](https://github.com/espressif/esp-lwip/) and link that PR to this one. Can update the LWIP submodule commit as part of this PR.

david-cermak:
Hi @ryankurte 

>Current approach uses an application task to feed the SLIP interface, a default impl of the task could be provided however parsing is likely to be application-specific due to common command extensions to SLIP (see: contiki and !P, ?P).

You can provide a default implementation of some/common commands in the component and as an example project overwrite/add application specific commands, something like a virtual method in C++ for parsing commands

> What is the best way to patch LWIP to work with this?

I think there's no need to change the lwip submodule:
* `SLIP_RX_FROM_ISR` could be put in /port/opt.h (maybe based on Kconfig)
* `slipif_output()` may be called using netif->output (the same as for other netifs)

ryankurte:
Thank you for the clarifications! I am afraid I am likely to need a bit more assistance to get this working (to be clear, it currently does not) 🙃

***Progress:***
- Added the LWIP flags to `lwip_opts.h` and the kconfig so the correct components are exposed
- Added an [SIO module](https://github.com/espressif/esp-idf/pull/4985/files#diff-b7bb7f551f93606b53a7dc76841e3196) works for sending / receiving SLIP using a lwip netif
- Implemented a [SLIP netif](https://github.com/espressif/esp-idf/pull/4985/files#diff-c81681cd2c2038525b8d956c8a8ebff0) loosely based on the PPP one
- Propagated the SLIP interface methods through [esp_netif_lwip.c](https://github.com/espressif/esp-idf/pull/4985/files#diff-7371b5e54a943658131b92bcf01d6af8) and added [defaults](https://github.com/espressif/esp-idf/pull/4985/files#diff-ca60b6408cabd47fc164c7ae5f313b7c) so this can be initialised

***Questions:***
- The ethernet implementation has a separate driver, though other LWIP components do not seem to need this. At the moment I have the UART thread and management in the `lwip_slip_ctx_t` module. What's the aim for splitting between the driver and the `esp_netif`/`lwip_slip_ctx_t` components?
- Is there a more useful way to update configurations than creating a default object then later calling `esp_netif_slip_set_params`? The anonymous types inside the `esp_netif_config_t` seem to make this tricky

**Usage:**
```c
// Create default config
esp_netif_config_t cfg = ESP_NETIF_DEFAULT_SLIP();

// Create network interface
esp_netif_t *slip_netif = esp_netif_new(&cfg);

// Build new slip config
esp_netif_slip_config_t slip_config = {
    .uart_dev = UART_NUM_2,
    .uart_tx_pin = 4,
    .uart_rx_pin = 36,
    .uart_baud = 115200,

    .rx_filter = slip_rx_filter,
    .rx_buffer_len = 1024,
};

// Apply slip config
esp_netif_slip_set_params(slip_netif, &slip_config);

// Create new driver
void* slip_driver = esp_slip_new_netif_glue(slip_netif);

// TODO: do we need to do any of this?
// We _could_ have the uart / thread be part of the driver perhaps
//ESP_ERROR_CHECK(esp_slip_driver_install(&config, &slip_ctx));

// Attach interface and driver
ESP_ERROR_CHECK(esp_netif_attach(slip_netif, slip_driver));
```

I _think_ I'm on roughly the right track but the internals are pretty, complex, so, any feedback / guidance is definitely appreciated ^_^

ryankurte:
Alright so this is working now 🎉🎈

I've cleaned up a bunch of the logging and added a lil contiki udp example in `examples/protocols/slip` that is tested and working here (though may be difficult to test without a working slip gateway and network and/or a test harness to inject into the slip interface).

Output from the example:
```
I (372) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (384) SLIP_EXAMPLE: Initialising SLIP interface
I (394) esp-netif_lwip-slip: esp_netif_new_slip: Created SLIP interface (netif 0x3ffb96d8, slip_ctx: 0x3ffb98d4)
I (404) esp-netif_lwip-slip: esp_netif_start_slip: Starting SLIP connection (slip_ctx: 0x3ffb98d4, addr: FD00::1)
I (414) uart: queue free spaces: 10
I (414) SLIP_EXAMPLE: Socket bound, port 5678
I (72014) SLIP_EXAMPLE: Prefix request
I (72014) SLIP_EXAMPLE: slip_set_prefix: prefix set (fd000000:00000000)
I (91614) SLIP_EXAMPLE: Received 'hello 0' from 'FD00::280:E107:5018:1C26'
```

The outstanding questions still stand, I'm not super sure about the split between the `lwip_slip` component and the `lwip_slip_driver` component, and the requirement to re-configure seems slightly less than ideal, but otherwise this might be ready for a review? (also please feel free to directly edit this branch whenever it's more efficient than the back-and-forth)

ryankurte:
Running this requires patches for #5046, #4744, #5096

david-cermak:
> The outstanding questions still stand, I'm not super sure about the split between the lwip_slip component and the lwip_slip_driver component, 

You can check the current pppos-client example. It uses a private component which includes both the IO driver and netif glue. Also the `esp-netif-slip` should be really a thin translation layer, no threads running. 

> and the requirement to re-configure seems slightly less than ideal

Could you be more specific? What reconfiguration are you talking about?

Briefly reviewed, this PR LGTM very generally, after moving the driver/glue to the correct place. 
This can also be resolved internally, if you prefer, but may take some time and result in more "dramatic" refactoring...

Thank you for posting this PR!

ryankurte:
Thanks for the review!

> You can check the current pppos-client example. It uses a private component which includes both the IO driver and netif glue. 

Are you talking about the `esp_modem_dte_t` in `esp_modem.c`? I'm not sure I understand why the netif glue stuff is in the application if it's common for all devices? (maybe this is not the case for PPPOE but with the filter abstraction I put in is for slip).

And would it, in the SLIP case, make sense to move this functionality to a `slip_modem` component or something containing the slip equivalent of this so it need not be implemented by all consumers?

> Also the esp-netif-slip should be really a thin translation layer, no threads running.

Yeah I wasn't sure where to put the thread eh.

> Could you be more specific? What reconfiguration are you talking about?

Ahh yeah this wasn't super clear sorry, annotated from `slip_client_main.c`

```c
// Create the default config, since the internals are abstract you can't change
// any slip-specific settings here
esp_netif_config_t cfg = ESP_NETIF_DEFAULT_SLIP();

// Create the network interface
esp_netif_t *slip_netif = esp_netif_new(&cfg);

// Generate the slip configuration you actually want
esp_netif_slip_config_t slip_config = {
    ...
};

// Apply this configuration to the existing network interface
esp_netif_slip_set_params(slip_netif, &slip_config);
```

It _seems_ like it should be possible to pass the configuration you want in at init time, but, because the types are hidden for the default object I can't see how to do it without the later `esp_netif_slip_set_params`.

> Briefly reviewed, this PR LGTM very generally, after moving the driver/glue to the correct place.
This can also be resolved internally, if you prefer, but may take some time and result in more "dramatic" refactoring...

I'm happy to do another pass or two to get closer and see from there? Ultimately y'all have to be happy with / with maintaining it so if/when you do decide to pick it up and refactor heavily just lmk so we're not wasting effort.

david-cermak:
About the `esp-modem` component: The reason this component lives with examples is just that it's not considered "very" universal to be moved under IDF component folder. Not important though, the project sees it as a standard component anyway.
I referred this component because it contains both the netif glue layer and the peripheral driver in a similar way the SLIP example should use
and are both outside the `esp-netif` component.

The same applies to the slip component, no matter if the component stays with examples or goes into component folder.
Users would just consume it using `EXTRA_COMPONENT_DIRS` if not located in the default $IDF_PATH/components

Let's move the threads into the slip component, as well, maybe `slip_modem` is not a bad name.

> It seems like it should be possible to pass the configuration you want in at init time...

Okay, thanks for explaining. The thing is that the `esp_netif_config_t` is a generic network interface configuration,
 application code can update the IP addresses, events, stuff that's common for all netifs.
 Update slip specific options should be done separately. Maybe here this could be wrapped within some `slip_modem_init()`.

> I'm happy to do another pass or two

Sounds good, here are my general comments about moving the layers around:
* Move components/esp_netif/lwip/esp_netif_lwip_slip_sio.c -> slip component
* Move components/esp_netif/lwip/esp_netif_lwip_slip_driver.c -> slip component
* Reduce components/esp_netif/lwip/esp_netif_lwip_slip.c to only lwip interaction and keep that part in esp-netif
* Move everything else under slip component

I would also suggest placing the slip/slip-modem component into example folder the same way `esp-modem` is used as a private component.

Please also clean up the git history a git, squash the relevant commits.

components/esp_netif/lwip/esp_netif_lwip_slip_driver.c
```diff
@@ -0,0 +1,120 @@
...
+    esp_slip_driver_start(driver->slip_driver);
+    
+    return ESP_OK;
+}
+
+esp_err_t esp_slip_driver_netif_set_default_handlers(esp_netif_slip_driver_t *driver, esp_netif_t * esp_netif)
+{
+    ESP_LOGD(TAG, "%s", __func__);
+
+    // TODO: do we need to attach any handlers here?
```
david-cermak, line 87: Yes, we typically create handlers for start/stop actions to set the corresponding netif up/down. For that we also need someone to post into that event id, could be the driver code after setting up the UART.

ryankurte:
looks good to me! i've got some time scheduled to do this next week.

> Move components/esp_netif/lwip/esp_netif_lwip_slip_sio.c -> slip component

for this to be moved the linker needs to resolve symbols expected in the `lwip` module and exposed in the `slip_modem` module, i had a dig through the docs and tried this when i started (building this as an out-of-tree component) but couldn't work it out, is this possible within the build system?

david-cermak:
> for this to be moved the linker needs to resolve symbols expected in the lwip module...

It is possible to configure build system this way (see [this post](https://github.com/espressif/esp-idf/issues/3581#issuecomment-519940878) on out-of-tree linker dependency), but not recommended. What I meant was moving just the implementation part and keep the sio_read/open/witer wrappers in `esp_netif`. Sorry, it wasn't very clear previously. We have the `esp-netif` component to bridge between application and driver code and the lwip.

components/esp_netif/include/esp_netif.h
```diff
@@ -119,6 +119,15 @@ esp_err_t esp_netif_set_driver_config(esp_netif_t *esp_netif,
...
 esp_err_t esp_netif_attach(esp_netif_t *esp_netif, esp_netif_iodriver_handle driver_handle);
 
+/**
+ * @brief This function sets default netif no matter which implementation used
+ *
+ * @param esp_netif handle to network interface
+ *
+ * @note: This function must be called from lwip thread
+ */
+void esp_netif_set_default_netif(esp_netif_t *esp_netif);
```
ryankurte, line 129: this needs to be public as if you have SLIP and WIFI, and the WIFI connection drops, slip becomes the default interface, causing any non-slip network stuff to fail for the rest of time.

in some cases this could be mitigated by `bind`ing on the correct interface, but, this option is not always exposed (for example, with the MQTT client), and, you may wish for either ethernet or wifi to work (or to prioritise one or the other) which is not easily achieved with `bind`.

exporting this allows the slip + wifi case to be resolved with something like the following on `IP_EVENT_STA_GOT_IP` or `IP_EVENT_ETH_GOT_IP` events:
```c
ip_event_got_ip_t* ip_event = (ip_event_got_ip_t*) event_data;
esp_netif_set_default_netif(ip_event->esp_netif);
```

components/esp_netif/include/esp_netif_slip.h
```diff
@@ -0,0 +1,122 @@
...
+ * @brief  Fetch IP6 address attached to the SLIP interface
+ *
+ * @param[in]    esp_netif handle to slip esp-netif instance
+ * @param[in]    address index (unused)
+ *
+ * @return
+ *         - pointer to the internal ip6 address object
+ */
+const ip6_addr_t* esp_slip_get_ip6(esp_netif_t *slip_netif);
+
```
jeandudey, line 121:
```suggestion
#ifdef __cplusplus
} /* extern "C" */
#endif
```

components/lwip/port/esp32/include/lwipopts.h
```diff
@@ -697,37 +718,82 @@
 /**
  * ETHARP_DEBUG: Enable debugging in etharp.c.
  */
-#define ETHARP_DEBUG                    LWIP_DBG_OFF
+#ifdef CONFIG_LWIP_ETHARP_DEBUG
```
jeandudey, line 721: These Kconfig changes to lwIP debug can be a separate PR that can be merged first, as they might be useful when implementing other network interfaces or debugging lwIP.

components/esp_netif/lwip/esp_netif_lwip_slip.c
```diff
@@ -0,0 +1,335 @@
...
+#if 1
+    // Attach interface
+    netif_add_noaddr(slip_ctx->esp_netif->lwip_netif, (void*)slip_ctx->uart_dev, &slipif_init, &ip_input);
+
+    // Bind address
+    int8_t addr_index = 0;
+    //netif_add_ip6_address(slip_ctx->esp_netif->lwip_netif, &slip_ctx->addr, &addr_index);
+
+    netif_ip6_addr_set(slip_ctx->esp_netif->lwip_netif, addr_index, &slip_ctx->addr);
+    netif_ip6_addr_set_state(slip_ctx->esp_netif->lwip_netif, addr_index, IP6_ADDR_VALID);
```
jeandudey, line 142:
```c
int8_t addr_index;
netif_add_ip6_address(slip_ctx->esp_netif->lwip_netif, &slip_ctx->addr, &addr_index);
netif_ip6_addr_set_state(slip_ctx->esp_netif->lwip_netif, addr_index, IP6_ADDR_VALID);
```

[`netif_ip6_addr_set`](http://www.nongnu.org/lwip/2_0_x/group__netif__ip6.html#gae21572fdbd8664d22a1b281a6c31c9bbf) is used to change an existing address, `netif_add_ip6_address` could do the trick

components/esp_netif/lwip/esp_netif_lwip_slip.c
```diff
@@ -0,0 +1,335 @@
...
+ */
+esp_err_t esp_netif_slip_set_params(esp_netif_t *netif, const esp_netif_slip_config_t *slip_config) {
+
+    lwip_slip_ctx_t* slip_ctx = netif->lwip_slip_ctx;
+
+    ESP_LOGD(TAG, "%s (slip_ctx: %p)", __func__, slip_ctx);
+
+    if (slip_ctx->running) {
+        ESP_LOGE(TAG, "Cannot set parameters while SLIP interface is running");
+        return -1;
```
jeandudey, line 200:
```suggestion
        return ESP_ERR_INVALID_STATE;
```

or if -1 makes more sense here, `ESP_FAIL` can be used.

components/esp_netif/include/esp_netif_slip.h
```diff
@@ -0,0 +1,122 @@
...
+/** @brief SLIP event base */
+ESP_EVENT_DECLARE_BASE(NETIF_SLIP_STATUS);
+
+typedef enum {
+    ESP_SLIP_EVENT_START = 1,
+    ESP_SLIP_EVENT_STOP = 2,
+} esp_netif_slip_event_e;
+
+typedef bool slip_rx_filter_f(void* ctx, uint8_t* data, uint32_t len);
```
jeandudey, line 33:
```suggestion
typedef bool slip_rx_filter_cb_t(void* ctx, uint8_t* data, uint32_t len);
```

components/esp_netif/include/esp_netif.h
david-cermak: We have to be careful if we want to make this function public, as you can see from the note it needs to be called in lwIP context. We would need the api version of that function (similar to e.g. setup/down, start/stop -- call using `esp_netif_lwip_ipc_call()`

Understand the point, it make sense to control the routing from user code. Currently the routing priorities are assigned to every interface and the one with the highest number is chosen to be the default one. Default priorities are set here: https://github.com/espressif/esp-idf/blob/master/components/esp_netif/include/esp_netif_defaults.h#L73 and could be easily changed from user code (see that I've set the PPP  prio too high, will fix).

Would suggest lowering the default `routeprio` of the SLIP interface first. I can make setting the default interface more explicit/public API in some future update.

ryankurte: ahh yeah dropping the defaults should do it, and the STA/ETH priority is the solution to another bug i've been dealing with, will revert and fix priorities / put a note in, thanks!

components/esp_netif/lwip/esp_netif_lwip_slip.c
ryankurte: When I tested it calling `netif_ip6_addr_set` resulted in the address state being set to `IP6_ADDR_TENTATIVE` and then never upgraded so it didn't work... not sure why this is

components/esp_netif/lwip/esp_netif_lwip_slip.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/**
+ * @brief Stops the SLIP interface
+ */
+esp_err_t esp_netif_stop_slip(lwip_slip_ctx_t *slip_ctx)
+{
+    ESP_LOGI(TAG, "%s: Stopped SLIP connection: %p", __func__, slip_ctx);
+
+    // TODO: stop / unbind interface
```
ryankurte, line 122: Should we be calling the opposite of `netif_set_link_up` and `netif_set_up` here?

components/lwip/port/esp32/include/lwipopts.h
ryankurte: thanks but, i'm not interested in splitting out or managing more prs at this time

components/esp_netif/lwip/esp_netif_lwip_slip_driver.c
ryankurte: Alrighty, so instead of calling the slip start / stop functions here:

https://github.com/espressif/esp-idf/pull/4985/files#diff-a02d6b6299ad381db96e4ff380108574R163

I should emit start/stop events and let the default handlers manage em?

ryankurte:
Alrighty, second pass.

I moved everything that can be moved out to `examples/slip` and fixed all the comments I could. A couple of open comments left, and I couldn't clearly see the line between the glue and the modem so I put them together, not sure if y'all have opinions on that?

components/esp_netif/lwip/esp_netif_lwip_slip.c
david-cermak: I think using the `netif_ip6_addr_set()` is okay in this context as we actually set an IP before connecting bypassing all the IPv6 negotiation/ND.

components/esp_netif/lwip/esp_netif_lwip_slip.c
david-cermak: Yes please, `netif_set_link_down()` seems appropriate.

components/lwip/port/esp32/include/lwipopts.h
david-cermak: agree that this is a useful changeset, unrelated to the SLIP support. would suggest dissecting this change into a separate commit.

david-cermak:
In general still two more concerns:
* `LWIP_SLIP_RX_FROM_ISR` AFAIK this doesn't use interrupts, correct? Might be misleading to call this option this way.
* `esp_netif` (sio) directly references uart driver

Not sure how to deal with these issues at the moment, but OK to leave it as is and will address it somehow, but please expect some rework in this area.

ryankurte:
thanks for the review! i should be able to get to these changes in the next week or so.

> LWIP_SLIP_RX_FROM_ISR AFAIK this doesn't use interrupts, correct? Might be misleading to call this option this way.

yeeah, i left this as it reflects what it's called internally in LWIP so changing it would be inconsistent with the other config options, not sure how to do this better.

> esp_netif (sio) directly references uart driver

yeah, so we only have a single u8 to pass context into the sio module. super unfortunate lwip doesn't take a fake-vtable / callback function object for constructing sios.

short of a pr to lwip to fix this, i had considered:
- adding an `sio_cb_s` object with callback pointers for each of the `sio` functions
- adding a function to bind a single static `sio_cb_s` instance to the sio driver
- adding a list of `void*` context objects indexed in the sio module using the `u8` port option
- adding a function to bind `void*` context objects to a given slot in the list

but this is a lot of work to work around a fundamental silliness in lwip, maybe it would be simpler to pull SIO out and configure the build system to externally locate these symbols if LWIP is enabled?

david-cermak:
@ryankurte 
I'm sorry, have been a bit busy lately and couldn't reply in time.

> LWIP_SLIP_RX_FROM_ISR

Okay to just document this, a note in the "help" section of Kconfig is enough I think.

> sio uses uart driver

The problem seems to be in this conversion from netif pinter to a number https://github.com/espressif/esp-lwip/blob/2.1.2-esp////src/netif/slipif.c#L368 (I believe we could have even kept the vanilla netif for SLIP, if the `netif->state` were used to call `sio_open()`).
One option to work this around is to pass an interface number (`netif_get_index()`) to the `sio_open()` function and return related `esp-netif` pointer, which contains everything what we need. Then, to avoid using the uart number directly suggest calling `esp_netif_receive()/transmit()` with this `esp_netif` instance from sio_read/write.

What I don't quite understand after looking at the code again is that `sio_read()` and `sio_write()` reads/writes data to UART directly (sio_read is supposed to be called from lwip thread periodically), but at the same time, esp_netif_receive and transmit are also called from the `slip_modem` component -- again reading UART data in a thread and passing it over to the lwIP. Also the `esp_slip_modem_transmit()` doesn't seem to be ever called, or is it?

ryankurte:
> I'm sorry, have been a bit busy lately and couldn't reply in time.

hey no worries, it all works for me so i'm happy.

> The problem seems to be in this conversion from netif pointer to a number

> One option to work this around is to pass an interface number (netif_get_index()) to the sio_open()

huh, so we could have a context and an _abstract_ sio (set of function pointers) in the netif that got called out? seems viable though, complicated?

> What I don't quite understand after looking at the code again is...

it's entirely possible i have misunderstood something, it's not a _simple_ arrangement 😅

`sio_read()` and `sio_write()` use depends on whether it's interrupt driven or poll based SLIP, there's some documentation in the lwip slip headers but it's not a particularly clean delineation.
rather than using sio_read() i elected to poll on the UART because it's simpler to sleep on that and then feed data to the slip interface, imo either approach is reasonable.

we also need to expose a mechanism to let apps [write to the slip interface](https://github.com/espressif/esp-idf/pull/4985/files#diff-237b0af849c4289dd869ddbceecf0fe3R129) so support the extensions for setting addresses etc. that some operating systems use.

i think my intent was that this should be refactored to call the `esp_slip_modem_transmit ` function rather than the `esp_netif_lwip_slip_raw_output` one currently used, and that the `esp_slip_modem_receive` function is technically not required.

`esp_slip_modem_transmit ` is (afaik) required as a `transmit` function for the generic driver object `esp_netif_driver_ifconfig_t` to adapt between the function signatures / related contexts, which is bound in [here](https://github.com/espressif/esp-idf/pull/4985/files#diff-a02d6b6299ad381db96e4ff380108574R217).

david-cermak:
> huh, so we could have a context and an abstract sio

I don't think this is too complicated, seems like the SLIP has a bit different concept than other lwIP interfaces, so we have to adapt it's IO objects to plug them onto esp-netif IO objects. We have to abstract the `sio` somehow on this level. The general idea of esp-netif is to keep an interface independent from any IO specifics and normalize its IO functions to have them easily attached to any IO driver.

>  it's not a simple arrangement

It is certainly not simple! The issue I've been getting at is the two threads simultaneously reading from UART and posting data to the lwIP netif -- bad code reading on my end :-) -- the [lwIP read thread](https://github.com/espressif/esp-lwip/blob/2.1.2-esp/src/netif/slipif.c#L413) indeed gets created (as `SLIP_USE_RX_THREAD` is 1!) but never succeeds (as `SLIPIF_THREAD_STACKSIZE` is 0). Suggest setting the `SLIP_USE_RX_THREAD` to 0 in `lwipopts.h` to make it clear.

> we also need to expose a mechanism to let apps write to the slip interface

Oh yes, forgot about that! This is also where the slip differs from other lwIP interfaces. Agree with your approach to keep the `esp_netif_lwip_slip_raw_output()` in public headers for slip section of `esp-netif`

> this should be refactored to call the esp_slip_modem_transmit

Not really, the modem-transmit function should output data to the IO driver in general - all packets should come and go via this end and the `esp_netif_transmit()` consequently. Conversely the `esp_netif_receive()` (i.e. modem-receive) is to pass all the incoming packets from UART to lwIP netif input (so the reading thread in the slip-modem should call the `esp_netif_receive()` instead of `esp_netif_lwip_slip_input()`). Not sure if this helps making things any clear... 

Agree that this is quite complex, plus we don't document the general ideas on what's preferred to be *abstracted* or *internal only* and what's okay to keep as public. I think I understand the concept much better now, thanks for explaining your intentions in the last post! At this point I would suggest accepting the PR as is (unless you have any other preference?) and rework as outlined above internally (this refactor would just connect the IOs on different layers and abstract others while keeping internals the same).

david-cermak:
@ryankurte 
Here's a preliminary version of the update: 
[0001-esp-netif-slip-interface-update.patch.txt](https://github.com/espressif/esp-idf/files/4785955/0001-esp-netif-slip-interface-update.patch.txt)
(applied on top of your changes)

It would be great if you could test it with your device and use-case (have just tested on plain linux slip interface on IPv4 -- added the config option)
Also, please comment on the changes if you have any concerns (tried to simplify that but maybe have gone too far...?)

This would still need some adjustments, but overall just cosmetic changes to be expected, such as:
* remove lwip dependencies from everything but `esp-netif`
* slip modem stop/restart
* event processing -- thinking of removing slip events completely as the slip_modem could do without it.

ryankurte:
Sweet as, I'll have a look at this / bash at running it probably next week.

ryankurte:
oops, pushed this to the external mirror instead of the internal one.

sorry it's taken a while, have been playing with this recently and there seems to be an issue with routing over the slip connection with the patch. still working on whether that might be on our side or part of the patch.

(as an aside, is the new `esp_netif_init()` behaviour markedly different to the old `tcpip_adaptor_init()`? because making this switch, and the required patches to init the netifs manually seems to render routing inoperable.)

ryankurte:
righty so changing the bare-minimum to get the patch compiling against the example (not moving from `tcpip_adapter_init`) is definitely not working. I still see the prefix request / response so the _serial_ component is working, but on incoming messages i see:

```
slipif: Got packet (115 bytes)
ip6_input: packet accepted on interface XX
ip6_input: packet not for us.
```

which i think suggests there is no available `netif` for routing the incoming packet.

i note we're no longer calling [`netif_ip6_addr_set` or `netif_set_up`](https://github.com/espressif/esp-idf/commit/522adc81a2db6fb02176271cdca9a16f72e51cb3#diff-c81681cd2c2038525b8d956c8a8ebff0L110) on start, is there something different i should be doing to bring this up? (also adding a manual call to `netif_ip6_addr_set` does not _appear_ to help).

david-cermak:
Thanks for checking for after the patch. In fact the related MR has already been merged internally (IPv4 part of the example worked okay) but haven't yet pushed to GitHub. I couldn't really reproduce the issue locally, testing on linux where SLIP doesn't work well with IPv6. I don't think we have to call `netif_set_up()`, should be enough to bring the link up, which is done by calling `esp_netif_start()`.

>  manual call to netif_ip6_addr_set

Maybe that's the problem, please try to call `esp_netif_slip_set_ipv6()` from esp_netif_slip.h from the main. Besides setting the address we might want to set ipv6 state:
```
    netif_ip6_addr_set(netif->lwip_netif, addr_index, (ip6_addr_t *)&slip_ctx->addr);
    netif_ip6_addr_set_state(netif->lwip_netif, addr_index, IP6_ADDR_VALID);
```
Possibly please also try `IP6_ADDR_PREFERRED` state.

Thanks again for testing this after these updates. I will look at the differences between initialization using tcpip_adapter and esp_netif, but I believe that's not the issue. (seems like we have to push fixing commits now)

ryankurte:
> In fact the related MR has already been merged internally

ooh exciting! i am counting the days until we don't have to build forked docker images and i can drop CI times by like, 90% 😂

> I couldn't really reproduce the issue locally, testing on linux where SLIP doesn't work well with IPv6

huh, we've had this running in the past, though i don't have physical hardware that would work for this anymore. i wonder whether it's possible to cross the rx/tx lines and use it as a _physical_ loopback? or i guess to setup two esp slip interfaces on one device that talk to eachother.

> Thanks again for testing this after these updates

thanks for all the support with implementing this! i'll try have another bash at this on monday / tuesday work schedule depending. fingers crossed one of these works but if not i'd be happy to do a live co-op debug on real hardware if it'd help / you have the time to do so.

on a not entirely unrelated front, would you expect NAPT (NAT) to work over a v6 interface? i had a go at bridging the IPv6-over-slip network via the eth/wifi interfaces but, seem to be missing something, and the examples would all _appear_ to be for IPv4.

Alvin1Zhang:
Thanks for your contribution, please help refer to https://github.com/espressif/esp-idf/commit/e9bce145cfbaa7412faf9777d10f0951516b04e8

ryankurte:
> Maybe that's the problem, please try to call esp_netif_slip_set_ipv6() from esp_netif_slip.h from the main. Besides setting the address we might want to set ipv6 state

unfortunately no success playing with these options :-(

as an aside, given that `esp_netif_slip_set_params` only copies the address where `esp_netif_slip_set_ipv6` both copies and initializes it should one of these be removed (or, call out to the other?)

david-cermak:
@ryankurte I was able to reproduce the issue using linux `tun` interface. I wanted to make the esp-netif initialisation for the slip interface similar to the generic netif, but didn't realize `netif_add()` actually cleans out the ipv6 structs (which we setup in set-params). 

The workaround is to run `esp_netif_slip_set_ipv6()` just after `esp_netif_attach()` in main (and making it pass the *already started* condition).

In addition to that I made a stupid (alignment) mistake in [`IP6_ADDR ()`](https://github.com/espressif/esp-idf/blob/master/examples/protocols/slip/slip_udp/main/slip_client_main.c#L201) making the actual address `00fd::1` instead of `fd::1`.

I will post a fix later this week.

PS: As for the NAPT, this works only for IPv4

Alvin1Zhang:
Thanks for your contribution, changes merged by https://github.com/espressif/esp-idf/commit/5dae28069f5c0995a08997008bd954e3efb6bdfd.

ryankurte:
awesome, thanks! any idea which release branch this is likely to land in?

david-cermak:
The changes have just been merged to master so will only be available in the upcoming release `v4.3`.

webworxshop:
Hi @david-cermak, I found a bug in the esp-lwip component related to this MR. PR is here: https://github.com/espressif/esp-lwip/pull/23

Please advise how to proceed in getting this merged and the submodule updated here. Thanks!

david-cermak:
@webworxshop Thanks for the report, good catch! And thanks for the proposed fix too, let's discuss the details in the linked PR -- I think we might try to address the issue without lwip modification if possible.

PR #4992, opened 2020-03-25
driver/gpio: Change log level to debug for gpio_config() (IDFGH-2959)

fri-sch:
Some applications need a frequent reconfiguration of gpios via
gpio_config(), for example to switch between input and output, etc.
In that case the console is spammed with debug messages. Silence
these message by changing the log level to "debug".

### DIFF
--- a/components/driver/gpio.c
+++ b/components/driver/gpio.c
@@ -361,7 +361,7 @@ esp_err_t gpio_config(const gpio_config_t *pGPIOConfig)
                 gpio_pulldown_dis(io_num);
             }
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@fri-sch Thanks for your contribution.

xiongyumail:
@fri-sch 
Hi,
We usually use `gpio_config()` when the program is initialized. If we want to switch input and output frequently, why not use `gpio_set_direction()`   ?

fri-sch:
> @fri-sch
> Hi,
> We usually use `gpio_config()` when the program is initialized. If we want to switch input and output frequently, why not use `gpio_set_direction()` ?

That is a very good objection. You're right indeed. I makes much more sense to use `gpio_set_direction()` instead of `gpio_config()` for such cases. I'll change my code and let's just forget about this PR.

PR #4904, opened 2020-03-07
esp_modem: Ensure uart_param_config and uart pins are set before uart… (IDFGH-2844)

AxelLin:
…_driver_install

Fixes Guru Meditation Error: Core 0 panic'ed (LoadProhibited) when config with
CONFIG_PM_ENABLE=y && CONFIG_PM_DFS_INIT_AUTO=y.

Signed-off-by: Axel Lin <axel.lin@gmail.com>

Issue #4830, opened 2020-02-24
Can't read any byte from serial  (IDFGH-2763)


romocitto88:
## Environment

- Development Kit:      ESP32-Dev-Kit
- Module or chip used:  ESP32-WROOM-32D
- IDF version latest stable version in PlatformIO
- Build System: idf.py
- Operating System: Windows
- Using an IDE?: Yes PlatformIO
- Power Supply: USB

## Problem Description

I tried all possible example code but I can't get any byte read from uart (I'm using uart2). 
Trasmitting is ok.
Below the code that i'm using.

### Code to reproduce this issue

```cpp

#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "esp_system.h"
#include "esp_log.h"
#include "driver/uart.h"
#include "string.h"
#include "driver/gpio.h"

static const int RX_BUF_SIZE = 1024;

#define TXD_PIN (GPIO_NUM_4)
#define RXD_PIN (GPIO_NUM_5)

void init(void) {
    const uart_config_t uart_config = {
        .baud_rate = 115200,
        .data_bits = UART_DATA_8_BITS,
        .parity = UART_PARITY_DISABLE,
        .stop_bits = UART_STOP_BITS_1,
        .flow_ctrl = UART_HW_FLOWCTRL_DISABLE,
        //.source_clk = UART_SCLK_APB,
    };
    // We won't use a buffer for sending data.
    uart_driver_install(UART_NUM_1, RX_BUF_SIZE * 2, 0, 0, NULL, 0);
    uart_param_config(UART_NUM_1, &uart_config);
    uart_set_pin(UART_NUM_1, TXD_PIN, RXD_PIN, UART_PIN_NO_CHANGE, UART_PIN_NO_CHANGE);
}

int sendData(const char* logName, const char* data)
{
    const int len = strlen(data);
    const int txBytes = uart_write_bytes(UART_NUM_1, data, len);
    ESP_LOGI(logName, "Wrote %d bytes", txBytes);
    return txBytes;
}

static void tx_task(void *arg)
{
    static const char *TX_TASK_TAG = "TX_TASK";
    esp_log_level_set(TX_TASK_TAG, ESP_LOG_INFO);
    while (1) {
        sendData(TX_TASK_TAG, "Hello world");
        vTaskDelay(2000 / portTICK_PERIOD_MS);
    }
}

static void rx_task(void *arg)
{
    static const char *RX_TASK_TAG = "RX_TASK";
[...]

PR #5017, opened 2020-03-28
Call uart_param_config before uart_driver_install (IDFGH-2985)

eliabieri:
In many examples, uart_driver_install is called before uart_param_config.
This leads to problems as can be seen in issues like #4830.

This will close PR #4904


Alvin1Zhang:
@eliabieri Thanks for contribution.

koobest:
Hi, @eliabieri 
Can you provide the IDF release version that encountered this issue? I have tried your example (https://github.com/espressif/esp-idf/issues/4830) on master branch but can not reproduce this issue. However, we want to make the two functions without the order of calling.

thanks ~

eliabieri:
I'm using ESP-IDF Release v4.0

PR #5052, opened 2020-04-02
Fix format string in websocket example (IDFGH-3025)

eliabieri:
Remove duplicate format character from websocket example_test python script.


Alvin1Zhang:
@eliabieri Thanks for your contribution.

PR #5106, opened 2020-04-12
[mbedtls] add configuration options for EC-JPAKE (IDFGH-3085)

wgtdkp:
This PR adds mbedTLS component configuration for `EC-JPAKE`. The `EC-JPAKE` definitions are by default commented out in mbedTLS's [default config.h](https://github.com/ARMmbed/mbedtls/blob/e62bdefce164cbdd44bcc8cacff5ee7d92819c40/include/mbedtls/config.h#L1107-L1124), but it is still able to turn it on with a user config file.

For the mbedTLS component built with esp-idf, we need the kconfig options.

## Background
EC-JAPKE stuffs are needed by the [Thread protocol](https://www.threadgroup.org/), and this PR helps support Thread on ESP32 boards.

### DIFF
--- a/components/mbedtls/Kconfig
+++ b/components/mbedtls/Kconfig
@@ -389,6 +389,13 @@ menu "mbedTLS"
             help
                 Enable to support ciphersuites with prefix TLS-ECDHE-RSA-WITH-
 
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@wgtdkp Thanks for contribution.

PR #5109, opened 2020-04-12
Missing SENDER_EMAIL in snprintf (428) (IDFGH-3088)

svenbieg:
I did not try the example, but snprintf should crash!

### DIFF
--- a/examples/protocols/smtp_client/main/smtp_client_example_main.c
+++ b/examples/protocols/smtp_client/main/smtp_client_example_main.c
@@ -425,7 +425,7 @@ static void smtp_client_task(void *pvParameters)
                    "From: %s\r\nSubject: mbed TLS Test mail\r\n"
                    "To: %s\r\n"
                    "MIME-Version: 1.0 (mime-construct 1.9)\n",
-                   "ESP32 SMTP Client", RECIPIENT_MAIL);
+                   "ESP32 SMTP Client", SENDER_MAIL, RECIPIENT_MAIL);
 
     /**
      * Note: We are not validating return for some ssl_writes.


Alvin1Zhang:
@svenbieg Thanks for noting this and the contribution, we will look into. Thanks.

vikramdattu:
@svenbieg The example already uses "ESP32 SMTP Client" as Sender.

Not needed to add actual email address in header. MAIL FROM already takes care of that.

svenbieg:
Yes, i know. The point is that the snprintf-function expects two string-parameters and only one is provided. The function will print whatever is on the stack. The program compiles but because variable arguments are not checked.

svenbieg:
I could successfully send an e-mail with this example via web.de, but i had to send another EHLO after the TLS-handshake. The line-endings are inconsistent too. I'm always using \r\n, and \r\n twice after the header.

vikramdattu:
@svenbieg 

```
    len = snprintf((char *) buf, BUF_SIZE,
                   "From: %s\r\nSubject: mbed TLS Test mail\r\n"
                   "To: %s\r\n"
                   "MIME-Version: 1.0 (mime-construct 1.9)\n",
                   "ESP32 SMTP Client", RECIPIENT_MAIL);
```

vikramdattu:
> I had to send another EHLO after the TLS-handshake.

This looks like some server specific behaviour to me.

> The line-endings are inconsistent too. I'm always using \r\n, and \r\n twice after the header.

Please let us know where you faced this issue. Please send an MR for this if you wish to.

svenbieg:
I'm sorry, i didn't see the comma between the strings.

PR #5135, opened 2020-04-16
Update idf_tools.py - double negation in info (IDFGH-3116)

lenhart:
Remove double negation in warning that tool is not supported.

(for me that is currently: "Not using an unsupported version of tool openocd-esp32 found in PATH: v0.10.0-esp32-20200309" after pulling from git today and ./install.sh)

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -1005,7 +1005,7 @@ def action_export(args):
                 continue
 
         if tool.version_in_path and tool.version_in_path not in tool.versions:
-            info('Not using an unsupported version of tool {} found in PATH: {}.'.format(
+            info('Not using a supported version of tool {} found in PATH: {}.'.format(
                  tool.name, tool.version_in_path) + prefer_system_hint, f=sys.stderr)
 
         version_to_use = tool.get_preferred_installed_version()


igrr:
I think the warning was correct... Looks like what happened is:

- you had a terminal where you executed . ./export.sh with an older version of IDF
- that step has added OpenOCD 20200309 to PATH
- you have pulled new version of IDF, ran install.sh. This has downloaded and installed the new version, v0.10.0-esp32-20200406
- you ran export.sh again. Then idf_tools.py noticed that there is already some 'openocd' in PATH, but it is not of a supported version (v0.10.0-esp32-20200406), and printed a warning about the fact that this unsupported version is not going to be used. It has then added new OpenOCD version's path the the beginning of PATH.

lenhart:
Hi igrr, thank you for the reply! Yes I can now see that reading it that way also makes sense.
There are two versions of openOCD in the .espressif folder and it happend after a git pull with python dependencies updated.
The warning came every time when starting a new terminal but has disappeared now. There is however only one openOcd version on PATH.

I wonder in which cases this error occurs. When using an old toolchain (maybe before restart/sourcing, missing parts?) with updated packages? If that is the single case it maybe could reflect that in the message, but I cannot come up with a general message which is a bit more precise... so thanks for your time and I'll close this then!

igrr:
I agree that the message should be different in two cases:

* there is a system-wide copy of the specific tool in the PATH (such as OpenOCD, if it is installed globablly) and we don't want to use it
* there is another version of the same tool, installed by install.sh, which we no longer want to use because the version has changed.

In the second case the message can be different or can be skipped entirely. I have recorded this suggestion and will add this in the next idf_tools update.

PR #5148, opened 2020-04-18
Fix typo with sigmadelta.h #ifdef (IDFGH-3129)

dowster:
I was experiencing a build error with an unmatched { bracket when including the sigmadelta.h file on a C++ project. It looks like a simple typo where instead of two underscores only one was provided. 

A quick scan of the rest of the esp-idf revealed this was the only occurrence of the typo.


Alvin1Zhang:
@dowster Thanks for contribution.

PR #5161, opened 2020-04-21
Update linenoise.c: calloc returning NULL is not handled (IDFGH-3144)

motla:
Calloc function tries to allocate 4096 bytes.
If such memory is not available, it returns NULL pointer.
This exception was not handled in the code, causing a dirty crash.

### DIFF
--- a/components/console/linenoise/linenoise.c
+++ b/components/console/linenoise/linenoise.c
@@ -1119,6 +1119,9 @@ int linenoiseHistoryLoad(const char *filename) {
         if (p) *p = '\0';
         linenoiseHistoryAdd(buf);
     }
+
+    free(buf);
     fclose(fp);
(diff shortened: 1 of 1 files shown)


boarchuz:
Perhaps take this opportunity to update here with the same calloc and null check:
https://github.com/espressif/esp-idf/blob/5214f40356c923189934ad6f15a20b1a9e58f307/components/console/linenoise/linenoise.c#L1105

Currently that's 4k on the stack.

igrr:
@motla Thanks for the PR, will get take it for merging!

motla:
@boarchuz I didn't saw that, great remark thank you. 
It seems that the original linenoise was not suited for embedded chips...
I just updated the file.
Thanks

Issue #2083, opened 2018-06-21
esp_ota_begin() starving network task of cpu time with large partitions and SPI RAM (IDFGH-261)


markwj:
We use 16MB ESP32 modules with 4MB OTA partitions and external SPI RAM. Currently our firmware image is approximately 2.8MB in size. The flow of our code is to make an http GET request for the firmware image, read the headers (in particular download size), call esp_ota_begin() with the firmware download size, then download chunk by chunk and call esp_ota_write for each chunk, and a final esp_ota_end when done. Our networking buffers are in INTERNAL RAM, not SPI RAM.

After our firmware exceeded about 2MB in size, and we started to use SPIRAM more in our application, we started to see random crashes during OTA firmware updates over wifi.

> abort() was called at PC 0x401b8e84 on core 0
> 0x401b8e84: pm_on_beacon_rx at ??:?
> 
> Backtrace: 0x40091e6b:0x3ffcc4a0 0x40091fc3:0x3ffcc4c0 0x401b8e84:0x3ffcc4e0 0x401b94ef:0x3ffcc520 0x401b9bd1:0x3ffcc550 0x40089e62:0x3ffcc5a0
> 
> 0x40091e6b: invoke_abort at /Users/mark/esp/esp-idf/components/esp32/panic.c:669
> 0x40091fc3: abort at /Users/mark/esp/esp-idf/components/esp32/panic.c:669
> 0x401b8e84: pm_on_beacon_rx at ??:?
> 0x401b94ef: ppRxProtoProc at ??:?
> 0x401b9bd1: ppRxPkt at ??:?
> 0x40089e62: ppTask at ??:?

We narrowed down the issue to using networking functions (reading from the TCP/IP socket) after calling esp_ota_begin with large image sizes (over approximately 2MB). The Espressif code calls a single esp_partition_erase_range() which disables the SPI RAM cache and blocks any task trying to access that. If the system networking task is blocked for too long, then it seems to get messed up in handling wifi beacons, and panics when it finally gets some CPU time?

If we change esp_ota_begin() to use a loop to erase the partition in 256KB chunks (calling esp_partition_erase_range() multiple times), with a 1 tick vTaskDelay between each chunk, the problem goes away and OTA works again. The vTaskDelay is required as without it the panic in pm_on_beacon_rx still happens.

[...]

PR #5171, opened 2020-04-23
Cherry-pick missing commit 7a2885885c25821d8985b6a97798956f12907b5a to master (IDFGH-3157)

Hallot:
See: https://www.esp32.com/viewtopic.php?f=13&t=14978

spi_flash: Add vTaskDelay while a long erasing
Added Kconfig options to enable yield operation during flash erase

Closes: https://github.com/espressif/esp-idf/issues/2083
Closes: IDFGH-261

### DIFF
--- a/components/spi_flash/Kconfig
+++ b/components/spi_flash/Kconfig
@@ -76,6 +76,28 @@ menu "SPI Flash driver"
         config SPI_FLASH_DANGEROUS_WRITE_ALLOWED
             bool "Allowed"
     endchoice
+   
+    config SPI_FLASH_YIELD_DURING_ERASE
+        bool "Enables yield operation during flash erase"
+        default y
+        help
+            This allows to yield the CPUs between erase commands.
+            Prevents starvation of other tasks.
+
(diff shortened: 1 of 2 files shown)


Hallot:
Signing CLA does not seem to work, I've done it multiple times already, to no avail.

mahavirj:
@Hallot We have fix for this already submitted in our internal merge requests queue. It will soon reflect to github as well. You can probable close this PR.

projectgus:
@Hallot ,

Thanks again for this, the internal fix for this should be pushed to GitHub very soon. (We cherry-picked this commit but also implemented this feature for the new SPI Flash API added in V4.0).

For the record:

> Signing CLA does not seem to work, I've done it multiple times already, to no avail.

It doesn't matter in this case, but for future reference this didn't work because the email address used for the commit isn't associated with your GitHub account (if you look at the commit shown above you can see that there's anonymous GitHub user shown as the committer). So the CLA bot has no way to know who the committer for this change is.

PR #5179, opened 2020-04-24
Fixes for -Wundef and -Wsigned-compare

tannewt:
Also ignores tools folders created during install.


tannewt:
I just updated this again.

Alvin1Zhang:
Thanks for the updates.

PR #5183, opened 2020-04-25
esp-netif: esp_netif_is_netif_up should also check netif_is_link_up (IDFGH-3171)

AxelLin:
Otherwise esp_netif_update_default_netif_lwip() may select wrong interface
as default route. e.g. calling esp_netif_is_netif_up() for ppp interface before
dialing up returns true without this patch then it's selected as the default
route due to higher priority.


Alvin1Zhang:
@AxelLin Thanks for your contribution. @david-cermak

PR #5216, opened 2020-04-30
mdns: Allow config mDNS task stack size (IDFGH-3219)

AxelLin:
When I check uxTaskGetStackHighWaterMark with my application I found mdns
task wastes some memory but the stack size is not configurable now.
Thus add this patch to allow config mdns task stack size.


Alvin1Zhang:
@AxelLin Thanks for your contribution.

PR #5228, opened 2020-05-01
Http sse example (IDFGH-3231)

nkorber:
Example http server implementing server-sent event


Alvin1Zhang:
@nkorber Thanks for your contribution.

shahpiyushv:
@nkorber , sorry for the late reply. This example looks a bit too complicated and could be hard for users to understand. Can you make the following changes
- Use C instead of C++ to be consistent with other examples
- Remove hardware dependent code, and instead use timers to generate events and prints to show that command was received from the browser.

I will add other comments on the actual code.

shahpiyushv:
@nkorber , sorry for the late reply. This example looks a bit too complicated and could be hard for users to understand. Can you make the following changes
- Use C instead of C++ to be consistent with other examples
- Remove hardware dependent code, and instead use timers to generate events and prints to show that command was received from the browser.

I will add other comments on the actual code.

nkorber:
@shahpiyushv : Thanks for your comments.
* Use timers : I have to think how to make the example still demonstrative without any hardware. Should be possible but I have not so much time now ; will need delay.
* Use C : Consistency, I understand but on the other hand we are in 21st century ; it will be time to evolve and C++ is IMHO the best successor since it provides a lot of improvements without loosing the flexibility of C.

shahpiyushv:
@nkorber , by timer, I just mean having a [periodic esp timer](https://github.com/espressif/esp-idf/tree/7d75213/examples/system/esp_timer) which will keep incrementing and reporting some counter, and the webpage will show the live counter value. You could even have an option to reset the counter from the browser, so that 2 way communication, like your original example had, can be demonstrated.

Regarding the point about C++, I had a discussion with IDF maintainers, and was told that CPP examples are fine.

Please also have a look at the contribution guidelines [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/creating-examples.html). I see that your PR has GPL license. As per the contribution guide

> All original example code has a license header saying it is “in the public domain / CC0”, and a warranty disclaimer clause. Alternatively, the example is licensed under Apache License 2.0. See existing examples for headers to adapt from.

Alvin1Zhang:
@nkorber Thanks for your contribution, any further updates regarding the PR? Thanks.

Alvin1Zhang:
Thanks for your contribution again, feel free to reopen when the suggested changes are done, thanks.

nkorber:
Hello, I know, it's a pity, but I have no time for this thing for the
moment. And since the work is quite old it's probably no more up to
date...
Sorry

Alvin1Zhang a écrit :

nkorber:
You're welcome.

Le 2023-07-17 16:44, Michael Hillpot a écrit :

PR #5246, opened 2020-05-06
Modify the OTA implementation to incrementally erase flash (IDFGH-3250)

mkellner:
This patch modifies the OTA implementation to incremental erase flash sectors as they are written, rather than bulk erasing them all at once when OT begins. This solves several problems:

- The bulk flash erase can take so long that it triggers the watchdog timer. Disabling the watchdog is an undesirable risk.
- The bulk erase interferes with the responsiveness of the system to user input.
- Incremental erasing eliminates the need to pass the firmware image size to esp_ota_begin to avoid erasing more flash sectors than required.
- When installing an OTA received via HTTP, the size of the firmware image is known when the content-length header is received. Calling esp_ota_begin at that time causes a long delay to do the bulk erase which causes the TCP transfer speed to be measurably slowed when the transfer resumes after the erase. Incremental erasing eliminates this TCP slowdown.

### DIFF
--- a/components/app_update/esp_ota_ops.c
+++ b/components/app_update/esp_ota_ops.c
@@ -48,7 +48,6 @@
 typedef struct ota_ops_entry_ {
     uint32_t handle;
     const esp_partition_t *part;
-    uint32_t erased_size;
     uint32_t wrote_size;
     uint8_t partial_bytes;
     uint8_t partial_data[16];
@@ -131,7 +130,6 @@ static esp_ota_img_states_t set_new_state_otadata(void)
 esp_err_t esp_ota_begin(const esp_partition_t *partition, size_t image_size, esp_ota_handle_t *out_handle)
 {
     ota_ops_entry_t *new_entry;
-    esp_err_t ret = ESP_OK;
 
     if ((partition == NULL) || (out_handle == NULL)) {
         return ESP_ERR_INVALID_ARG;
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@mkellner Thanks for your contribution.

mahavirj:
@mkellner Is it possible to keep this under some config option, like `CONFIG_APP_ERASE_DURING_WRITE` or similar? Reason being, there are some use-cases where OTA happens in non-sequential order (firmware blocks from cloud may receive out-of-order) and that time having (entire) partition erased prior to individual block write is desirable to have (other option being to handle that in `esp_ota_write` using some bit-map).

phoddie:
Hello. Thank you for the feedback.

I just learned of `esp_ota_write_with_offset` which appears to be part of a later IDF. That would indeed be incompatible with these changes. The documentation states:

> Use of esp_ota_write_with_offset() in combination with esp_ota_write() is not recommended

Given the presence of `esp_ota_write_with_offset`, a special value for the `image_size` argument to `esp_ota_begin` to indicate writes will be contiguous is a workable solution that will allow applications to opt-in. I'll prepare an updated patch for that.

phoddie:
Hello. We've made the changes requested. If there is any further impediment to integrating this change, please let me know. Otherwise, please let us know when we can expect to see this PR accepted. Thank you.

mahavirj:
@phoddie This change reverted earlier fix from https://github.com/espressif/esp-idf/commit/c82a4d59ee1564a2abad69b12f5a95ada8efdeed#diff-9135df8adbcedc507846e133b75e9f62. However, this now taken care along with one additional fix in `esp_ota_write_with_offset` API. Change will be integrated soon and that will auto update this PR as well. Thank you for contribution.

phoddie:
@mahavirj - That's great news. Thank you.

mahavirj:
Change has been merged with https://github.com/espressif/esp-idf/commit/3a768f51c3f932663af99384210d7d618ae885ea. Thank you.

PR #3290, opened 2019-04-10
Populate the TaskStatus_t.pxStackBase field (IDFGH-952)

rojer:

PR #5252, opened 2020-05-07
Set pxStackBase value in TaskStatus_t when filling task info (IDFGH-3258)

Hallot:
The value was never set, thus resulting in garbage when accessed.


Alvin1Zhang:
@Hallot Thanks for your contribution.

Hallot:
I just realised there already was a PR for this [here](https://github.com/espressif/esp-idf/pull/3290), but it has never been merged.
Feel free to close this one if needed.

AxelLin:
Current master tree is using freertos 10, so this does not apply to current master tree now.
Please re-check if the master tree still needs fix or not.

Hallot:
@AxelLin Master now properly sets it, l4134: `pxTaskStatus->pxStackBase = pxTCB->pxStack;`

Closing this now.

Issue #5193, opened 2020-04-26
idf.py double-loads "idf_ext.py" in project directory. (IDFGH-3193)


BrianPugh:
Starting [here](https://github.com/espressif/esp-idf/blob/master/tools/idf.py#L635), `idf.py` begins searching for and loading extensions. If the environment variable `IDF_EXTRA_ACTIONS_PATH` doesn't exist, it defaults to an empty string. When searching for python files ending in `'_ext'`, an empty string for a directory is interpretted as the current directory (which for many cases is the project directory). So, if the project has an `idf_ext.py`, it gets loaded here. A few lines down (under commend `# Load extensions from project dir`), `idf_ext.py` is searched for again and reloaded, duplicating the extension.

So there are a few solutions (I didn't open a PR because I'm not sure of external implications/backwards compatibility). I personally thing solution (1) is the cleanest and actually reduces the current amount of code:

1. Only append `extra_paths` if it's a non-empty. Not sure if other esp-idf elements relied on the current directory being added to the search path. I think this may have been the intended implementation, because it doesn't seem correct that we should only load the current directory if the environment variable `IDF_EXTRA_ACTIONS_PATH` is not set. Additionally, by default, the current directory is already in the python path.

```
# Make a set to automatically handle duplicate paths
extension_dirs = set([os.path.abspath(os.path.normpath(idf_py_extensions_path))])
extra_paths = os.environ.get("IDF_EXTRA_ACTIONS_PATH")
if extra_paths is not None:
    extension_dirs.update([os.path.abspath(os.path.normpath(path)) for path in extra_paths.split(';')])
```

In this event, its simpler and cleaner to append the `project_dir` to `extension_dirs` and get rid of the code under   `# Load extensions from project dir`:

```
extension_dirs.add(project_dir)
```

And then finally, convert it back to a list:

```
extension_dirs = list(extension_dirs)
```

2. If we do intend importing the current directory (which may be the project directory, this solution is absolutely backwards compatible), make sure all paths are absolute and normalized (i.e. `os.path.abspath(os.path.normpath(some_extension_dir))`). Prior to loading an extension, check if the absolute-normalized-path to the file is in a newly-created list-of-loaded-paths. If the extension is already loaded, don't reload it.

BrianPugh:
totally unrelated: the `run` method of `Task` would be more appropriate as the magic `__call__` method

Alvin1Zhang:
@BrianPugh Thanks for reporting, we will look into.

[...]

PR #5278, opened 2020-05-12
Fix double project idf_ext.py loading (IDFGH-3287)

BrianPugh:
Fixes the issue described in #5193 . I did not combine the project extension loading with the generic built-in extension loading since it'd be slightly backward incompatible (but would have reduced ~10 lines of code).

### DIFF
--- a/tools/idf.py
+++ b/tools/idf.py
@@ -635,10 +635,15 @@ def init_cli(verbose_output=None):
     all_actions = {}
     # Load extensions from components dir
     idf_py_extensions_path = os.path.join(os.environ["IDF_PATH"], "tools", "idf_py_actions")
-    extra_paths = os.environ.get("IDF_EXTRA_ACTIONS_PATH", "").split(';')
-    extension_dirs = [idf_py_extensions_path] + extra_paths
-    extensions = {}
+    extension_dirs = [realpath(idf_py_extensions_path)]
+    extra_paths = os.environ.get("IDF_EXTRA_ACTIONS_PATH")
+    if extra_paths is not None:
+        for path in extra_paths.split(';'):
+            path = realpath(path)
+            if path not in extension_dirs:
+                extension_dirs.append(path)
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@BrianPugh Thanks for your contribution.

kumekay:
Hi @BrianPugh 
Thank you for the PR!

I thought twice and realised that making path as a set is not a good idea, 
because extensions from IDF_EXTRA_ACTIONS_PATH may depend on some built-in extensions, but in a set order is not determined, so it's better to keep it as list to keep the order:

```python
    extension_dirs = [realpath(idf_py_extensions_path)]
    extra_paths = os.environ.get("IDF_EXTRA_ACTIONS_PATH")
    if extra_paths is not None:
        for path in extra_paths.split(';'):
            path = realpath(path) 
            if path not in extension_dirs:
                extension_dirs.append(path)

    extensions = {}
```

Could you please update the PR?

BrianPugh:
updated!

tools/idf.py
```diff
@@ -635,10 +635,16 @@ def parse_project_dir(project_dir):
     all_actions = {}
     # Load extensions from components dir
     idf_py_extensions_path = os.path.join(os.environ["IDF_PATH"], "tools", "idf_py_actions")
-    extra_paths = os.environ.get("IDF_EXTRA_ACTIONS_PATH", "").split(';')
-    extension_dirs = [idf_py_extensions_path] + extra_paths
-    extensions = {}
+    # Make a set to automatically handle duplicate paths
```
kumekay, line 638: This comment doesn't make sense anymore,
and could you remove it and squish all commits into one?

BrianPugh: fixed and squashed

BrianPugh:
merged in https://github.com/espressif/esp-idf/commit/3a6b8bb6e3116c8780b6a5c8638489451bbca51a

PR #5307, opened 2020-05-17
Added const for pointer in master_write method. (IDFGH-3324)

schaeferpm:
This time a clean commit.

Data shall not be modified in write method therfor the pointer should be marked as const.

Solved some const issues with the write method.


Alvin1Zhang:
@schaeferpm Thanks for your contribution.

Alvin1Zhang:
Thanks for contribution and sorry for slow turnaround, please help refer to https://github.com/espressif/esp-idf/commit/1aabab0160e8c87abc11c5e97ed88e50cb0b4461

PR #5309, opened 2020-05-18
Add include guards gatt mtu size (IDFGH-3327)

BMichiel:
I'm working on a bluetooth mesh application with the ESP32 (wroom32).
Because the app I am using does not handle segmentation properly during the provisioning I needed to change GATT_DEF_BLE_MTU_SIZE.

This was not possible without include guards.

### DIFF
--- a/components/bt/host/bluedroid/api/include/api/esp_gatt_common_api.h
+++ b/components/bt/host/bluedroid/api/include/api/esp_gatt_common_api.h
@@ -26,10 +26,14 @@ extern "C" {
 #endif
 
 // Maximum Transmission Unit used in GATT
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@BMichiel Thanks for contribution.

gengyuchao:
Why not use the function `esp_ble_gatt_set_local_mtu` to set the local mtu? This function is called to set local MTU before BLE connection.

BMichiel:
You are right, this was a misunderstanding on my part. I have now tried the `esp_ble_gatt_set_local_mtu` function and it seems to do what I needed. Thank you.

PR #5310, opened 2020-05-18
lwip: Add sio port for esp32 (IDFGH-3328)

jeandudey:
This adds a `sio` implementation that can be used by the SLIP interface on lwip (`slipif`), it's very basic, doesn't support aborting blocking reads (it's fine for it to be unimplemented).

It adds two additional functions that can be used to configure the `sio` UART port:

- `sio_set_baudrate`.
- `sio_set_pin`.

This was taken from #4985 to reduce the amount of code to review and also addresses some issues.

### DIFF
--- a/components/lwip/CMakeLists.txt
+++ b/components/lwip/CMakeLists.txt
@@ -89,6 +89,7 @@ set(srcs
     "port/esp32/debug/lwip_debug.c"
     "port/esp32/freertos/sys_arch.c"
     "port/esp32/netif/dhcp_state.c"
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
@jeandudey Thanks for your contribution.

david-cermak:
@jeandudey Nice work! Thanks for helping out with #4985, makes sense to address the sio port separately, CC @ryankurte.

I would prefer to integrate both changes at once, so will not merge it beforehand. Please keep this open until it gets merged/integrated, thanks!

Issue #5313, opened 2020-05-18
On windows using git bash or cmd, ninja monitor just terminate without any error message (IDFGH-3332)


callalilychen:
## Environment

- Development Kit: not relevant
- Kit version (for WroverKit/PicoKit/DevKitC): not relevant
- Module or chip used:  not relevant
- IDF version (run ``git describe --tags`` to find it): v4.0
- Build System:         CMake
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): not relevant
- Operating System:    Windows
- (Windows only) environment type:  ESP Command Prompt|Plain Command Prompt|git bash.
- Using an IDE?:  No
- Power Supply:  not relevant

## Problem Description

`ninja monitor` in build folder just terminate without any error message. No terminal for serial interface is started.
`idf.py monitor` in root folder works

### Expected Behavior
`ninja monitor` in build folder should also works like `idf.py monitor` in root folder 

### Actual Behavior
`ninja monitor` in build folder just terminate without any error message. No terminal for serial interface is started.

### Steps to reproduce
```
mkdir build
cd build
cmake ..
ninja flash
ninja monitor
```

### Code to reproduce this issue

It is reproducible to use a esp idf example

Alvin1Zhang:
@callalilychen Thanks for the report and letting us know about this. We will look into.

ESP-Marius:
As far as I can see `ninja monitor` works fine on master now. 

If anyone still has issues with this feel free to reopen the issue

PR #5314, opened 2020-05-18
call python explicitly and use command echo to stdout (IDFGH-3333)

callalilychen:
a possible solution to solve the issue #5313


Alvin1Zhang:
@callalilychen Thanks for your contribution.

renzbagaporo:
Hi @callalilychen, this seems to have lots of changes?

renzbagaporo:
Hi @callalilychen, this seems to have lots of unintended changes?

renzbagaporo:
Hi @callalilychen, this PR seems unintentionally includes other changes?

renzbagaporo:
Hi @callalilychen, unfortunately we can't use `COMMAND_ECHO` as it seems to be present on CMake 3.15 onwards  only. We currently guarantee support for version as low as CMake 3.5.

PR #5327, opened 2020-05-20
Bugfix: Casting int to size_t (IDFGH-3351)

FredrikFornstad:
Without this patch CMake complains and build fails since x is of type int and size is of type size_t .
See here: [libwebsockets commit](https://github.com/warmcat/libwebsockets/commit/fe7b6179901ddd592294391da5b6957cf8a1c285)


FredrikFornstad:
Commited from a temporary user… CLAassistant does not approve… Will make a new PR

PR #5328, opened 2020-05-20
Bugfix: Casting int to size_t (IDFGH-3352)

FredrikFornstad:
Without this patch CMake complains and build fails since x is of type int and size is of type size_t .
See here: [libwebsockets commit](https://github.com/warmcat/libwebsockets/commit/fe7b6179901ddd592294391da5b6957cf8a1c285)

### DIFF
--- a/components/soc/src/esp32/include/hal/cpu_ll.h
+++ b/components/soc/src/esp32/include/hal/cpu_ll.h
@@ -116,7 +116,7 @@ static inline void cpu_ll_set_watchpoint(int id,
 
     //We support watching 2^n byte values, from 1 to 64. Calculate the mask for that.
     for (int x = 0; x < 7; x++) {
-        if (size == (1 << x)) {
+        if (size == (size_t)(1 << x)) {
             break;
         }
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
@FredrikFornstad Thanks for your contribution.

projectgus:
Hi @FredrikFornstad,

Thanks for sending the PR.

ESP-IDF builds with [-Wextra -Wno-sign-compare](https://github.com/espressif/esp-idf/blob/master/tools/cmake/build.cmake#L106) (`-Wsign-compare` is enabled for C by `-Wextra` and for C++ by `-Wall`), which is why this warning doesn't appear in our build system. I'm guessing this flag is enabled by the libwebsockets build system.

However, in this case `cpu_ll.h` is included from public header files (FreeRTOS port-specific macros), so we'll add a CI check that all public headers can be fully `-Wextra`-clean.

Angus

renzbagaporo:
Thanks @FredrikFornstad. This PR has been submitted internally.

PR #5333, opened 2020-05-22
refactor: shortened expression (IDFGH-3357)

sio13:
- replaced with oneliner
- removed default arguments

I replaced the nested if statement with an "or assignment". This also reduces calling method getenv.  I also removed argument None, since it is default in getenv.


Alvin1Zhang:
@sio13 Thanks for your contribution.

Issue #352, opened 2017-02-18
undefined reference to spi_flash_disable_cache  undefined reference to spi_flash_restore_cache


ESP32DE:
hi 

try to use
https://github.com/espressif/esp-idf/blob/master/components/spi_flash/cache_utils.h
`#include "../cache_utils.h`

and try to use 
`spi_flash_disable_cache`
`spi_flash_restore_cache`

that are in cache_utils.c
https://github.com/espressif/esp-idf/blob/master/components/spi_flash/cache_utils.c#L35

```
uint32_t cpuid = 1;
uint32_t saved_state = 0;


extern void IRAM_ATTR spi_flash_disable_cache(uint32_t cpuid, uint32_t* saved_state);
extern void IRAM_ATTR spi_flash_restore_cache(uint32_t cpuid, uint32_t saved_state);

extern void spi_flash_disable_interrupts_caches_and_other_cpu();
extern void spi_flash_enable_interrupts_caches_and_other_cpu();

extern void spi_flash_disable_interrupts_caches_and_other_cpu_no_os();
extern void spi_flash_enable_interrupts_caches_no_os();



```
> CC sd_card.o
> AR libmain.a
> LD sd_card.elf
> C:/nod/27_sd_Card_iram/build/main\libmain.a(sd_card.o):(.iram1.literal+0x14): undefined reference to `spi_flash_disable_cache'
> C:/nod/27_sd_Card_iram/build/main\libmain.a(sd_card.o):(.iram1.literal+0x18): undefined reference to `spi_flash_restore_cache'
> C:/nod/27_sd_Card_iram/build/main\libmain.a(sd_card.o): In function `app_main':
> C:/nod/27_sd_Card_iram/main/sd_card.c:98: undefined reference to `spi_flash_disable_cache'
> C:/nod/27_sd_Card_iram/main/sd_card.c:100: undefined reference to `spi_flash_restore_cache'
> 

best wishes
rudi ;-)

ESP32DE:
```
extern void spi_flash_disable_interrupts_caches_and_other_cpu();
extern void spi_flash_enable_interrupts_caches_and_other_cpu();

extern void spi_flash_disable_interrupts_caches_and_other_cpu_no_os();
extern void spi_flash_enable_interrupts_caches_no_os();




/**
 * The following two functions are replacements for Cache_Read_Disable and Cache_Read_Enable
 * function in ROM. They are used to work around a bug where Cache_Read_Disable requires a call to
[...]

PR #5336, opened 2020-05-22
Changed ParseException message argument (IDFGH-3360)

the-cat-that-codes:
The 'message' attribute seems to be not usable since python 3.0 (see PEP #352 : https://www.python.org/dev/peps/pep-0352/#exception-hierarchy-changes)
but 'msg' is a valid attribute for pyparsing.ParseException.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -151,6 +151,7 @@ exclude =
         components/tinyusb,
         components/unity/unity,
         examples/build_system/cmake/import_lib/main/lib/tinyxml2,
+        examples/peripherals/secure_element/atecc608_ecdsa/components/esp-cryptoauthlib,
     # other third-party libraries
         tools/kconfig_new/kconfiglib.py,
         tools/kconfig_new/menuconfig.py,
--- a/tools/ldgen/generation.py
+++ b/tools/ldgen/generation.py
@@ -591,7 +591,7 @@ class SectionsInfo(dict):
         try:
             results = parser.parseString(first_line)
         except ParseException as p:
-            raise ParseException("Parsing sections info for library " + sections_info_dump.name + " failed. " + p.message)
+            raise ParseException("Parsing sections info for library " + sections_info_dump.name + " failed. " + p.msg)
 
         archive = os.path.basename(results.archive_path)
         self.sections[archive] = SectionsInfo.__info(sections_info_dump.name, sections_info_dump.read())


Alvin1Zhang:
@the-cat-that-codes Thanks for contribution.

renzbagaporo:
Hi @the-cat-that-codes, thanks for the PR! The changes here look good to me. Although the commit eaf45773f37564440379caf144f86b2c50e0bdbd seem to have the commit description cut off? 

Thanks for the heads up about PEP #352. I will probably end up fixing some more instances of `.message` use on top of your changes here!

the-cat-that-codes:
Hello @renzbagaporo,

I am sorry for the mistakes, it is my first collaboration on a public base. I will take more care in the future.
The commit eaf4577 was not planned to be in this PR. Initially, I wanted to propose another PR for this "issue" which is not really an issue at all. 
For the story, I have found the .message issue because my development system is in french and so there are some accented characters in some file paths. Due to these accented characters, the ParseException was raised but also an other one, the attribute error. I found that the attribute error has to be corrected. That was the goal of my first PR.
For the second (the commit eaf4577), I am not sure that supporting accented characters in path of libraries is required because it seems that it is not very common to use accented characters in directories and file paths (although in my opinion, in 2020, using any characters should not be a problem anymore :-). The message of this commit is : 

**_Allow accented characters in full filepaths of archives used in the generation._**

renzbagaporo:
@the-cat-that-codes Cleaning up this PR to only include the `.message()` fix and opening another PR for the accented characters fix is OK with me.

I see no harm in the accented characters change, and frankly, it's something that I haven't considered. So thanks for bringing this into our attention!

the-cat-that-codes:
@renzbagaporo Ok I will do the changes but  have you an advice to do it well (and in order to not introduced other errors in this PR). Maybe you have an hyperlink to follow for having a procedure to make the clean up of this PR safely ?

the-cat-that-codes:
Hi @renzbagaporo,

I have cleaned up the PR but I think there is an other problem since I am not the author of the commit c8f3ca7.
What I have done is :
- git pull in my fork
- rebase to drop the bad commits (the one with wrong email)
- and reset --soft the commit about the accented character (in order to keep the changes for an other PR)
- finally push on my own fork
Any ideas on what happened ?

renzbagaporo:
I'm not sure either. I can drop `c8f3ca7` in our internal submission if that's ok with you @the-cat-that-codes.

the-cat-that-codes:
@renzbagaporo Oh yes, sorry the late answer ... of course, you can drop the c8f3ca7 if you want.

PR #5338, opened 2020-05-23
ldgen: Support flags on mappings in linker fragments (IDFGH-3362)

TobleMiner:
This PR adds support for specifying per mapping flags in linker fragments.

The supported flags are:

* emit(symbol)
* align(alignment)
* keep

The following section describes the effect of each flag during linking:

#### emit(symbol)
Specifying `emit(symbol)` for a mapping causes ld to emit the symbols `__start_<symbol>` and `__stop_<symbol>` for start and end of the mapping, allowing them to be referenced in C source. This is useful for referencing mappings where the size is not known in advance. That could e.g. be a mapping constructed from object sections emitted by the compiler through the use of the `section(sectionname)` attribute.

#### align(alignment)
Adding `align(alignment)` to a mapping will cause its start to be aligned to `alignment` bytes. This is useful for all types of data requiring special alignment. It can also be used in conjunction with `emit` to ensure that an emitted symbol does actually line up with the start of the mappings data.

#### keep
The `keep` flag causes the generated mapping to be guarded by a `KEEP(...)` in the linker script. This can be important since `--gc-sections` will remove sections that do not contain any referenced symbols from the link.

All of the above is particularly useful for implementing dynamic initialization systems. I've set up a sample project showcasing this use here:
https://github.com/TobleMiner/dynamic-init-demo

### DIFF
--- a/docs/en/api-guides/linker-script-generation.rst
+++ b/docs/en/api-guides/linker-script-generation.rst
@@ -488,6 +488,39 @@ Example:
     entries:
         * (noflash)
 
+.. _ldgen-mapping-flags :
+
+**Mapping flags**
+
+Each mapping entry can be followed by a set of mapping flags, influencing linker script
+generation.
+Supported flags are:
+
+.. code-block:: none
+
+    keep
+    emit(symbol)
+    align(alignment)
+
+The :code:`keep` flag ensures that all sections matched by a mapping will be guarded with a
+:code:`KEEP(...)` in the resulting linker script. This protects the sections from being
+garbage collected, even if they do not specify any symbols required by the application.
+
+The :code:`emit(symbol)` flag causes the generated linker script to emit symbols for start
+and end of a mapping. The symbols will be called :code:`__start_symbol` and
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
@TobleMiner Thanks for contribution.

renzbagaporo:
Thanks for this substantial PR @TobleMiner! I am not able to review this immediately, but will be putting this in my review queue as this is also an internal `ldgen` improvement task.

TobleMiner:
@renzbagaporo Hey, any news on this topic?

renzbagaporo:
Hi @TobleMiner, I am currently working on implementing this feature. Unfortunately, I can't accept the PR as is, as the implementation requires other changes to pass internal tests and also for our other use case (I mentioned in my previous comment that this was also something we wanted internally).

Will be closing this, since there is already an issue that tracks this task internally.

TobleMiner:
Thanks for the update! Is there also a public issue tracking this feature?

PR #5373, opened 2020-06-01
Fix reserved psram region (IDFGH-3407)

negativekelvin:
If the user selects 8mb psram size, reserve region will fail because there is only 4mb available. There is a fixed address space for external ram, just reserve all of it.


Alvin1Zhang:
Thanks for your contribution.

Issue #5363, opened 2020-05-29
build: undefined reference to `_Unwind_Resume' (Arch Linux v4.0.1) (IDFGH-3394)


dshil:
## Environment

- Module or chip used:  ESP32-WROVER-B
- IDF version: v4.0.1
- Build System: idf.py (cmake - the same)
- Compiler version: xtensa-esp32-elf-gcc (crosstool-NG esp-2019r2) 8.2.0
- Operating System: Linux
- Using an IDE: No
- Power Supply: USB

## Debug Logs

```
/home/dshil/.espressif/tools/xtensa-esp32-elf/esp-2019r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/bin/ld: /home/dshil/.espressif/tools/xtensa-esp32-elf/esp-2019r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/lib/no-rtti/libstdc++.a(functexcept.o):(.literal._ZSt19__throw_logic_errorPKc+0x8): undefined reference to `_Unwind_Resume'
/home/dshil/.espressif/tools/xtensa-esp32-elf/esp-2019r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/bin/ld: /home/dshil/.espressif/tools/xtensa-esp32-elf/esp-2019r2-8.2.0/xtensa-esp32-elf/bin/../lib/gcc/xtensa-esp32-elf/8.2.0/../../../../xtensa-esp32-elf/lib/no-rtti/libstdc++.a(functexcept.o): in function `std::__throw_logic_error(char const*)':
/builds/idf/crosstool-NG/.build/xtensa-esp32-elf/src/gcc/libstdc++-v3/src/c++11/functexcept.cc:66: undefined reference to `_Unwind_Resume'
[...]

PR #5380, opened 2020-06-02
cxx: properly disable C++ exceptions mechanism (IDFGH-3417)

dshil:
The problem was appeared in #5363 

First I explored the git blame and find that in e0f7e196f `throw` stub were disabled for gcc8. I am not sure if disabling C++ exception was implemented correctly, probably nobody use C++ with ESP-IDF a lot, anyway we need to somehow disable exceptions since we are moving from v3.3.1 to latest stable release.

---- 

I used the workflow suggested by @igrr - to compile the source with exception enabled and see which symbols are used in project.map and then define missed symbols. It helped a little bit, but linker continued to fail.
 
Then I moved to the underlying of C++ exceptions and realized what else should be defined.
During exploration I referred to toolchain headers and the following links:

- [libc++ abi spec from LLVM](https://libcxxabi.llvm.org/spec.html)
- [C++ exception handling internals](https://monoinfinito.wordpress.com/series/exception-handling-in-c/)
- [Itanium C++ ABI: Exception handling](https://itanium-cxx-abi.github.io/cxx-abi/abi-eh.html)

I attached links to PR, thus somebody can use it later.

----

Relates: #459

### DIFF
--- a/components/cxx/cxx_exception_stubs.cpp
+++ b/components/cxx/cxx_exception_stubs.cpp
@@ -36,6 +36,36 @@ extern "C" void __cxx_fatal_exception_int(int i)
     abort();
 }
 
+namespace std {
+
+void __throw_bad_exception(void) __attribute__((alias("__cxx_fatal_exception")));
+void __throw_bad_alloc(void) __attribute__((alias("__cxx_fatal_exception")));
+void __throw_bad_cast(void) __attribute__((alias("__cxx_fatal_exception")));
+void __throw_bad_typeid(void) __attribute__((alias("__cxx_fatal_exception")));
+void __throw_logic_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_domain_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_invalid_argument(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_length_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_out_of_range(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_out_of_range_fmt(const char*, ...) __attribute__((alias("__cxx_fatal_exception_message_va")));
+void __throw_runtime_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_range_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_overflow_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_underflow_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_ios_failure(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void __throw_system_error(int) __attribute__((alias("__cxx_fatal_exception_int")));
+void __throw_bad_function_call(void) __attribute__((alias("__cxx_fatal_exception")));
+void __throw_future_error(int) __attribute__((alias("__cxx_fatal_exception_int")));
+
+void unexpected(void) __attribute__((alias("__cxx_fatal_exception")));
+void set_unexpected(void) __attribute__((alias("__cxx_fatal_exception")));
+bool uncaught_exception() __attribute__((alias("__cxx_fatal_exception_bool")));
+
+void terminate(void) __attribute__((alias("__cxx_fatal_exception")));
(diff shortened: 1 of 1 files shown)


components/cxx/cxx_exception_stubs.cpp
```diff
@@ -36,6 +36,36 @@ extern "C" void __cxx_fatal_exception_int(int i)
...
+void std::__throw_runtime_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void std::__throw_range_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void std::__throw_overflow_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void std::__throw_underflow_error(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void std::__throw_ios_failure(const char*) __attribute__((alias("__cxx_fatal_exception_message")));
+void std::__throw_system_error(int) __attribute__((alias("__cxx_fatal_exception_int")));
+void std::__throw_bad_function_call(void) __attribute__((alias("__cxx_fatal_exception")));
+void std::__throw_future_error(int) __attribute__((alias("__cxx_fatal_exception_int")));
+
+namespace std {
```
igrr, line 58: since you are adding `namespace std` here, perhaps you can fold the `std::__throw_*` definitions above into the namespace as well?

dshil: Done

dshil:
Hi @igrr, could you please check the latest version? The last question, since I am in process of moving from v3.3.1 to v4.0.1, in which version could I expect to see this patch?

0xjakob:
@dshil The change looks good. We'll add it to our internal CI system and proceed as with your other PR. We'll backport it to the older still supported versions (v4.0 is definitely in there). This normally doesn't take longer than a few weeks.

dshil:
@0xjakob or @igrr, any updates?

dshil:
@igrr the interesting thing, the linker error is happened only when I use cmake 3.17, with cmake 3.6 it links correctly. I did not try another cmake versions.

0xjakob:
@dshil Thanks for the update and sorry for the long response time. Indeed, I couldn't quite reproduce the linker error on my system. We would like to create a unit-test like source which exposes this linker error to reproduce the error before the fix and ensure that the change fixes the error and it stays fixed in the future.

Would you have a minimal working example which exposes this issue?

dshil:
Hi @0xjakob, thanks for coming back. I tried to prepare the minimal failed example but I wasn't succeeded. Since project is quite big it is not so trivial to figure out which part of code is responsible for linker error.

I understand your point about having the reproducible test to be able to avoid the same issues in the future.

I tried to build several examples with vanila v4.0.1 and I did not get any linker errors but when tried to build my own project I got the error.

Any ideas, suggestions?

0xjakob:
@dshil Could you try to compare the actual linker commands when running with cmake 3.6 and cmake 3.17?

0xjakob:
@dshil 
We decided to add your changes since another user has a related issues.

dshil:
Hi @0xjakob,

sorry for such a long response, did not have enough time to back to this issue. Another interesting thing is that I was only able to reproduce this LD failure when manually specify a list of used components as:

```cmake
set(COMPONENTS bootloader esptool_py ${MY_CUSTOM_PROJECT})
```

and all other components specify a list of required components as:

```cmake
set(COMPONENT_REQUIRES freertos bt esp_wifi bla bla bla)
```

When I removed it and allowed build tool to do all the stuff automatically (and magically :) ) the LD errors moved away...

I hope I will take look into this issue a little bit later.

dshil:
Hi @0xjakob, any updates? Last time you said that you would take this change? Have you changed the decision? If no do we still need to have this PR open?

0xjakob:
Hi @dshil unfortunately, we had some problems when integrating this, some other link errors emerged and we need to check them before we can add this PR. Sorry for the long delay.

0xjakob:
Hi @dshil. It's been a long time. We found another solution to the problem which is safer because it excludes the libgcc-unwind code responsible for the size increase directly, instead of trying the mock out functions which call that code as it was done before. So unfortunately, we couldn't merge your PR. Let us know if you need a patch, we can provide it until the fix gets uploaded to github.

If you have ideas for IDF in the future, however, you are more than welcome to submit more pull requests.

dshil:
Hi @0xjakob, thanks for coming back to this issue. Glad, that you finally was able to resolve the issue. I will close the PR.

PR #5439, opened 2020-06-12
Enable nesting of sub-components (IDFGH-3480)

KerryRJ:
In response to https://esp32.com/viewtopic.php?f=13&t=13991.


Alvin1Zhang:
Thanks for your contribution.

AshUK:
+1 would like to see this merged, we have projects with 100s of components and using PRIV_REQUIRES would not solve this issue

nicklasb:
@Alvin1Zhang @AshUK @KerryRJ 

I agree about nesting components being an important feature, it would make ESP-IDF way more scalable.
Right now, providing and configuring a larger and dynamic set of components, especially in conjunction with getting Menuconfig to work nicely, seems difficult, if even possible.

nicklasb:
BTW, I am having no problems adding nested components manually except for that not being maintainable ad redistributable, so I don't see the problem with adding this PR. 
Let's not hamper the growth of ESP-IDF by not adding features like this.

o-marshmallow:
Hello @KerryRJ ,

Thanks for your contribution, after discussing with the team, we decided to not merge this feature (under this form at least) because we think it raises more questions than it actually solves.

Indeed, currently in IDF, each component is treated and compiled as a C library. Let's say you have `components/comp1` in your project, after compilation this will actually compile into `libcomp1.a`. Now, if you have a nested component, let's say `components/comp1/components/subcomp1`, with your change, this will generate a new library named `libsubcomp1.a`, which is not related whatsoever to its parent component `comp1`.

However, if you happen to use a component that also have such sub-component, `components/comp2/components/subcomp1`, the generated library will also be named `libsubcomp1.a`, even thought the path was different and the "parent" component is different. Thus, you get a conflict.

The fact is even if we consider that IDF has such feature and prefixes the sub-components to get something like `libcomp1_subcomp1.a` and `libcomp2_subcomp1.a` after compilation, the new question that is raised is, do we consider sub-components as simply having a different "namespace" than regular libraries or do we consider that they depend on their parent component? In other words, do we consider that `libcomp1_subcomp1.a` is usable as a standalone library and `comp1` is only here as the name of the library? Or do we consider that `libcomp1_subcomp1.a` depends on its parent component and thus must only be linked if `libcomp1.a` is also linked into the final binary?


For these reasons, we decided not to implement this feature for the moment. You can still use the variable `EXTRA_COMPONENT_DIRS` offered by IDF CMake environment in order to manually add new components to the build.

Issue #5459, opened 2020-06-16
tinyusb component does not compile with c++ source files (IDFGH-3504)


sAlexander:
# Summary

The example provided in esp-idf/examples/peripherals/usb/tusb_sample_descriptor is great for getting started with tinyusb on the esp32s2. However, the example does not compile with C++ source files. 

# How to reproduce

Specifically, if you rename tusb_sample_descriptor.c to tusb_sample_descriptor.cc, add an `extern "C" { void app_main(void);}`, and update `CMakeList.txt`, the linking stage will fail due to duplicate symbols:

```
ld: esp-idf/main/libmain.a(tusb_sample_descriptor.cc.obj): multiple definition of `descriptor_str_kconfig';
ld: esp-idf/main/libmain.a(tusb_sample_descriptor.cc.obj): multiple definition of `descriptor_kconfig';
...
collect2: error: ld returned 1 exit status
```
These variables are all declared in `components/tinyusb/port/common/include/usb_descriptors.h`.

# The potential root cause

Running nm on the c vs cc generated objects shines some light onto the issue:

*c*:
```
xtensa-esp32s2-elf-nm tusb_sample_descriptor.c.obj 
00000000 T app_main
00000012 C descriptor_kconfig
0000001c C descriptor_str_kconfig
0000001c C descriptor_str_tinyusb
00000012 C descriptor_tinyusb
         U _esp_error_check_failed
         U esp_log_timestamp
         U esp_log_write
00000000 r __func__$6692
         U tinyusb_driver_install
         U tud_task
00000001 C tusb_desc_set
00000000 t usb_device_task
         U xTaskCreatePinnedToCore
```

*cc*:
```
xtensa-esp32s2-elf-nm tusb_sample_descriptor.cc.obj 
00000000 T app_main
00000000 B descriptor_kconfig
00000000 B descriptor_str_kconfig
00000000 B descriptor_str_tinyusb
00000000 B descriptor_tinyusb
         U _esp_error_check_failed
         U esp_log_timestamp
         U esp_log_write
         U tinyusb_driver_install
         U tud_task
00000000 B tusb_desc_set
         U xTaskCreatePinnedToCore
00000000 t _ZL15usb_device_taskPv
00000000 r _ZZ8app_mainE19__PRETTY_FUNCTION__

```

[...]

PR #5460, opened 2020-06-16
Add extern qualifier to variables in tinyusb/.../usb_descriptors.h (IDFGH-3505)

sAlexander:
As described in #5459, the tinyusb port does not correctly support c++ source files. Specifically, the linking stage fails due to duplicate symbols from usb_descriptors.h.

This change adds the extern qualifier to all variables in usb_descriptors.h to clarify that all definitions come from elsewhere (specifically, usb_descriptors.c). This is not required when all includers of usb_descriptors.h are c files, as c is compiled with -fcommon, but is required if any includer of usb_descriptors.h is a cc file, which is compiled with -fnocommon.


Alvin1Zhang:
@sAlexander Thanks for your contribution.

Alvin1Zhang:
Thanks for your contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/e97e7281b260cdd8b963cb503ad296d38cba348f.

PR #5516, opened 2020-06-29
Adds http_auth empty algorithm header check. (IDFGH-3569)

klisevich:
Receiving a response without an algorithm header leads to a crash.

### DIFF
--- a/components/esp_http_client/lib/http_auth.c
+++ b/components/esp_http_client/lib/http_auth.c
@@ -95,7 +95,7 @@ char *http_auth_digest(const char *username, const char *password, esp_http_auth
     }
 
     ESP_LOGD(TAG, "%s %s %s %s\r\n", "Digest", username, auth_data->realm, password);
-    if (strcasecmp(auth_data->algorithm, "md5-sess") == 0) {
+    if (auth_data->algorithm && strcasecmp(auth_data->algorithm, "md5-sess") == 0) {
         if (md5_printf(ha1, "%s:%s:%016llx", ha1, auth_data->nonce, auth_data->cnonce) <= 0) {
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97:
Hello @klisevich, Can you share a test URL where application would crash without this change?

klisevich:
> Hello @klisevich, Can you share a test URL where application would crash without this change?

Hi! Unfortunately, I can't because it is a corporative resource. But I can quote standard description where mentioned that the "algorithm" header is optional. 

>  **algorithm**
      A string indicating an algorithm used to produce the digest and an
      unkeyed digest.  If this is not present, it is assumed to be
      "MD5".  If the algorithm is not understood, the challenge SHOULD
      be ignored (and a different one used, if there is more than one).

[https://tools.ietf.org/html/rfc7616](url)

shubhamkulkarni97:
@klisevich Thanks for the info!

shubhamkulkarni97:
@klisevich, RFC states that if algorithm header is not present it should assumed to be MD5. Change in this PR is not according to what RFC states.

I'm attaching a patch below, can you check if it works for your use case.

[0001-Fix-issue-when-algotithm-header-is-not-present.patch.zip](https://github.com/espressif/esp-idf/files/5100907/0001-Fix-issue-when-algotithm-header-is-not-present.patch.zip)

If it works, we can close this PR.

Thanks,
Shubham

klisevich:
@shubhamkulkarni97, it works. Thanks.

Alvin1Zhang:
Thanks for contribution, will close now.

PR #5523, opened 2020-07-01
Improve error log description UART rx buffer size. (IDFGH-3579)

joncmaloney:
The under the error condition of rx buffer size is <=128 an error log is printed that reads uart rx buffer length error(>128). Propose an update to better describe the error condition as uart rx buffer length error(<=128).


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for contribution, please refer to https://github.com/espressif/esp-idf/commit/cef10fdfef533d8c95e555c440a1a45b407c9d8b

PR #5626, opened 2020-07-22
Release/v4.0 (IDFGH-3698)

elasticdotventures:
A repository referenced by libcoap hosted by eclipse is broken/down (for the past 23 days or so)
The libcoap repository needs to be fixed, however since the esp-idf release v4.0 points at a specific commit, even IF the repo is fixed it won't fix the IDF.

The esp-idf v4 needs to update it's reference to libcoap (which in turn needs to update it's reference to tinydtls)

https://github.com/eclipse/tinydtls

as a work around, i've forked the libcoap repo and fixed the upstread tinydtls url.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 [submodule "components/coap/libcoap"]
 	path = components/coap/libcoap
 	url = ../../elasticdotventures/libcoap.git
-        branch = dtls
+        branch = release-4.2.0
 
 [submodule "components/nghttp/nghttp2"]
 	path = components/nghttp/nghttp2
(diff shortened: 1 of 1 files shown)


mahavirj:
@elasticdotventures Can you please elaborate on which repository is broken or down? I am unable to see any problem with either `libcoap` or `tinydtls` repositories.

mahavirj:
@elasticdotventures Looking at this again, we decided to move submodule pointer for `libcoap` in `release/v4.0` to include fix from https://github.com/obgm/libcoap/commit/017528d7d96dca2318ed587b5fd3c0d57a89e49a. This should address the problem. I will go ahead and close this PR, `release/v4.0` will have this fix soon.

PR #5682, opened 2020-08-02
FreeRTOS: add mutex hold count to task status info (IDFGH-3762)

dexterbg:
This change enables applications to track down and debug stuck mutex priority inheritance and missing mutex release issues.

If `uxCurrentPriority` != `uxBasePriority` and `uxMutexesHeld` > 0, the task has taken a mutex which was then required by a higher priority task. That's normal FreeRTOS mutex priority inheritance, but should only occur temporarily. If `uxMutexesHeld` continues rising, the application has a bug in mutex management.

### DIFF
--- a/components/freertos/include/freertos/task.h
+++ b/components/freertos/include/freertos/task.h
@@ -181,6 +181,7 @@ typedef struct xTASK_STATUS
 	eTaskState eCurrentState;		/*!< The state in which the task existed when the structure was populated. */
 	UBaseType_t uxCurrentPriority;	/*!< The priority at which the task was running (may be inherited) when the structure was populated. */
 	UBaseType_t uxBasePriority;		/*!< The priority to which the task will return if the task's current priority has been inherited to avoid unbounded priority inversion when obtaining a mutex.  Only valid if configUSE_MUTEXES is defined as 1 in FreeRTOSConfig.h. */
+	UBaseType_t uxMutexesHeld;		/*!< The number of mutexes currently held by the task. Only valid if configUSE_MUTEXES is defined as 1 in FreeRTOSConfig.h. */
 	uint32_t ulRunTimeCounter;		/*!< The total run time allocated to the task so far, as defined by the run time stats clock.  See http://www.freertos.org/rtos-run-time-stats.html.  Only valid when configGENERATE_RUN_TIME_STATS is defined as 1 in FreeRTOSConfig.h. */
 	StackType_t *pxStackBase;		/*!< Points to the lowest address of the task's stack area. */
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/freertos/include/freertos/task.h
```diff
@@ -181,6 +181,7 @@ typedef struct xTASK_STATUS
 	eTaskState eCurrentState;		/*!< The state in which the task existed when the structure was populated. */
 	UBaseType_t uxCurrentPriority;	/*!< The priority at which the task was running (may be inherited) when the structure was populated. */
 	UBaseType_t uxBasePriority;		/*!< The priority to which the task will return if the task's current priority has been inherited to avoid unbounded priority inversion when obtaining a mutex.  Only valid if configUSE_MUTEXES is defined as 1 in FreeRTOSConfig.h. */
+	UBaseType_t uxMutexesHeld;		/*!< The number of mutexes currently held by the task. Only valid if configUSE_MUTEXES is defined as 1 in FreeRTOSConfig.h. */
```
uLipe, line 184: @dexterbg , could you please put this field around a configUSE_MUTEXES ifdef? Since this is more a debug feature, release applications may not use it, and it could be good to save more memory as possible on the TCBs.

dexterbg:
@uLipe I would already have done so, if `uxBasePriority` was conditional. Why isn't it? Shall I include it in the ifdef?

uLipe:
@dexterbg , the `uxBasePriority` were used to stuff related to the SMP and priority changing functions besides the mutex priority inheritance algorithm. So even if mutexes are not actually being used, these functions needs of this TCB field if I'm not missing something.

dexterbg:
@uLipe Take a look at the TCB definition:
https://github.com/espressif/esp-idf/blob/master/components/freertos/tasks.c#L192

There is no `uxBasePriority` even in the TCB without `configUSE_MUTEXES`. My assumption would be, this condition has been forgotten on the initial inclusion of `uxBasePriority` into `struct xTASK_STATUS`.

uLipe:
@dexterbg , I was also afraid this was left behind to its usage by espcoredump component (which currently has very strict rules about data structure size and depends on freertos tasks). 

After looking better, the implementation, leaving both fields without #ifdef will keep the freertos code close of the upstream version, so please just ignore my suggestion.

Thanks again for your contribution

sudeep-mohanty:
Hi [dexterbg](https://github.com/dexterbg),
Thank you for your contribution. Since this change modifies the task status structure, which is primarily used for debugging, we prefer not to incorporate it directly into our fork of FreeRTOS. We kindly request you to open a pull request to the upstream FreeRTOS repository with this change. Once it is accepted there, we will integrate the update into ESP-IDF accordingly. Closing this PR as of now.

PR #5683, opened 2020-08-02
pthread: fix pthread_mutex_destroy() breaking priority inheritance (IDFGH-3764)

dexterbg:
Using https://github.com/espressif/esp-idf/pull/5682, I found this bug in the POSIX thread mutex implementation: `pthread_mutex_destroy()` locks the mutex prior to deleting it, without doing an unlock before deletion. This behaviour breaks the FreeRTOS mutex priority disinheritance, the mutex hold count will get stuck.

### DIFF
--- a/components/pthread/pthread.c
+++ b/components/pthread/pthread.c
@@ -593,7 +593,12 @@ int pthread_mutex_destroy(pthread_mutex_t *mutex)
         return EBUSY;
     }
 
+    // FreeRTOS mutex must not be deleted while taken (breaks priority inheritance):
+    vTaskSuspendAll();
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

uLipe:
Hi @dexterbg , I found this issue was already fixed and is being tested internally and will be added to master branch soon. I'm sorry for did not notice this earlier.

Again thank you very much for your contribution :)

Alvin1Zhang:
The fix is here https://github.com/espressif/esp-idf/commit/202e6c8984255b551e29e3182e091cf7cd0153b0, thanks for contribution.

PR #5688, opened 2020-08-03
Update main.c (IDFGH-3769)

felmue:
Fix to make example work multiple times, i.e. every time GPIO0 changes from low to high or high to low the main processor wakes from deep sleep

### DIFF
--- a/examples/system/ulp_riscv/main/ulp/main.c
+++ b/examples/system/ulp_riscv/main/ulp/main.c
@@ -33,7 +33,6 @@ int main (void)
         if(gpio_level != gpio_level_previous) {
             gpio_level_previous = gpio_level;
             ulp_riscv_wakeup_main_processor();
-            break;
         }
 
     }


Alvin1Zhang:
Thanks for your contribution.

examples/system/ulp_riscv/main/ulp/main.c
```diff
@@ -33,10 +33,9 @@ int main (void)
         if(gpio_level != gpio_level_previous) {
             gpio_level_previous = gpio_level;
             ulp_riscv_wakeup_main_processor();
-            break;
```
uLipe, line 36: @felmue , the only problem by doing this is actually the ULP will neve enter in its low power mode. The current example only illustrates how the flow works by wakeup and then put all the systems in deep sleep.

I suggest you place a `ulp_riscv_shutdown()` somewhere inside the loop every time the processing ends.

felmue:
Hi @uLipe

thank you very much for getting back to me. I appreciate it. I tried your suggestion and added `ulp_riscv_shutdown()` instead of the `break` statement. The result is the same as with the original, unmodified example - the main processor is woken up once, then ULP enters low power mode and the main processor goes into deep sleep.

Since the example is meant to do exactly that, my pull request is obsolete and I will close it.

Thanks
Felix

Issue #3784, opened 2019-07-16
WPA2: Incorrect initialisation behind slow WiFi connection times? (IDFGH-1520)


boarchuz:
Slow WPA2 connection times have long been an issue for many ESP32 users. I think I may have a small clue as to where to look for the problem, and a bandaid to fix it for now.
Today I was able to reduce connection time on my home WPA2 network from the usual ~1100ms down to ~350ms (240MHz, CONFIG_LWIP_DHCP_DOES_ARP_CHECK=0). Those times are using Arduino's millis(), the debug output below might have more meaningful timestamps.
All it took was this seemingly redundant fix:
```
wifi_config_t conf;
esp_wifi_get_config(WIFI_IF_STA, &conf);
esp_wifi_set_config(WIFI_IF_STA, &conf);
```

Often I get this additional debug output (the indicated line):
Without fix:
```
I (722) wifi: n:11 0, o:1 0, ap:255 255, sta:11 0, prof:1
D (1472) nvs: nvs_set_blob sta.apinfo 700[1B][0m           <-------
I (1522) wifi: state: init -> auth (b0)
I (1532) wifi: state: auth -> assoc (0)
I (1542) wifi: state: assoc -> run (10)
I (1582) wifi: connected with MYSSID, channel 11
D (1582) nvs: nvs_set sta.chan 1 11[1B][0m
D (1582) nvs: nvs_set_blob sta.apinfo 700[1B][0m
I (1592) wifi: pm start, type: 1

D (1592) event: SYSTEM_EVENT_STA_CONNECTED
```
With fix, for comparison:
```
I (722) wifi: n:11 0, o:1 0, ap:255 255, sta:11 0, prof:1
I (722) wifi: state: init -> auth (b0)
I (732) wifi: state: auth -> assoc (0)
I (732) wifi: state: assoc -> run (10)
I (782) wifi: connected with MYSSID, channel 11
D (782) nvs: nvs_set sta.chan 1 11[1B][0m
D (782) nvs: nvs_set_blob sta.apinfo 700[1B][0m
I (792) wifi: pm start, type: 1

D (792) event: SYSTEM_EVENT_STA_CONNECTED
```
That additional "nvs_set_blob sta.apinfo" doesn't seem to occur every time though, so maybe a red herring. The ~750ms mysterious delay unfortunately does. 

[...]

PR #5708, opened 2020-08-07
[WIP] Keep PHY calibration data in RTC memory (IDFGH-3790)

boarchuz:
In designing an ultra low power sensor node that periodically transmits data to a gateway via ESP-NOW, I found that the active time could be effectively _halved_ by avoiding the need for NVS initialization and/or PHY calibration on wakeup by keeping PHY calibration data in RTC memory. In combination with existing config optimisations, the total time for an ESP-NOW transmission, from deep sleep to deep sleep, can be reduced to as little as 39ms (from 85ms without this change), allowing _up to 50% longer battery life_ (@ 1 Tx/5mins).

ESP-NOW transmission:

rtc_phy_data

Applications targeting very fast WiFi connection times could also see some benefit from this.

For demonstration, I've got one implementation here, where esp_wifi claims a chunk of RTC memory internally, however it is incomplete (eg. no way to access the data to save it manually or via esp_phy_store_cal_data_to_nvs) and I'm open to better ways.

### DIFF
--- a/components/esp_wifi/Kconfig
+++ b/components/esp_wifi/Kconfig
@@ -326,7 +326,7 @@ menu "PHY"
         
     config ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC
         bool "Keep PHY calibration data in RTC memory"
-        default y
+        default n
         help
             Enabling this option keeps PHY calibration data in RTC memory where it will persist across deep sleep 
             wakeup cycles, allowing optimization of WiFi initialization upon wakeup by avoiding the need to 
--- a/components/esp_wifi/Kconfig
+++ b/components/esp_wifi/Kconfig
@@ -323,6 +323,35 @@ menu "PHY"
             1.If your board is easy to be booted up with antenna disconnected.
             2.Because of your board design, each time when you do calibration, the result are too unstable.
             If unsure, choose 'y'.
+        
+    config ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC
+        bool "Keep PHY calibration data in RTC memory"
+        default y
+        help
+            Enabling this option keeps PHY calibration data in RTC memory where it will persist across deep sleep 
+            wakeup cycles, allowing optimization of WiFi initialization upon wakeup by avoiding the need to 
+            initialize NVS and load PHY data into memory.
+
+    choice ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC_LOC
+        depends on ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC
+        prompt "PHY calibration RTC memory location"
+        default ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC_LOC_SLOW
+        help
+            Select whether to keep PHY calibration data in RTC Slow Memory or RTC Fast Memory.
+
+            As RTC memory is limited, the region may be selected depending on the pressures of the particular 
+            application:
+
+            - Select RTC Slow Memory if available Fast Memory is low, or APP CPU may require access to PHY data.
+            
+            - Select RTC Fast Memory if available Slow Memory is low (eg. in use by ULP), and/or all accesses
+            (including WiFi initialization) are restricted to PRO CPU.
+
+        config ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC_LOC_SLOW
+            bool "RTC Slow Memory"
+        config ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC_LOC_FAST
+            bool "RTC Fast Memory"
+    endchoice
 
     menuconfig ESP32_PHY_INIT_DATA_IN_PARTITION
         bool "Use a partition to store PHY init data"
--- a/components/esp_wifi/src/phy_init.c
+++ b/components/esp_wifi/src/phy_init.c
@@ -846,7 +846,7 @@ void esp_phy_load_cal_and_init(phy_rf_module_t module)
     uint8_t sta_mac[6];
 
     bool rtc_cal_data_valid =   rtc_get_reset_reason(0) == DEEPSLEEP_RESET &&
-                                ESP_OK == phy_check_rf_cal_version(&cal_data->version) &&
+                                ESP_OK == phy_check_rf_cal_version((uint32_t*)&cal_data->version) &&
                                 ESP_OK == phy_check_rf_cal_mac(cal_data->mac);
     
 #ifdef CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_STORAGE // RTC + NVS
--- a/components/esp_wifi/src/phy_init.c
+++ b/components/esp_wifi/src/phy_init.c
@@ -147,6 +147,19 @@ static phy_country_to_bin_type_t s_country_code_map_type_table[] = {
     {"US",  ESP_PHY_INIT_DATA_TYPE_FCC},
 };
 #endif
+
+#if CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_IN_RTC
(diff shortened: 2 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

AshUK:
@boarchuz would this optimisation also work for wifi STA to AP ?

boarchuz:
@AshUK I think so, but it's a lot more complexity for a proportionally much smaller benefit.
As I understand, a lot of the hard work in authenticating with a secured access point is saved to flash upon initial connection, making every subsequent connection much faster by loading it back up rather than doing all that processing from scratch every time. (ie. NVS actually saves a lot of time.)
So if it's an open network, then I don't see any issues; otherwise, it might be possible to initialize WiFi and start connecting, then initialize NVS while still scanning for the access point. Hacky and probably not worth the 30-40ms shaved off, though.

boarchuz:
This can already be achieved with the current API by using esp_phy_rf_init before WiFi initialisation.

eg.

```
static esp_err_t init_phy(esp_phy_calibration_mode_t cal_mode)
{
    static RTC_DATA_ATTR esp_phy_calibration_data_t s_rtc_cal_data;
    const esp_phy_init_data_t* init_data = esp_phy_get_init_data();
    assert(init_data);
    return esp_phy_rf_init(init_data, cal_mode, &s_rtc_cal_data, PHY_WIFI_MODULE);
}

void app_main(void)
{
  if (esp_reset_reason() != ESP_RST_DEEPSLEEP) {
    // Full calibration on reset
    ESP_ERROR_CHECK( init_phy(PHY_RF_CAL_FULL) );
  }
  else
  {
    // No calibration on deep sleep wakeup, just set phy data
    ESP_ERROR_CHECK( init_phy(PHY_RF_CAL_NONE) );
  }
  // WiFi initialisation, ESP NOW initialisation, etc
}
```

FruitieX:
esp_phy_rf_init no longer exists in recent ESP-IDF (v4.3), if somebody winds up here looking for a solution for fast ESP-NOW init after deep sleep I have a working example here:

Calibration on bootup:
https://github.com/FruitieX/esp32-volume-control-sender/blob/22aec694bc168d854eaab43edd34de7ebd4a88f4/main/ulp_adc_example_main.c#L288-L297

Load calibration data from RTC memory after deep sleep wakeup:
https://github.com/FruitieX/esp32-volume-control-sender/blob/22aec694bc168d854eaab43edd34de7ebd4a88f4/main/ulp_adc_example_main.c#L312-L330

You also need to disable `CONFIG_ESP32_WIFI_NVS_ENABLED`, `CONFIG_ESP32_PHY_CALIBRATION_AND_DATA_STORAGE` and `CONFIG_ESP32_PHY_INIT_DATA_IN_PARTITION` in sdkconfig.

2opremio:
Did you manage to fully cache the Wifi STA info into RTC?

I want to fully disable NVS to save energy (its initialization and access take a big chunk of time from the deep-sleep wakeups of my low-power application)

It seems like supplying the bssid and channel to `wifi_init_config_t` isn't enough. After reading https://github.com/espressif/esp-idf/issues/3784 from @boarchuz, it seems like the missing piece is somehow caching the PMK in RTC and restoring it on every wakeup. 

I bet that that information is in the infamous `apinfo` NVS key, but there doesn't seem to be an API which would allow to store it in RTC memory.

boarchuz:
@2opremio 
Yes, I went the route of modifying all nvs_* pointers in g_wifi_osi_funcs to custom functions that store and retrieve from RTC memory instead.

2opremio:
Ah, nice! Would you be able to share the code?

2opremio:
Also, it would be awesome if it could be upstreamed!

boarchuz:
Sure, you can find the component "bo_wsc" here: https://github.com/boarchuz/bo_wsc
You might run into some minor issues as I only had my own project in mind when writing/using it. In particular, I'm on ESP-IDF 5.1.0 and set CONFIG_BO_WSC_NVS_DISABLED.

You just need to do `ESP_ERROR_CHECK(bo_wsc_set(&g_wifi_osi_funcs));` before any WiFi initialisation to replace the default functions, and then later (eg. when returning to sleep) use the `bo_wsc_nvs_...` functions to commit changes to flash if desired.

2opremio:
@boarchuz wow! such a nice generic solution.

In the meantime I have managed to solve my specific problem in a simpler way. As I expected, the issue was coming from the WPA PMK not being cached. This patch (against idf 5.3) solves it:

```diff
diff --git a/components/wpa_supplicant/src/rsn_supp/wpa.c b/components/wpa_supplicant/src/rsn_supp/wpa.c
index 4e4e189363..bb0fc4488c 100644
--- a/components/wpa_supplicant/src/rsn_supp/wpa.c
+++ b/components/wpa_supplicant/src/rsn_supp/wpa.c
@@ -2483,6 +2483,11 @@ wpa_set_passphrase(char * passphrase, u8 *ssid, size_t ssid_len)
         sm->key_mgmt == WPA_KEY_MGMT_DPP)
         return;
 
+    // PMK-cache in RTC memory (so that it survives deep-sleeps)
+    static RTC_DATA_ATTR u8 cached_ssid[SSID_MAX_LEN] = {0};
+    static RTC_DATA_ATTR size_t cached_ssid_len = 0;
+    static RTC_DATA_ATTR u8 cached_pmk[PMK_LEN] = {0};
+
     /* This is really SLOW, so just re cacl while reset param */
     if (esp_wifi_sta_get_reset_param_internal() != 0) {
         // check it's psk
@@ -2491,8 +2496,13 @@ wpa_set_passphrase(char * passphrase, u8 *ssid, size_t ssid_len)
                            esp_wifi_sta_get_ap_info_prof_pmk_internal(), PMK_LEN) != 0)
                 return;
         } else {
-            pbkdf2_sha1((char *)esp_wifi_sta_get_prof_password_internal(), sta_ssid->ssid, (size_t)sta_ssid->len,
-                        4096, esp_wifi_sta_get_ap_info_prof_pmk_internal(), PMK_LEN);
+            if (cached_ssid_len == ssid_len && memcmp(ssid, cached_ssid, ssid_len) == 0) {
+                // Use RTC-cached pmk
+                memcpy(esp_wifi_sta_get_ap_info_prof_pmk_internal(), cached_pmk, PMK_LEN);
+            } else {
+                pbkdf2_sha1((char *)esp_wifi_sta_get_prof_password_internal(), sta_ssid->ssid, (size_t)sta_ssid->len,
+                            4096, esp_wifi_sta_get_ap_info_prof_pmk_internal(), PMK_LEN);
+            }
         }
         esp_wifi_sta_update_ap_info_internal();
         esp_wifi_sta_set_reset_param_internal(0);
@@ -2503,6 +2513,10 @@ wpa_set_passphrase(char * passphrase, u8 *ssid, size_t ssid_len)
     } else {
         memcpy(sm->pmk, esp_wifi_sta_get_ap_info_prof_pmk_internal(), PMK_LEN);
         sm->pmk_len = PMK_LEN;
+        // Save pmk to cache
+        memcpy(cached_ssid, ssid, ssid_len);
+        cached_ssid_len = ssid_len;
+        memcpy(cached_pmk, sm->pmk, PMK_LEN);
     }
 #ifdef CONFIG_IEEE80211R
     /* Set XXKey to be PSK for FT key derivation */
```

It sucks that there isn't (or I havent'been able to find) a way to set/cache the PMK through `wifi_config_t` 

I may upstream my change in a PR.

@boarchuz Nonetheless I may still use your solution. However, I am weary of it's RTC memory consumption since it stores all keys (I assume your approach is all-or-nothing, since the driver won't likely accept only certain keys being present)

PR #5744, opened 2020-08-17
Correct export.sh failure exit code and cleanup

jtchitty:
In failure cases, when idf_export_main() returns 1 (failure),
the export.sh script returns 0 (success). This makes it very
difficult to detect failure when writing scripts that use
export.sh. Furthermore, idf_export_main() does not clean up all
internal variables and functions in failure cases.

Move all cleanup steps into a cleanup function and pass
the return value from idf_export_main() to the cleanup
function so it can return with that same return value.

### DIFF
--- a/export.sh
+++ b/export.sh
@@ -118,17 +118,6 @@ idf_export_main() {
         echo "  ${PATH}"
     fi
 
-    # Clean up
-    unset old_path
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

PR #5761, opened 2020-08-20
Added option to use datetime stamp during logging.

Molorius:
Added option to use datetime stamp during logging.

Also changed log timestamp formatting to use unsigned int, allowing for smaller buffer.

### DIFF
--- a/components/log/Kconfig
+++ b/components/log/Kconfig
@@ -70,4 +70,12 @@ menu "Log output"
             bool "System Time"
     endchoice
 
+    config LOG_TIMESTAMP_SOURCE_SYSTEM_DATETIME
+        bool "Log Date and Time"
+        default "n"
+        depends on LOG_TIMESTAMP_SOURCE_SYSTEM
+        help
+            Use full date and time in logs, rather than just time.
+            e.g. (1990-01-31 00:01:30.000)
+
 endmenu
--- a/components/log/log_freertos.c
+++ b/components/log/log_freertos.c
@@ -52,7 +52,11 @@ void esp_log_impl_unlock(void)
 
 char *esp_log_system_timestamp(void)
 {
-    static char buffer[18] = {0};
+#if CONFIG_LOG_TIMESTAMP_SOURCE_SYSTEM_DATETIME
+    static char buffer[24] = {0};
+#else
+    static char buffer[13] = {0};
+#endif
     static _lock_t bufferLock = 0;
 
     if (xTaskGetSchedulerState() == taskSCHEDULER_NOT_STARTED) {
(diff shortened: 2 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/log/Kconfig
```diff
@@ -70,4 +70,12 @@ menu "Log output"
             bool "System Time"
     endchoice
 
+    config LOG_TIMESTAMP_SOURCE_SYSTEM_DATETIME
```
projectgus, line 73: Relatively minor, but for ease of use suggest adding this as a "choice" so the choices become "Milliseconds Since Boot", "System Time", "System Date and Time"

components/log/log_freertos.c
```diff
@@ -52,7 +52,11 @@ void esp_log_impl_unlock(void)
 
 char *esp_log_system_timestamp(void)
```
projectgus, line 53: The description of this function in esp_log.h needs to be updated to match

projectgus (changes requested):
Hi @Molorius ,

Thanks for sending this. I have a couple of minor comments but the change looks very useful to me, look forward to integrating it.

components/log/log_freertos.c
Molorius: Would you prefer changing the name of the function or splitting it into esp_log_system_timestamp and esp_log_system_datetimestamp? The xTaskGetSchedulerState portion would be put into a static function for use between both.

projectgus: @Molorius Good point! That is probably a cleaner solution, thank you for pointing it out.

However if it becomes less clean to implement in some way, I'm OK with the current implementation - it just needs to be noted in the header that it might return a date+time depending on config.

components/log/Kconfig
Molorius: It is now another choice in the kconfig.

components/log/log_freertos.c
Molorius: I separated the datetime elements into a separate function and updated esp_log.h to match.

projectgus:
Hi @Molorius ,

Thanks for being patient while I got back to you. This looks great to me, I've squashed the commits and pushed it into our internal review & merge queue. The PR will be updated once it has merged.

Angus

projectgus (approved):

projectgus:
As it happens, at the last minute we've decided to try and implement this in a different way. Rather than having to add a whole new function and kconfig item for each new format, we're going to try to find an easier way to customize the log timestamp.

You don't need to do anything, this PR will still be updated once this functionality is available.

PR #5777, opened 2020-08-23
Avoid unnecessary repetition of RTC slow clock initialisation for faster wakeup (IDFGH-3876)

boarchuz:
With ESP32S2 especially, most of the time from deep sleep wakeup to app_main entry is spent in delays, waiting for RTC slow clock calibration (regardless of CONFIG_ESP32S2_RTC_CLK_CAL_CYCLES). The main culprit is rtc_clk_set_xtal_wait - 
https://github.com/espressif/esp-idf/blob/c77c4ccf6c43ab09fd89e7c907bf5cf2a3499e3b/components/soc/src/esp32s2/rtc_clk.c#L170

2000 clock cycles at 90kHz is 22ms each, and that runs several times, in addition to the user configured calibration cycles. This is entirely unnecessary: I doubt that xtal_wait needs that level of calibrated precision, slow clock is incapable of such precision anyway, there is no need for it to be repeated multiple times in quick succession, and there's no pressing need to repeat every wakeup.

By calibrating only when necessary, ESP32 wakeup to app_main entry is reduced from **27ms to 17ms**; more significantly, ESP32-S2 is reduced from **90ms to 17ms**.

Additionally, this will set pll_wait and ck8m_wait when RTC slow clock changes (there was only a reference to 1ms for xtal_wait; I used 200us for the other two values based loosely on defaults).

(I have not tested with external 32k crystal)


Alvin1Zhang:
Thanks for your contribution.

boarchuz:
Closing as https://github.com/espressif/esp-idf/commit/622593220102e88860a722a25c1180c3e5e54793 adds weak linkage to esp_clk_init, which coincidentally offers a neat way to provide a custom init as well.

PR #5788, opened 2020-08-25
esp_hid: make stack size adjustable using Kconfig (IDFGH-3889)

v1993:
This allows users to specify stack size for HID application callbacks. Default value is based on one that previously was hardcoded.

Note: it is possible that default value is too small for esp_hid_host example. I can't test it right now, but it is worth double checking.

### DIFF
--- a/components/esp_hid/Kconfig
+++ b/components/esp_hid/Kconfig
@@ -1,0 +1,7 @@
+menu "HID Support"
+
+    config HID_TASK_STACK_SIZE
+        int "HID event (callback to application) stack size"
+        default 2048
+
+endmenu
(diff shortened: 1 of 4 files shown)


me-no-dev:
Could you please change it to `HID_EVENT_TASK_STACK_SIZE`?

v1993:
@me-no-dev No problem, but FYI I've based this naming off Bluetooth's `BT_BTC_TASK_STACK_SIZE` option.

me-no-dev:
Thanks @v1993 :)

igrr:
@v1993 sorry that it took so long to handle this. We have merged https://github.com/espressif/esp-idf/pull/6385 which is very similar to your change, but makes run-time configuration of the stack size possible, rather than keeping it compile-time only. It also fixes the "default value too small" issue you have mentioned.

Issue #5787, opened 2020-08-25
VSCode extension installs incompatible version of gdbgui (IDFGH-3888)


istokm:
- IDF version (run ``git describe --tags`` to find it): **v4.3-dev-907-g6c17e3a64**
- Build System: **idf.py**
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): **xtensa-esp32-elf-gcc.exe (crosstool-NG esp-2020r2) 8.2.0**
- Operating System: **Windows**
- Environment type: **ESP Command Prompt** | **Powershell**
- Using an IDE?: **Visual Studio Code w/ ESP-IDF extension**

## Problem Description

After a clean installation through the ESP-IDF's onboarding the builtin VSCode IDF terminal can't run anything related to `idf.py` due to a package error:
```PS D:\Infotech\Projects\ESP32-IDF\Projects\flowrack-mechanical> idf.py
The following Python requirements are not satisfied:
gdbgui>=0.13.2.0
To install the missing packages, please run "D:\Infotech\Projects\ESP32-IDF\esp-idf\install.bat"
Diagnostic information:
    IDF_PYTHON_ENV_PATH: D:\Infotech\Projects\ESP32-IDF\.espressif\python_env\idf4.3_py3.8_env
    Python interpreter used: D:\Infotech\Projects\ESP32-IDF\.espressif\python_env\idf4.3_py3.8_env\Scripts\python.exe

ESP-IDF v4.3-dev-907-g6c17e3a64

PS D:\Infotech\Projects\ESP32-IDF\Projects\flowrack-mechanical> pip show gdbgui
Name: gdbgui
Version: 0.14.0.0
Summary: Browser-based frontend to gdb. Debug C, C++, Go, or Rust.
Home-page: https://github.com/cs01/gdbgui
Author: Chad Smith
Author-email: grassfedcode@gmail.com
License: License :: GNU GPLv3
Location: d:\infotech\projects\esp32-idf\.espressif\python_env\idf4.3_py3.8_env\lib\site-packages
Requires: Flask-SocketIO, eventlet, pygdbmi, gevent-websocket, Flask-Compress, Pygments, gevent, Flask
Required-by:

PS D:\Infotech\Projects\ESP32-IDF\Projects\flowrack-mechanical> pip show pygdbmi
Name: pygdbmi
Version: 0.9.0.2
Summary: Parse gdb machine interface output with Python
Home-page: https://github.com/cs01/pygdbmi
[...]

Issue #5782, opened 2020-08-24
requirements.txt (esp-idf/requirements.txt) has a dependency issue that cause the build to fail (IDFGH-3882)


hasantrq:
For the latest master branch.

![dependencies](https://user-images.githubusercontent.com/54283746/91042000-7558f300-e62a-11ea-8a66-bb1a53061cb8.png)

The requirements.txt pins **gdbgui>=0.13.2.0** which requires **pygdbmi<=0.9.0.2**. 

However the latest version of **gdbgui, 0.14.0.0** requires **pygdbmi<0.11, >=0.10.0.0b0**

Please take a look into this

skieast:
Looks like pygdmi has a pre-release 10.0.0 which should fix this. I have changed my requirements.txt to gdbgui==0.13.2.0 for now. 
I guess once released both dependencies will change.

leeebo:
I gdbgui>=0.13.2.0 pygdbmi<=0.9.0.2.

hasantrq:
> Fixed the version gdbgui==0.13.2.0 pygdbmi==0.9.0.2 can temporarily skip the problem.

I am running a CD pipeline and it automatically replicates the esp-idf repo. and I its fairly difficult to change the requirements.txt over there. Can I use a pip command to somehow downgrade after the newest package has been installed?

leeebo:
> > Fixed the version gdbgui==0.13.2.0 pygdbmi==0.9.0.2 can temporarily skip the problem.
> 
> I am running a CD pipeline and it automatically replicates the esp-idf repo. and its fairly difficult to change the requirements.txt over there. Can I use a pip command to somehow downgrade after the newest package has been installed?

Yes, using `pip install` you can install any available version. eg, change dir to `/home/{user}/.espressif/python_env/idf4.3_py3.6_env/bin` then run `./pip3 install gdbgui==0.13.2.0 pygdbmi==0.9.0.2`

Francois-Belanger:
Same issue on Mac OsX running "pip3 install gdbgui==0.13.2.0 pygdbmi==0.9.0.2". fixed the issue.  Thanks

feklee:
Same issue with branch `release/v4.2`:

    gdbgui 0.14.0.0 has requirement pygdbmi<0.11,>=0.10.0.0b0, but you'll have pygdbmi 0.9.0.2 which is incompatible.

rafasimionato:
[...]

PR #5794, opened 2020-08-26
fix windows incompatibility with gdbgui>=0.14 (IDFGH-3896)

istokm:
gdbgui intentionally (https://github.com/cs01/gdbgui/pull/346) introduced dependencies that are incompatible with windows, and don't seem to have plans to resolve them anytime soon

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ future>=0.15.2
 cryptography>=2.1.4
 pyparsing>=2.0.3,<2.4.0
 pyelftools>=0.22
-gdbgui>=0.13.2.0
+gdbgui<=0.13.2.1,>=0.13.2.0
 pygdbmi<=0.9.0.2
 # The pygdbmi required max version 0.9.0.2 since 0.9.0.3 is not copatible with latest gdbgui (>=0.13.2.0)
 reedsolo>=1.5.3,<=1.5.4


istokm:
fixes #5787

david-cermak (approved):
Thanks for addressing this issue! The changes are pending the internal review/test pipeline and will be merged (We might possibly be more strict and enforce the `0.13.2.0` version only)

david-cermak:
@istokm Thank you for your help, Linking also https://github.com/espressif/esp-idf/issues/5782

v1993:
Relevant gdgbui issue: https://github.com/cs01/gdbgui/issues/348

Note that it is not true that they have no plans to resolve it at all, just not right now:
> Screenshot_20200826_231507

It also wasn't done "intentionally" as to make windows unusable, only because new UI is much better and happens not to work properly on windows.

Related question: would it be possible for users on other platforms to use new version? If API is the same it would be great to take advantage of updated UI on Linux (and OS X?).

requirements.txt
```diff
@@ -11,7 +11,7 @@ future>=0.15.2
 cryptography>=2.1.4
 pyparsing>=2.0.3,<2.4.0
 pyelftools>=0.22
-gdbgui>=0.13.2.0
+gdbgui<=0.13.2.1,>=0.13.2.0
 pygdbmi<=0.9.0.2
```
cs01, line 15: gdbgui calls out which pygdbmi version it requires, so it shouldn't be included here. 

This project should consider using [pip-tools](https://github.com/jazzband/pip-tools#without-setuppy), which lets you specify things that should be installed at the top level, and then generates a requirements.txt lock file:

```
# requirements.in 
gdbgui==0.13.2.0
```

then compile the top level requirements into a lock file:

```
> pip-compile 
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    pip-compile
#
brotli==1.0.9             # via flask-compress
click==7.1.2              # via flask
dnspython==2.0.0          # via eventlet
eventlet==0.25.2          # via gdbgui
flask-compress==1.5.0     # via gdbgui
flask-socketio==2.9.6     # via gdbgui
flask==0.12.5             # via flask-compress, flask-socketio, gdbgui
gdbgui==0.14.0.0          # via -r requirements.in
gevent-websocket==0.10.1  # via gdbgui
gevent==1.5.0             # via gdbgui, gevent-websocket
greenlet==0.4.16          # via eventlet, gevent
itsdangerous==1.1.0       # via flask
jinja2==2.11.2            # via flask
markupsafe==1.1.1         # via jinja2
monotonic==1.5            # via eventlet
pygdbmi==0.10.0.0         # via gdbgui
pygments==2.6.1           # via gdbgui
python-engineio==3.13.2   # via python-socketio
python-socketio==4.6.0    # via flask-socketio
six==1.15.0               # via eventlet, python-socketio
werkzeug==0.16.1          # via flask
```

It also includes a dependency resolver, which ensures there are no conflicting requirements in the installed versions.

cs01:
Hi I came across this from the gdbgui issue (I am the creator/maintainer). I am hoping to get windows support added back someday with https://bugs.python.org/issue41663. Until then, I wonder if there is a way to install the latest version of gdbgui for non-windows users, and use a frozen version for windows users. As @v1993 mentioned, the new UI offers a better experience for users.

requirements.txt
igrr: Hi @cs01, pygdbmi is also a "top-level" requirement in ESP-IDF, this is why it is mentioned here explicitly. Thanks for the pointer about pip-tools, it looks very useful!

Alvin1Zhang:
Thanks for your contribution, changed merged by https://github.com/espressif/esp-idf/commit/96cfdf3184f5fac4c744fa9b8bc76a3e69fe2b65.

PR #5816, opened 2020-08-31
add __verbose()

rwalker-apple:
#### Problem
 export.sh is noisy for projects that invoke the script for every build

 #### Proposed changes
 * move verbose echoes to __verbose()
 * namespace other functions that get unset


Alvin1Zhang:
Thanks for contribution.

igrr:
Thanks for the PR @rwalker-apple! I will combine this with the other proposed change to export.sh, which fixes the return code, if you don't mind: https://github.com/espressif/esp-idf/pull/5744.

export.sh
```diff
@@ -11,8 +11,12 @@ realpath_int() {
     echo "$scriptdir"
 }
 
+__verbose() {
+    [[ -n ${IDF_QUIET} ]] && return
```
igrr, line 15: Sorry, i just ran shellcheck and realized that this `[[` won't work in `dash`. Can fix this up while rebasing on the other PR i've mentioned.

igrr: One more thing, how do you feel about using IDF_EXPORT_QUIET or IDF_TOOLS_QUIET instead of IDF_QUIET? I'm afraid that IDF_QUIET would imply to users that *all* the tools would disable output. Meeting that expectation would require quite a bit more changes to be added.

rwalker-apple: @igrr please take it from here and do as you see fit.  Just some suggestions.  Thanks for the portability check.

rwalker-apple: @igrr are you going to assume responsibility for this?

igrr: Sorry, have missed your previous comment. Sure, can take it from here. Thank you.

rwalker-apple:
@igrr, do you intend to work on this branch, or should I close the PR?

igrr:
Let's keep it open, and it should get closed automatically when the combined changes from the PR and #5744 are merged into master.

PR #5838, opened 2020-09-07
[freertos] Silence sign-conversion warning

sU8U7SfkcwTJVH7PjaVmej7D:
This a a header file, so including it from a project using -Wsign-conversion generates a warning.

`cpu_hal_get_core_id()` will never be negative nor very large, so let's just explicitly cast it to the good type.


Alvin1Zhang:
Thanks for your contribution.

AxelLin:
This does not apply to current master tree.

sU8U7SfkcwTJVH7PjaVmej7D:
@AxelLin : updated

PR #5856, opened 2020-09-11
Enable use of the ESP-IDF with a non-LWIP (and non-BSD-style) IP stack. (IDFGH-3971)

cnlohr:
Not sure how the process for getting things into the main IDF goes, but I do have my TCP/IP stack working (as part of another closed-source commercial project, but I am in the process of open-sourcifying it) which gives a significant performance boost for websockets-based and other ultra-low-latency applications.  It gets almost to the performance of the ESP8266.

Here's two issues I ran into:
(1) The use of lwip seems integrally tied in a few places, so I still allow linking to some of the headers.
(2) mbedtls is configured to be heavily dependent upon it in the default build, but it's not necessary.

I'm confident some of these should be changed up a little.

### DIFF
--- a/components/esp_netif/loopback/esp_netif_loopback.c
+++ b/components/esp_netif/loopback/esp_netif_loopback.c
@@ -18,6 +18,7 @@
 
 #include "esp_netif.h"
 #include "esp_netif_private.h"
+#include "esp_netif_sta_list.h"
 
 #if CONFIG_ESP_NETIF_LOOPBACK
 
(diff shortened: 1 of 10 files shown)


Alvin1Zhang:
Thanks for your contribution.

cnlohr:
Do I have to take any additional actions in order to facilitate this PR?

Alvin1Zhang:
Thanks for the updates and sorry for slow reply, we are checking the PR and will provide feedback later. Thanks.

cnlohr:
Thank you.

Alvin1Zhang:
Thanks for your contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/507c08251e87e5981d09ef79bbe67e24d77769c0

PR #4944, opened 2020-03-16
Add export and install scripts for fish shell (IDFGH-2899)

eliabieri:
I added a fish shell version of the export.sh and install.sh scripts.
This will close #4459 

Looking forward to your feedback.
Cheers

PR #5860, opened 2020-09-13
export.fish: Remove source check (IDFGH-3976)

kelvie:
This prevents starting a new shell with the IDF environment like this:

    env IDF_PATH=/my/idf/path fish -C 'source "$IDF_PATH"/export.fish'

as `status current-command` returns `fish` in this case, but this check expects
`source`.

This check does nothing anyway, as you can't even execute that file even if it
was marked executable, as fish rejects files without a shebang, and it's not
valid bash.

### DIFF
--- a/export.fish
+++ b/export.fish
@@ -1,15 +1,6 @@
 # This script should be sourced, not executed.
 
 function idf_export_main
-    # The file doesn't have executable permissions, so this shouldn't really happen.
-    # Doing this in case someone tries to chmod +x it and execute...
-    set cmd (status current-command)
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

igrr:
@eliabieri since you've added export.fish in #4944, could you please review this PR?

eliabieri (approved):
Seems totally reasonable to me. I approve the changes

igrr:
Thanks, have put this into the internal merge queue.

PR #5886, opened 2020-09-18
Bug with input parametr enum in function esp_wifi_80211_tx (IDFGH-4007)

gleb1536:
You have an error in input parametr in function "esp_wifi_80211_tx".
The docs say, that input parametr may be "WIFI_IF_STA" ( = 0 ) . But it parameter staying into enum: esp_interface_t. I use the parametr "WIFI_MODE_STA" ( = 1 ), that staying into wifi_interface_t, and have an error while use this function. Please apply this change in DOCs, i don't find document inc/esp_wifi.inc.

### DIFF
--- a/components/esp_wifi/include/esp_wifi.h
+++ b/components/esp_wifi/include/esp_wifi.h
@@ -992,7 +992,7 @@ esp_err_t esp_wifi_get_event_mask(uint32_t *mask);
   *    - ESP_ERR_WIFI_NO_MEM: out of memory
   */
 
-esp_err_t esp_wifi_80211_tx(wifi_interface_t ifx, const void *buffer, int len, bool en_sys_seq);
+esp_err_t esp_wifi_80211_tx(esp_interface_t ifx, const void *buffer, int len, bool en_sys_seq);
 
 /**
   * @brief The RX callback function of Channel State Information(CSI)  data. 


projectgus:
Hi @gleb1536,

Thanks for explaining the issue you found.

`esp_interface_t` and `wifi_interface_t` are the same type, see here:
https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi_types.h#L33

`WIFI_INTERFACE_STA` is defined to the same value as `ESP_IF_WIFI_STA`:
https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi_types.h#L35

(`wifi_interface_t` and`WIFI_INTERFACE_STA` are kept for compatibility reasons, we may change them in a future version.)

Because the two types are the same, changing the type in the header shouldn't change anything from the point of view of the compiler.

Can you post the code you are using, and any error message(s)?

YouDONG-ESP:
Hi @gleb1536 , sorry for the late reply. I am not sure that I totally understand your point, because WIFI_MODE_STA belongs to wifi_mode_t instead of wifi_interface_t in master, or can you show me your code or the snippet?

Alvin1Zhang:
Thanks for your contribution, feel free to reopen.

PR #5900, opened 2020-09-23
Changed to C linkage in netdb.h for fixing bug when using mixed C/C++ code (IDFGH-4027)

mathiasbredholt:
I got an error when trying to compile a project with mixed C/C++ code. This fixes the linkage issues. The error message is:
```
error: conflicting declaration of 'int getnameinfo(const sockaddr*, socklen_t, char*, socklen_t, char*, socklen_t, int)' with 'C' linkage
 int getnameinfo(const struct sockaddr *addr, socklen_t addrlen,
     ^~~~~~~~~~~
```

### DIFF
--- a/components/lwip/port/esp32/include/netdb.h
+++ b/components/lwip/port/esp32/include/netdb.h
@@ -32,9 +32,17 @@
 
 #include "lwip/netdb.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+  
 #ifdef ESP_PLATFORM
 int getnameinfo(const struct sockaddr *addr, socklen_t addrlen,
                 char *host, socklen_t hostlen,
                 char *serv, socklen_t servlen, int flags);
 
+#endif
+
+#ifdef __cplusplus
+}
 #endif


Alvin1Zhang:
Thanks for your contribution.

components/lwip/port/esp32/include/netdb.h
```diff
@@ -33,8 +33,17 @@
 #include "lwip/netdb.h"
 
 #ifdef ESP_PLATFORM
+
```
0xjakob, line 36: I would prefer the C++ guards outside the `#ifdef`. Not very likely but there could be some platform independent code as well which should also be guarded.

0xjakob (commented):
@mathiasbredholt The change is a good catch! Thanks for providing this to us. I just have a tiny comment regarding the order of `#ifdef` and C++ guards.

components/lwip/port/esp32/include/netdb.h
mathiasbredholt: @0xjakob  You're welcome! I agree and I moved the C++ guards outside of the ifdef.

0xjakob (approved):

Alvin1Zhang:
Thanks for your contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/8a102926f8bc74d180c61d913a69310f652dd1d6

PR #5913, opened 2020-09-25
brownout: shutdown rather than reboot if brownout occurs during boot.

adrien-n:
On battery, if the esp32 is woken up and the input voltage is too low, the brownout detector triggers, reboots and since the voltage is obviously not higher, the detector triggers again. This goes on for a long time and fully depletes the battery at a rate of more than 20mA. A better alternative is to shutdown. There will still be some energy usage but far less, giving far more chances that the battery is changed before being damaged.

Some context: I'm using an ESP32 for temperature monitoring and wanted to put the system on battery. The system spends most of its time deep-sleeping and only wakes briefly to get a new measurement and store it in RTC memory (and even less often synchronizes with a server). I've found that a li-ion battery directly connected to the 5V input would work fairly well if I could protect against deep discharge (brownouts happen below 3.6V input which is close to perfect for battery health).

I tried to install my own brownout handler but my code doesn't even get a chance to run: as soon as the current handler is installed, it gets run and begins the reboot loop.

I now install a handler for shutdown and right after I install the regular handler. The handler that shutdowns has enough time to trigger in the situation I've mentionned above and if it doesn't, the code continues with the regular behaviour.

I'm absolutely not sure I've done this right: I'm not even sure this is the best way to shutdown and I don't know if there's a way to check for brownout without installing the handler. I thought a PR was the best way to start the discussion but I'm ready to completely change the code.


Alvin1Zhang:
Thanks for your contribution.

igrr:
Hi @adrien-n,

> I'm not even sure this is the best way to shutdown and I don't know if there's a way to check for brownout without installing the handler.

This should be possible, and would be a cleaner solution than to use an interrupt handler.

The core related to brownout detector is in components/soc/src/esp32/brownout_hal.c (and also esp32s2, esp32s3). You can add a new function there, `bool brownout_hal_intr_raw()`, which would check rtc_brown_out field of RTCCNTL.int_raw register.

Then, after you call `brownout_hal_config`, you can call `brownout_hal_intr_raw` to check if the brownout detector has already triggered. If yes, proceed to shutdown. If no, install the regular interrupt handler and enable the interrupt via `brownout_hal_intr_enable`.

One more thing, could you please add a Kconfig option to enable this shutdown feature? I think to avoid unintended side effects for existing applications/users, we need to keep the behavior same as it is now, by default. The risk of this shutdown option is that if board suffers from a lengthy (but not permanent) brownout, the application will enter deep sleep and never wake up. The Kconfig option can be added next to the other brownout options in $IDF_PATH/components/{esp32,esp32s2,esp32s3}/Kconfig.

components/esp_common/src/brownout.c
```diff
@@ -42,19 +43,34 @@
...
+     * at the same time as the following ets_printf.
+     */
+    esp_cpu_stall(!xPortGetCoreID());
+    esp_reset_reason_set_hint(ESP_RST_BROWNOUT);
+    ets_printf(s);
+}
+
+static void shutdown_on_brownout(void *arg)
+{
+    warn("\r\nBrownout detector was triggere during bootd: shutting down.\r\n\r\n");
```
atanisoft, line 58: triggere -> triggered.

adrien-n:
Thanks for the feedback. I've updated the PR: I'm directly checking for brownouts rather than using the interrupt handler and I've added configuration options in all three Kconfig files with a fairly detailled description.

components/esp32s2/Kconfig
```diff
@@ -368,6 +368,27 @@ menu "ESP32S2-specific"
         # high resolution timekeeping anymore. Instead the esp_timer API, implemented
         # using FRC2 timer, is used.
         # FRC1 name in the option name is kept for compatibility.
+
+    config ESP32_BROWNOUT_SHUTDOWN_IF_EARLY
+        bool "Shutdown if a brownout is detected during boot"
+        depends on ESP32_BROWNOUT_DET
```
atanisoft, line 374: this should be `ESP32S2_BROWNOUT_DET`

components/esp32s2/Kconfig
```diff
@@ -368,6 +368,27 @@ menu "ESP32S2-specific"
         # high resolution timekeeping anymore. Instead the esp_timer API, implemented
         # using FRC2 timer, is used.
         # FRC1 name in the option name is kept for compatibility.
+
+    config ESP32_BROWNOUT_SHUTDOWN_IF_EARLY
```
atanisoft, line 372: probably should rename to `ESP32S2_BROWNOUT_SHUTDOWN_IF_EARLY` for consistency with other Kconfig entries.

components/esp32s3/Kconfig
```diff
@@ -418,6 +418,27 @@ menu "ESP32S3-Specific"
         # high resolution timekeeping anymore. Instead the esp_timer API, implemented
         # using FRC2 timer, is used.
         # FRC1 name in the option name is kept for compatibility.
+
+    config ESP32_BROWNOUT_SHUTDOWN_IF_EARLY
+        bool "Shutdown if a brownout is detected during boot"
+        depends on ESP32_BROWNOUT_DET
```
atanisoft, line 424: this should be `ESP32S3_BROWNOUT_DET`

components/esp_common/src/brownout.c
```diff
@@ -70,6 +81,21 @@ void esp_brownout_init(void)
 
     brownout_hal_config(&cfg);
 
+#ifdef CONFIG_ESP32_BROWNOUT_SHUTDOWN_IF_EARLY
```
atanisoft, line 84: with the renames in S2 and S3 Kconfig files this should be expanded to check for those as well ie:
```

#if CONFIG_ESP32_BROWNOUT_SHUTDOWN_IF_EARLY || CONFIG_ESP32S2_BROWNOUT_SHUTDOWN_IF_EARLY || CONFIG_ESP32S3_BROWNOUT_SHUTDOWN_IF_EARLY
```
Note that defined() or ifdef is optional here since these are boolean values.
Optionally

adrien-n:
I've updated the PR. 

@atanisoft Thanks for the review, I think I've addressed all your comments.

I've also removed an empty function I had left over and update the comment about the installation of the interrupt handler.

diablo1281:
Hi, I am trying to find this functionality in v5.0.1 but I can't see it anywhere. Is it gone or is there any way to accomplish system shutdown after the brownout?

PR #5919, opened 2020-09-27
rmt: Add user context pointer for use with RMT translator callbacks (IDFGH-4048)

outlandnish:
## Overview
This PR allows the user to pass in a context pointer to ` rmt_translator_init`. This data is provided to the translator callback in `rmt_write_sample`. This is useful for situations when you need a reference to the RMT channel being used in the translator function (or other context like timing data for the RMT pulses). 

## Example
An example of this usage is shown in https://github.com/intentfulmotion/AddressableLED/blob/master/src/OneWireLED.cpp. In this example, timing data for different addressable LED strips are passed to the translation function. Since this object is dynamically created and the timing is not known at compile time, this gives the developer flexibilty to write a single translator callback for different timing usecases.


suda-morris:
Hi @outlandnish First of all, thanks for your nice contribution. I left my consideration here.

On one hand, I agree to add the ability to pass user arguments to callback functions. But due to the historical reasons, the `rmt_translator_init` function is not a well designed API. 

On the other hand, this is a breaking change. We have to keep this API during the whole 4.x life cycle.

One of the workaround could be, introduce another API like "rmt_set_translator_arguments".

outlandnish:
@suda-morris that's fair, I was trying to find a way to avoid a breaking change. I think the challenge with either my solution above or with the alternate API call is that the callback signature, `sample_to_rmt_t` would have to change to accommodate the user arguments.

Is that acceptable or would there have to be an alternate callback signature that accepts user arguments?

suda-morris:
Hi @outlandnish Sorry for the late response. 
You're right, the main issue is about the callback function signature. If you just want the RMT driver to store the user_ctx for you, then we can add two more APIs to achieve it.
For example, invoke `rmt_set_translator_user_ctx(channel, user_ctx)` near  `rmt_translator_init`. And in the callback, you can get that context by invoking `rmt_get_translator_user_ctx(channel)`. In this way, we can still make the legacy callback live, at least in the v4.x lifetime.
What's your option? I'm open to discussion.

outlandnish:
@suda-morris yeah, I think that would work as a great interim solution for IDF 4.x

I'll update the code and let you know when that's ready!

outlandnish:
@suda-morris one challenge I'm facing as I write this code is we don't even have reference to the channel number in the current callback. Unless there's a way to do so that I'm missing?

An alternate solution is that when you call `rmt_write_sample`, it will call a new callback function signature that has that context if the context pointer is not NULL. If it is NULL, then it will call the existing callback. That way it maintains the existing functionality.

How does that sound?

boarchuz:
You could use the rmt_item32_t pointer in the callback to request what you need.

eg. callback
```
void my_rmt_cb(void *src, rmt_item32_t *dest, size_t src_size, size_t wanted_num, size_t *translated_size, size_t *item_num){
    my_rmt_ctx_t *ctx = rmt_get_buffer_owner_ctx(dest);
    // etc
}
```

eg. rmt.c
```
void *rmt_get_buffer_owner_ctx(rmt_item32_t *find_ptr) {
    for(int i = 0; i < RMT_CHANNEL_MAX; ++i) {
      if(p_rmt_obj[i] && p_rmt_obj[i]->tx_buf == find_ptr) {
        return p_rmt_obj[i]->tx_context;
      }
    }
    return NULL;
}
```
I'm not suggesting it's preferable, but it's an option.

outlandnish:
@boarchuz ah that's a clever workaround! Yeah that could work as long as there's a good example highlighting how to do so.

It seems a bit extra to have to do that with each transmission though :/ but we're a bit limited in the API signature

outlandnish:
@suda-morris the code is updated and preserves the original callback signature. I think @boarchuz's solution will have to work until IDF 5.x

It might be helpful to add an example that demonstrates how to do that so it lives outside of this PR

suda-morris:
Thanks @outlandnish for your patience in work around the limited RMT API. I apologize for not noticing the callback even lacking the RMT channel reference.

> However, we need to handle code in the translator differently for each instance.

Maybe register different callback for each channel can help.

> An alternate solution is that when you call rmt_write_sample, it will call a new callback function signature that has that context if the context pointer is not NULL. If it is NULL, then it will call the existing callback. That way it maintains the existing functionality.

Sounds nice to me. So will you add `rmt_translator_init_with_context` in this MR? Due to limitation of current RMT callback, I agree to introduce a new callback type and keep the legacy one, they're distinguished by if the context is NULL or not.

Thanks @boarchuz your usage demo looks smart and hacky. But in this way, we will have to introduce two new APIs to get it work as expected (i.e. `get/set_user_context`). Sorry for my misleading before.

So I now, I'm more prefer to the way @outlandnish  suggested, i.e. providing `rmt_translator_init_with_context`.

outlandnish:
@suda-morris sorry for the delay! I was having issues with IDF. I've re-opened this PR over here instead https://github.com/espressif/esp-idf/pull/6002

PR #5928, opened 2020-09-29
Allow for slip input of longer than 256 bytes (alternative fix) (IDFGH-4057)

webworxshop:
Alternative fix to https://github.com/espressif/esp-lwip/pull/23, which does not change lwip.

Implements batch processing of incoming slip data, with a maximum batch size of 255 bytes.

### DIFF
--- a/components/esp_netif/lwip/esp_netif_lwip_slip.c
+++ b/components/esp_netif/lwip/esp_netif_lwip_slip.c
@@ -165,7 +165,7 @@ void esp_netif_lwip_slip_input(void *h, void *buffer, unsigned int len, void *eb
     const int max_batch = 255;
     int sent = 0;
     while(sent < len) {
-        int batch = len > max_batch ? max_batch : len;
+        int batch = (len - sent) > max_batch ? max_batch : (len - sent);
         slipif_received_bytes(slip_ctx->esp_netif->lwip_netif, buffer+sent, batch);
         sent += batch;
     }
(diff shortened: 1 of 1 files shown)


components/esp_netif/lwip/esp_netif_lwip_slip.c
```diff
@@ -162,7 +162,13 @@ void esp_netif_lwip_slip_input(void *h, void *buffer, unsigned int len, void *eb
     ESP_LOG_BUFFER_HEXDUMP(TAG, buffer, len, ESP_LOG_DEBUG);
 
     // Update slip netif with data
-    slipif_received_bytes(netif->lwip_netif, buffer, len);
+    const int max_batch = 255;
+    int sent = 0;
+    while(sent < len) {
+        int batch = len > max_batch ? max_batch : len;
```
david-cermak, line 168: I think we have to compare with the *remaining length* here rather than the total `len` itself, since it would always be greater than... Sorry must have confused you in the lwip PR.

webworxshop: Yep, good catch. Fixed.

david-cermak (approved):
Thanks for the fixes! LGTM!

Alvin1Zhang:
Thanks for your contribution again, changes merged as https://github.com/espressif/esp-idf/commit/a92ae03b1e026af352a4d16ff1044c41c5f131a7

PR #5929, opened 2020-09-30
Properly clear the interrupt status in pcnt_intr_service(). (IDFGH-4058)

cjeker:
status is 0 when pcnt_hal_clear_intr_status() is called and so the
interrupt will be triggered immediatly again. Store the original
status and pass this as mask to pcnt_hal_clear_intr_status().


suda-morris:
@cjeker Nice found. We will merge it as soon as possible. Thanks again for your contribution.

suda-morris:
Your code has been merged into the master branch and also got backported to earlier release. Thanks.

PR #5942, opened 2020-10-05
fix call to usart_set_pin with 4 params (IDFGH-4071)

stbnrivas:
remove duplicate UART_PIN_NO_CHANGE I thing


igrr:
The declaration of uart_set_pin has 5 arguments, so the code snippet seems to be correct?
https://github.com/espressif/esp-idf/blob/8bc19ba893e5544d571a753d82b44a84799b94b1/components/driver/include/driver/uart.h#L407
(also in https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html#_CPPv412uart_set_pin11uart_port_tiiii)

Did you meet any issue when using the code snippet from the docs?

stbnrivas:
sorry I'm wrong

Issue #4537, opened 2019-12-22
How to read UART bytes in queue after break?


luksal:
I'm currently working on a [DMX512](https://en.wikipedia.org/wiki/DMX512#Protocol) library for the ESP32 https://github.com/luksal/ESP32-DMX-RX .
DMX512 is a protocol based on RS485 where after a break signal 512 bytes are transmitted.

The main concept is running an infinite loop and call `xQueueReceive` periodically and then checking the `event.type`.
When a `UART_DATA` is received, I'm copying the received data to an array and waiting for the next event. That way I'm getting 120 bytes each event, in total 480 until the break occurs.

How can I receive the last 32 bytes? Because when `event.type` is `UART_BREAK`, then calling `uart_get_buffered_data_len` equals 0 and trying to use `uart_read_bytes` fails at this point.

xiongyumail:
Can you provide the timing diagram of your UART port through a logic analyzer or oscilloscope?Especially before and after the UART_BREAK event. We want to know if there is a UART_DATA event from the UART_RXFIFO_TOUT after the UART_BREAK event

macdroid53:
I have attached a csv file of the data/timing on the DMX input port. (I had to change the file name, just rename without '.txt'.

I have also provided the Saleae Logic capture file as well. You can download their software to look at it. (Same with the name, just rename without '.txt'.

[DMX_capture.logicdata.txt](https://github.com/espressif/esp-idf/files/4010186/DMX_capture.logicdata.txt)



I have no idea how to capture UART_DATA event. Isn't that internal to the UART and reported by interrupt?
[DMX_capture.csv.txt](https://github.com/espressif/esp-idf/files/4010171/DMX_capture.csv.txt)

xiongyumail:
@luksal 
Thank you for the wave file provided, I have a general understanding of what kind of protocol DMX-512 is. I think we can use the UART register directly, which can solve your problem faster and provide it as a solution. Before that, if you have implemented ESP32 as the source to send DMX-512, I can use another ESP32 for debugging.

macdroid53:
DMX protocol is very simple. It starts with a break (the break is a low signal for approximately 88us). It then sends 513, 8bit values (each with 1 start bit, two stop bits, for a total of 11 bits per byte). This is sent at 250k baud.

I have no transmit code to offer, I'm working on a receive only application. (though I believe any serial source that could send 513 bytes @250k baud should be sufficient. See my reasoning below...)

[...]

PR #5959, opened 2020-10-09
Ensure serial break event is generated in the correct place in the serial byte stream

acf-bwl:
Currently the UART_BREAK event generated upon detection of a serial break is delivered asynchronously to the bytes received by the UART. An event containing bytes received following a serial break is sent prior to the event for the corresponding break. If the serial break is used as a frame delimiter, as it is in many protocols, this makes determining where to delineate frames challenging.

This patch was proposed by @negativekelvin in #4537, and ensures UART_BREAK events are delivered in the correct place in the stream.

### DIFF
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -799,10 +799,11 @@ static void UART_ISR_ATTR uart_rx_intr_handler_default(void *param)
                 }
             }
         }
-        else if ((uart_intr_status & UART_INTR_RXFIFO_TOUT)
+        else if (((uart_intr_status & UART_INTR_RXFIFO_TOUT)
                 || (uart_intr_status & UART_INTR_RXFIFO_FULL)
                 || (uart_intr_status & UART_INTR_CMD_CHAR_DET)
-                ) {
+                ) || ( (uart_intr_status & UART_INTR_BRK_DET) 
+                      && (uart_ll_get_rxfifo_len(uart_context[uart_num].hal.dev) > 0) )) {
             if(pat_flg == 1) {
                 uart_intr_status |= UART_INTR_CMD_CHAR_DET;
                 pat_flg = 0;


Alvin1Zhang:
Thanks for your contribution.

AshUK:
@Alvin1Zhang Any chance of getting this reviewed and merged any time soon, we are also wanting this fix to be merged ?

dzarda:
Bump, this is important

AshUK:
Any update on why this has not been approved and merged yet ?

moefear85:
how is this not resolved yet?

AshUK:
@negativekelvin is this PR still relevant with the latest v5 changes ?

tore-espressif:
Hi @acf-bwl @moefear85 

Sorry this has sat around for an extraordinary long time without further feedback.
Currently we hit regression on some of our UART driver unit tests with this change. I'll update this PR as soon as they are resolved.

moefear85:
ok thx for acknowledging the issue, it helps to know I'm not doing something wrong (or atleast not simultaneously).

SuGlider:
This PR didn't pass CI with RS-485 test case...

A alternative solution for this issue is very simple: read byte by byte from UART instead of reading 120 bytes when its FIFO is full.

BREAK can happen any time.... just make sure that IDF driver always gets every single byte from UART and copy it to its internal RingBuffer. Doing so, when BREAK happens, the RingBuffer would be already complete and it will be possible to catch the BREAK event and then process the data already copied.

In order to read byte by byte, just initialize the UART IDF Driver and then execute `uart_set_rx_full_threshold(uart_port_t uart_num, int threshold)`, with correct uart_num and threshold = 1.

moefear85:
As far as I extensively experimented, "hardware" RS-485 is broken anyways. I only got it working for multi-drop by implementing collision detection/avoidance in software. RS-232 is far more basic and important, because RS-485 is build on RS-232, meaning if uart breaks aren't correctly detected in RS-232, then neither are they in RS-485, so that's broken too!

Besides, reading byte by byte is not a solution, especially not at 2Mbps. I hate when people assume everyone is just trying to run some simple arduino project. It's supposed to be a professional-grade product. Anyways I moved on a long time ago and will unsubscribe here. Plz don't mention me.

johnboiles:
Tested on IDF 5.1.4 and it fixes an issue I was seeing where the bytes send before a BREAK were lost (until more data was sent later).

johnboiles (approved):
Tested on esp-idf 5.1.4 and it solved an issue I was seeing where bytes sent immediately before the break would not issue a `UART_DATA` event.

KenthJohan:
Tested this on v5.3.1 and it receives LIN frames successful.

PR #6002, opened 2020-10-18
rmt: Add user context pointer for use with RMT translator callbacks (IDFGH-4135)

outlandnish:
## Overview
_**This is a continuation of https://github.com/espressif/esp-idf/pull/5919**_
This PR allows adds a temporary function signature, `rmt_init_translator_with_context` so that users can pass in a context pointer to use with `rmt_write_sample`. This is useful for situations when you need a reference to the RMT channel being used in the translator function (or other context like timing data for the RMT pulses). 

## Example
An example of this usage is shown in https://github.com/intentfulmotion/AddressableLED/blob/master/src/OneWireLED.cpp. In this example, timing data for different addressable LED strips are passed to the translation function. Since this object is dynamically created and the timing is not known at compile time, this gives the developer flexibilty to write a single translator callback for different timing usecases.

### DIFF
--- a/components/driver/include/driver/rmt.h
+++ b/components/driver/include/driver/rmt.h
@@ -154,6 +154,22 @@ typedef struct {
 typedef void (*sample_to_rmt_t)(const void *src, rmt_item32_t *dest, size_t src_size, size_t wanted_num, size_t *translated_size, size_t *item_num);
 
 /**
+* @brief IDF 4.x Workaround callback function for sample_to_rmt_t that lets user pass in context for rmt_write_sample
+*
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
Thanks @outlandnish this PR is in our internal review list now, will merge it as soon as possible :+1:

Alvin1Zhang:
Thanks for contribution, changes merged with https://github.com/espressif/esp-idf/commit/21bb6f2334f76dd809c5499a85839bbd83fef9db.

PR #6023, opened 2020-10-25
fixed debug spi args (IDFGH-4161)

sdrabb:
Fixed debug slave arguments


Alvin1Zhang:
Thanks for your contribution.

Icarus113:
Thanks for following up. But we don't need this debug code any more, it'll be removed.

sdrabb:
> Thanks for your contribution.

It's a pleasure

sdrabb:
> Thanks for following up. But we don't need this debug code any more, it'll be removed.

You're welcome... I think it still useful during SPI debugging

PR #6034, opened 2020-10-27
Add weak symbol support for LwIP hooks (IDFGH-4173)

pidpawel:
Hi!
This is a PR following one at https://github.com/espressif/esp-lwip/pull/22

As stated in the linked PR - it is practically impossible to add extra hooks for LwIP from projects using `arduino-esp32` while it may be beneficial in some cases. This PR aims to improve that by defining weak implementations for them to they can be overridden during linking.

I took a look at https://github.com/espressif/esp32-arduino-lib-builder and in my opinion adding those changes there would be unnecessarily messy and leaving them here seems like a reasonable compromise.

This is kind of a proof of concept. It may be freely extended/generalized to other hooks if needed. 

I'm open to any feedback, even if it means making a mess in `esp32-arduino-lib-builder` :-P


Alvin1Zhang:
Thanks for your contribution.

david-cermak:
Hi, I think this is a better place for the lwip hooks. Might be useful for other IDF users, as well.
Alternatively, we can define Kconfg options for adding hooks (just to be more explicit, undesired linkage with the *weak* rather than the *strong* implementation might be harder to track than simple unresolved symbol error). But both options are equally okay with me.

This LGTM overall, thanks for posting!

Probably should also think of linkage order, if we have the *strong* implementation in another component (very likely, but that's user responsibility) and make the changes in line with the recently added [`TCP_ISN` hook](https://github.com/espressif/esp-idf/commit/58c33f3bdfa)

pidpawel:
Hi! Thanks for the feedback! I've added some changes to the PR namely: added the Kconfig and moved some stuff around to align the TCP_ISN hook and those new ones. Hopefully the pattern I came up with will be extensible enough to suit needs of multiple projects.
I hope that this addresses all of your comments. Please do let me know if I missed anything.

david-cermak:
Thanks for the updates! This is ready for adding into our internal queue of merge requests, but please expect some minor changes in the names of options, etc, which we preferably address internally, amending your commits.

Also, some of the files are missing a copyright header. @pidpawel Please feel free to add *your own* copyright notice if you wish to, otherwise we are going to add standard Espressif "Apache v2.0" header.

david-cermak (approved):
LGTM!

pidpawel:
That's great news! Feel free to amend the changes in any way you like and to include the standard copyright header. I'm not attached to those changes in any significant way :-P

AxelLin:
This PR causes a regression https://github.com/espressif/esp-idf/issues/6261 because it defines LWIP_HOOK_FILENAME.
It probably needs to add CONFIG_LWIP_HOOK_IP4_ROUTE_SRC_CUSTOM option so users can still
implement a customized LWIP_HOOK_IP4_ROUTE_SRC.

david-cermak:
@pidpawel Thanks for the contribution, cherry-picked as https://github.com/espressif/esp-idf/commit/cf60ec0ffc7e057f9ef66e4b6fc2abe3339b29e0

@AxelLin Fix (posted here: https://github.com/espressif/esp-idf/issues/6261#issuecomment-745999569) is pending the internal review/test process and will be merged soon.

Alvin1Zhang:
Thanks for contribution.

PR #6044, opened 2020-10-29
Make the UserAgent string in esp-http-client configurable (IDFGH-4184)

scottsh:
This change creates a new configurable parameter (through idf.py menuconfig), which is able to change the Http user agent string in esp_http_client.

The default value of the config parameter is exactly the same as the previous hard coded value, so people who don't change the config will see no difference in behavior.

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -136,7 +136,7 @@ static esp_err_t _clear_connection_info(esp_http_client_handle_t client);
 #define ASYNC_TRANS_CONNECTING 0
 #define ASYNC_TRANS_CONNECT_PASS 1
 
-static const char *DEFAULT_HTTP_USER_AGENT = "ESP32 HTTP Client/1.0";
+static const char *DEFAULT_HTTP_USER_AGENT = CONFIG_ESP_HTTP_CLIENT_USER_AGENT;
 static const char *DEFAULT_HTTP_PROTOCOL = "HTTP/1.1";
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97 (approved):

components/esp_http_client/esp_http_client.c
```diff
@@ -136,7 +136,7 @@ static esp_err_t _clear_connection_info(esp_http_client_handle_t client);
 #define ASYNC_TRANS_CONNECTING 0
 #define ASYNC_TRANS_CONNECT_PASS 1
 
-static const char *DEFAULT_HTTP_USER_AGENT = "ESP32 HTTP Client/1.0";
+static const char *DEFAULT_HTTP_USER_AGENT = CONFIG_ESP_HTTP_CLIENT_USER_AGENT;
```
mahavirj, line 139: How about we add an additional member to `esp_http_client_config_t` as `const char *user_agent`?

And then during initialization we can use something like:

```
const char *user_agent = config->user_agent == NULL ? DEFAULT_HTTP_USER_AGENT : config->user_agent;
esp_http_client_set_header(client, "User-Agent", user_agent);
```
```

scottsh: Sounds good to me. :)

mahavirj:
Merged with https://github.com/espressif/esp-idf/commit/8c09fbf6f1f71fd636061fb65129433574dbb0c6

PR #6055, opened 2020-11-02
[mdns] fix wrong mdns source address (IDFGH-4196)

gjc13:
The struct definition of ip6_addr_t in lwip and esp_ip6_addr_t
differs since zone is not enabled in lwip. Using `memcpy` to copy the
address will cause wrong source address. Copy the entries manually
instead.

### DIFF
--- a/components/mdns/mdns_networking.c
+++ b/components/mdns/mdns_networking.c
@@ -127,7 +127,8 @@ static void _udp_recv(void *arg, struct udp_pcb *upcb, struct pbuf *pb, const ip
         packet->tcpip_if = MDNS_IF_MAX;
         packet->pb = this_pb;
         packet->src_port = rport;
-        memcpy(&packet->src, raddr, sizeof(ip_addr_t));
+        packet->src.type = raddr->type;
+        memcpy(&packet->src.u_addr, raddr->u_addr, sizeof(ip6_addr_t));
         packet->dest.type = packet->src.type;
 
         if (packet->src.type == IPADDR_TYPE_V4) {


Alvin1Zhang:
Thanks for your contribution.

david-cermak:
Hi @gjc13 

Yes, these two `struct` definitions differ, but the scope_id is enabled in IDF, assuming it's not forced into lwip component otherwise. In that case it might be possible the addr-type would not be copied correctly, is this what you're trying to address? if so, i'd suggest copying `u_addr` field and assigning the `type` manually.

components/mdns/mdns_networking.c
```diff
@@ -127,7 +127,12 @@ static void _udp_recv(void *arg, struct udp_pcb *upcb, struct pbuf *pb, const ip
         packet->tcpip_if = MDNS_IF_MAX;
         packet->pb = this_pb;
         packet->src_port = rport;
-        memcpy(&packet->src, raddr, sizeof(ip_addr_t));
+        packet->src.type = raddr->type;
+        if (raddr->type == IPADDR_TYPE_V4) {
```
david-cermak, line 131: Is this condition correct? seems to me we would like to copy the IPv4 address if the type is `IPADDR_TYPE_V4`

david-cermak (changes requested):
Please elaborate on the condition or/and think of an easier way to address the issue, such as
``` c
memcpy(packet->src.u_addr, raddr->u_addr, sizeof(raddr->u_addr));
packet->src.type = raddr->type;
```

gjc13:
> Please elaborate on the condition or/and think of an easier way to address the issue, such as
> 
> ```c
> memcpy(packet->src.u_addr, raddr->u_addr, sizeof(raddr->u_addr));
> packet->src.type = raddr->type;
> ```

Updated

david-cermak:
@gjc13 Thanks for the update, this LGTM in general. Could you please also click trough the CLA agreement?
Just two nitpicks which could be applied internally:
* copy from the address of the union
* copy `sizeof()` of the union, the `ip6` size is the same but i'd just find it more readable
``` c
memcpy(&packet->src.u_addr, &raddr->u_addr, sizeof(raddr->u_addr));
```

PR #6117, opened 2020-11-13
esp_crt_bundle: Allow verify_callback to correct BADCERT_BAD_MD (IDFGH-4266)

jepler:
This allows CircuitPython on ESP32S2 to successfully connect to io.adafruit.com, like standard browsers and desktop https clients like openssl s_client, curl, and mbedtls ssl_client2.  This problem also affects at least api.thingspeak.com.

io.adafruit.com serves up a slightly interesting certificate chain.  The part returned to the client includes 3 certificates, and the third one uses the SHA1 signature algorithm.  Since this has turned up on two IOT platforms, I suspect (but have not verified) that this must be to improve compatibility with some older deployed IOT devices.

In standard Linux SSL implementations which use `ca-certificates.crt` which contains DigiCert Global Root CA, the second certificate is marked as trusted and the SHA1 signature algorithm of the third certificate is not considered.  This is also true of mbedtls when compiled for Linux and using ca-certificates.crt.

esp-idf uses a different method (esp_crt_verify_callback) to determine whether a certificate chain is trusted.  This leads to slightly different semantics when items further into the certificate chain than the first trusted certificate would signal problems.  esp_crt_verify_callback is executed after the whole certificate chain has been considered, and in this case the third certificate in the chain has had the flag MBEDTLS_X509_BADCERT_BAD_MD set.

Update  esp_crt_verify_callback so that it can override either NOT_TRUSTED or BAD_MD results when the certificate is present in the esp crt bundle.  If no flags are set, or flags other than these two are set, then it does not influence the result.

The certificate chain presently served by io.adafruit.com, for posterity: [io.txt](https://github.com/espressif/esp-idf/files/5539403/io.txt)

Our discussion of this issue: https://github.com/adafruit/circuitpython/issues/3424

cc: @ladyada @tannewt

### DIFF
--- a/components/mbedtls/esp_crt_bundle/esp_crt_bundle.c
+++ b/components/mbedtls/esp_crt_bundle/esp_crt_bundle.c
@@ -97,7 +97,10 @@ int esp_crt_verify_callback(void *buf, mbedtls_x509_crt *crt, int data, uint32_t
 {
     mbedtls_x509_crt *child = crt;
 
-    if (*flags != MBEDTLS_X509_BADCERT_NOT_TRUSTED) {
+    if (!*flags) {
+        return 0;
+    }
+    if (*flags & ~(MBEDTLS_X509_BADCERT_NOT_TRUSTED | MBEDTLS_X509_BADCERT_BAD_MD)) {
         return 0;
     }
 


Alvin1Zhang:
Thanks for your contribution.

ESP-Marius:
@jepler I see your issue, and I agree that it shouldnt fail just because the signature of a trusted root certificate is weak.

I'll make an internal MR for this. Thanks for the contribution!

ESP-Marius:
@jepler The MR is running tests in our CI now. 

I took the liberty to add a commit changing the check to:

``` 
uint32_t flags_filtered = *flags & ~(MBEDTLS_X509_BADCERT_BAD_MD);

if (flags_filtered != MBEDTLS_X509_BADCERT_NOT_TRUSTED) {
     return 0;
}
```

As I don't think we should be verifying the certificate if the only `MBEDTLS_X509_BADCERT_BAD_MD` is set and not `MBEDTLS_X509_BADCERT_NOT_TRUSTED`. Do you agree with this?

Also added a unit test with the `io.adafruit.com` certificate chain.

jepler:
Thanks for your attention to this problem!  I may have reasoned incorrectly about the bitflags, so any improvement there is welcome.

ESP-Marius:
@jepler Just to keep you informed: this change has already been merged into our master. So it will be available on github the next time master is synced! Sorry for keeping you waiting.

tannewt:
Thanks @ESP-Marius! Would you mind backporting it to release/4.2 as well?

ESP-Marius:
> Thanks @ESP-Marius! Would you mind backporting it to release/4.2 as well?

Will do.

tannewt:
Thank you!

jepler:
As far as I can tell there's no reason to leave this PR open.

PR #6187, opened 2020-12-01
Allow KConfig to configure LWIP checksum validation (IDFGH-4349)

ryankurte:
for some reason (?!) the ESP LWIP configuration defaults to disabling checksum validation for IP, UDP, and ICMP packets. this PR adds KConfig options to configure LWIP checksums and passes these through to the platform config.

the `Checksums` submenu is there because while there are `ICMP` and `UDP` menus that these could be nested under, there is not one for `IP`.

cc. @webworxshop

### DIFF
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -574,13 +574,13 @@ menu "LWIP"
 
     config LWIP_CHECKSUM_CHECK_IP
         bool "Enable LWIP IP checksums"
-        default y
+        default n
         help
             Enable checksum checking for received IP messages
 
     config LWIP_CHECKSUM_CHECK_UDP
         bool "Enable LWIP UDP checksums"
-        default y
+        default n
         help
             Enable checksum checking for received UDP messages
 
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -570,6 +570,28 @@ menu "LWIP"
 
     endmenu # UDP
 
+    menu "Checksums"
+
+    config LWIP_CHECKSUM_CHECK_IP
+        bool "Enable LWIP IP checksums"
+        default y
+        help
+            Enable checksum checking for received IP messages
+
+    config LWIP_CHECKSUM_CHECK_UDP
+        bool "Enable LWIP UDP checksums"
+        default y
+        help
+            Enable checksum checking for received UDP messages
+
+    config LWIP_CHECKSUM_CHECK_ICMP
+        bool "Enable LWIP ICMP checksums"
+        default y
+        help
+            Enable checksum checking for received ICMP messages
+
+    endmenu # Checksums
+
     config LWIP_TCPIP_TASK_STACK_SIZE
         int "TCP/IP Task Stack Size"
         default 3072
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/lwip/Kconfig
```diff
@@ -570,6 +570,31 @@ menu "LWIP"
 
     endmenu # UDP
 
+    menu "Checksums"
+
+    config LWIP_CHECKSUM_CHECK_IP
+        int "Enable LWIP IP checksums"
+        range 0 1
```
david-cermak, line 577: We usually prefer boolean type for these enable/disable options.

components/lwip/Kconfig
```diff
@@ -570,6 +570,31 @@ menu "LWIP"
...
+        int "Enable LWIP UDP checksums"
+        range 0 1
+        default 0
+        help
+            Enable checksum checking for received UDP messages
+
+    config LWIP_CHECKSUM_CHECK_ICMP
+        int "Enable LWIP ICMP checksums"
+        range 0 1
+        default 0
```
david-cermak, line 592: Current master has the `ICMP` checksums turned `ON` if i'm not mistaken, so suggest keeping the compatibility.

david-cermak (commented):
Thanks for posting this useful update! Please see my comments to the changes.

david-cermak:
Thanks for the update. I'm sorry, didn't answer the question:

> for some reason (?!) the ESP LWIP configuration defaults to disabling checksum validation

This configuration is used in esp-lwip from its initial version. TBH not sure what was the primary reason, but I think this makes some sense, since:
* IDF developers always place importance on performance and network thoughput and UDP/IP checksum validation has an impact on the UDP and TCP bandwidth (see [iperf](https://github.com/espressif/esp-idf/tree/master/examples/wifi/iperf) example) unlike the ICMP checksum (which defaults to *ON*)
* Checksums on IP and UDP layers are useful, but since the frame check sequence is validated on lower layers we have certain confidence about the packets and the upper layer protocols being valid.

components/lwip/Kconfig
```diff
@@ -570,6 +570,28 @@ menu "LWIP"
 
     endmenu # UDP
 
+    menu "Checksums"
+
+    config LWIP_CHECKSUM_CHECK_IP
+        bool "Enable LWIP IP checksums"
+        default y
```
david-cermak, line 577: This should default to "no", as mentioned before.

components/lwip/Kconfig
```diff
@@ -570,6 +570,28 @@ menu "LWIP"
...
+        help
+            Enable checksum checking for received IP messages
+
+    config LWIP_CHECKSUM_CHECK_UDP
+        bool "Enable LWIP UDP checksums"
+        default y
```
david-cermak, line 583: Same here.

david-cermak (changes requested):
I think this change would break our internal throughput tests (unlike the previous version). We have to maintain backward compatibility in the default settings.

david-cermak (approved):
LGTM!

Alvin1Zhang:
Thanks for contribution, changes merged with https://github.com/espressif/esp-idf/commit/5d91a27a1eed9f3b4602e0874b7b216750f6cdd9

PR #6216, opened 2020-12-07
Fix invalid clock source in i2c_self_test example (IDFGH-4382)

c-dill:
Fixed issue where the check at `components/driver/i2c.c:667` would fail due to `clk_flags` not being specified in the self test example


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for your contribution, merged with https://github.com/espressif/esp-idf/commit/a3b59aeeb5fe78d768667636f135eaf1470fd750, thanks.

PR #6231, opened 2020-12-08
Fixed register to RTC_CNTL_ULP_CP_TIMER_1_REG in ulp.c (IDFGH-4398)

zdovcj:
I was initially working on release 4.2 and noticed that ULP example doesn't work. One of the main reasons was that the RTC_CNTL_ULP_CP_TIMER_REG was rewritten with a constant value so that the start address of the ULP program was corrupt. This happened on two places in 4.2. 

I managed to get it working in version 4.2. However, the master version has so many changes that it broke the example again. I'm still struggling to find the reason why it fails. 
Unfortunately, this correction is not enough to fix the problem, but at least it's one bug less :) Maybe some of you folks will figure it out.


Alvin1Zhang:
@zdovcj Thanks for your contribution.

sudeep-mohanty:
@zdovcj There have been fixes done for ULP FSM init code in f163ab26a965789740bc5b07b6782c21ff1b8ee1. We have verified that the ULP FSM examples work well too. The change submitted in this PR is subsumed by the other fixes. In case you are still interested, you may switch to the top-of-tree master branch to evaluate ULP FSM.

PR #6330, opened 2020-12-29
Issue 6316 Rudimentary session handling for the wifi (SoftAP) provisioning (IDFGH-4505)

william-ferguson-au:


Alvin1Zhang:
Thanks for contribution.

william-ferguson-au:
NB The real solution here would be to separate sockets and http_sessions.

A more complete solution than what I have offered (but still a hack that confuses sockets with sessions) would be to update the LRU for the socket whenever a http_session maps back to the original socket. This would ensure that the socket (and it's security ctx) is not destroyed while the session is still in use.

mahavirj:
@william-ferguson-au 

Hopefully fix from https://github.com/espressif/esp-idf/commit/72fa8d07507045ddd982a37f6e3b3adc254c541f helps in this case. Closing this PR, however if you think that we should accept this as generic feature in HTTP server, please feel free to resolve merge conflicts and reopen. Thank you.

PR #6347, opened 2021-01-01
Prepend (esp-idf) to shell prompt when the idf is entered. (IDFGH-4523)

jack-davidson:
Added function in `export.sh` to update the shell prompt such that the
programmer knows that they are in an esp-idf environment. This works
just like when you enter a python virtual environment and the name of
the virtualenv is prepended to your shell PS1 in parens.

esp-idf_prompt_pr_example

### DIFF
--- a/export.sh
+++ b/export.sh
@@ -146,8 +146,13 @@ enable_autocomplete() {
     fi
 }
 
+update_prompt() {
+	export PS1="(esp-idf) ${PS1}"
+}
+
 idf_export_main
 enable_autocomplete
+update_prompt
 
 unset realpath_int
 unset idf_export_main
--- a/export.sh
+++ b/export.sh
@@ -146,13 +146,10 @@ enable_autocomplete() {
     fi
 }
 
-update_prompt() {
-    export PS1="(esp-idf) ${PS1}"
-}
-
 idf_export_main
 enable_autocomplete
-update_prompt
+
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

cgbahk:
I was about to make exactly same PR, but found this. 

IMO, to make a consistent visual notification is essential feature to have.

Python venv module also has such feature in long time with such many users. I'm pretty sure most user are quite appreciate with it

So could we apply this change ? :smile:

jack-davidson:
I really think we should apply this change. Haha, I have forgotten about
this PR.

jack-davidson:
Shall I close this PR? It seems this has just been ignored?

Kiznoh:
Thanks for your contribution.
Unfortunately, the altering of PS1 is beyond the scope of the export script. 
It's advisable to use [direnv](https://direnv.net/#getting-started) to automatically invoke the export script when entering a project folder.

You are, however, free to add the altering of PS1 to the alias you use to invoke the export script:
```bash
alias get_idf='. $HOME/esp/esp-idf/export.sh; source ~/.bashrc; export PS1="(esp-idf) ${PS1}"'
```
Please note that `source ~/.bashrc` is used to reset PS1 to a clean state when running the alias multiple times to refresh the esp-idf environment. You may need to update it when a different shell than bash.

PR #6389, opened 2021-01-13
app_update: fix incorrect first byte from esp_ota_get_app_elf_sha256 (IDFGH-4572)

itayperl:
At -O2 optimization level, GCC seems to optimize out the copying of the
first byte of the checksum, assuming it is zero. This "miscompilation"
happens because the esp_app_desc struct is declared const, but then modified
post-compilation. Casting to volatile disables the optimization.

### DIFF
--- a/components/app_update/esp_app_desc.c
+++ b/components/app_update/esp_app_desc.c
@@ -89,7 +89,7 @@ int IRAM_ATTR esp_ota_get_app_elf_sha256(char* dst, size_t size)
     static bool first_call = true;
     if (first_call) {
         first_call = false;
-        const uint8_t* src = esp_app_desc.app_elf_sha256;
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

KonstantinKondrashov:
Hi @itayperl.
Thank you for the fix. Really the first byte is 0. 

```
I (175) cpu_start: ELF file SHA256:  005a3a4c72ec7bd0...
...
ELF file SHA256: 005a3a4c72ec7bd0

> sha256sum  build/unit-test-app.elf
e25a3a4c72ec7bd0ff4afeaa832b449e74ec699b256c32634360bfad95c86c1a  build/unit-test-app.elf
```

We will merge this fix into master and create backports.

PR #6391, opened 2021-01-13
ESP HTTPS OTA: send POST request body if set (IDFGH-4574)

### LINKED ISSUE #6390: esp_https_ota does not send POST request body (IDFGH-4573)
## Environment

- Module or chip used:  ESP32-WROVER-E
- IDF version (run ``git describe --tags`` to find it): 2bfdd036b2dbd07004c8e4f2ffc87c728819b737

--

Hello,

The OTA backend in my project requires a POST request with a JSON body. I tried implementing this by adding a custom handler in the `http_client_init_cb` field of the esp_https_ota config. The handler calls `esp_http_client_set_post_field` to set the request body.

It seems that this does not work. Looking at the code, esp_https_ota calls `esp_http_client_open` followed by `esp_http_client_fetch_headers`. This skips the call to `esp_http_client_send_post_data` (which is not even a public API) which sends the request body.

itayperl:
Fixes #6390.

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -165,7 +165,6 @@ static const char *HTTP_METHOD_MAPPING[] = {
 
 static esp_err_t esp_http_client_request_send(esp_http_client_handle_t client, int write_len);
 static esp_err_t esp_http_client_connect(esp_http_client_handle_t client);
-static esp_err_t esp_http_client_send_post_data(esp_http_client_handle_t client);
 
 static esp_err_t http_dispatch_event(esp_http_client_t *client, esp_http_client_event_id_t event_id, void *data, int len)
 {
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97:
Hi @itayperl,

Instead of making changes in esp_http_client component, please check if the attached patch can work for your use case.

[0001-Send-post-data-if-post-field-is-set.patch.zip](https://github.com/espressif/esp-idf/files/5917317/0001-Send-post-data-if-post-field-is-set.patch.zip)

Thanks,
Shubham

itayperl:
@shubhamkulkarni97, thank you for looking into this.
Your patch might work, but it looks like esp_http_client_write is allowed to perform a partial write, so it might be better to have something like this (untested):
```c
while (post_len > 0) {
    int write_len = esp_http_client_write(http_client, post_data, post_len);
    if (write_len <= 0) {
        ESP_LOGE(TAG, "Write failed");
        return ESP_FAIL;
    }
    post_len -= write_len;
    post_data += write_len;
}

```

shubhamkulkarni97:
Hi @itayperl, which server are you using to perform OTA. Is it open source or your custom implementation?

Also, please share how does the server work (does it send firmware image in response to POST request or it sends OTA URL to download image using GET request ?).

itayperl:
@shubhamkulkarni97, it's a custom server. The server gets a POST request, with some parameters passed in the body in JSON format, and returns the firmware binary directly.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/807bf9688c87098bc4b578adbb002392a2f7a618

PR #6396, opened 2021-01-13
uart: Add critical section barriers to uart_get_buffered_data_len() (IDFGH-4579)

### LINKED ISSUE #6397: uart: uart_get_buffered_data_len() very rarely fails
## Environment

- Module or chip used:  [ESP32-WROOM-32]
- IDF version: found in `v4.0.1`, also present in current "master" (`v4.3-dev-2399-gf3adbbb8e` as of 2021-01-13)

## Problem Description

I am developing a project using the ESP32 port of MicroPython; the main task has to process chars from the serial port when they are available. MicroPython does not have non-blocking file I/O, but offers a `select()` operation that can be used to query if a stream is ready for reading; the assumption being that when so, at least a single byte can be read without blocking. 

Very rarely, though, this `select()` was returning the UART stream as ready, even though a subsequent read operation had 0 bytes available. This caused the Python application to hang until data was actually sent to the serial port.

Tracing the issue in the Micropython and then ESP-IDF sources led to identification of an issue in the current implementation of the `uart_get_buffered_data_len()` API function.

### Expected Behavior

- `uart_get_buffered_data_len()` should always fill its `size` parameter either with 0 or with the number of bytes available in the requested UART.

### Actual Behavior

- `uart_get_buffered_data_len()` may return a garbage value in `size` if a different execution thread updates the internal state of the UART while this function is being run.

### Steps / code to reproduce

I assume the following snippet for a given `uart_num` would eventually show the issue if left running long enough (no input is required):

```cpp
    while (1) {
        size_t data_len;
        if (uart_get_buffered_data_len(uart_num, &data_len) == ESP_OK) {
            uint8_t buf;
            int ret = uart_read_bytes(uart_num, &buf, 1, 0);
            if (ret != 1) ESP32_LOGW("data_len %u but read_bytes %i?", data_len, ret);
        }
}
```

### Proposed fix

I will shortly provide a pull request adding the missing critical section barriers; these are already present in all other functions accessing the UART state in `uart.c`.

pillo79:
The missing barriers caused the function to (very rarely) return a garbage
value. When used in MicroPython, though, this caused select() to return
and a subsequent read() to stall indefinitely until a char was actually
available.

Fixes #6397.

### DIFF
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -1251,7 +1251,9 @@ esp_err_t uart_get_buffered_data_len(uart_port_t uart_num, size_t* size)
 {
     UART_CHECK((uart_num < UART_NUM_MAX), "uart_num error", ESP_FAIL);
     UART_CHECK((p_uart_obj[uart_num]), "uart driver error", ESP_FAIL);
+    UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
     *size = p_uart_obj[uart_num]->rx_buffered_len;
+    UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
     return ESP_OK;
 }
 
@@ -1281,12 +1283,15 @@ esp_err_t uart_flush_input(uart_port_t uart_num)
         }
         data = (uint8_t*) xRingbufferReceive(p_uart->rx_ring_buf, &size, (portTickType) 0);
         if(data == NULL) {
+            UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
             if( p_uart_obj[uart_num]->rx_buffered_len != 0 ) {
-                ESP_LOGE(UART_TAG, "rx_buffered_len error");
                 p_uart_obj[uart_num]->rx_buffered_len = 0;
+                UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
+                // this must be called outside the critical section
+                ESP_LOGE(UART_TAG, "rx_buffered_len error");
+                UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
             }
             //We also need to clear the `rx_buffer_full_flg` here.
-            UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
             p_uart_obj[uart_num]->rx_buffer_full_flg = false;
             UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
             break;


Alvin1Zhang:
Thanks for your contribution.

AxelLin:
Hi @Alvin1Zhang 

When there is no reponse on a PR, the submitter can only just wait and polling the status.
I'm wondering if you have standard procedure to let people know if a PR is accepted or rejected.

Note, I'm not talking about the timing to get things merged. I know it takes time.
I'm talking about the timing to let people know if the PR is ok(accepted) and will be merged latter,
or something still needs fix.

AxelLin:
Can someone help to review it first? (Is this fix correct? Is this fix required?)
If it really takes time to pull the fix, at least people can pick up the fix manually if it got reviewed.

AxelLin:
> Thanks for your contribution.

No comments for such long time? It really discourages people to contribute fixes.

components/driver/uart.c
```diff
@@ -1281,12 +1283,15 @@ esp_err_t uart_flush_input(uart_port_t uart_num)
         }
         data = (uint8_t*) xRingbufferReceive(p_uart->rx_ring_buf, &size, (portTickType) 0);
         if(data == NULL) {
+            UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
             if( p_uart_obj[uart_num]->rx_buffered_len != 0 ) {
-                ESP_LOGE(UART_TAG, "rx_buffered_len error");
                 p_uart_obj[uart_num]->rx_buffered_len = 0;
+                UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
+                // this must be called outside the critical section
+                ESP_LOGE(UART_TAG, "rx_buffered_len error");
```
o-marshmallow, line 1291: It would be better to set a flag in this case and print this log message after `UART_EXIT_CRITICAL` and before `break;` in case the flag is set. Entering and exiting critical sections is rather expensive compared to setting and testing a flag

o-marshmallow:
@pillo79 Thank you for you your contribution and sorry for the delay, I am going to take care of this until it gets merged. I just added a small comment to improve performance.

components/driver/uart.c
AxelLin: Why do you need to care about performance in an unlikely error path?

AxelLin:
Hi @o-marshmallow 

Similar locking issue by checking below functions:
uart_get_stop_bits
uart_get_parity
uart_get_baudrate

I'm wondering why in uart_get_baudrate(), it needs
UART_ENTER_CRITICAL/UART_EXIT_CRITICAL around uart_hal_get_baudrate.

But uart_hal_get_stop_bits and uart_hal_get_parity calls do not need UART_ENTER_CRITICAL/UART_EXIT_CRITICAL protect?

components/driver/uart.c
o-marshmallow: You are right about performance, in fact, I meant that we would prefer to have both variable set atomically, in order to avoid being pre-empted in-between

o-marshmallow:
> Hi @o-marshmallow
> 
> Similar locking issue by checking below functions:
> uart_get_stop_bits
> uart_get_parity
> uart_get_baudrate
> 
> I'm wondering why in uart_get_baudrate(), it needs
> UART_ENTER_CRITICAL/UART_EXIT_CRITICAL around uart_hal_get_baudrate.
> 
> But uart_hal_get_stop_bits and uart_hal_get_parity calls do not need UART_ENTER_CRITICAL/UART_EXIT_CRITICAL protect?

Hi @AxelLin ,

This is because `uart_hal_get_baudrate` underneath is not atomic, it is not a single instruction but multiple instructions (this function is in fact called: https://github.com/espressif/esp-idf/blob/1cb31e50943bb757966ca91ed7f4852692a5b0ed/components/hal/esp32/include/hal/uart_ll.h#L128), whereas `uart_get_stop_bits` and `uart_get_parity` are atomic.

components/driver/uart.c
AxelLin: > ), whereas `uart_get_stop_bits` and `uart_get_parity` are atomic.

You mean this is atomic?
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32/include/hal/uart_ll.h#L381
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32/include/hal/uart_ll.h#L347

AxelLin:
> ), whereas `uart_get_stop_bits` and `uart_get_parity` are atomic.

You mean this is atomic?
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32/include/hal/uart_ll.h#L381
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32/include/hal/uart_ll.h#L347

o-marshmallow:
I guess I overlooked the ESP32 implementation, then we should enter critical sections in these cases also

AxelLin:
@o-marshmallow 

I notice this fix is **not** in recent esp-idf update, so I have no idea what's going on.
It's really does not make sense taking so long time to fix the bug.
(The PR was sent 5 months ago!!!)
IMHO, even the pull request as-is is good enough to fix the reported bug.
I have no idea what causes the delay...

o-marshmallow:
Hello @AxelLin ,

Thanks for your interest, I am going to perform the modifications requested before and open an internal merge request.

AxelLin:
Recent master tree update still does not include the fix.

AxelLin:
@o-marshmallow @Alvin1Zhang 
This is not a complex fix and it fixed real issue https://github.com/espressif/esp-idf/issues/6397
It's a terrible experience that such fix takes more than 6 months and not yet available now.
Stable branches also needs the fix and it will take even longer time to reach stable branches.

AxelLin:
@Alvin1Zhang  @o-marshmallow 

The esp-idf v4.3 and v4.2 still need the fix (It has been a long wait...)

o-marshmallow:
@AxelLin I am still working on this, this is still being processed. The fix has just been backported to v4.2.
It will be synchronized to Github once the CI tests pass. Regarding v4.3, it may be a bit longer.

AxelLin:
@o-marshmallow 
There is a conflict in  3691fb9 , you fixed it in  60ce887.
This could causes build error when people do bisect to find other problems.
Please make sure you fixed the conflict in right commit. (Just FYI)

o-marshmallow:
@AxelLin Thanks for the info! These two commits are part of the same internal MR, still, I'll be careful.

PR #6399, opened 2021-01-14
websocket: add configurable ping interval (IDFGH-4582)

zakkie:
add the ability to change the Websocket PING interval.

### DIFF
--- a/components/esp_websocket_client/include/esp_websocket_client.h
+++ b/components/esp_websocket_client/include/esp_websocket_client.h
@@ -88,7 +88,7 @@ typedef struct {
     char                        *headers;                   /*!< Websocket additional headers */
     int                         pingpong_timeout_sec;       /*!< Period before connection is aborted due to no PONGs received */
     bool                        disable_pingpong_discon;    /*!< Disable auto-disconnect due to no PONG received within pingpong_timeout_sec */
-
+    int                         ping_interval_sec;          /*!< Websocket ping interval */
 } esp_websocket_client_config_t;
 
 /**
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

david-cermak (approved):
Thanks for adding this config option.
Your changes LGTM!

components/esp_websocket_client/include/esp_websocket_client.h
```diff
@@ -88,7 +88,7 @@ typedef struct {
     char                        *headers;                   /*!< Websocket additional headers */
     int                         pingpong_timeout_sec;       /*!< Period before connection is aborted due to no PONGs received */
     bool                        disable_pingpong_discon;    /*!< Disable auto-disconnect due to no PONG received within pingpong_timeout_sec */
-
+    int                         ping_interval_sec;          /*!< Websocket ping interval */
```
septatrix, line 91: Why is it given is seconds here but stored in milliseconds above? I am new to this IDF so I do not know if this may be the default among other components too...

zakkie: - Because `pingpong_timeout_sec` was already there. And 
I think I'd never to make detailed settings on the order of milliseconds.
- In esp_websocket_client.c, comparing with _tick_get_ms(), it is milliseconds

Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/9ff9137e7a8b64e956c1c63e95a48f4049ad571e.

PR #6416, opened 2021-01-16
Make install.fish not dependent on being run from the root dir (IDFGH-4601)

iconara:
Using `$PWD` as the basedir only works when the script is run from the root directory. The Bash version determines the location of the root directory from the script, and this change makes the Fish script behave the same.

### DIFF
--- a/install.fish
+++ b/install.fish
@@ -1,6 +1,6 @@
 #!/usr/bin/env fish
 
-set basedir $PWD
+set basedir "$PWD/"(dirname (status -f))
 
 set -x IDF_PATH $basedir
 


Alvin1Zhang:
Thanks for your contribution.

ntwb:
Thanks for this, resolved a failed install for me using Fish

I also noticed that only `export.fish` is mentioned in the docs, would be helpful to get the `install.fish` details added to the docs below and any other locations that may be required:

> https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/index.html#step-3-set-up-the-tools

Now onto a Python debug to fix a conflict with Python 3.9 installed via home brew...

Alvin1Zhang:
Thanks for contribution, the PR is already in the merge queue.

igrr:
Based on https://github.com/espressif/esp-idf/pull/6718#issuecomment-800749697, we have cherry-picked a slightly different change: https://github.com/espressif/esp-idf/commit/f5f0648608e2a8dfb96dc1b082d0c6bb87318696. Hope this solves the issue for you as well @iconara. Thank you for contributing!

PR #6420, opened 2021-01-17
Weak linkage for esp_phy_load_cal_and_init (IDFGH-4605)

boarchuz:
Weakly link esp_phy_load_cal_and_init so that the application can optionally override initialisation/loading/storing of calibration data.

nb: esp_phy_rf_init previously offered this functionality but this API has been removed in 4.3 (and was probably not the intended usage regardless)

### DIFF
--- a/components/esp_wifi/src/phy_init.c
+++ b/components/esp_wifi/src/phy_init.c
@@ -523,7 +523,7 @@ static void __attribute((unused)) esp_phy_reduce_tx_power(esp_phy_init_data_t* i
 }
 #endif
 
-void esp_phy_load_cal_and_init(void)
+__attribute__((weak)) void esp_phy_load_cal_and_init(void)
 {
     char * phy_version = get_phy_version_str();
     ESP_LOGI(TAG, "phy_version %s", phy_version);


Alvin1Zhang:
Thanks for your contribution.

HarveyRong-Esp:
Hi @boarchuz , 
I didn’t find this function in the previous version. 
This function be used by modules such as WiFi BT, if it is modified into a weakly link, it may cause unpredictable issue.
If you want to refactor this API, it is recommended to modify the IDF directly for testing.

boarchuz:
Hi @HarveyRong-Esp 
Do you mean esp_phy_rf_init? It was here in release/4.2:
https://github.com/espressif/esp-idf/blob/28371f45a4ab7e1e340c36177ea02e30a9d810bf/components/esp_wifi/include/esp_phy_init.h#L195-L211

But there has been a slight refactor for 4.3 and this is removed. There is an API exposed for NVS of calibration data - load, store, and erase - but no way to actually _set_ the data any more. (Except with custom OSI function pointers in the WiFi/BT init config, which is a lot of added complexity for this one simple purpose.)

The use case here is very fast initialisation out of deep sleep. There was a PR here - https://github.com/espressif/esp-idf/pull/5708 (a bit hacky though) and I understand I can modify the component locally. I was hoping this very small change would be a more acceptable solution?

HarveyRong-Esp:
Hi @boarchuz, can it be solved using the following methods?

-  [esp_phy_store_cal_data_to_nvs(const esp_phy_calibration_data_t* cal_data)](https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_phy_init.h#L142)
- [esp_phy_enable()](https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/src/phy_init.c#L198)
-  [esp_phy_load_cal_and_init()](https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/src/phy_init.c#L526)

boarchuz:
Hi @HarveyRong-Esp , the objective is to only initialise NVS in the rare event that data needs to be loaded/saved (or even not at all - a full calibration could be performed first time, then persisted in RTC memory).
That works, but there isn't any benefit as long as `esp_phy_load_cal_and_init` still requires NVS and is the only way to set cal data.

Overriding `esp_phy_load_cal_and_init` would be ideal but something new like this might be an alternative?

```
esp_phy_set_cal_data(esp_phy_calibration_data_t* user_cal_data) {
  esp_err_t err = register_chipv7_phy(init_data, user_cal_data, MODE_NONE);
  if(err == ESP_OK) {
    s_is_phy_calibrated = true;
  }
  return err;
}
```

HarveyRong-Esp:
@boarchuz, Good way.

HarveyRong-Esp:
@boarchuz，Thank you very much for your contribution. We will not modify the API at this time. I think the above method will meet your needs. Suggest to close this PR.

boarchuz:
@HarveyRong-Esp Thanks for considering, I'll use wifi_osi_funcs_t

smhc:
Sorry to dig this up, but I think there may have been a misunderstanding here. The ["above method"](https://github.com/espressif/esp-idf/pull/6420#issuecomment-762213193) that is suggested isn't possible due to register_chipv7_phy and s_is_phy_calibrated being private.

It seems like a pretty simple and backwards compatible suggestion that would have solved this issue (although perhaps keys are still a problem). As it stands there's no API way to initialise wifi without a significant performance hit of either NVS or full calibration when it could have been easily avoided by this simple solution of using RTC memory instead.

The use of wifi_osi_funcs_t is a bit of an ugly hack, but thanks @boarchuz for your contribution as it's currently the only method available without IDF modifications.

PR #6448, opened 2021-01-22
Update esp_wifi.rst

maya2003:
Fixed typo


YouDONG-ESP:
Hi @maya2003 , thanks for point out the typos. We will add this fix with some others documentations update within one commit, so this PR will probably not be merged.

PR #6455, opened 2021-01-23
tinyusb: add 'extern C' bit to tusb_console.h (IDFGH-4641)

jesboat:
Public headers need the "if __cplusplus, extern C" boilerplate. Otherwise, C++
sources which include the header will look for a name-mangled symbol and fail
at link time.


Alvin1Zhang:
Thanks for your contribution.

PR #6461, opened 2021-01-25
fix(esp_tls): prevent freeing global CA store after each request (IDFGH-4647)

SiarheiVolkau:
Identical PR for ESP8266 https://github.com/espressif/ESP8266_RTOS_SDK/pull/1037
Not tested on ESP32.

### DIFF
--- a/components/mbedtls/port/dynamic/esp_mbedtls_dynamic_impl.c
+++ b/components/mbedtls/port/dynamic/esp_mbedtls_dynamic_impl.c
@@ -509,7 +509,9 @@ void esp_mbedtls_free_cacert(mbedtls_ssl_context *ssl)
         mbedtls_ssl_config *conf = (mbedtls_ssl_config *)ssl->conf;
 
 #ifdef CONFIG_ESP_TLS_USING_MBEDTLS
-        if (conf->ca_chain != esp_tls_get_global_ca_store())
+        if (conf->ca_chain == esp_tls_get_global_ca_store()) {
+            return;
+        }
 #endif
         mbedtls_x509_crt_free(conf->ca_chain);
         conf->ca_chain = NULL;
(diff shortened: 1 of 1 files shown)


components/mbedtls/port/dynamic/esp_mbedtls_dynamic_impl.c
```diff
@@ -505,6 +508,9 @@ void esp_mbedtls_free_cacert(mbedtls_ssl_context *ssl)
     if (ssl->conf->ca_chain) {
         mbedtls_ssl_config *conf = (mbedtls_ssl_config *)ssl->conf;
 
+#ifdef CONFIG_ESP_TLS_USING_MBEDTLS
+        if (conf->ca_chain != esp_tls_get_global_ca_store())
```
AdityaHPatwardhan, line 512:
```suggestion
        if (conf->ca_chain == esp_tls_get_global_ca_store()) {
            return;
        }
```

AdityaHPatwardhan:
@SiarheiVolkau , sorry for the late turnaround
We had internal discussions about this. The thing is `esp_tls` is a high level wrapper around `mbedtls` and `wolfssl`. So it is not appropriate to use `esp_tls` API in `mbedtls`. ( My bad, I should have mentioned this at the time of the reviewing this PR). Hence this PR can no longer be accepted.
I have fixed this issue with a different approach and appropriate MR has been raised in our internal code base. It should be merged in a couple of days in our internal codebase and then it will be synced to GitHub repo after some time.
Thank you for bringing our attention to this issue and your contribution in fixing it.

mahavirj:
@SiarheiVolkau As mentioned above this is now fixed (with different approach) in https://github.com/espressif/esp-idf/commit/bf513b6f31a4dac992b5cf0fd389b1389e46402c#diff-267dab0fdb0a487503dba92eb56e659f8889988d82c7608231db43604a5bbd79

Than you for your contribution and patience!

Issue #6421, opened 2021-01-17
gen_esp32part.py and deprecation of Python 2 (IDFGH-4606)


Martin-Laclaustra:
Due to deprecation of Python 2 some Linux distributions have decided to remove the generic "python" executable from the path, forcing all code to explicitly declare which version of Python they require (A wise solution in my opinion, as it serves as "canary" for undefined code and prevents obscure errors, resolving the longstanding ambiguity, and still allows the use of older python 2 code). This change has been annoying when one's mind is in other tasks, but one must adapt to new times and deprecation was announced years ago.
I was proposing to change the shebang to `#!/usr/bin/env python3` here https://github.com/espressif/arduino-esp32/issues/4717 but me-no-dev suggested that a global upstream action might be suitable.

Best regards.

Alvin1Zhang:
Thanks for reporting.

dobairoland:
Hi @Martin-Laclaustra. Thank you for the suggestion.

> Due to deprecation of Python 2 some Linux distributions have decided to remove the generic "python" executable from the path,

Which Linux distributions are you talking about? I think all of them provide some possibility to have a "python" defined.

I personally really don't like hardcoding "python3" into all of the scripts. (ESP-IDF have aprox. 160 scripts!). The right way is to setup the desired Python at the distribution level. We have recommendations regarding that in the documentation, for example for Ubuntu: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-setup.html#setting-up-python-3-as-default-for-ubuntu-and-debian Please take a look at them to see if any of them are suitable for you.

Another argument against is that the upcoming release (v4.3) still supports Python 2. Forcing Python 3 in the scripts would mean a breaking change for developers who stuck with Python 2.

The default Python is used only when you run `install.sh`. A virtual Python environment is created and `. export.sh` switches to that environment and "python" will become the interpreter installed in that virtual environment. Hardcoding sheebangs also could mess things up and use another Python for which `install.sh` didn't install the required packages.

dobairoland:
I also commented in the linked https://github.com/espressif/arduino-esp32/issues/4717. I'm closing this now because currently a couple of workarounds are documented and we are tracking related future improvements elsewhere.

adam-ah:
> I personally really don't like hardcoding "python3" into all of the scripts. (ESP-IDF have aprox. 160 scripts!). The right way is to set up the desired Python at the distribution level. 

[...]

PR #6471, opened 2021-01-28
Update idf_tools.py to use python3 instead of python2 (IDFGH-4659)

adam-ah:
The header should explicitly ask for python3 execution.

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # coding=utf-8
 #
 # This script helps installing tools required to use the ESP-IDF, and updating PATH


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Thank you @adam-ah for your contribution but I would like to achieve the same in a different way.

Python 2 is still supported in the current master branch (upcoming v4.3) and will be dropped in one of the next releases. However, a lot of developers are still stuck with old systems without Python 3. Adding this would break their system.

`idf_tools.py` is invoked from the install scripts. I would like to check there if `python3` exist in the PATH then invoke `idf_tools.py` with it. If you are interested then please feel free to follow-up on this idea. Otherwise, I will address this issue in the near future.

This topic has been discussed also in  https://github.com/espressif/esp-idf/issues/6421 and https://github.com/espressif/arduino-esp32/issues/4717.

adam-ah:
I see your points. However, requiring Python 2 is not correct and throws an exception on MacOS (as it's missing key components, like pip):
```
./tools/idf_tools.py install-python-env 
Creating a new Python environment in /Users/adam/.espressif/python_env/idf4.3_py2.7_env
Installing virtualenv
/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: No module named pip
  File "./tools/idf_tools.py", line 1622, in <module>
    main(sys.argv[1:])
  File "./tools/idf_tools.py", line 1618, in main
    action_func(args)
  File "./tools/idf_tools.py", line 1335, in action_install_python_env
    stdout=sys.stdout, stderr=sys.stderr)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 190, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python', '-m', 'pip', 'install', '--user', 'virtualenv']' returned non-zero exit status 1
```

The header-change I made is valid given Python 2 is end-of-life with support ended over a year ago (2020-01-01).

However, it might be worthwhile investigating modifying the runner bash script to fall back to Python 2 is 3 is not available, with a warning message.

I'm happy to look into modifying the wrapper .sh, if you accept that change?

dobairoland:
@adam-ah

> The header-change I made is valid given Python 2 is end-of-life with support ended over a year ago (2020-01-01). 

Yes this is true and I very much agree with you. I'm using Python 3 as my default interpreter in the last 10 years and I don't like neither that in 2021 we are still talking about supporting Python 2. It is even worse that Python 2 is not supported but we still have to support for our tools. However, I cannot agree to change the shebang right now.  

ESP-IDF uses virtual Python environment generated by the interpreter used with `idf_tools.py`. The rest of the scripts is invoked in the virtual environment where `python` exists and is based on the interpreter used with `idf_tools.py`. So changing the sheebang in that file would be really easy and I would like to do it if I could. 

Just to mention one possible issue against implementing the "perfect way" and hardcoding `python3` into our scripts: in Windows command line in our virtual environment `python` is of version 3 and there is no `python3`. 

`install.sh` invokes `${IDF_PATH}/tools/idf_tools.py install-python-env`. My idea is to change it to `${PYTHON} ${IDF_PATH}/tools/idf_tools.py install-python-env` and set PYTHON to `python3` if it exists in the PATH or use `python` as fallback. I would print `which ${PYTHON}` for diagnostic reasons. The same has to be done in the other install shell scripts as well (and probably in the export scripts as well). So probably a new python script for detecting the value of PYTHON and printing warnings would be useful. Or a Python wrapper for `idf_tools.py`. The documentation for the affected OSs has to be updated as well.

I'm planning to start to work on this the next week. However, If you find this acceptable I would pretty much appreciate your help. Feel free to update this PR and start the implementation. I would continue and do the rest on top of your contribution.

Thank you for your help with this and sharing your opinion. I'm sorry for being so rejecting regarding the update of the shebang.

dobairoland:
This is just a note that I'm starting to work on this. It will be closed with the related commit.

adam-ah:
Thanks for letting me know, @dobairoland. Give me a shout if you need help with any of these (e.g., testing it or whatnot)

PR #6491, opened 2021-02-02
Update MQTT-Client example (IDFGH-4679)

sebastianstampfel:
Fixes possible LoadStoreError by initializing the config-struct in the mqtt-example properly.

If - like now - only the uri field is set, values for other fields are undefined, which ends in a LoadStoreError on my ESP32-WROOM-32 board.  Fixed by initially setting all fields to zero and **then** adding custom configuration (uri, in the case of this example).

See https://github.com/espressif/arduino-esp32/issues/4647 for further information.

### DIFF
--- a/examples/protocols/mqtt/tcp/main/app_main.c
+++ b/examples/protocols/mqtt/tcp/main/app_main.c
@@ -104,9 +104,9 @@ static void mqtt_event_handler(void *handler_args, esp_event_base_t base, int32_
 
 static void mqtt_app_start(void)
 {
-    esp_mqtt_client_config_t mqtt_cfg = {
-        .uri = CONFIG_BROKER_URL,
-    };
+    esp_mqtt_client_config_t mqtt_cfg = {0};
+    mqtt_cfg.uri = CONFIG_BROKER_URL;
+    
 #if CONFIG_BROKER_URL_FROM_STDIN
     char line[128];
 


KaeLL:
This doesn't make sense. Have you read [this](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/fatal-errors.html#loadstoreerror)? Also...
> C11 6.7.9p19
> -- all subobjects that are not initialized explicitly shall be initialized implicitly the same as objects that have static storage duration.

Whatever it is that's causing that error for you, I'm not sure it's related to what your change addresses. If anything, it's just redundant.

sebastianstampfel:
@KaeLL  Haha alright, by fiddling around with struct-initialization in gcc/gdb I just figured out that apparently not all initializations are equal, meaning that
```
esp_mqtt_client_config_t mqtt_cfg;
mqtt_cfg.uri = SOME_VALUE;
```

is apparently not the same as

```
esp_mqtt_client_config_t mqtt_cfg = {
    .uri = SOME_VALUE
};
```

Thx for the heads up, the more you know!

david-cermak:
Let me add a note about designated initializers, just for completeness. They work in C++ too, but need to be careful about the order of fields and aggregate types. IDF users could also use this macro:

https://github.com/espressif/esp-idf/blob/master/components/esp_common/include/esp_compiler.h#L33-L49

It seems like the trouble in the referenced issue was incorrect use of designated initializers in C++/Arduino.

PR #6541, opened 2021-02-15
fix: infinite loop on closed socket (IDFGH-4735)

SiarheiVolkau:
In some scenarios returning 0 there result in infinite loop on closed socket (esp_http_server in particul).

### DIFF
--- a/components/mbedtls/port/dynamic/esp_ssl_tls.c
+++ b/components/mbedtls/port/dynamic/esp_ssl_tls.c
@@ -89,7 +89,7 @@ int __wrap_mbedtls_ssl_read(mbedtls_ssl_context *ssl, unsigned char *buf, size_t
     ret = esp_mbedtls_add_rx_buffer(ssl);
     if (ret == MBEDTLS_ERR_SSL_CONN_EOF) {
         ESP_LOGD(TAG, "fail, the connection indicated an EOF");
-        return 0;
+        return ret;
     } else if (ret < 0) {
         ESP_LOGD(TAG, "fail, error=-0x%x", -ret);
         return ret;


Alvin1Zhang:
Thanks for your contribution.

donghengqaz:
@SiarheiVolkau Thanks for your PR, but I think this is not an issue, we can see https://github.com/espressif/mbedtls/blob/99c88bb7bd7c2d91c6873abd6e6d5ee04f4f164c/library/ssl_tls.c#L8451, when `mbedtls_ssl_read` get a error code `MBEDTLS_ERR_SSL_CONN_EOF`, it will also return 0 to application.

donghengqaz:
When you don't use the mbedtls dynamic buffer, is there the same issue catched ?

SiarheiVolkau:
I see that documentation is updated for that behavior (0 as EOF) but not all code in codebase modified to this rule .
For example there you can catch an infinite loop:
https://github.com/espressif/esp-idf/blob/master/components/protocomm/src/transports/protocomm_httpd.c#L138
and there:
https://github.com/espressif/esp-idf/blob/master/components/esp_http_server/src/httpd_parse.c#L809

shubhamkulkarni97:
> I see that documentation is updated for that behavior (0 as EOF) but not all code in codebase modified to this rule .
> For example there you can catch an infinite loop:
> https://github.com/espressif/esp-idf/blob/master/components/protocomm/src/transports/protocomm_httpd.c#L138
> and there:
> https://github.com/espressif/esp-idf/blob/master/components/esp_http_server/src/httpd_parse.c#L809

Hi @SiarheiVolkau,

Thanks for bringing up this issue. Instead of changing return value incase of `MBEDTLS_ERR_SSL_CONN_EOF`, I think it is better to fix the check for return value in esp_http_server and protocomm components (this will also fix the issue with non TLS use cases)

SiarheiVolkau:
@shubhamkulkarni97,  yep, i think so.

shubhamkulkarni97:
@SiarheiVolkau, can you share test code to reproduce this issue and verify the fix?

SiarheiVolkau:
No, i can't, since the connection must be closed on server side to reproduce the issue.
You can try to use OTA example modified to use controlled HTTP server as file source. Then stop server during file download.

shubhamkulkarni97:
Suggestion to check recv length looks correct. But we are not able to recreate a scenario to verify this fix.

shubhamkulkarni97:
@SiarheiVolkau 

Can you try applying attached patch and check if it fixes the issue?
[0001-esp_http_server-protocomm-Consider-a-connection-as-c.patch.zip](https://github.com/espressif/esp-idf/files/6520025/0001-esp_http_server-protocomm-Consider-a-connection-as-c.patch.zip)

SiarheiVolkau:
@shubhamkulkarni97

The patch you propose looks well.  
Seems like it should fix problem mentioned in this PR.  
Thanks.

PR #6552, opened 2021-02-16
singular form of quanta is quantum (IDFGH-4748)

crackwitz:
it's like bacteria/bacterium, phenomena/phenomenon


Alvin1Zhang:
Thanks for your contribution.

Issue #6609, opened 2021-02-25
Missing documentation / example for factory and test app (IDFGH-4810)


0xFEEDC0DE64:
Maybe I'm missing something, but I do not find any documentation or examples on how to create the factory and test app.

I've prepared the partition table with a factory app partition but where should I place the c/cpp sources for it? I've enabled the GPIO triggeres in the bootloader and I get output like this:

    I (167) boot: Defaulting to factory image       
    E (172) esp_image: image at 0x860000 has invalid magic byte
    W (178) esp_image: image at 0x860000 has invalid SPI mode 255                                                                                                  
    W (185) esp_image: image at 0x860000 has invalid SPI size 15

I'm looking for commands like `idf.py factory factory-flash`

I need to share some code between the main app and factory app to read/write the configs, so I thought about using common components?

**EDIT:** Since adding the factory partition I also cannot flash the main app anymore as `idf.py app-flash` tries to flash the main app into the factory partition, which is way smaller that the main app's partition. I would like to expand my question on how to specify the partition during app-flash, so I can decide if I want to flash the factory or main app partition.

0xFEEDC0DE64:
I'm following the bootloader subproject structure in combination with my branch which allows flashing the app to different partitions, it looks like it is a lot of work to get a factory app up and running

projectgus:
Hi @0xFEEDC0DE64,

You're right that there's not a lot of documentation or tooling for this, thanks for pointing it out.

Our assumption is that this will usually be a .bin file built from a separate project, that's flashed in the production process rather than flashed from the main project's build system. But of course we can still provide some steps to make it easier to work with flashing and testing these different apps in development with a common partition table (rather than having to run `esptool.py` manually). Your PR looks like a great step towards this - thanks.

> I've prepared the partition table with a factory app partition but where should I place the c/cpp sources for it?

Just to check: for the "factory" app, by default this the partition which is flashed by the build system. For example, in our simplest default partition table we only have a factory app:
https://github.com/espressif/esp-idf/blob/master/components/partition_table/partitions_singleapp.csv

and for our "full" OTA pre-built partition table  we have one factory app and two OTA app slots:

https://github.com/espressif/esp-idf/blob/master/components/partition_table/partitions_two_ota.csv

[...]

PR #6613, opened 2021-02-25
Allow flashing main app to different partition that default

0xFEEDC0DE64:
Added Kconfig options to select different partition then default selected which is select incorrectly with OTA and Factory.


0xFEEDC0DE64:
Helps implementing #6609

Alvin1Zhang:
Thanks for your contribution.

0xFEEDC0DE64:
Ok is this PR already in consideration? Are any questions open? Can I help somehow to get this merged?

0xFEEDC0DE64:
@Alvin1Zhang ping?

projectgus:
Hi @0xFEEDC0DE64 ,

Thanks for being patient while someone got back to you about this. This looks like a very useful thing to support!

The only unfortunate thing is that any change in project config requires a full rebuild of the entire project - whereas this change is just at flashing time and can use the same binary.

As an alternative, what would you think if we added some new flash targets to the build system instead. i.e. in addition to `app-flash` (which stays the same) we can add `app-flash-ota_0`, `app-flash-ota_1`, `app-flash-test`, etc?

This would require running a different target to `idf.py flash` in order to flash the desired partition, but it would allow doing this without triggering a rebuild.

igrr:
As one more alternative, how about an argument of `idf.py flash`, such as `idf.py flash --to ota1`? This means that flashing to a non-default partition won't be possible by invoking a CMake target (such as, from some IDE), but it seems to be an acceptable tradeoff.

projectgus:
> As one more alternative, how about an argument of idf.py flash, such as idf.py flash --to ota1? 

I agree this would be a good user experience. Unfortunately it's a lot more complex to implement, because `idf.py flash` currently operates by calling the build system target "flash". When CMake runs, it reads the partition table and generates args files like `flash_project_ args` in the build directory, and it creates targets that use those args files. Then the flash command in idf.py just runs `ninja flash` which does the actual flashing without needing to be aware of how the arguments were generated.

So adding additional app flashing targets (which can then be exposed via idf.py commands also) is quite straightforward in this model, but making idf.py aware of the partition table details would require some parts of the implementation to move to "idf.py runtime" or "build runtime" rather than "CMake runtime" where they currently are.

igrr:
Theoretically the information required for idf.py to implement flashing is available in build/flasher_args.json, so we could move  this logic to idf.py. As I understand, the reason it was moved to CMake run time was to support flashing by executing a CMake target in an IDE (such as CLion). To support both use cases, we could move the implementation out of CMake again, into a helper utility, and invoke this utility both from CMake and from idf.py.

PR #6631, opened 2021-02-28
Implemented esp_core_dump_image_erase() (IDFGH-4832)

0xFEEDC0DE64:
Implemented esp_core_dump_image_erase() and esp_core_dump_image_get() now returns ESP_ERR_NOT_FOUND when partition is blank/erased.

esp_core_dump_image_erase() calls esp_partition_erase_range() and additionally writes size 0xFFFFFFFF to avoid problems when flash encryption is enabled. Without explicitly writing 0xFFFFFFFF for size, esp_core_dump_image_get() will read a garbage size when the erased bytes are being decrypted.

With size 0xFFFFFFFF esp_core_dump_image_get() now returns ESP_ERR_NOT_FOUND instead of ESP_ERR_INVALID_SIZE

I needed to know if an error happened during reading the coredump or if it has simply been not found / been invalid or erased (because never crashed or coredump has been transmitted already successfully)

### DIFF
--- a/components/espcoredump/include/esp_core_dump.h
+++ b/components/espcoredump/include/esp_core_dump.h
@@ -92,6 +92,6 @@ esp_err_t esp_core_dump_image_get(size_t* out_addr, size_t *out_size);
  *
  * @return ESP_OK on success, otherwise \see esp_err_t
  */
-esp_err_t esp_core_dump_image_erase();
+esp_err_t esp_core_dump_image_erase(void);
 
 #endif
--- a/components/espcoredump/include/esp_core_dump.h
+++ b/components/espcoredump/include/esp_core_dump.h
@@ -85,4 +85,13 @@ void esp_core_dump_to_uart(panic_info_t *info);
  */
 esp_err_t esp_core_dump_image_get(size_t* out_addr, size_t *out_size);
 
+/**
+ * @brief  Erases coredump data in flash. esp_core_dump_image_get() will then return
+ *         ESP_ERR_NOT_FOUND. Can be used after a coredump has been transmitted successfully.
+ *         This function is always available, even when core dump is disabled in menuconfig.
+ *
+ * @return ESP_OK on success, otherwise \see esp_err_t
+ */
+esp_err_t esp_core_dump_image_erase();
+
 #endif
--- a/components/espcoredump/src/core_dump_flash.c
+++ b/components/espcoredump/src/core_dump_flash.c
@@ -19,6 +19,8 @@
 #include "esp_flash_encrypt.h"
 #include "esp_rom_crc.h"
 
+#define BLANK_COREDUMP_SIZE 0xFFFFFFFF
+
 const static DRAM_ATTR char TAG[] __attribute__((unused)) = "esp_core_dump_flash";
 
 #if CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH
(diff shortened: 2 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/espcoredump/src/core_dump_flash.c
```diff
@@ -455,3 +455,37 @@ esp_err_t esp_core_dump_image_get(size_t* out_addr, size_t *out_size)
...
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "Failed to erase core dump partition (%d)!", err);
+        return err;
+    }
+
+    // on encrypted flash esp_partition_erase_range will leave encrypted
+    // garbage instead of 0xFFFFFFFF so overwriting again to safely signalize
+    // deleted coredumps
+    const uint32_t invalid_size = 0xFFFFFFFF;
+    err = esp_partition_write(core_part, 0, &invalid_size, sizeof(invalid_size));
```
KaeLL, line 485: This should be wrapped by an `ifdef` directive checking if NVS_ENCRYPTION is enabled.

gerekon (approved):

components/espcoredump/include/esp_core_dump.h
```diff
@@ -85,4 +85,13 @@ void esp_core_dump_to_uart(panic_info_t *info);
...
 esp_err_t esp_core_dump_image_get(size_t* out_addr, size_t *out_size);
 
+/**
+ * @brief  Erases coredump data in flash. esp_core_dump_image_get() will then return
+ *         ESP_ERR_NOT_FOUND. Can be used after a coredump has been transmitted successfully.
+ *         This function is always available, even when core dump is disabled in menuconfig.
+ *
+ * @return ESP_OK on success, otherwise \see esp_err_t
+ */
+esp_err_t esp_core_dump_image_erase();
```
o-marshmallow, line 95: Please, can you explicitly put `void` in the parenthesis to forbid passing arguments to the function?

components/espcoredump/src/core_dump_flash.c
```diff
@@ -455,3 +455,37 @@ esp_err_t esp_core_dump_image_get(size_t* out_addr, size_t *out_size)
     *out_addr = core_part->address;
     return ESP_OK;
 }
+
+esp_err_t esp_core_dump_image_erase()
```
o-marshmallow, line 459: As said above, can you please add `void` in the parenthesis ?

components/espcoredump/src/core_dump_flash.c
o-marshmallow: In both cases (encryption enabled or not), I find it is a good idea to explicitly set this field. Thus, the multi-line comment can be deleted (or replaced by a comment such as "Mark core dump as deleted by setting field size"...) and a macro can be added to hold the value `0xFFFFFFFF`:
```
#define ESP_CORE_DUMP_BLANK 0xFFFFFFFF
```

o-marshmallow (changes requested):
I added few comments, mostly details, please feel free to give me your opinion on them.

0xFEEDC0DE64:
anything I can do for this PR?

o-marshmallow:
Hello @0xFEEDC0DE64 ,

Sorry for the delay (again), your PR has been accepted and is currently passing internal reviews. Once merged, it will be synced to Github after passing the internal tests.

Issue #6633, opened 2021-02-28
Problem with CPP and V4.3 (IDFGH-4834)


william-ferguson-au:
## Environment

- Development Kit:      none
- Module or chip used:  ESP32-S2-WROVER
- IDF version  v4.3-beta1-24-g0bbc721a6
- Build System:         CMake idf.py
- Compiler version:  xtensa-esp32-elf-gcc (crosstool-NG esp-2020r3) 8.4.0
- Operating System:     Windows
- (Windows only) environment type: ESP Command Prompt
- Using an IDE?:   Yes CLion but build from Cmd line
- Power Supply:         Battery

## Problem Description

Fails to compile with:
```cpp
In file included from E:\Dev\Espressif\esp-idf\components\driver\include/driver/gpio.h:20,
                 from E:\Dev\Espressif\esp-idf\components\driver\include/driver/adc_common.h:21,
                 from E:\Dev\Espressif\esp-idf\components\driver\esp32s2\include/driver/adc.h:16,
                 from E:\Dev\Espressif\esp-idf\components\esp_adc_cal\include/esp_adc_cal.h:24,
                 from E:\Source\wylas\beamatouch\main\main.cpp:1:
E:\Dev\Espressif\esp-idf\components\xtensa\include/esp_attr.h:111:34: error: conflicting declaration of C function 'constexpr touch_pad_intr_mask_t operator~(touch_pad_intr_mask_t)'
 FORCE_INLINE_ATTR constexpr TYPE operator~ (TYPE a) { return (TYPE)~(INT_TYPE)a; } \
                                  ^~~~~~~~
E:\Dev\Espressif\esp-idf\components\xtensa\include/esp_attr.h:123:29: note: in expansion of macro 'FLAG_ATTR_IMPL'
 #define FLAG_ATTR_U32(TYPE) FLAG_ATTR_IMPL(TYPE, uint32_t)
                             ^~~~~~~~~~~~~~
E:\Dev\Espressif\esp-idf\components\xtensa\include/esp_attr.h:124:19: note: in expansion of macro 'FLAG_ATTR_U32'
 #define FLAG_ATTR FLAG_ATTR_U32
                   ^~~~~~~~~~~~~
E:\Dev\Espressif\esp-idf\components\hal\include/hal/touch_sensor_types.h:166:1: note: in expansion of macro 'FLAG_ATTR'
 FLAG_ATTR(touch_pad_intr_mask_t)
 ^~~~~~~~~
[...]

PR #6634, opened 2021-02-28
Fix for #6633 (IDFGH-4835)

william-ferguson-au:
Moving #includes above #ifdef __cplusplus extern "C" { #endif
So that we can compile with CPP.

This is a fix for https://github.com/espressif/esp-idf/issues/6633
NB this PR fixes in the v4.3 branch. I presume there are other branches (like master) to which this should also be applied.

### DIFF
--- a/components/driver/esp32s2/include/driver/touch_sensor.h
+++ b/components/driver/esp32s2/include/driver/touch_sensor.h
@@ -13,11 +13,11 @@
 
 #pragma once
 
+#include "driver/touch_sensor_common.h"
+
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


0xjakob (approved):
Looks good. Thanks for this PR! We'll integrate this into our internal master and CI. Should be on github during this or next week.

Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Changes merged with https://github.com/espressif/esp-idf/commit/13bc5feb0a787a10a09a63184d7273292fd4d9c7

PR #6652, opened 2021-03-02
Explain the need to enable BT and BTDM BLE in SDK (IDFGH-4854)

mjcross:
If users may try to use this (excellent) example code in their own projects, without understanding the need to enable the BT stack and BTDM settings in the SDK. If they do that then their builds will fail with linker errors but they may not understand why.

### DIFF
--- a/examples/provisioning/wifi_prov_mgr/README.md
+++ b/examples/provisioning/wifi_prov_mgr/README.md
@@ -22,6 +22,8 @@ Right after provisioning is complete, BLE is turned off and disabled to free the
 
 This example can be used, as it is, for adding a provisioning service to any application intended for IoT.
 
+> Note: If you use this example code in your own project, in BLE mode, then remember to enable the BT stack and BTDM BLE control settings in your SDK configuration (e.g. by using the `sdkconfig.defaults` file from this project).
+
 ## How to use example
 
 ### Hardware Required
--- a/examples/provisioning/wifi_prov_mgr/main/app_main.c
+++ b/examples/provisioning/wifi_prov_mgr/main/app_main.c
@@ -237,6 +237,10 @@ void app_main(void)
             0xb4, 0xdf, 0x5a, 0x1c, 0x3f, 0x6b, 0xf4, 0xbf,
             0xea, 0x4a, 0x82, 0x03, 0x04, 0x90, 0x1a, 0x02,
         };
+        
+        /* if your build fails with linker errors at this point, then you may have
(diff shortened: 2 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

mahavirj:
@mjcross Instead of this PR, if we select additional config options at https://github.com/espressif/esp-idf/blob/master/examples/provisioning/wifi_prov_mgr/main/Kconfig.projbuild#L12 will that help?

mjcross:
@mahavirj Yes I think that's an improvement - thanks 👍 
The challenge is that new users (like me) read the API documentation but don't understand why/how to configure the SDK until they get hit with a page full of linker errors; and for many people provisioning will be their first contact with BT...

mahavirj:
>  The challenge is that new users (like me) read the API documentation but don't understand why/how to configure the SDK until they get hit with a page full of linker errors; and for many people provisioning will be their first contact with BT...

I agree, however we do have some part of this covered in documentation at https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html. Additionally, just would like to mention that we can also keep BT enabled by default, but that has some impact on static footprint (RAM usage) of application and hence choice has been left to end application regarding this.

> Yes I think that's an improvement

Could you pleas help to update PR in that case? I will pull it in our internal review queue post that.

Thank your for your patience!

mjcross:
@mahavirj - I've had a look at modifying the PR to move the BT and BTDM config options from `sdkconfig.defaults` to `Kconfig.projbuild` - but to be honest I think it is a bit beyond my capability. I am not sufficiently familiar or confident with either the menu config system or the SDK BLE options, so there is a risk that I do more harm than good (!)

Also I can see it's going to be necessary to test a modification like that against the different provisioning options (BT / SoftAP / Nimble / Bluedroid) and I don' have the opportunity to do that here.

So my apologies that I can't help more - I think the best I can suggest is that adding a comment to the code as per the original PR might be the simplest and 'safest' approach; and it has the advantage that it points users in the right direction to be able to use Provisioning - which is after all the objective!

mahavirj:
@mjcross 

> I am not sufficiently familiar or confident with either the menu config system or the SDK BLE options, so there is a risk that I do more harm than good (!)

Sure, no problem.

> I think the best I can suggest is that adding a comment to the code as per the original PR might be the simplest and 'safest' approach; and it has the advantage that it points users in the right direction to be able to use Provisioning - which is after all the objective!

Makes sense. I will put in this change in our internal review queue.

Thank you!

PR #6658, opened 2021-03-04
Can't use spiram.h functions when building with cpp. (IDFGH-4861)

mongonta0716:
Functions in spiram.h cannot be used when built with cpp because extern "C" is not defined in spiram.h.

### DIFF
--- a/components/esp32/include/esp32/spiram.h
+++ b/components/esp32/include/esp32/spiram.h
@@ -21,6 +21,10 @@
 #include <stdbool.h>
 #include "esp_err.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 typedef enum {
     ESP_SPIRAM_SIZE_16MBITS = 0,   /*!< SPI RAM size is 16 MBits */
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

ginkgm:
Hi @mongonta0716 ,

Thanks for your contribution!

Generally speaking, it's non-public API, so we don't suggest you use that header directly, and haven't add CPP guard for it. But since you have your use case, we will accept this.

mongonta0716:
Hi @ginkgm ,Thanks for your reply.

I noticed this API while doing research on devices with 8MB of PSRAM. I'm sorry I didn't know it was non-public.

Is there any other API to know the total amount of PSRAM?

igrr:
Hi @mongonta0716, thank you for the PR. Your fix has been cherry-picked as 897496d with a subsequent commit applying similar fix for other chips. `esp_spiram_get_chip_size` function indeed looks like the correct one to use for the purpose you have mentioned.

PR #6665, opened 2021-03-05
component: adc/i2s: support multi channel ADC DMA

karlp:
Allow configuring ADC DMA via the i2s peripheral for multiple channels,
instead of hard limiting it to a single channel.

See-also: https://github.com/espressif/esp-idf/pull/1991

Signed-off-by: Karl Palsson <karlp@tweak.net.au>

This is very heavily inspired by #1991, simply updated to current 4.2 release.  Should we leave the original single channel API in place and add a _new_ api? or change it like this?

### DIFF
--- a/components/driver/esp32/adc.c
+++ b/components/driver/esp32/adc.c
@@ -84,19 +84,22 @@ esp_err_t adc_set_i2s_data_source(adc_i2s_source_t src)
     return ESP_OK;
 }
 
-esp_err_t adc_i2s_mode_init(adc_unit_t adc_unit, adc_channel_t channel)
+esp_err_t adc_i2s_mode_init(adc_unit_t adc_unit, size_t pattern_count, adc_digi_pattern_table_t *patterns)
 {
+    ADC_CHECK(pattern_count <= 16, "ADC only supports max 16 patterns", ESP_ERR_INVALID_ARG);
     if (adc_unit & ADC_UNIT_1) {
         ADC_CHECK((SOC_ADC_SUPPORT_DMA_MODE(ADC_NUM_1)), "ADC1 not support DMA for now.", ESP_ERR_INVALID_ARG);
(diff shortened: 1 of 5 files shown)


Alvin1Zhang:
Thanks for contribution.

ginkgm:
Hi @karlp ,

Due to back-compatible reason, we can't make such API change. However, we have added a set of new ADC-DMA APIs to replace the old ones. Currently it's only for C3,  But we have a plan to add that support to earlier chips (ESP32, ESP32-S2). 

You can find the new API here: 
https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/peripherals/adc.html?highlight=adc_digi_controller_config#_CPPv426adc_digi_controller_configPK17adc_digi_config_t. 
Could you please kindly have a look to see whether it meets your requirements?

karlp:
I honestly can't tell from the api docs I'm afraid.  Having DMA from ADC channels be available via ADC apis seems far more sensible and easier to understand than using i2s apis though, yes.

I do know this breaks the api, would you consider it as an _extra_ api?  adc_i2s_mode_init_multi() or similar?  the existing single hard channel api could be kept, and could use the same implementation internally.

karlp:
See #6685 for a an alternative that doesn't break any api compatibility

ginkgm:
Hi @karlp ,
Controlling ADC by i2s is historical design, which seems not happening to future chips. Since we have the ADC-DMA driver, the overall plan is to keep the ADC part in ADC-DMA driver, and gradually remove the ADC support in I2S.

karlp:
That's completely fine with me, I submitted this because I needed it to even evaluate whether I could use the part.  multi channel adc dma via adc without having to think about i2s makes a lot more sense to me, API wise.

ginkgm:
Thanks for this PR. Close it for the above reasons.

Issue #6670, opened 2021-03-06
idf.py build system fails on 4.3 and 4.2 due to missing referece to esptool_py_flash_target function in component/partition_table/CMakeLists.txt (IDFGH-4874)


marcidy:
## Environment

- Module or chip used:  ESP32
- IDF version (run ``git describe --tags`` to find it): v4.3, v4.2
- Build System:         idf.py
- CMake version:  3.13.4
- Operating System:     Linux
- Using an IDE?: No

## Problem Description

create a project with a component that REQUIRES app_update

idf,py build fails with error:
```
CMake Error at /toolchains/esp32/esp-idf/components/partition_table/CMakeLists.txt:101 (esptool_py_flash_target):
  Unknown CMake command "esptool_py_flash_target".

```
upgrading from v4.1 to v4.2 or v4.3 resulted failure to build a simple project. (The project doesn't compile at the moment, though this is not relevant to the build failure, which doesn't get to the compile stage)

I removed the build directory and rebuilt the toolchain a number of times with different versions of the idf, all with the same result.

I deleted esp-idf repo clone, removed ~/.espressif/ directories, cloned from scratch per instructions,  and attempted a number of fresh installations (install.sh, export.sh) using a few versions.  All failed with the same error. 

multiple `idf.py commands` failed similarly.

### Steps to reproduce

run any invocation of idf.py <command> on a fresh installation

## Debug Logs

```
$ idf.py build
Executing action: all (aliases: build)
Running make in directory /simple/build
Executing "make -j 6 all"...
-- Project is not inside a git repository, or git repository has no commits; will not use 'git describe' to determine PROJECT_VER.
-- Building ESP-IDF components for target esp32
-- Project sdkconfig file /simple/sdkconfig
CMake Error at /toolchains/esp32/esp-idf/components/partition_table/CMakeLists.txt:101 (esptool_py_flash_target):
  Unknown CMake command "esptool_py_flash_target".


-- Configuring incomplete, errors occurred!
See also "/simple/build/CMakeFiles/CMakeOutput.log".
make: *** [Makefile:1220: cmake_check_build_system] Error 1
make failed with exit code 2

```

## Possible Solution Indication

`esptool_py_flash_target` is defined in the `esptool_py` component, in `project_include.cmake`.

Inspecting the `partition_table` CMakeList.txt, I didn't see a reference to the esptool_py component.  I changed line 1 in `components/partition_table/CMakeLists.txt` to:
```
[...]

PR #6672, opened 2021-03-07
partition_table requires esptool_py for esptool_py_flash_target (IDFGH-4876)

marcidy:
possible fix for #6670 

Not sure if this is the right approach.

A sample project is attached and posted to the issue which shows the behavior.  The project will trigger the issue without this fix.  With this fix, the project will build (but will not compile, which is fine).

Including component app_update triggers the issue, but i believe that's due to alphabetical component building.  The root cause appears to be partition_table doesn't explicitly require esptool_py.

[simple.zip](https://github.com/espressif/esp-idf/files/6096346/simple.zip)

### DIFF
--- a/components/partition_table/CMakeLists.txt
+++ b/components/partition_table/CMakeLists.txt
@@ -1,4 +1,4 @@
-idf_component_register()
+idf_component_register(PRIV_REQUIRES esptool_py)
 
 if(BOOTLOADER_BUILD)
     return()


Alvin1Zhang:
Thanks for your contribution.

projectgus:
This is the right approach, thanks! MR will be updated when this merges.

Issue #5897, opened 2020-09-22
portGET_ARGUMENT_COUNT() defintion broken for C++ (IDFGH-4021)


seijikun:
## Environment

- Development Kit:      none
- Module or chip used:  ESP32-WROOM-32
- IDF version (run ``git describe --tags`` to find it): v4.3-dev-1197-g8bc19ba89
- Build System:  idf.py
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it):  xtensa-esp32-elf-gcc (crosstool-NG esp-2020r3) 8.4.0
- Operating System:     Linux
- Using an IDE?: No
- Power Supply:         USB

## Problem Description
The macro `portGET_ARGUMENT_COUNT()` in [portmacro.h line 331](https://github.com/espressif/esp-idf/blob/master/components/freertos/xtensa/include/freertos/portmacro.h#L331) seems to be broken when used with C++.
Have a look at the demonstration [here](https://onlinegdb.com/HkxCTPDBP). With `0` arguments, it returns `1` instead of `0`. This triggers the first static assertion directly below the definition:
```c
_Static_assert(portGET_ARGUMENT_COUNT() == 0, "portGET_ARGUMENT_COUNT() result does not match for 0 arguments");
```
As can be seen [here](https://onlinegdb.com/HyjM1_DSD), this works when instead run with C.

### Code to reproduce this issue

```cpp
#include <iostream>

#define portGET_ARGUMENT_COUNT(...) portGET_ARGUMENT_COUNT_INNER(0, ##__VA_ARGS__,1,0)
#define portGET_ARGUMENT_COUNT_INNER(zero, one, count, ...) count

int main() {
    std::cout << portGET_ARGUMENT_COUNT() << std::endl;
    std::cout << portGET_ARGUMENT_COUNT(1) << std::endl;
    std::cout << portGET_ARGUMENT_COUNT(1,2) << std::endl;
    std::cout << "###" << std::endl;

    return 0;
}
```

## Log
I am using a main.cpp in my main component. The include tree looks as follows:
```
In file included from <esp32_path>/tools/tools/xtensa-esp32-elf/esp-2020r3-8.4.0/xtensa-esp32-elf/xtensa-esp32-elf/sys-include/stdlib.h:19,
[...]

PR #6692, opened 2021-03-11
Fix portGET_ARGUMENT_COUNT macro (IDFGH-4898)

b1ackviking:
portGET_ARGUMENT_COUNT uses a GCC extension `##__VA_ARGS__`. It forces the user to compile the code with GNU extensions enabled instead of ISO language standard, i.e. use `-std=gnuXX` and `-std=gnu++XX` instead of `-std=cXX` and `-std=c++XX`. The suggested change is to replace `##__VA_ARGS__` with `__VA_OPT__(,) __VA_ARGS__` which is supported by the current version of GCC (8.4.0) used in ESP-IDF for both C and C++ ISO standards.

This fix would enable ESP-IDF users to compile their code with ISO C++20 standard in future releases.

Resolves https://github.com/espressif/esp-idf/issues/5897


Alvin1Zhang:
Thanks for your contribution.

0xjakob:
@b1ackviking I will take a look today or tomorrow. If this passes all our internal build tests, I don't see any reason why we shouldn't apply your patch.

jamieparkinson:
Is there any progress on this or work that needs to be done to unblock it?

b1ackviking:
@0xjakob I'm looking forward to getting your feedback.

This particular fix is only necessary to pass the static asserts in the following lines after it. If you can confirm that it safely applies and passes tests, I will submit a pull request replacing all other occurrences of the `##__VA_ARGS__` construct.

0xjakob:
Hi @jamieparkinson @b1ackviking, sorry for the long delay. We have tested and found some issue with some of our applications. Need to check whether we can work around them.

0xjakob:
Maybe that issue is unrelated. We'll keep you updated!

b1ackviking:
@0xjakob Is there any progress with testing? As I mentioned, if everything works, I would like to proceed to replace other occurrences of `##__VA_ARGS__` with `__VA_OPT__(,) __VA_ARGS__`.

projectgus:
As I understand, the sticking point is that we compile some code with Clang for static analysis (clang-tidy), and it doesn't accept `__VA_OPT__` when compiling C.

@0xjakob @b1ackviking Would it be realistic to structure this change as:

#ifdef __VA_OPT__
#define portGET_ARGUMENT_COUNT(...) portGET_ARGUMENT_COUNT_INNER(0 __VA_OPT__(,) __VA_ARGS__,1,0)
#else
#define portGET_ARGUMENT_COUNT(...) portGET_ARGUMENT_COUNT_INNER(0, ##__VA_ARGS__,1,0)
#endif

... I don't know how many other places you are thinking of applying this change, though...?

b1ackviking:
@projectgus 
> Clang for static analysis (clang-tidy), and it doesn't accept `__VA_OPT__` when compiling C.

Yes, you are right. I have checked that Clang issues an error for both C and C++. It is now fixed in the latest version of Clang (12.0.0).

> ... I don't know how many other places you are thinking of applying this change, though...?

There are 150+ places where `##__VA_ARGS__` is used. Ideally, it would be great to change all of them.

This particular fix is required to pass static asserts with `-std=c++XX` instead of `-std=gnu++XX` here (because `##__VA_ARGS__` is a GNU extension):
https://github.com/espressif/esp-idf/blob/639e7ad494d93fc82159f9fe854041bc43a96d5f/components/freertos/port/xtensa/include/freertos/portmacro.h#L311-L319

igrr:
@b1ackviking Am I right that you only require -std=c++XX rather than -std=gnu++XX applied to the source files of your components, not to ESP-IDF source files? If that is the case, we need to make this change only for the cases where `__VA_ARGS__` is used in public header files of ESP-IDF components. I found the following offenders:
* esp_log.h
* portmacro.h
* esp_check.h

So the `ifdef __VA_OPT__` approach proposed by @projectgus looks viable, i think.

b1ackviking:
@igrr
> Am I right that you only require -std=c++XX rather than -std=gnu++XX applied to the source files of your components, not to ESP-IDF source files?

Yes, that is what I need.

Using `#ifdef __VA_OPT__` produces the following `warning: __VA_OPT__ can only appear in the expansion of a variadic macro [-Wvariadic-macros]`.

I don't think it is a good idea to use that workaround in the long term. It also breaks the build with the `-Werror` flag, which should be enabled by default.

I think the best way would be to suspend this PR for now, so later we could remove `##__VA_ARGS__` completely.

igrr:
> Using #ifdef `__VA_OPT__` produces the following warning: `__VA_OPT__` can only appear in the expansion of a variadic macro [-Wvariadic-macros].

Right, okay, what about something like (pseudocode):
  ifdef (compiling with GCC) || (compiling with LLVM && version >= 12.0)

projectgus:
One other possibility, we could ignore the (otherwise convenient) fact `__VA_OPT__` is defined by GNU C in gcc[*] and detect the C++ version:

```
#if defined(__cplusplus) && __cplusplus >  201703L
..
```
([__cplusplus values are documented here](https://gcc.gnu.org/onlinedocs/gcc-8.4.0/cpp/Standard-Predefined-Macros.html#Standard-Predefined-Macros), in our current ESP32 toolchain I see 201709L.)

[*] Seemingly the GNU C variant in Clang doesn't define this.

b1ackviking:
@igrr @projectgus
I have tried the options that you proposed.

First:
>ifdef (compiling with GCC) || (compiling with LLVM && version >= 12.0)

is equivalent to:
```c
#if defined(__GNUC__) || defined(__GNUG__) || (__clang_major__ >= 12)
```

This approach works.

I have stepped a little bit further and enabled `-pedantic` warnings. As before, Clang works starting from version 12, but GCC says `warning: __VA_OPT__ is not available until C++20`. Although it works, the warning reminds us that the `__VA_OPT__` is a C++20 feature, and we should not use it with earlier standards or in C code.

Following [this](https://stackoverflow.com/a/48045656) answer on StackOverflow, I have found that the warning above is considered a bug. But I feel uncomfortable using `__VA_OPT__` in headers that are included in C source files.

So I think the right way would be to check for the C++ standard version as @projectgus proposed. Also, here is another [link](https://en.cppreference.com/w/cpp/preprocessor/replace#Predefined_macros) to look for __cplusplus values.

If the C++ standard is greater than 17:
```cpp
#if defined(__cplusplus) && (__cplusplus > 201703L)
```
Or if the C++ standard is greater or equal to 20:
```cpp
#if defined(__cplusplus) && (__cplusplus >= 202002L)
```

I would rather use the second option because the first one does not guarantee that `__VA_OPT__` is available, but the current GCC 8.4 toolchain does not support C++20. The support for C++20 is expected in future toolchain releases https://github.com/espressif/crosstool-NG/issues/11#issuecomment-803864897. For now, I do not see how to make `__VA_OPT__` work in a robust standard way.

PR #6718, opened 2021-03-15
Modify method to get base directory. (IDFGH-4926)

blakjak44:
Current method assumes that you are in the esp-idf directory when executing install. This method will get the correct path regardless of where you execute from.

### DIFF
--- a/install.fish
+++ b/install.fish
@@ -1,6 +1,6 @@
 #!/usr/bin/env fish
 
-set basedir $PWD
+set basedir (realpath (dirname (status -f)))
 
 set -x IDF_PATH $basedir
 


Alvin1Zhang:
Thanks for your contribution.

igrr:
Hi @blakjak44, will #6416 work for you to solve the issue? That PR is already in the merge queue, if it's okay for you we will merge it instead.

blakjak44:
Hi @igrr, personally this method is better, as #6416 adds a period into the generated path:

`set basedir "$PWD/"(dirname (status -f))` = `/Users/<user>/./espressif/esp-idf`
`set basedir (dirname (realpath (status -f)))` = `/Users/<user>/espressif/esp-idf`

igrr:
Thank you for the PR @blakjak44, it was cherry-picked as https://github.com/espressif/esp-idf/commit/f5f0648608e2a8dfb96dc1b082d0c6bb87318696!

PR #6727, opened 2021-03-16
reset DCE pointer in DTE if we failed to init modem (IDFGH-4936)

fromeijn:
so we can e.g. retry later


Alvin1Zhang:
Thanks for your contribution.

david-cermak (approved):
Thanks for this useful change! Indeed we shouldn't keep this dangling pointer if dce's initialization fails.

This PR has been moved to the internal pipeline for reviews & tests and will be merged.

igrr:
@fromeijn Thanks again for the PR, the fix has been cherry-picked as 2610e9c!

PR #6768, opened 2021-03-23
Fix ignored return value. (IDFGH-4976)

bugadani:
sdio_ringbuf_send ignores the return value of the callback.


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:

igrr:
The fix has been cherry-picked as c8b697a, thanks again for the PR @bugadani!

PR #6769, opened 2021-03-23
Possible memory leak in wpa_supplicant. (IDFGH-4977)

bugadani:


Alvin1Zhang:
Thanks for your contribution.

components/wpa_supplicant/src/common/sae.c
```diff
@@ -1183,10 +1183,10 @@ static int sae_parse_password_identifier(struct sae_data *sae,
 			wpa_printf(MSG_DEBUG,
 				   "SAE: No Password Identifier included, but expected one (%s)",
 				   sae->tmp->pw_id);
+		    os_free(sae->tmp->pw_id);
```
sagb2015, line 1186: @bugadani  Good catch. This actually can be contribution to upstream wpa_supplicant since this code came from there. We will definitely include your contribution here until then.

bugadani: You might be right. I haven't considered the original intent of this code, just the fact that a success branch called free on null, and an error branch did not, which looked incorrect. Your analysis is much more comprehensive. Still, I've sent the same patch to the original author and maybe he will clean this up.

PR #6770, opened 2021-03-23
Fix to documentation misstating the units. (IDFGH-4978)

jsinglet:
The documentation for `uxTaskGetStackHighWaterMark` appears to use the FreeRTOS convention of words instead of bytes; in IDF the return type here appears to actually be in bytes like `xTaskCreate.`

https://www.esp32.com/viewtopic.php?f=13&t=20043

### DIFF
--- a/components/freertos/include/freertos/task.h
+++ b/components/freertos/include/freertos/task.h
@@ -1448,8 +1448,8 @@ TaskHandle_t xTaskGetHandle( const char *pcNameToQuery ) PRIVILEGED_FUNCTION; /*
  * this function to be available.
  *
  * Returns the high water mark of the stack associated with xTask.  That is,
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

projectgus:
Hi @jsinglet,

Thanks for submitting this PR. Around the same you did this, I raised the same fix internally - sorry!

It's merged but is not on GitHub yet, will update the PR when it is.

Angus

jsinglet:
thanks @projectgus!

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/11e2620accbaac4d07c3ce6f422c48e45d1be80a#diff-ba4acba365081f48ebfffffc0f57f0b4f4e0c65e81995012184441cc1873b2e4

PR #6808, opened 2021-04-01
Interrupt allocation dedicated interrupt support (IDFGH-5022)

cnlohr:
In some situations users may want to place code in the interrupt vector and attach an interrupt generating mechanism.  This allows for very low latency, low overhead interrupts, and it allows users to use the existing code and systems for peripherals, just when specifying an interrupt flag, they can pick a dedicated interrupt which they can write code in the interrupt vector for.

I have not contributed in a very long time, so please let me know if I have done something wrong.

### DIFF
--- a/components/esp_rom/CMakeLists.txt
+++ b/components/esp_rom/CMakeLists.txt
@@ -21,10 +21,10 @@ else()
                         "patches/esp_rom_efuse.c")
 
 
-# Override regi2c implementation in ROM
-if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/patches/esp_rom_regi2c_${target}.c")
-    list(APPEND sources "patches/esp_rom_regi2c_${target}.c")
-endif()
+    # Override regi2c implementation in ROM
+    if(CONFIG_ESP_ROM_HAS_REGI2C_BUG)
+        list(APPEND sources "patches/esp_rom_regi2c_${target}.c")
+    endif()
 
     if(CONFIG_HEAP_TLSF_USE_ROM_IMPL AND CONFIG_ESP_ROM_TLSF_CHECK_PATCH)
         # This file shall be included in the build if TLSF in ROM is activated
--- a/components/esp_rom/esp32c6/Kconfig.soc_caps.in
+++ b/components/esp_rom/esp32c6/Kconfig.soc_caps.in
@@ -54,3 +54,7 @@ config ESP_ROM_HAS_LAYOUT_TABLE
 config ESP_ROM_HAS_SPI_FLASH
     bool
     default y
+
+config ESP_ROM_HAS_REGI2C_BUG
+    bool
+    default y
--- a/components/esp_rom/esp32c6/esp_rom_caps.h
+++ b/components/esp_rom/esp32c6/esp_rom_caps.h
@@ -19,3 +19,4 @@
 #define ESP_ROM_HAS_HEAP_TLSF               (1) // ROM has the implementation of the tlsf and multi-heap library
 #define ESP_ROM_HAS_LAYOUT_TABLE            (1) // ROM has the layout table
 #define ESP_ROM_HAS_SPI_FLASH               (1) // ROM has the implementation of SPI Flash driver
+#define ESP_ROM_HAS_REGI2C_BUG              (1) // ROM has the regi2c bug
--- a/components/esp_rom/esp32s2/Kconfig.soc_caps.in
+++ b/components/esp_rom/esp32s2/Kconfig.soc_caps.in
@@ -14,3 +14,7 @@ config ESP_ROM_HAS_MZ_CRC32
 config ESP_ROM_NEEDS_SWSETUP_WORKAROUND
     bool
     default y
+
+config ESP_ROM_HAS_REGI2C_BUG
+    bool
+    default y
--- a/components/esp_rom/esp32s2/esp_rom_caps.h
+++ b/components/esp_rom/esp32s2/esp_rom_caps.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
-#define ESP_ROM_HAS_CRC_LE            (1) // ROM CRC library supports Little Endian
-#define ESP_ROM_HAS_MZ_CRC32          (1) // ROM has mz_crc32 function
+#define ESP_ROM_HAS_CRC_LE                  (1) // ROM CRC library supports Little Endian
+#define ESP_ROM_HAS_MZ_CRC32                (1) // ROM has mz_crc32 function
 #define ESP_ROM_NEEDS_SWSETUP_WORKAROUND    (1) // ROM uses 32-bit time_t. A workaround is required to prevent printf functions from crashing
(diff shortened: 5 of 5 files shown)


Alvin1Zhang:
Thanks for your contribution.

cnlohr:
I just realized here is a 1-line change that makes it possible to use the IDF without any modifications, should I update this PR or make a new one?  I am unfamiliar with this process.

As it stands, I thought local IDF changes were needed but that's not true.

cnlohr:
Excellent!  I was able to push to my local branch and it updated the PR.  Ok, this should be everything needed to do custom interrupt work without any modifications to the IDF.

cnlohr:
Sorry, I updated without rebase.  But at least now the history should be clean and only have that one commit.

projectgus:
Hi @cnlohr,

This is a nicely thought out way to support this!

For RISC-V, we actually have some ideas to support this in a different way via link-time allocation. Specifically, adding placeholder symbol entries `_interrupt_vector0`, `_interrupt_vector1`, etc that are all weak-aliased to `_interrupt_handler` by default, and then rewriting [the vector table](https://github.com/espressif/esp-idf/blob/master/components/riscv/vectors.S#L124) to be `j _interrupt_vector0`, `j _interrupt_vector1`, etc.

This way anyone can write an ISR in their project with the name `_interrupt_vectorN` and `__attribute__(("interrupt"))` to generate the necessary prelude/prologue, and this will replace the matching entry in the vector table at link-time.

This also has the advantage of being a little lower latency than calling via the ESP-IDF ISR `_interrupt_handler` wrapper, if the routine is lightweight.

We haven't actually implemented this yet, it's just an idea at this stage. Would it work for your use case as well?

cnlohr:
Somehow I missed the alert on this.  I think that's good for most people, but because the hardware interrupts aren't already tied to specific interrupts it gets a little weird.  Either way is OK... But... I think this still is a decent way of going regardless, to have a flag to say "I want to pin this interrupt to this interrupt number." Regardless of the outcome.

I think most of the time I would want to use the interrupt override like you point out instead of a custom table.  All of the changes in this PR still apply even for that situation.  The only caveat is you would be able to get rid of the code check I removed.  But, It seems very strange to assert because users chose to put executable code in the interrupt vector.  It's rare someone would want to do this but why stop them?

luzidchris:
Hi,

I know, this issue had been closed quite a while ago, but I'd like to come back to this.

At some point, there's been the following statement, presumably posted by a member of the IDF maintenance/contributor team:

> For RISC-V, we actually have some ideas to support this in a different way via link-time allocation. Specifically, adding placeholder symbol entries `_interrupt_vector0`, `_interrupt_vector1`, etc that are all weak-aliased to `_interrupt_handler` by default, and then rewriting [the vector table](https://github.com/espressif/esp-idf/blob/master/components/riscv/vectors.S#L124) to be `j _interrupt_vector0`, `j _interrupt_vector1`, etc.
> 
> This way anyone can write an ISR in their project with the name `_interrupt_vectorN` and `__attribute__(("interrupt"))` and this will replace the matching entry in the vector table at link-time.
> 
> We haven't actually implemented this yet, it's just an idea at this stage. Would it work for your use case as well?

I might be wrong, but I've not found any changes in this regard in the master and latest release branches. That's why I'd like to ask if this idea has been dropped or still in the making?

I'm facing a pretty similar problem like @cnlohr whose projects and a link to this issue I found during my research. In my project, I need to react on a input signal easy quickly as possible (ESP32-C3). I've already modified the corresponding code to register the EXTI/GPIO input ISR "directly" (via `gpio_isr_register`), so I'm not using the ISR service. Still, I'm observing about 1 us latency, between the rising/falling edge of the input signal and the rising/edge of a GPIO output I'm setting immediately after entering the ISR (written in assembly). I'll now proceed a step further, and write my ISR into the vector table to get rid of any additional overhead.

Although I don't know whether I'll achieve an acceptable latency or not, I still think it might be useful to have "weak" _interrupt_vector routines, to replace them for specific vectors, instead of replacing the full vector table. The "weak"-approach seems to be more robust and easier to maintain, considering to future modifications in the vector table.

ESP-Marius:
> I might be wrong, but I've not found any changes in this regard in the master and latest release branches. That's why I'd like to ask if this idea has been dropped or still in the making?

Hi, you are correct that this hasn't been implemented yet. We do agree that this would be useful though, and we do have a task for this in our backlog.

Recently we've been putting some effort into cleaning up our interrupt allocation code (e.g. a79c6f7f), so hopefully the feature requested here is also something we can take a look at soon. But I cannot make any promises on when this will be available.

luzidchris:
Thanks for the quick response, especially in this case as this issue had been closed a while ago, and for pointing me to the most recent changes related to interrupt allocation. I've not yet noticed them and they look very interesting!

I've made quite some progress and could successfully realize and test an GPIO ISR which is called by a custom trap vector which part of a custom vector table. The custom vector table I've used for testing relies on a couple handlers with weak implementation. I override one of them with protocol-specific implementation, so basically tried to follow the idea described by this thread, just to help me get a better picture of the underlying mechanisms of interrupt handling of ESP32C3 and related functionality.
Besides of routing the peripheral interrupt source to the desired vector, setting its priority and type (basically following what's done by `intr_alloc.c` of component `esp_hw_support` ), I used `esp_intr_reserve` to block other drivers from allocating the trap blocked my non-IDF-managed ISR to avoid issues. At no point had there been any need to modify or extend any IDF implementation. I could strictly rely on the existing interfaces provided by IDF which is fantastic. 

Having made this experience, I'm not sure whether my argument regarding effort and maintenance still makes a lot of sense. Even with weak handlers, there'd be some custom low-level implementation which could come with risks and be difficult to maintain. When it comes to latency though, I can already say that it was definitely worth spending the effort. Before, it took about 1 us after the GPIO signal state transition to enter the ISR. Now, the first ISR instruction is executed after about 400 ns, thus fulfilling my project's timing requirements.

Keep up the great work, and thank you very much for the support!

Regards,
Chris

PS: I initially faced some issues because the strong wouldn't replace the weak implementation, until I found [this thread in the ESP32 forum](https://esp32.com/viewtopic.php?t=9270), which pointed to [a work-around described in the official documentation of the build system](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c3/api-guides/build-system.html#advanced-workaround-undefined-symbols). It might sound exaggerated but this has once again shown to me how great your support and the quality of the official documentation actually is. By replying to literally every single community request and competently providing useful and detailed information, and by obviously spending a lot of effort to deliver great documentation and test/example projects you really have become a role model.

ESP-Marius:
> Having made this experience, I'm not sure whether my argument regarding effort and maintenance still makes a lot of sense. Even with weak handlers, there'd be some custom low-level implementation which could come with risks and be difficult to maintain.

I think we still wanna give it a try. If we do add some basic support + an example for this then at least it will be easier for any future user that want to achieve the same as what you did, without having to do all the research themselves :smile: 

> By replying to literally every single community request and competently providing useful and detailed information, and by obviously spending a lot of effort to deliver great documentation and test/example projects you really have become a role model.

Appreciate the feedback! We do try our best to help all of our users with the time we have available.

PR #6841, opened 2021-04-07
fix #6840: compare auth scheme in case insensitive fashion (IDFGH-5056)

### LINKED ISSUE #6840: esp_http_client basic auth not comparing scheme in case-insensitive fashion (IDFGH-5055)
## Environment

- Development Kit:      none
- Module or chip used:  does not apply
- IDF version (run ``git describe --tags`` to find it): v4.4
- Build System:         idf.py
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): not relevant
- Operating System:     not relevant
- (Windows only) environment type: not relevant
- Using an IDE?: not relevant
- Power Supply:         not relevant

## Problem Description

HTTP basic auth does not work with all servers with esp_http_client. This is due to the authentication-scheme in the `WWW-Authenticate`-header being compared in a case-sensitive fashion against the string `Basic` in `esp_http_client_add_auth()`. Most servers use the capitalization of `Basic` given in the RFC example. However, some don't and use e.g. `BASIC` or `basic` instead. And they are rightly doing som because [RFC 2617 Secion 1.2](https://tools.ietf.org/html/rfc2617#section-2) clearly states "It uses an extensible, /case-insensitive/ token to identify the authentication scheme,..." (emphasis mine).

### Expected Behavior

Basic auth working with servers using a capitalization other than `Basic`.

### Actual Behavior

Basic auth only works with servers using `Basic`.

### Steps to reproduce

1. find a server using a `WWW-Authenticate`-header with a scheme capitalization other than `Basic`.
2. try http with basic auth
3. see it fail

Any sample http client code with authentication will do.

## Debug Logs

```
[  9512][D][test.cpp] debugHTTPEventHandler(): HTTP_EVENT_ON_HEADER, key=WWW-Authenticate, value=BASIC realm="my realm"
[ 16670][E][HTTP_CLIENT]: This authentication method is not supported: BASIC realm="my realm"
```

hennejg:
Fixes #6840: use case-insensitive comparison for the auth scheme.


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97:
Hi @hennejg,

Thanks for the PR. Changes look good to me. However, there is no need to create new API for insensitive comparison. It can be fixed in existing API. Also, according to your comment in https://github.com/espressif/esp-idf/issues/6840 I have updated http_utils_get_string_between to use `strcasestr`.

Please find patch attached below and check if it looks fine for your use case.
[0001-esp_http_client-Make-http_utils_get_string_between-a.patch.zip](https://github.com/espressif/esp-idf/files/6279309/0001-esp_http_client-Make-http_utils_get_string_between-a.patch.zip)

Thanks,
Shubham

hennejg:
@shubhamkulkarni97 thanks a lot! You are right, of course - I had assumed that `http_utils_str_starts_with` was used in other places where case-sensitivity was desireable, but as this is not the case, it is fine to just fix `http_utils_str_starts_with` directly. The patch looks fine.

PR #6846, opened 2021-04-08
Delete i2c cmd_mux semaphore more cleanly (v4.3 branch) (IDFGH-5061)

makermelissa:
See https://github.com/adafruit/esp-idf/pull/1 for more info.

### DIFF
--- a/components/driver/i2c.c
+++ b/components/driver/i2c.c
@@ -399,7 +399,9 @@ esp_err_t i2c_driver_delete(i2c_port_t i2c_num)
     p_i2c->intr_handle = NULL;
 
     if (p_i2c->cmd_mux) {
+	// Let any command in progress finish.
         xSemaphoreTake(p_i2c->cmd_mux, portMAX_DELAY);
+	xSemaphoreGive(p_i2c->cmd_mux);
         vSemaphoreDelete(p_i2c->cmd_mux);
     }
     if (p_i2c_obj[i2c_num]->cmd_evt_queue) {
--- a/components/driver/i2c.c
+++ b/components/driver/i2c.c
@@ -399,9 +399,9 @@ esp_err_t i2c_driver_delete(i2c_port_t i2c_num)
     p_i2c->intr_handle = NULL;
 
     if (p_i2c->cmd_mux) {
-	// Let any command in progress finish.
+        // Let any command in progress finish.
         xSemaphoreTake(p_i2c->cmd_mux, portMAX_DELAY);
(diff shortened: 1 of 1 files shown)


dhalbert:
This change fixes an problem we have seen in CircuitPython between `i2c_driver_delete()` and `esp_wifi_start()`. It is not a hard crash, but in our case we lose the USB CDC connection, and things stop running: it may be a runaway set of interrupts or some other loop. The usual IDF threads that are running when the issue happens look the same as in the good case.

The essence of the problem is this sequence:
```c
i2c_driver_delete(NULL);   // does nothing
i2c_param_config();
i2c_driver_install();
i2c_driver_delete();   // This is crucial.

esp_netif_init();               
esp_event_loop_create_default();
xEventGroupCreateStatic();
esp_event_handler_instance_register();
esp_event_handler_instance_register();
esp_wifi_init();
esp_wifi_set_mode();
esp_wifi_start();
```
I arrived at this change by divide and conquer, seeing which part of the `i2c.c` was causing the issue.Unfortunately `esp_wifi_start()`, which is where the problem seems to occur, is in the closed-source wifi stack.

 The critical piece of the original `i2c_driver_delete()` code is:
```c
xTakeSemaphore();
vSemaphoreDelete;
```
If the `xTakeSemaphore();` is completely removed, or if the `xGiveSemaphore()` is added, then we don't see the problem. I arrived at adding the `xGiveSemaphore()` by examining other semaphore cleanup code in the IDF.

I have tried to create a minimal example ESP-IDF. I have one I, but the only differences I see immediately is that the `DEBUG` traces between the "good" and the "bad" cases. It would be great if someone with access could look inside `esp_wifi_start()` to see what the actual issue might be.

Other drivers (e.g., SPI, etc.) do not affect `esp_wifi_start()` in this way. As mentioned, there is more information in https://github.com/adafruit/esp-idf/pull/1 and a longer discussion in https://github.com/adafruit/circuitpython/pull/4387.

Alvin1Zhang:
Thanks for your contribution.

dhalbert:
See https://github.com/espressif/esp-idf/pull/6846#issuecomment-816280296 for more information about the motivation for this PR.

Alvin1Zhang:

L-KAYA:
Thanks for your contribution again! This PR has been merged into release/v4.3 (commit:`2ac59cc8`). Could you help to check whether the issue has been fixed.

Alvin1Zhang:
Thanks for your contribution.

PR #6853, opened 2021-04-10
fix gpio interrupts lost (IDFGH-5069)

toybox01:
fix race condition


Alvin1Zhang:
Thanks for your contribution.

toybox01:
Is this PR under consideration?

ginkgm:
Hi @toybox01 

I think you are right about the interrupt may be accidently cleared without being properly handled. It's really a bug.

But clearing the interrupt before the handler is called is also a breaking change, we are still thinking about how to deal with it properly...

tommy1394:
Hi @toybox01,
I am new to ESP32, and have just discovered this issue in my application.  Is there a way to apply this patch to the Arduino IDE or do I need to move my code to the ESP-IDF?

PR #6869, opened 2021-04-14
Typo pthread doc: create->get (IDFGH-5085)

a2800276:
I believe that:

    esp_create_default_pthread_config

is a typo (no such function exists), the intention was almost certainly to use:

     esp_pthread_get_default_config

### DIFF
--- a/docs/en/api-reference/system/esp_pthread.rst
+++ b/docs/en/api-reference/system/esp_pthread.rst
@@ -26,7 +26,7 @@ Example to tune the stack size of the pthread:
     {
         pthread_t t1;
 
-        esp_pthread_cfg_t cfg = esp_create_default_pthread_config();
+        esp_pthread_cfg_t cfg = esp_pthread_get_default_config();
         cfg.stack_size = (4 * 1024);
         esp_pthread_set_cfg(&cfg);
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

projectgus:
Hi @a2800276 ,

Thanks for this. You are totally correct.

As it happens, I'm in the process of reworking this documentation page and moving these code snippets into examples (which means they'll be compiled by our CI system and can't break in this way in future.)

I've cherry-picked your commit into this branch so it will be applied at the same time. The PR will be updated when that happens.

Thanks again,

Angus

a2800276:
I've noticed a couple of similiar issues, e.g. in the [uart](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html) the examples use `int uart_num` instead of `uart_port_t`. 

Is it helpful for you if I raise issues for these sorts of things?

krzychb:
> Is it helpful for you if I raise issues for these sorts of things?

Hi @a2800276, yes, it is definitely helpful!
To raise an issue you can also use "Provide feedback about this document" link at the bottom of each documentation page.

PR #6879, opened 2021-04-15
Allow to fully integrate the esptool-ftdi utility in the idf.py build/flash system (IDFGH-5095)

M-Bab:
To use this set the environment variable ESPTOOL_FTDI to the path to the utility.

This utility allows to flash an ESP event simpler/cheaper with a standard FTDI USB-serial cable.
The lack of RTS/DTR is compensated by bitbanging RTS/CTS.

Implements https://github.com/jimparis/esptool-ftdi/issues/3

I am not sure, if you like this hacky approach. But after all, it _is_ very comfortable to use a simple FTDI cable for flashing.

### DIFF
--- a/components/esptool_py/project_include.cmake
+++ b/components/esptool_py/project_include.cmake
@@ -12,7 +12,7 @@ if(target STREQUAL "esp32s3")
     endif()
 endif()
 
-set(ESPTOOLPY ${python} "$ENV{ESPTOOL_FTDI}" "${CMAKE_CURRENT_LIST_DIR}/esptool/esptool.py" --chip ${chip_model})
+set(ESPTOOLPY ${python} "$ENV{ESPTOOL_WRAPPER}" "${CMAKE_CURRENT_LIST_DIR}/esptool/esptool.py" --chip ${chip_model})
 set(ESPSECUREPY ${python} "${CMAKE_CURRENT_LIST_DIR}/esptool/espsecure.py")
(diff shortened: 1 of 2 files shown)


M-Bab:
Or alternatively, we could use a more generic environment variable like ESPTOOL_WRAPPER which is prepended, if it set?

Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Hi @M-Bab. You have a nice project and thank you for your contribution.

Personally, I think this can be accepted but I don't promise anything because it will go through an internal review process.
Could you please correct the variable name to `ESPTOOL_WRAPPER` as you've suggested?

M-Bab:
Renaming is done. Thanks a lot for your positive feedback.

The esptool-ftdi itself is not my development, this is the work of @jimparis .

Some background information on how to use it in practice (if anyone is curious):

Integrate in the eclipse environment:

image

Schematic implementation (works directly without the usual RTS/DTR transistor circuit):

image

Be aware that _regularly_ opening the COM port with an FTDI fully connected will keep the ESP in reset because RTS is handled on open which keeps the ESP disabled. This can be handled by certain serial terminals which are able to toggle RTS like gtkterm.

This schematic with pin header can be combined with the cheap USB FTDI adapter: FTDI TTL-232R-3V3

Or even for footprint only production designs with e.g. tag-connect TC2030-FTDI-TTL-232R-3V3

PR #6880, opened 2021-04-15
Added esp_https_ota_get_image_size() to retrieve total size of OTA update (IDFGH-5096)

0xFEEDC0DE64:
I use `esp_https_ota_get_image_len_read()` to retrieve the succeeded progress of the update and now I can use the new method `esp_https_ota_get_image_size()` get the total size and create some sort of progress indication.

When the OTA update uses chunked encoding I think there is no way of telling the total size so I return -1 instead.


Alvin1Zhang:
Thanks for your contribution.

PR #6882, opened 2021-04-16
esp_https_ota: Fix build warning when CONFIG_OTA_ALLOW_HTTP is set (IDFGH-5098)

AxelLin:
Fix below build warning when CONFIG_OTA_ALLOW_HTTP is set:
esp_https_ota.c:165:13: warning: 'is_server_verification_enabled' defined but not used [-Wunused-function]

While at it, also simplify the is_server_verification_enabled logic.

Fixes: 48987ad9df35 ("esp_https_ota: Enable option of using global_ca_store and x509_crt_bundle")

### DIFF
--- a/components/esp_https_ota/src/esp_https_ota.c
+++ b/components/esp_https_ota/src/esp_https_ota.c
@@ -162,11 +162,13 @@ static esp_err_t _ota_write(esp_https_ota_t *https_ota_handle, const void *buffe
     return err;
 }
 
+#if !CONFIG_OTA_ALLOW_HTTP
 static bool is_server_verification_enabled(esp_https_ota_config_t *ota_config) {
     return  (ota_config->http_config->cert_pem
             || ota_config->http_config->use_global_ca_store
-            || !(ota_config->http_config->crt_bundle_attach == NULL));
+            || ota_config->http_config->crt_bundle_attach != NULL);
 }
+#endif
 
 esp_err_t esp_https_ota_begin(esp_https_ota_config_t *ota_config, esp_https_ota_handle_t *handle)
 {


Alvin1Zhang:
Thanks for your contribution.

mahavirj (approved):

negativekelvin:
Wouldn't it be better to fix https://github.com/espressif/esp-idf/blob/1f54d17503ff17d830de75eed4eb47bc6a3aafc3/components/esp_https_ota/src/esp_https_ota.c#L183-L189
so that it still prints a warning but doesn't return an error?

mahavirj:
@negativekelvin 

We can do that, but if we look at plain HTTP config option and its help:

https://github.com/espressif/esp-idf/blob/1f54d17503ff17d830de75eed4eb47bc6a3aafc3/components/esp_https_ota/Kconfig#L4

It is clearly stated that this is "test" only configuration and there is security risk associated with it. So if someone is enabling it then they are aware about associated risk.

Given this, current PR looks good to me. However open to suggestions as well.

negativekelvin:
I think it's better to have the runtime logs for confirmation and understanding than silently remove the function.  I'm sure it will happen at least once that somebody doesn't know it's enabled.

AxelLin:
IMHO, one topic per patch. The fix does what the subject line says.

negativekelvin:
@AxelLin the other way would also fix build warning by always calling is_server_verification_enabled.  If this is merged just have to undo it later.

AdityaHPatwardhan:
Hi @AxelLin  I have added a followup commit which incorporates changes suggested by @negativekelvin, It is merged in the internal codebase, should be available on GitHub shortly. Thank you for your contribution in fixing this issue.

PR #6895, opened 2021-04-18
Fixed error in "touch_ll_get_sleep_time" function for ESP32 (IDFGH-5111)

FoxKeys:
Because of this bug touch_hal_get_sleep_time() returns wrong value (both, sleep_cycle and meas_cycle filled with meas_cycle). As result, touch_pad_get_meas_time returns wrong sleep_cycle too.
Additional, bad side effect, touch_pad_config() makes wrong vTaskDelay() (because delay based on values returned by touch_pad_get_meas_time())


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for contribution again, changes merged with https://github.com/espressif/esp-idf/commit/850de61e34c3d1b09dcd70dd8514cb98687920a9.

PR #6927, opened 2021-04-23
fix buffer overrun bug (IDFGH-5150)

zakkie:
`output_buffer` isn't necessarily NUL-terminated

### DIFF
--- a/examples/protocols/esp_http_client/main/esp_http_client_example.c
+++ b/examples/protocols/esp_http_client/main/esp_http_client_example.c
@@ -603,7 +603,7 @@ static void http_native_request(void)
                 ESP_LOGI(TAG, "HTTP GET Status = %d, content_length = %d",
                 esp_http_client_get_status_code(client),
                 esp_http_client_get_content_length(client));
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

mahavirj:
@zakkie 

> output_buffer isn't necessarily NUL-terminated

https://github.com/espressif/esp-idf/blob/a6de15a602a2c3f9803093d152206a9c4d67c761/examples/protocols/esp_http_client/main/esp_http_client_example.c#L584

Looks like this PR should not be required, any specific issue your are observing without this change?

zakkie:
https://github.com/espressif/esp-idf/blob/a6de15a602a2c3f9803093d152206a9c4d67c761/examples/protocols/esp_http_client/main/esp_http_client_example.c#L601

`esp_http_client_read_response()` reads up to MAX_HTTP_OUTPUT_BUFFER.
So, if the HTTP server returns a large response, `output_buffer[MAX_HTTP_OUTPUT_BUFFER - 1] != '\0'`

mahavirj:
@zakkie 

Aha, right!

mahavirj (approved):

PR #6949, opened 2021-04-27
Fixed esp_core_dump_image_erase() for flash encryption with 16byte long write buffers (IDFGH-5173)

0xFEEDC0DE64:
esp_partition_write() kept failing in esp_core_dump_image_erase() when I enabled flash encryption, as the write function then requires multiple of 16 byte long write buffers.

I changed the erase method so it always writes 16 bytes and the blank coredump size at the beginning.

Also I added a compile time static assert to make sure the struct has a valid size.

### DIFF
--- a/components/espcoredump/src/core_dump_flash.c
+++ b/components/espcoredump/src/core_dump_flash.c
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #include <string.h>
+#include <assert.h>
+#include "esp_core_dump.h"
 #include "esp_partition.h"
 #include "esp_log.h"
 #include "esp_core_dump_types.h"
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
@0xFEEDC0DE64 Thanks for your contribution.

0xFEEDC0DE64:
Is this fix already merged? or is there anything I can do to help?

o-marshmallow:
Hello @0xFEEDC0DE64,

Sorry for the delay of response, thank you for your contribution once again :smile: !
I am going to handle this

components/espcoredump/src/core_dump_flash.c
```diff
@@ -484,9 +486,21 @@ esp_err_t esp_core_dump_image_erase(void)
         return err;
     }
 
+    // helper to create (multiple of) 16 byte long write buffers
```
o-marshmallow, line 489: There is a simpler way to do that I would prefer, it's to add at the beginning of this function the following snippet:
```
/* If flash is encrypted, we can only write blocks of 16 bytes. */
uint32_t helper[4] = { BLANK_COREDUMP_SIZE };
```
You can add the static assert right after (same as now, `sizeof(helper)` will return 16).

components/espcoredump/src/core_dump_flash.c
```diff
@@ -484,9 +486,21 @@ esp_err_t esp_core_dump_image_erase(void)
...
+
+    _Static_assert(sizeof(helper) % 16 == 0, "esp_partition_write() needs multiple of 16 byte long buffers");
+
     // Mark core dump as deleted by setting field size
-    const uint32_t blank_size = BLANK_COREDUMP_SIZE;
-    err = esp_partition_write(core_part, 0, &blank_size, sizeof(blank_size));
+    helper.size = BLANK_COREDUMP_SIZE;
+
+    // fill the remaining bytes
+    memset(&helper.buf, '\0', sizeof(helper.buf));
```
o-marshmallow, line 501: If you follow my comment above, no need to `memset` the memory

components/espcoredump/src/core_dump_flash.c
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #include <string.h>
+#include <assert.h>
+#include "esp_core_dump.h"
```
o-marshmallow, line 16: Was these includes strictly required?  `_Static_assert` can be used without them

o-marshmallow (changes requested):
I just added few comments to simplify what you did, please check them

PR #6951, opened 2021-04-27
Patched sprintf usage (IDFGH-5175)

PBearson:


AxelLin:
Why not using strcpy instead?

PBearson:
> Why not using strcpy instead?

Good point, I guess strcpy is better in this case. I updated to strncpy and used the defined WPS macros for the bounds lengths. For example:

` strncpy(dev->manufacturer, s_factory_info->manufacturer, WPS_MAX_MANUFACTURER_LEN);`

Alvin1Zhang:
@PBearson Thanks for your contribution.

kapilkedawat:
Hi @PBearson , we use snprintf now which checks for bound limitation, therefore this change is not needed anymore.

Alvin1Zhang:
Thanks for contribution again.

PR #6974, opened 2021-05-03
tcp_transport: Fix NULL pointer dereference if esp_transport_init ret… (IDFGH-5201)

AxelLin:
…urns NULL

Add missing NULL test for esp_transport_init() call.
Otherwise, it will hit NULL pointer dereference when assign t->_get_socket.


Alvin1Zhang:
@AxelLin Thanks for your contribution.

david-cermak (approved):
Thanks for the fix!

PR #7032, opened 2021-05-17
More SPI Master frequencies

patrykkrz:
Sometimes slower SPI frequencies are needed.


Alvin1Zhang:
Thanks for your contribution.

PR #7048, opened 2021-05-19
Implement session ticket support in case mbedtls is used as ssl library (IDFGH-5288)

dbahrdt:
This PR implements session ticket support in case mbedtls is selected as the ssl library and sever side session ticket support is enabled.

We're still using 4.2 and consider upgrading to 4.3. Is there any chance this change could get into 4.3?

### DIFF
--- a/components/esp-tls/Kconfig
+++ b/components/esp-tls/Kconfig
@@ -40,9 +40,16 @@ menu "ESP-TLS"
             Enable support for creating server side SSL/TLS session, available for mbedTLS
             as well as wolfSSL TLS library.
 
+    config ESP_TLS_SERVER_SESSION_TICKETS
+        bool "Enable session tickets"
+        depends on ESP_TLS_SERVER && ESP_TLS_USING_MBEDTLS && MBEDTLS_SERVER_SSL_SESSION_TICKETS
+        default n
+        help
+            Enable session ticket support as specified in RFC5077
+
     config ESP_TLS_SERVER_SESSION_TICKET_TIMEOUT
         int "Server session ticket timeout in seconds"
-        depends on ESP_TLS_SERVER && ESP_TLS_USING_MBEDTLS && MBEDTLS_SERVER_SSL_SESSION_TICKETS
+        depends on ESP_TLS_SERVER_SESSION_TICKETS
         default 86400
         help
             Sets the session ticket timeout used in the tls server.
--- a/components/esp-tls/Kconfig
+++ b/components/esp-tls/Kconfig
@@ -40,6 +40,13 @@ menu "ESP-TLS"
             Enable support for creating server side SSL/TLS session, available for mbedTLS
             as well as wolfSSL TLS library.
 
+    config ESP_TLS_SERVER_SESSION_TICKET_TIMEOUT
+        int "Server session ticket timeout in seconds"
+        depends on ESP_TLS_SERVER && ESP_TLS_USING_MBEDTLS && MBEDTLS_SERVER_SSL_SESSION_TICKETS
+        default 86400
+        help
+            Sets the session ticket timeout used in the tls server.
+
     config ESP_TLS_PSK_VERIFICATION
         bool "Enable PSK verification"
         select MBEDTLS_PSK_MODES if ESP_TLS_USING_MBEDTLS
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -570,50 +570,27 @@ mbedtls_x509_crt *esp_tls_get_global_ca_store(void)
 #endif /* CONFIG_ESP_TLS_USING_MBEDTLS */
 #ifdef CONFIG_ESP_TLS_SERVER
 
+int esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t * cfg) {
 #if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
-
-int esp_tls_session_ticket_ctx_init(esp_tls_session_ticket_ctx_t * ctx) {
-
-    mbedtls_ctr_drbg_init(&ctx->ctr_drbg);
-    mbedtls_entropy_init(&ctx->entropy);
-    mbedtls_ssl_ticket_init(&ctx->ticket_ctx);
-    int ret;
-    if ((ret = mbedtls_ctr_drbg_seed(&ctx->ctr_drbg,
-                                     mbedtls_entropy_func, &ctx->entropy, NULL, 0)) != 0) {
-        ESP_LOGE(TAG, "mbedtls_ctr_drbg_seed returned -0x%x", -ret);
-        return ESP_ERR_MBEDTLS_CTR_DRBG_SEED_FAILED;
+    if (cfg->ticket_ctx) {
+        return ESP_ERR_INVALID_ARG;
     }
-
-    if( ( ret = mbedtls_ssl_ticket_setup( &ctx->ticket_ctx,
-                    mbedtls_ctr_drbg_random, &ctx->ctr_drbg,
-                    MBEDTLS_CIPHER_AES_256_GCM,
-                    CONFIG_ESP_TLS_SERVER_SESSION_TICKET_TIMEOUT ) ) != 0 )
-        {
-            ESP_LOGE(TAG, "Failed mbedtls_ssl_ticket_setup with error code %d", ret);
-            return ESP_ERR_MBEDTLS_SSL_SESSION_TICKET_SETUP_FAILED;
-        }
+    cfg->ticket_ctx = calloc(1, sizeof(esp_tls_session_ticket_ctx_t));
+    if (!cfg->ticket_ctx) {
+        return ESP_ERR_NO_MEM;
+    }
+    if (esp_tls_session_ticket_ctx_init(cfg->ticket_ctx) != ESP_OK) {
+        return ESP_FAIL;
+    }
     return ESP_OK;
-}
-
-void esp_tls_session_ticket_ctx_free(esp_tls_session_ticket_ctx_t * ctx) {
-    mbedtls_ssl_ticket_free(&ctx->ticket_ctx);
-    mbedtls_ctr_drbg_init(&ctx->ctr_drbg);
-    mbedtls_entropy_free(&ctx->entropy);
-}
-
-#endif
-
-int esp_tls_cfg_server_init(esp_tls_cfg_server_t * cfg) {
-#if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
-    return esp_tls_session_ticket_ctx_init(&cfg->ticket_ctx);
 #else
-    return ESP_OK;
+    return ESP_ERR_NOT_SUPPORTED;
 #endif
 }
 
-void esp_tls_cfg_server_free(esp_tls_cfg_server_t * cfg) {
+void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t * cfg) {
 #if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
-    esp_tls_session_ticket_ctx_free(&cfg->ticket_ctx);
+    esp_tls_session_ticket_ctx_free(cfg->ticket_ctx);
 #endif
 }
 
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -575,17 +575,14 @@ mbedtls_x509_crt *esp_tls_get_global_ca_store(void)
 int esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
 {
 #if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
-    if (cfg->ticket_ctx) {
+    if (!cfg || cfg->ticket_ctx) {
         return ESP_ERR_INVALID_ARG;
     }
     cfg->ticket_ctx = calloc(1, sizeof(esp_tls_session_ticket_ctx_t));
     if (!cfg->ticket_ctx) {
         return ESP_ERR_NO_MEM;
     }
-    if (_esp_tls_session_ticket_ctx_init(cfg->ticket_ctx) != ESP_OK) {
-        return ESP_FAIL;
-    }
-    return ESP_OK;
+    return _esp_tls_session_ticket_ctx_init(cfg->ticket_ctx);
 #else
     return ESP_ERR_NOT_SUPPORTED;
 #endif
@@ -594,7 +591,9 @@ int esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
 void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
 {
 #if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
-    _esp_tls_session_ticket_ctx_free(cfg->ticket_ctx);
+    if (cfg && cfg->ticket_ctx) {
+        _esp_tls_session_ticket_ctx_free(cfg->ticket_ctx);
+    }
 #endif
 }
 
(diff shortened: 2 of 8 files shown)


components/esp-tls/Kconfig
```diff
@@ -40,6 +40,13 @@ menu "ESP-TLS"
             Enable support for creating server side SSL/TLS session, available for mbedTLS
             as well as wolfSSL TLS library.
 
+    config ESP_TLS_SERVER_SESSION_TICKET_TIMEOUT
```
dbahrdt, line 43: Should I add another option to explicitly enable server side session ticket support instead of using CONFIG_MBEDTLS_SERVER_SSL_SESSION_TICKETS?

components/esp-tls/esp_tls_errors.h
```diff
@@ -39,18 +39,19 @@ extern "C" {
...
-#define ESP_ERR_WOLFSSL_SSL_SET_HOSTNAME_FAILED           (ESP_ERR_ESP_TLS_BASE + 0x13)  /*!< wolfSSL api returned error */
-#define ESP_ERR_WOLFSSL_SSL_CONF_ALPN_PROTOCOLS_FAILED    (ESP_ERR_ESP_TLS_BASE + 0x14)  /*!< wolfSSL api returned error */
-#define ESP_ERR_WOLFSSL_CERT_VERIFY_SETUP_FAILED          (ESP_ERR_ESP_TLS_BASE + 0x15)  /*!< wolfSSL api returned error */
-#define ESP_ERR_WOLFSSL_KEY_VERIFY_SETUP_FAILED           (ESP_ERR_ESP_TLS_BASE + 0x16)  /*!< wolfSSL api returned error */
-#define ESP_ERR_WOLFSSL_SSL_HANDSHAKE_FAILED              (ESP_ERR_ESP_TLS_BASE + 0x17)  /*!< wolfSSL api returned failed  */
-#define ESP_ERR_WOLFSSL_CTX_SETUP_FAILED                  (ESP_ERR_ESP_TLS_BASE + 0x18)  /*!< wolfSSL api returned failed */
-#define ESP_ERR_WOLFSSL_SSL_SETUP_FAILED                  (ESP_ERR_ESP_TLS_BASE + 0x19)  /*!< wolfSSL api returned failed */
-#define ESP_ERR_WOLFSSL_SSL_WRITE_FAILED                  (ESP_ERR_ESP_TLS_BASE + 0x1A)  /*!< wolfSSL api returned failed */
+#define ESP_ERR_MBEDTLS_SSL_SESSION_TICKET_SETUP_FAILED   (ESP_ERR_ESP_TLS_BASE + 0x12)  /*!< mbedtls api returned failed  */
```
dbahrdt, line 42: Changing all value in case a new error code is needed seems strange. Is there a reason why an enum is not an option?

Alvin1Zhang:
Thanks for your contribution.

AdityaHPatwardhan:
Hi @dbahrdt, Thank you for the contribution.
I had a few questions/suggestions on this one.

1. I think it would be a good idea to create a new config option for session tickets and have the support depend on that instead on the mbedtls config option.

2. Instead of giving functions out to users to initialise the session context, can we just keep a boolean in the `esp_tls_cfg_server_t`, something like 'use_session_tickets', and then depending on the boolean ( + esp_tls cfg option for session tickets ) initialise the
session_tickets_context inside the `set_server_config` function of `esp_tls_mbedtls.c`. As done for [ALPN](https://github.com/espressif/esp-idf/blob/c13afea635adec735435961270d0894ff46eef85/components/esp-tls/esp_tls_mbedtls.c#L373)
That way we won't have to call any function and only set something like `use_session_tickets = true` in the `esp_tls_cfg_server_t`.

3. I think we can use the tls->ctr_drbg instead of defining a new one.

dbahrdt:
> 1. I think it would be a good idea to create a new config option for session tickets and have the support depend on that instead on the mbedtls config option.

I'll add the config option.

> 2\. Instead of giving functions out to users to initialise the session context, can we just keep a boolean in the `esp_tls_cfg_server_t`, something like 'use_session_tickets', and then depending on the boolean ( + esp_tls cfg option for session tickets ) initialise the
>     session_tickets_context inside the `set_server_config` function of `esp_tls_mbedtls.c`. As done for [ALPN](https://github.com/espressif/esp-idf/blob/c13afea635adec735435961270d0894ff46eef85/components/esp-tls/esp_tls_mbedtls.c#L373)
>     That way we won't have to call any function and only set something like `use_session_tickets = true` in the `esp_tls_cfg_server_t`.

The way I understand it `esp_tls_t` is used per session (https://github.com/espressif/esp-idf/blob/c13afea635adec735435961270d0894ff46eef85/components/esp_https_server/src/https_server.c#L104).
The rng is also initialized for each session (https://github.com/espressif/esp-idf/blob/c13afea635adec735435961270d0894ff46eef85/components/esp-tls/esp_tls_mbedtls.c#L77).
However I think we need to have the session ticket context to be persistent across sessions since we need to keep the key that is used to encrypt the session ticket.
I'm not really happy with the current solution either since I'm storing state data in a cfg struct.

components/esp-tls/esp_tls.c
```diff
@@ -566,6 +566,54 @@ mbedtls_x509_crt *esp_tls_get_global_ca_store(void)
 
 #endif /* CONFIG_ESP_TLS_USING_MBEDTLS */
 #ifdef CONFIG_ESP_TLS_SERVER
+
+#if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
+
+int esp_tls_session_ticket_ctx_init(esp_tls_session_ticket_ctx_t * ctx) {
+
+    mbedtls_ctr_drbg_init(&ctx->ctr_drbg);
```
AdityaHPatwardhan, line 574: I think this code can be moved to `esp_tls_mbedtls.c` , maybe add a similar function in `esp_tls_mbedtls.h` and include it here.

components/esp-tls/esp_tls.h
```diff
@@ -28,6 +28,9 @@
 #include "mbedtls/ctr_drbg.h"
 #include "mbedtls/error.h"
 #include "mbedtls/certs.h"
+#ifdef CONFIG_ESP_TLS_SERVER_SESSION_TICKETS
+    #include "mbedtls/ssl_ticket.h"
```
AdityaHPatwardhan, line 32:
```suggestion
#include "mbedtls/ssl_ticket.h"
```
```suggestion
    #include "mbedtls/ssl_ticket.h"
```

components/esp-tls/esp_tls.h
```diff
@@ -227,7 +243,13 @@ typedef struct esp_tls_cfg_server {
     unsigned int serverkey_password_len;        /*!< String length of the password pointed to by
                                                      serverkey_password */
 
+#if defined(CONFIG_ESP_TLS_USING_MBEDTLS) && defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
+    esp_tls_session_ticket_ctx_t ticket_ctx;                          /*!< Session ticket generation context */
+#endif
 } esp_tls_cfg_server_t;
+
+int esp_tls_cfg_server_init(esp_tls_cfg_server_t * cfg);
```
AdityaHPatwardhan, line 251: As this is a public API, also add appropriate docs here.

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -352,8 +352,28 @@ static esp_err_t set_global_ca_store(esp_tls_t *tls)
...
 }
 
-
 #ifdef CONFIG_ESP_TLS_SERVER
+#ifdef CONFIG_ESP_TLS_SERVER_SESSION_TICKETS
+int esp_tls_session_ticket_write(void *p_ticket, const mbedtls_ssl_session *session, unsigned char *start, const unsigned char *end, size_t *tlen, uint32_t *lifetime) {
+    int ret = mbedtls_ssl_ticket_write(p_ticket, session, start, end, tlen, lifetime);
+#ifndef NDEBUG
+    if (ret != 0) {
+        ESP_LOGD(TAG, "Writing session ticket resulted in error code %d", ret);
```
AdityaHPatwardhan, line 361:
```suggestion
        ESP_LOGE(TAG, "Writing session ticket resulted in error code %d", ret);
```
```suggestion
        ESP_LOGD(TAG, "Writing session ticket resulted in error code %d", ret);
```

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -404,6 +424,16 @@ esp_err_t set_server_config(esp_tls_cfg_server_t *cfg, esp_tls_t *tls)
...
         return ESP_ERR_INVALID_STATE;
     }
+
+#ifdef CONFIG_ESP_TLS_SERVER_SESSION_TICKETS
+    ESP_LOGD(TAG, "Enabling server-side tls session ticket support");
+
+    mbedtls_ssl_conf_session_tickets_cb( &tls->conf,
+            esp_tls_session_ticket_write,
+            esp_tls_session_ticket_parse,
+            &cfg->ticket_ctx.ticket_ctx );
```
AdityaHPatwardhan, line 434: I think we need to check here if the `&cfg->ticket_ctx.ticket_ctx` is valid or not.
Which I think will also give the user the option to not use server session ticket even if they are enabled in menuconfig.

AdityaHPatwardhan:
Hi @dbahrdt sorry for the late reply, I was looking for any alternate implementation, but I think this is possibly the only way we can keep it persistent across the connections.
I have a few suggestions with this code which have been raised above.
Also, since it is related, do you plan to add the support for session tickets for client as well, otherwise I would be happy to add that !
Thank you

dbahrdt:
Hi @AdityaHPatwardhan,
Thank you for your comments. Hopefully I can address them tomorrow and push an improved version.
I'd be delighted if you implemented the client side part since I'm quite busy at the moment and client session tickets are not a priority for my company.

components/esp-tls/esp_tls_mbedtls.c
dbahrdt: I'm now using a pointer to store the ticket_ctx. This way one only pays for what one uses.
The pointer is initialized in `esp_tls_cfg_server_session_tickets_init` making the code backwards compatible since this function is optional now.
I've add an appropriate configuration option to `httpd_ssl_config`.

components/esp-tls/esp_tls.c
dbahrdt: I have moved the function definition and declaration to esp_tls_mbedtls.*

components/esp-tls/esp_tls.h
dbahrdt: Should be fixed now

components/esp-tls/esp_tls.h
dbahrdt: I have renamed the function to reflect that only the session ticket context is initialized.
Additionally one does not need to call the function if one does not want to use session tickets.
This way the changes should be compatible with the old API.

components/esp-tls/esp_tls_mbedtls.c
dbahrdt: Done.

dbahrdt:
Hi @AdityaHPatwardhan,
sorry for the extremely long delay. I've addressed your comments and added the ability to dynamically enable the session ticket support. The API should now be compatible to the old API. Users of session tickets now need to call additional functions.
Users of the esp_https_server only need to set a configuration option.

AdityaHPatwardhan:
Hi @dbahrdt No problems, It got delayed on my side as well ( sorry for that). Actually last week I had made required changes and was going to add the session tickets support for the client side. I see that you have made certain improvements to this. please have a look into my changes about moving the code to  `esp_tls_mbedtls.c` ?. I will rebase my changes on your new changes and submit a patch some time today so it will be easy to add them for you.  
[server_session_tickets.patch.zip](https://github.com/espressif/esp-idf/files/6766889/server_session_tickets.patch.zip)
and once again thank you for your contribution.

dbahrdt:
Hi @AdityaHPatwardhan,
I had a look at your patch and I'll merge your (new) patch as soon as you publish it.
However you may also push directly to the source branch of this PR.

AdityaHPatwardhan:
[session_tickets_pr_changes.patch.zip](https://github.com/espressif/esp-idf/files/6774112/session_tickets_pr_changes.patch.zip)

This is the patch, please check it once if it works in your case. Thanks

Umm, I don't think I can push directly to your branch in this PR. Maybe I need to push it to my GitHub first and then create a different  PR in your branch.

dbahrdt:
Hi @AdityaHPatwardhan,
I've checked the box `Allow edits and access to secrets by maintainers`.
To be honest I have no idea how you could push to my branch. I guess you could add my repository as another remote?
Anyway, I have applied your patch and pushed the changes.
I had to remove the references to `mbedtls_print_error_msg` since that function does not exist. Did you intend to print a human-readable error message?

Some timings with `openssl s_time` (esp32 rev1):
No session ticket support (disabled on runtime):

```bash
Collecting connection statistics for 30 seconds
**************

14 connections in 0.06s; 233.33 connections/user sec, bytes read 0
14 connections in 31 real seconds, 0 bytes read per connection


Now timing with session id reuse.
starting
***************

15 connections in 0.06s; 250.00 connections/user sec, bytes read 0
15 connections in 33 real seconds, 0 bytes read per connection

```

With session tickets:

```bash
Collecting connection statistics for 30 seconds
***************

15 connections in 0.07s; 214.29 connections/user sec, bytes read 0
15 connections in 32 real seconds, 0 bytes read per connection


Now timing with session id reuse.
starting
rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr

565 connections in 0.47s; 1202.13 connections/user sec, bytes read 0
565 connections in 31 real seconds, 0 bytes read per connection
```

The difference is tremendous. In my view this is an important building block for the usage of tls/https connections on iot devices.
A connection setup of 2 seconds to switch a lightblub is just too much. Persistent connections are not a good option either due to the memory usage (it's better with dynamic receive/send buffers).

AdityaHPatwardhan:
@dbahrdt Thanks for applying the changes,
Actually the `mbedtls_print_error_message` is a new API added in the master branch of the idf. for printing the debug info for the error messages.
I think you will need to rebase your current branch over master since it points to a very old master.

Thanks for sharing the results of your test case.  I agree this  is a good improvement

dbahrdt:
Hi @AdityaHPatwardhan,
that is strange. I've recently rebased to the current master branch. My fork says so too (https://github.com/dbahrdt/esp-idf/tree/mbedtls-session-tickets-in-master: This branch is 9 commits ahead of espressif:master.)

Searching for the function in the espressif/esp-idf repository does not yield a result (https://github.com/espressif/esp-idf/search?q=mbedtls_print_error_msg&type=code).
Could you point me to the file where that function is declared? Thanks.

AdityaHPatwardhan:
ohh, I understand. I am using our internal codebase in which the MR is merged on 28 June, But I think it has not been updated on GitHub till now so that's why you are unable to view it.
Never-mind, I will do that when I will open an internal MR for merging this PR

Alvin1Zhang:

dbahrdt:
Hi @AdityaHPatwardhan,
unfortunately my code contained several bugs. I did not check the validity of the session ticket context pointer in the free functions.
This resulted in a crash if session ticket support was enabled in Kconfig but not during runtime.

I also ported all changes to 4.3 (https://github.com/dbahrdt/esp-idf/tree/patch/4.3%2Fsession-tickets).
The only difference to this PR is the location of the error codes which are defined in an extra file in master but not in relase 4.3.

AdityaHPatwardhan:
@dbahrdt Thank you very much for fixing the issue.
I understand this MR has been long due for merge. This was blocked on a discussion where we would implement this in alternative along with the client. But that seems to be very complicated for the user and this is indeed an easier approach. 
Thank you for your patience. 

As for adding this feature in `release/v4.3`. I don't think this will be added in `release/v4.3` since this is a new feature and contains considerable amount of code change.

PR #7052, opened 2021-05-20
esp-netif: Improve esp_netif_is_netif_listed and esp_netif_get_handle… (IDFGH-5289)

AxelLin:
…_from_ifkey

Each esp_netif_next_unsafe() call needs to iterate the s_head list.
It is inefficient with a do-while loop + esp_netif_next_unsafe() call.
Use SLIST_FOREACH instead to simplify the code and speed-up the matching.

While at it, also fix returning NULL from esp_netif_is_netif_listed().


david-cermak (approved):
LGTM!

Thanks for your help here!

PR #7063, opened 2021-05-23
Enable support for armv6l (IDFGH-5303)

vlad-ivanov-name:
Used in Raspberry Pi Zero


dobairoland:
I'm sorry for taking so long to get back to you. We have difficulties to properly test all of our executables in every relevant scenarios for this architecture. Therefore, we are not able to accept it as supported. We would like to focus our resources more on recent architectures.

Thank you for your contribution and I'm sorry that I have to reject it.

PR #7085, opened 2021-05-30
Fix creation of Python2 virtualenv (IDFGH-5332)

weshouman:
`sys.executable` could be `python3` but the user has the default `virtualenv` interpreter set to `python2` which would result in creating python2 environment and failing with the message `ERROR: Package 'setuptools' requires a different Python: 2.7.12 not in '>=3.5'`.
Thus it's recommended to explicitly specify usage of `python3` as an argument to `virtualenv`.

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -1386,7 +1386,7 @@ def action_install_python_env(args):  # type: ignore
             subprocess.check_call([sys.executable, '-m', 'pip', 'install', '--user', 'virtualenv'],
                                   stdout=sys.stdout, stderr=sys.stderr)
 
-        subprocess.check_call([sys.executable, '-m', 'virtualenv', idf_python_env_path],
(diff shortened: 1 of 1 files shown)


weshouman:
Note: this patch does **not** handle minor version depreciation cases, for example, if the default interpreter is 3.5, which already [fails for version 4.3 beta3](https://github.com/espressif/esp-idf/issues/7086)

dobairoland:
> sys.executable could be python3 but the user has the default virtualenv interpreter set to python2

@weshouman I think this is cased by broken `pip` as I posted in the related issue. 

What you describe can happen if you invoke `virtualenv` directly. However, `sys.executable -m virtualenv` should use the `virtualenv` module installed for `sys.executable`.

We try to avoid hardcoding `python3` in our scripts. The same script is invoked from all platforms. So please let us find a suitable solution for your issue in another way.

PR #7113, opened 2021-06-07
Const corrections in websocket client config structs (IDFGH-5364)

0xFEEDC0DE64:
Fixes some compiler warning in C++, as we assign string literals, which should not be converted to a char*


Alvin1Zhang:
Thanks for your contribution.

david-cermak (changes requested):
Thanks for posting these changes! we should pay more attention to const correctness, during design (now this is a little bit breaking API, but still okay to be accepted)

Are the changes made in the `components/esp_websocket_client/esp_websocket_client.c` needed to get the cxx compiler issues fixed? If you revert these changes, the PR would LGTM.

Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/70b1247a47f4583fccd8a91bf6cc532e5741e632.

Issue #7129, opened 2021-06-09
Default argument to esp_prov.get_security causes a TypeError (IDFGH-5380)


alexhenrie:
The `get_security` function in [esp_prov.py](https://github.com/espressif/esp-idf/blob/master/tools/esp_prov/esp_prov.py) is defined as follows:

```
def get_security(secver, pop=None, verbose=False):
    if secver == 1:
        return security.Security1(pop, verbose)
    elif secver == 0:
        return security.Security0(verbose)
    return None
```

However, if attempting to use the function from an external script, the default `pop` argument results in a TypeError:

```
import sys
from os import environ
sys.path.append(environ['IDF_PATH'] + '/tools/esp_prov')
from esp_prov import *

security = get_security(secver=1)
```

```
TypeError: cannot convert 'NoneType' object to bytes
```

A workaround is to explicitly specify `get_security(secver=1, pop='')` when not using proof-of-possession.

Alvin1Zhang:
Thanks for reporting and patience, we have a fix which is now under internal reviewing, the issue will close automatically once the fix is available on GitHub. Thanks.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/5ed53c272b2fac5c052d2d097b2fd6e9957ffd01

PR #7130, opened 2021-06-09
Change default PoP argument of esp_prov.get_security from None to string (IDFGH-5381)

alexhenrie:
Closes [IDFGH-5380](https://github.com/espressif/esp-idf/issues/7129)

### DIFF
--- a/tools/esp_prov/esp_prov.py
+++ b/tools/esp_prov/esp_prov.py
@@ -51,7 +51,7 @@ def on_except(err):
         print(err)
 
 
-def get_security(secver, pop=None, verbose=False):
+def get_security(secver, pop='', verbose=False):
     if secver == 1:
         return security.Security1(pop, verbose)
     elif secver == 0:


Alvin1Zhang:
Thanks for your contribution.

shivani-tipnis:
Hi @alexhenrie

The esp_prov.py provides the option whether you want to use PoP or not as a CLI argument.

Could you tell us your usecase for calling the function separately from a different script ?

Thanks

alexhenrie:
My biggest reason for importing functions from esp_prov.py, but not running esp_prov.py directly, is to send data to a series of custom endpoints, none of which is named "custom-data". It's clear that esp_prov.py was designed to allow importing its functions into other scripts because its main logic is disabled by the `if __name__ == '__main__':` statement when importing.

shivani-tipnis:
Thank you.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/5ed53c272b2fac5c052d2d097b2fd6e9957ffd01, thank you for your contribution.

alexhenrie:
Thank you!

PR #7134, opened 2021-06-10
Add missing REG_SPI_BASE(x) macro

s-hadinger:
The `REG_SPI_BASE(x)` macro is missing for Esp32c3. Considering the address scheme, we can safely use the same formula as Esp32.

I can be further simplified since i can have only values 0/1/2, but for the sake of simplicity, re-using ESP32 formula also works.


Alvin1Zhang:
Thanks for your contribution.

PR #7147, opened 2021-06-14
Update i2c.h

iltis42:
Added  i2c_set_alive_interval() function prototype.

PR #7148, opened 2021-06-14
Update i2c.c: Support to configure the I2C alive_interval 

iltis42:
Add needed support to configure the I2C alive_interval to overcome a blocking issue with the 1000 ticks default, that equals to 1 second.

The issue is that I2C bus remains blocked by semaphore if one of several devices encounters a an error on reading, what leads to the situation that other devices cannot be reached during that period. A flexible timer will overcome that, and sporadic issues observed with some hardware chips and noisy buses, are mostly gone in the next reread several milliseconds later, so a timer much more less that 1000 ms behaves much more lean and efficient.

The change is 100% backward compatible, 
the default remains as is, but may optionally be overruled by set_i2c_alive_timeout( i2c_num, time );

### DIFF
--- a/components/driver/i2c.c
+++ b/components/driver/i2c.c
@@ -194,6 +194,12 @@ static i2c_obj_t *p_i2c_obj[I2C_NUM_MAX] = {0};
 static void i2c_isr_handler_default(void *arg);
 static void IRAM_ATTR i2c_master_cmd_begin_static(i2c_port_t i2c_num);
 static esp_err_t IRAM_ATTR i2c_hw_fsm_reset(i2c_port_t i2c_num);
+static int i2c_cmd_alive_interval[I2C_NUM_MAX] = { I2C_CMD_ALIVE_INTERVAL_TICK, I2C_CMD_ALIVE_INTERVAL_TICK };
+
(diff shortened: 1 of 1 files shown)


components/driver/i2c.c
```diff
@@ -194,6 +194,12 @@ static i2c_obj_t *p_i2c_obj[I2C_NUM_MAX] = {0};
 static void i2c_isr_handler_default(void *arg);
 static void IRAM_ATTR i2c_master_cmd_begin_static(i2c_port_t i2c_num);
 static esp_err_t IRAM_ATTR i2c_hw_fsm_reset(i2c_port_t i2c_num);
+static int i2c_cmd_alive_interval[I2C_NUM_MAX] = { I2C_CMD_ALIVE_INTERVAL_TICK, I2C_CMD_ALIVE_INTERVAL_TICK };
+
+static void i2c_set_alive_interval(i2c_port_t i2c_num, int interval)
```
atanisoft, line 199: this method should not be declared as `static` since it is exported via the i2c.h file (in your other PR: #7147)

iltis42: Sure, removed static qualifier.

Alvin1Zhang:
Thanks for your contribution.

PR #7166, opened 2021-06-18
Replaced void* with proper typed netif * (IDFGH-5423)

0xFEEDC0DE64:


Alvin1Zhang:
Thanks for your contribution.

david-cermak (changes requested):
One of the primary purpose of `esp_netif` is to decouple the IDF code from network stacks. This change would require users to pre-include lwip headers. This is currently enabled by default:
```
CONFIG_ESP_NETIF_TCPIP_ADAPTER_COMPATIBLE_LAYER=y
```

So the `netif_impl` pointer at this point should be something more universal, e.g. the opaque `void*` ptr as is, but perhaps it could be a `struct netif *`, but we would have to forward declare the `struct netif` if the lwip is not included.

david-cermak:
@0xFEEDC0DE64 Any idea how to use strong typing and be independent from the lwip at the same time?
We will close this PR if not, thanks.

Alvin1Zhang:
@0xFEEDC0DE64 Thanks for your contribution, any updates regarding @david-cermak 's comments? Thanks.

Alvin1Zhang:
@0xFEEDC0DE64 Thanks for your contribution, any updates regarding @david-cermak comments? Thanks.

Alvin1Zhang:
@0xFEEDC0DE64 Thanks for your contribution again, will close due to short of feedback, feel free to reopen.

PR #7200, opened 2021-06-28
Adding documentation --Displaying the board's port on Linux is more intuitive (IDFGH-5461)

tnoumar:
I added explanation to the available command, added more steps to it so it should be clear right now. All the commands used are POSIX and therefore work on most Linux flavours.

### DIFF
--- a/docs/en/get-started/establish-serial-connection.rst
+++ b/docs/en/get-started/establish-serial-connection.rst
@@ -62,7 +62,7 @@ Display the port of your board / dongle ::
 And then finally remove the two files ::
 
     rm ~/before_plugging.txt ~/after_plugging.txt
-    
+
 macOS ::
 
     ls /dev/cu.*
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

espwangning:
Thanks for your contribution! We are really sorry for the late reply.

After internal discussion, we've decided not to merge this PR. Here's our reasoning:

The proposed approach works, but it replaces a single command with a 4-step process. The single command to list available ports is a common practice, as you will find it in the [Ubuntu community wiki](https://help.ubuntu.com/community/DialupModemHowto/Setserial) and [Debian wiki](https://wiki.debian.org/HowToIdentifyADevice/Serial). We would like to keep our guide simple.

Your approach might be useful for users who have multiple serial ports connected and find visual comparison difficult. We may consider adding a simpler command for those users in the future:
```
sudo dmesg | grep tty
```

This command shows only the newly detected device after plugging in, without any temp files or cleanup needed.

Thanks again for taking the time to improve the docs. We appreciate it!

PR #7216, opened 2021-06-30
Avoid bootloop if chip is unknown

s-hadinger:
In case the PSIRAM chip is unknown, return an error and disable PSRAM
instead of calling abort() and causing a bootloop.

In Tasmota, we provide pre-compiled firmwares and we try to make the generic work on as many chips as possible. We have now enabled PSRAM by default, but when the chip is unknown, it bootloops because of the call to abort().

I don't see any downside in returning an error instead and letting the chip boot without PSRAM, and give control to the software as to how it should handle it.


Alvin1Zhang:
Thanks for your contribution.

ginkgm:
Hi @s-hadinger ,

Thanks for your contribution. We have a similar feature for this. Please enable SPIRAM_IGNORE_NOTFOUND in the menuconfig.

Indastri:
@ginkgm can we close this, since we have similar feature?

suda-morris:
can close now thanks to https://github.com/espressif/esp-idf/commit/33c26371269b76628ce82d1308f1750053ee23c6

PR #7233, opened 2021-07-06
esp_supplicant: Make esp_rrm_send_neighbor_rep_request return proper … (IDFGH-5503)

AxelLin:
…error

Current code always return 0 even though wpas_rrm_send_neighbor_rep_request()
fails. Return proper error so the caller can know what's wrong.


AxelLin:
Hi @kapilkedawat 
Please review this PR.

Without the fix, https://github.com/espressif/esp-idf/blob/master/examples/wifi/roaming/main/roaming_example.c#L316 cannot catch the error.

Also note v4.3 also needs the fix.

kapilkedawat:
Hi @AxelLin,

Thanks for contributing. This fix is already part of latest IDF code and should be available in github next push.

PR #7248, opened 2021-07-08
wpa_supplicant: Fix wps_free_pins to remove all pins (IDFGH-5519)

AxelLin:
Current code does not correctly free all pins in wps_free_pins due to the
semicolon at the end of dl_list_for_each_safe(). Fix it.


sagb2015:
@AxelLin  The change looks good. FYI we dont support registrar functionality yet. We intend sync up with upstream when we come to it. In the meantime we will merge this.

PR #7253, opened 2021-07-09
Linenoise - RAM optimization (IDFGH-5525)

patrykkrz:
Linenoise use default 4kB RAM memory. Maybe small default LINENOISE_MAX_LINE doesn't have to be that big.

### DIFF
--- a/components/console/linenoise/linenoise.c
+++ b/components/console/linenoise/linenoise.c
@@ -118,7 +118,7 @@
 #include "linenoise.h"
 
 #define LINENOISE_DEFAULT_HISTORY_MAX_LEN 100
-#define LINENOISE_MAX_LINE 4096
+#define LINENOISE_MAX_LINE 512
 
 static linenoiseCompletionCallback *completionCallback = NULL;
 static linenoiseHintsCallback *hintsCallback = NULL;


igrr (changes requested):
Thanks for the PR @patrykkrz!
As is, this has a chance of being a breaking change for some application that relies on such large line lengths. Although not very likely, we had seen cases when users have console commands e.g. to provision certificates, and base64-encoded certificates can be pretty long.
I agree that this is a useful optimization option to have, though.
How about making this runtime-configurable, same way as the max history length is?

Basically, you can introduce a new static variable (static int max_line_len), initialize it to LINENOISE_MAX_LINE, and then add new function `linenoiseSetMaxLineLen` which can modify `max_line_len`. This way existing applications won't see any change, and you will be able to adjust the maximum line length.

If you do add this new function, you can also add it to docs/en/api-reference/system/console.rst (see "Configuration" section for linenoise) and call it in the system/console example.

igrr:
One more request, ESP-IDF project only accepts pull requests made against the master branch. Please rebase your changes on top of `master` and update the PR. Once the changes are merged into master we can also backport them to release/v4.3, if you need them on that branch. This ensures that any improvements made now will not be missing from future versions of IDF.

patrykkrz:
> Thanks for the PR @patrykkrz!
> As is, this has a chance of being a breaking change for some application that relies on such large line lengths. Although not very likely, we had seen cases when users have console commands e.g. to provision certificates, and base64-encoded certificates can be pretty long.
> I agree that this is a useful optimization option to have, though.
> How about making this runtime-configurable, same way as the max history length is?
> 
> Basically, you can introduce a new static variable (static int max_line_len), initialize it to LINENOISE_MAX_LINE, and then add new function `linenoiseSetMaxLineLen` which can modify `max_line_len`. This way existing applications won't see any change, and you will be able to adjust the maximum line length.
> 
> If you do add this new function, you can also add it to docs/en/api-reference/system/console.rst (see "Configuration" section for linenoise) and call it in the system/console example.

Yes, it's good idea. My change isn't safe for existing projects. I will correct it.

patrykkrz:
@igrr I close this PR and open new. After change base to master, this PR have all changes beetwen 4.3 and master ...

PR #7254, opened 2021-07-09
Linenoise - RAM optimization (IDFGH-5527)

patrykkrz:
Linenoise use default 4kB RAM memory. Maybe small default LINENOISE_MAX_LINE doesn't have to be that big.

It's copy of the PR: Linenoise - RAM optimization (IDFGH-5525) #7253, but with correct base/compare branches and better changes (this change not will not affect existing projects).

### DIFF
--- a/components/console/linenoise/linenoise.c
+++ b/components/console/linenoise/linenoise.c
@@ -119,13 +119,15 @@
 #include "linenoise.h"
 
 #define LINENOISE_DEFAULT_HISTORY_MAX_LEN 100
-#define LINENOISE_MAX_LINE 4096
+#define LINENOISE_DEFAULT_MAX_LINE 4096
+#define LINENOISE_MINIMAL_MAX_LINE 64
 #define LINENOISE_COMMAND_MAX_LEN 32
 
 static linenoiseCompletionCallback *completionCallback = NULL;
 static linenoiseHintsCallback *hintsCallback = NULL;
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

ESP-YJM:
@patrykkrz I think it's better to use Kconfig to allow user to configure the length they want. It may look simple.

patrykkrz:
> @patrykkrz I think it's better to use Kconfig to allow user to configure the length they want. It may look simple.

I did it according to the description @igrr
https://github.com/espressif/esp-idf/pull/7253#pullrequestreview-702898837

@ESP-YJM Change it to Kconfig ?

igrr:
Generally Kconfig options are used only if something is hard to make run time configurable. In this case run-time configuration is trivial so let's keep that. This works better for environments where IDF is built with a fixed sdkconfig, such as Arduino or Micropython.

ESP-YJM:
@patrykkrz @igrr Sure, If it is configurable at runtime, offer API to configure is the best way.

igrr:
@patrykkrz Thanks again for the PR, the changes have been incorporated into https://github.com/espressif/esp-idf/commit/0fd74a43c807b3b5e94833c5820b3dab75a93c82. I am sorry that we didn't properly link back to your PR from the commit message and we squashed your changes into larger commit.

PR #7261, opened 2021-07-12
Configuration custom mac addresses for interfaces (overwrite base MAC address). (IDFGH-5534)

patrykkrz:
Sometimes you may need to define your own MAC addresses, not necessarily generated from the base MAC address.


Alvin1Zhang:
Thanks for your contribution.

KaeLL:
Shouldn't this be behind a kconfig option?

KonstantinKondrashov:
Hi!
We can add it to the master but it is outdated. I will create an MR in our internal repo based on these changes. Thank you for your contribution.

PR #7266, opened 2021-07-13
esp_modem_dce_service: Make esp_modem_dce_handle_* functions static (IDFGH-5542)

AxelLin:
These functions are unlikely to be reused, make them static.
Take esp_modem_dce_handle_cbc as an example, it's internal implementation
of esp_modem_dce_get_battery_status, expose esp_modem_dce_get_battery_status
function is enough.


david-cermak (changes requested):
Thanks for this PR. Yes, I think it's a good idea to make the functions for handling replies `static`.

Could you please just move the doxy comments next to the definitions. Even if it's not a public API, i think it's a good idea to keep the documentation, at least the first `@brief` line.

david-cermak (approved):
Thanks for the quick update. LGTM!

Alvin1Zhang:
Thanks for your contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/d6248eaf9b988c67dae6dafeaec523aea7428b35.

PR #7269, opened 2021-07-13
Add support for LAN874xA Ethernet PHYs (IDFGH-5544)

sauttefk:
The **LAN8740A**/**LAN8741A**/**LAN8742A** are enhanced versions of the well known **LAN8710A**/**LAN8720A** Ethernet PHYs. They are hardware and software drop-in-replacements for the existing designs. The additional features are enabled by software or by a different bootstrapping on some I/O pins.
The existing LAN8720 driver in esp-idf can be used as is, except for additional `id2.vendor_model` IDs that have to be accepted.
New features like _Wake on LAN_ (WOL), _Cable Diagnostics_ (CD) or Energy Efficient Ethernet (EEE) are not added by this PR.
New or modified vendor specific registers are added just for completeness.

Please refer to [SMSC/Microchip Application Note 25.3](http://ww1.microchip.com/downloads/en/AppNotes/en562713.pdf) for any questions regarding the migration to LAN874xA.


Alvin1Zhang:
Thanks for your contribution.

kostaond:
Thank you for your contribution! We already have internal update to add support of LAN874xA pending. However, we didn't added/updated registers layout for all chips since it was not needed from functional point of view. On the other hand it could be useful in the future. Therefore, we will consider to updating it based on your proposal.

PR #7296, opened 2021-07-20
Fix memory leaks identified by cppcheck (IDFGH-5573)

alexhenrie:
These memory leaks were identified by the [cppcheck](http://cppcheck.sourceforge.net/) static analyzer.


Alvin1Zhang:
Thanks for your contribution.

AdityaHPatwardhan:
@alexhenrie Thank you for the PR. We are in the process of merging it in our internal code base, it should be available on GitHub in a few days.

Issue #6172, opened 2020-11-27
esp_ota + encrypted firmware example (IDFGH-4334)


kshyshkin:
Hello. Have anybody example how make ota update with encrypted firmware using esp_ota_begin(), esp_ota_write(); esp_ota_end in release/v4.2 esp-idf ?

mahavirj:
@kshyshkin 

Default OTA examples shipped with ESP-IDF (ref: https://github.com/espressif/esp-idf/tree/master/examples/system/ota) should work with flash encryption enabled case as well.  `esp_ota_write` uses underlying partition manager API [esp_partition_write](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/spi_flash.html#_CPPv419esp_partition_writePK15esp_partition_t6size_tPKv6size_t), which transparently handles encrypted flash writes.

If you are facing any specific problem please feel to post detailed report per github issue template.

kshyshkin:
@mahavirj Thank you for your reply. Problem with ```esp_ota_write``` for pre-encrypted firmware. ```esp_ota_write``` check magic byte and etc as in non encrypted firmware. I try ```esp_ota_write_offset``` it works, but after in ```esp_ota_end``` i got same situation with checking magic byte. I cant find how solve problem with ota update with pre-encrypted firmware.

mahavirj:
@kshyshkin Can you please explain more details about your use-case? Do you have flash encryption enabled on device? If yes, any specific reason to flash pre-encrypted firmware?

kshyshkin:
@mahavirj Yes all secure features on controller on flash encryption (RELEASE), secure boot reflashable, all needed fuses on. But i have OTA (not through webserver), with intermediate device. When i need update by OTA, i make project, sighn fw with pem key, encrypt fw by pem key, and send to  intermediate device. Intermediate device send pre-encrypted fw to esp32 by custom interface and protocol  dividing the firmware by chunks. I need that from my host computer to esp32 fw goes pre-encrypted.

mahavirj:
@kshyshkin 

At this moment, pre-encrypted firmware write is not supported in `app_update` component.

> I see in documentation that esp_partition_write_raw function, i think it can help me solve this problem, bu i cant find in which release this function burn.

This API is part of latest `master` branch and it will reflect in ESP-IDF v4.3 release. 

[...]

Issue #3671, opened 2019-06-21
Does ESP32 support seamless Wi-Fi roaming (IEEE 802.11v , r, k) in the STA mode?  (IDFGH-1392)


flyoo:

sagb2015:
Currently ESP32 does not support these features. We will keep these requirements in mind, but it is unlikely that they will be available anytime soon. Among the three, we may prioritise 11r though.

MoffKalast:
It's been a year, any progress on this front?

sagb2015:
We have started working on 11r. However, it is still at preliminary stage.

g8ecj:
Is the 11r going to be available to resellers/packagers such as u-blox?

AxelLin:
@sagb2015 

I have noticed that 27101f94546bf054ae022fcb9c3acb673de0e270 added initial roaming support.
How about 802.11r support?

Also note, the comment of esp_wifi_set_rssi_threshold() says
This API needs to be called every time after WIFI_EVENT_STA_BSS_RSSI_LOW event is received.

But in the example code, it actually calls it when got WIFI_EVENT_STA_CONNECTED?

kapilkedawat:
Hi @AxelLin ,

11r support will be added in future.

'This API needs to be called every time after WIFI_EVENT_STA_BSS_RSSI_LOW event is received.'
This comment is to indicate that only one event will come after calling esp_wifi_set_rssi_threshold(), for getting a second event esp_wifi_set_rssi_threshold() needs to be called again.

policeman0077:
any updates?

sagb2015:
Hi @policeman0077, we had to put 11r in the backlog because of other higher priority action items. We will keep you updated when we restart 11r activity. We are targeting 2nd quarter of 2021.

policeman0077:
> Hi @policeman0077, we had to put 11r in the backlog because of other higher priority action items. We will keep you updated when we restart 11r activity. We are targeting 2nd quarter of 2021.

thank you :)

AxelLin:
Hi @kapilkedawat 

After checking the roaming_example.c code, it looks like esp_bss_rssi_low_handler can be a common implementation to me.
Why not provide a common implementation instead of adding the implementation in application (example) code?
Or do you expect people to implement different behavior in esp_bss_rssi_low_handler?

kapilkedawat:
Hi @AxelLin ,

[...]

PR #7305, opened 2021-07-22
app_update: Add esp_ota_write_raw function to write pre-encrypted data (IDFGH-5583)

AxelLin:
Use esp_ota_write_raw instead esp_ota_write to write pre-encrypted data.

### DIFF
--- a/components/app_update/esp_ota_ops.c
+++ b/components/app_update/esp_ota_ops.c
@@ -38,6 +38,7 @@ typedef struct ota_ops_entry_ {
     uint32_t handle;
     const esp_partition_t *part;
     bool need_erase;
+    bool raw_write;
     uint32_t wrote_size;
     uint8_t partial_bytes;
     WORD_ALIGNED_ATTR uint8_t partial_data[16];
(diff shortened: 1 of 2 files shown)


AxelLin:
Hi @mahavirj 
I understand you comments in https://github.com/espressif/esp-idf/issues/6172#issuecomment-884777914.
Before introducing new "pre-encrypted" OTA image format, this function allows
existing users to implement updating pre-encrypted ota data.(https://github.com/espressif/esp-idf/issues/6172#issuecomment-881801160)
Please review if this PR is acceptable.

mahavirj:
@AxelLin I am not quite clear on this change. So you think that application needs to take care about correct "pre-encrypted" image per flash offset per passive OTA update slot?

AxelLin:
Hi @mahavirj 
Yes, for the use case for some customers using device in closed netowrk.
I have no choice but to implement supporting "pre-encrypted" image by taking
care the correct logic in our application side.
Actually when we do OTA in an independent partition (e.g. test partition), we
can always update image to ota1 partition and set boot from ota1. (so choose image issue).

mahavirj:
@AxelLin Thanks for explaining. It will be difficult to accept this PR, unless we add lot of explanation in documentation (and probably application as well) to clarify usage scenario here. I would request that we wait for official approach as I had highlighted in https://github.com/espressif/esp-idf/issues/6172#issuecomment-884777914. Of course, if someone find this change useful they can always refer to PR here.

AxelLin:
And honestly, such requirement (ota with pre-encrypted image) has been asked so many times.
Waiting for a new "pre-encrypted" OTA image format could take too much time.
(It maybe not available this year)

If this PR is not acceptable, I have no choice but to maintain my on fork of esp-idf.
Thanks for your time anyway.

AxelLin:
> @AxelLin Thanks for explaining. It will be difficult to accept this PR, unless we add lot of explanation in documentation (and probably application as well) to clarify usage scenario here. I would request that we wait for official approach as I had highlighted in [#6172 (comment)](https://github.com/espressif/esp-idf/issues/6172#issuecomment-884777914). Of course, if someone find this change useful they can always refer to PR here.

On the second thought, by calling esp_ota_get_next_update_partition() the
firmware can easily know which encrypted ota image to download.

The only thing needs to comment or document is the feteched encrypted ota image needs
to be generated for the target partition address.

BTW, this is probably off-topic.
But I had bad experience about waitting new features ready.
e.g. regarding 802.11r support:
21 Jun 2019: we may prioritise 11r though. https://github.com/espressif/esp-idf/issues/3671#issuecomment-504357006
18 Dec 2020: 11r support will be added in future. https://github.com/espressif/esp-idf/issues/3671#issuecomment-747897503
25 Jan 2021: We are targeting 2nd quarter of 2021. https://github.com/espressif/esp-idf/issues/3671#issuecomment-766896837
Now, 11r is still not yet available. (It's more than 2 years since initial reply)

tsctrl:
Hi @AxelLin , interested with this pr change. few question:

1. this is not related to this pr, since we not able to distribute pre encrypted fw and rely on automatic fw updates with plaintext fw (correct me). how to make sure and to secure out plaintext fw from unauthorize download and access from the server.
2. is the ota update only work with a sequence next partition boot. means after ota_0, must ota_1 then ota_0 back. is it possible for example to just always boot/use and update ota_0 update current boot partition ota_0. if fail boot factory?
2. since the issue is with the correct address, is it possible we just use simpler approach. let say i have factory and ota_0, and ota_1. so every time i do ota update, instead of automatic next ota partition. we just check the file name encrypted_ota_1.bin if no match then no update. if match then update to ota_1.bin and set as boot?
3. for the application to manage the partition how that can be done? what is application btw? is it the website to upload firmware or the fw itsef? 

thanks!

sagb2015:
@AxelLin -  We have to keep changing priorities of the deliverables based on constantly changing demands/requirements.  For example, https://github.com/espressif/esp-idf/issues/3671#issuecomment-504357006 says that 11r will be prioritised over 11k, 11v, however, we had to make 11k and 11v ready before 11r for other prerequisites. Our intention of providing the tentative timeline is to give visibility to everyone based on the current inputs. This may or may not translate into hard deadlines for us. Hope you understand!

Issue #7272, opened 2021-07-14
wifi_apb80m_request (?) leads to crash and bricks device in field on Mikrotik routers (IDFGH-5547)


thecashit:
- Module or chip used: ESP32-WROOM-32D
- IDF version v4.3
- Build System:  idf.py
- Compiler version: xtensa-esp32-elf-gcc (crosstool-NG esp-2020r3) 8.4.0
- Operating System: macOS
- Power Supply: USB

## Problem Description
We encountered very unpleasant device behavior after upgrading to IDF 4.3. 
Moreover, this did not immediately manifest itself and the bug is floating.
Due to conditions that are still incomprehensible to us, sometimes the following output to the console appears:

Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.

Core  0 register dump:
PC      : 0x4000c52a  PS      : 0x00060033  A0      : 0x80093898  A1      : 0x3ffefe60  
A2      : 0x3ffefe80  A3      : 0xffffffff  A4      : 0x0000004b  A5      : 0x3ffefeca  
A6      : 0x00000001  A7      : 0x00000004  A8      : 0x00000000  A9      : 0x3ffeffbc  
A10     : 0xffffffff  A11     : 0x3ffeffbc  A12     : 0x3fff0070  A13     : 0x3fff0050  
A14     : 0x00000000  A15     : 0x00000001  SAR     : 0x00000000  EXCCAUSE: 0x0000001c  
EXCVADDR: 0xffffffff  LBEG    : 0x4000c46c  LEND    : 0x4000c477  LCOUNT  : 0x00000000  

Backtrace:0x4000c527:0x3ffefe60 0x40093895:0x3ffefe80 0x40094455:0x3ffeff00 0x40093434:0x3ffeff20 0x400933e6:0xffffffff |<-CORRUPTED
0x40093895: vPortExitCritical at /Firmware/Toolchain/esp-idf/components/freertos/port/xtensa/port.c:472
0x40094455: vTaskSwitchContext at /Firmware/Toolchain/esp-idf/components/freertos/tasks.c:3290 (discriminator 3)
0x40093434: _frxt_timer_int at /Firmware/Toolchain/esp-idf/components/freertos/port/xtensa/portasm.S:314
0x400933e6: _frxt_int_exit at /Firmware/Toolchain/esp-idf/components/freertos/port/xtensa/portasm.S:185

ELF file SHA256: b3ff32802362285b

Rebooting...
[...]

PR #7314, opened 2021-07-23
esp_flash_api fixes (IDFGH-5592)

negativekelvin:
As found in https://github.com/espressif/esp-idf/issues/7272, calling esp_flash_erase_region with len=0 would erase the whole flash because it would get to the erase loop and the length would underflow.  Fix to return ESP_OK to match behavior of other flash api functions.  Returning ESP_OK immediately without checking other arguments will mask other problems like memory corruption.  Add sanity checks to write/erase loops.


igrr:
Thanks for the fix @negativekelvin!

Alvin1Zhang:

PR #7325, opened 2021-07-25
Fix install.fish (IDFGH-5604)

kelvie:
Looks like in commit f6c2198a214b3b5a853904f9ceebfa0ba30f1418 there were
some unintended changes to this file.

### DIFF
--- a/install.fish
+++ b/install.fish
@@ -14,8 +14,7 @@ else
 end
 echo "Installing ESP-IDF tools"
 "$ESP_PYTHON" "$IDF_PATH"/tools/idf_tools.py install --targets=$TARGETS
-    exit 1
-end
+or exit 1
 
 echo "Installing Python environment and packages"
 "$ESP_PYTHON" "$IDF_PATH"/tools/idf_tools.py install-python-env


dobairoland:
Thank you @kelvie for letting us know and submitting a fix.

thomashinds:
It looks like the intended behavior is for the script to exit if [that command](https://github.com/espressif/esp-idf/pull/7325/commits/fb51e82c7316de6634f7e016e088a3d154b922ed#diff-6ec28791638d398f88f7b80f033ce516bfd141b0d0a356f82fdbd5020de56451L16) fails. In install.sh, `set -e` is used for the same effect. Can get the same effect in fish by leaving the `exit` and `end` and prefacing the previous command with `if not`

kelvie:
> It looks like the intended behavior is for the script to exit if [that command](https://github.com/espressif/esp-idf/pull/7325/commits/fb51e82c7316de6634f7e016e088a3d154b922ed#diff-6ec28791638d398f88f7b80f033ce516bfd141b0d0a356f82fdbd5020de56451L16) fails. In install.sh, `set -e` is used for the same effect. Can get the same effect in fish by leaving the `exit` and `end` and prefacing the previous command with `if not`

Fixed

PR #7336, opened 2021-07-27
lwip: menuconfig option for max number of SNTP servers (IDFGH-5616)

vortigont:
LWIP has two define's for setting max number of sntp servers:
 - total number of handled servers
 - max number of sntp's picked via DHCP

by default both values are equal to 1, but could be set separately.
This is a followup for espressif/arduino-esp32#5343. Having this menuconfig option allows building lwip for arduino-esp32 with multiple sntps's supported

### DIFF
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -837,7 +837,7 @@ menu "LWIP"
 
     menu "SNTP"
 
-        config LWIP_DHCP_MAX_NTP_SERVERS
+        config LWIP_SNTP_MAX_SERVERS
             int "Maximum number of NTP servers"
             default 1
             range 1 16
@@ -845,6 +845,14 @@ menu "LWIP"
                 Set maximum number of NTP servers used by LwIP SNTP module.
                 First argument of sntp_setserver/sntp_setservername functions
                 is limited to this value.
+
+        config LWIP_DHCP_MAX_NTP_SERVERS
+            int "Maximum number of NTP servers aquired via DHCP"
+            default 1
+            range 1 16
+            help
+                Set maximum number of NTP servers aquired via DHCP-offer.
+                Must be less or equal to total "Maximum number of NTP servers".
 
         config LWIP_SNTP_UPDATE_DELAY
             int "Request interval to update time (ms)"
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -846,13 +846,22 @@ menu "LWIP"
                 First argument of sntp_setserver/sntp_setservername functions
                 is limited to this value.
 
+        config LWIP_DHCP_GET_NTP_SRV
+            bool "Request NTP servers from DHCP"
+            default n
+            help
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/lwip/port/esp32/include/lwipopts.h
```diff
@@ -1074,6 +1074,15 @@ LWIP_FORWARD_DECLARE_C_CXX void sntp_sync_time(struct timeval *tv);
 
 LWIP_FORWARD_DECLARE_C_CXX uint32_t sntp_get_sync_interval(void);
 
+// Max number of SNTP servers handled (default equal to LWIP_DHCP_MAX_NTP_SERVERS)
+#if defined CONFIG_LWIP_SNTP_MAX_SERVERS && !defined SNTP_MAX_SERVERS
```
david-cermak, line 1078:
```suggestion
#if defined CONFIG_LWIP_SNTP_MAX_SERVERS
```
As we don't allow to override `CONFIG_...` macros defined in configuration menu.

components/lwip/Kconfig
```diff
@@ -846,6 +846,23 @@ menu "LWIP"
...
+                SNTP callback for such replies should be set accordingly (see sntp_servermode_dhcp() func.)
+
+        config LWIP_DHCP_MAX_NTP_SERVERS
+            int "Maximum number of NTP servers aquired via DHCP"
+            default 1
+            range 1 16
+            depends on LWIP_DHCP_GET_NTP_SRV
+            help
+                Set maximum number of NTP servers aquired via DHCP-offer.
+                Must be less or equal to total "Maximum number of NTP servers".
```
david-cermak, line 864: Question: Does the lwip throws a compile `#error` if it `> SNTP_MAX_SERVERS` ? maybe we should if it doesn't.

david-cermak (approved):
LGTM, just a question and an optional suggestion.

components/lwip/port/esp32/include/lwipopts.h
vortigont: OK, I will amend the change into last commit to keep it clean.

components/lwip/Kconfig
vortigont: No, it does not. Actually if I get this right any extra DHCP-servers would be just silently ignored. Not sure that throwing an error and stop here is a good idea.
@david-cermak I can make a test setup with dhcp-server and confirm this if you like?

vortigont:
@david-cermak I've updated notes for menuoption, also added NTP  debug option and sntp example to utilize SNTP via DHCP if build-in.
Would appreciate if you could review changes.
Thanks!

david-cermak:
@vortigont Thanks for the prompt update and addressing the concern of misconfiguration.
This PR looks very good to me and has been added to the internal round of testing and reviews (this usually takes some time though till it gets merged to master).

PR #7349, opened 2021-08-01
fix esp32-c3 usb connect pin number (IDFGH-5629)

acevest:


chegewara:
According to datasheet (page 11):
https://www.espressif.com/sites/default/files/documentation/esp32-c3_datasheet_en.pdf
Screenshot from 2021-08-01 17-23-24

Spritetm:
Aside from the fact that the pinout matches the datasheet, my devboard also would not work if that table was indeed wrong:
image
Closing this, if you really feel this is in error I'd love to hear why you think this.

acevest:
i bought esp32 dev board from ai-thinker.
But IO19 and IO18 on this development board are marked opposite.
sorry for that.


我是被安信可的ESP32-C3-12F-Kit开发板给坑了，他们把IO18、IO19标反了。所以才导致我提交这个MR。如果有人买的时候注意一下。我是2021年7月29号买的。
另外再补充一下，他家的ESP32-12H-Kit开发板也不靠谱，好多引脚都标错了。谨慎购买。



image

image

acevest:
> Aside from the fact that the pinout matches the datasheet, my devboard also would not work if that table was indeed wrong:
> image
> 
> Closing this, if you really feel this is in error I'd love to hear why you think this.

cause the board i bought marked IO18 as IO19, IO19 as IO18. 😂😂😂😭😭😭

0xjakob:
@acevest The problem is with the AIThinker board (AIThinker is a third party), not with the C3 chip. @Spritetm and me have both verified that these pins work according to the datasheet.

Issue #7339, opened 2021-07-29
timing for I2S DMA queue events via esp_timer_get_time (IDFGH-5619)


sdrsdr:
I'm in a project where ADC via I2S DMA is used to track an analog signal. The signal must be correlated to other events. As you an imagine the timing is important and with tight limits. Using the I2S driver`s capability to send events via queue is almost perfect with the only downside is that the events might get variable time to get delivered. This can be easily solved by punting a timestamp from a call to esp_timer_get_time()  in the queue event. The only problem I see is keeping the code compatible with older users of the event queue.  To keep it clean this can be made selectable from the configuration system with default to non-timestamped events.
   I can create a pull request with such feature if it is OK by the maintainers.  Any suggestions/guides are welcome :)

ginkgm:
Hi @sdrsdr ,

I'm afraid due to the random SW delay caused by higher level interrupts, any way of adding SW timestamp will not be as acurate as the HW solution - calculate the time by the point number read from the queue.

Seems that you have tried to add a timestamp in the ISR? Have you tested your project with Wi-Fi on?

Michael

sdrsdr:
There seems to be a lot of "problems" in the I2S DMA code   like for some reason I2S_IN_DONE_INT is not monitored but just IN_SUC_EOF and some handled descriptors are lost plus data consumers don't have any chance handling the "ready" data untill all buffers are filled by DMA

ginkgm:
Hi @sdrsdr ,

Yes, that's why I think we cannot solve the issue from the SW side. The HW will not immediately store the data, and the driver use a less-frequenty interrupt to handle the finished data. Even if you put the code in to the ISR, I can still see there are some unexpected reason for the SW timestamp to be in accurate.

Michael

sdrsdr:
Hi @ginkgm, 
SW timestams accuracy in ISR is a matter of IRQ priorities. I don't know any HW way to take a timestamp.  lldesc_t sctruct is all the DMA care for and I don't see any field about timestams

ginkgm:
@sdrsdr 
If you can ensure the I2S queue is never full, then each data point should be sampled at a fixed interval, this is guaranteed by the timer.

This can give you a good relative time among samples.

sdrsdr:
[...]

PR #7353, opened 2021-08-02
Add timstamps to i2s DMA events via sdkconfig (IDFGH-5633)

sdrsdr:
As per https://github.com/espressif/esp-idf/issues/7339:

I'm in a project where ADC via I2S DMA is used to track an analog signal. The signal must be correlated to other events. As you an imagine the timing is important and with tight limits. Using the I2S driver's capability to send events via queue is almost perfect with the only downside is that the events might get variable time to get delivered. This can be easily solved by punting a timestamp from a call to esp_timer_get_time() in the queue event. The only problem I see is keeping the code compatible with older users of the event queue. To keep it clean this is  made selectable from the configuration system with default to non-timestamped events.
Any suggestions/guides are welcome :)

### DIFF
--- a/components/driver/Kconfig
+++ b/components/driver/Kconfig
@@ -170,4 +170,17 @@ menu "Driver configurations"
 
     endmenu # GPIO Configuration
 
+    menu "I2S Configuration"
+        visible if IDF_TARGET_ESP32
+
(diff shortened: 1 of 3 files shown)


sdrsdr:
as per https://github.com/espressif/esp-idf/issues/7339#issuecomment-894058451  this is now obsolete

Issue #7374, opened 2021-08-05
TWAI alert for RX (like UART_DATA) (IDFGH-5652)


crackwitz:
Use case: I need to send and receive CAN frames. I use TX_SUCCESS to confirm that a frame was sent. My task tries `twai_receive()` and `twai_read_alerts()` in a loop (`twai_transmit` asynchronously from elsewhere). Currently I have to either call both with 0 timeout, which makes the idle task's watchdog angry, or I give one of them a positive timeout... which delays the other call: either I get TX_SUCCESS 10 ms later than necessary or I may *lose* CAN frames because I'm spending 10 ms waiting for alerts.

I'd rather not have to use separate tasks for this.

I think a TWAI alert would be fine that indicates messages waiting in the RX queue and to try reading them with `twai_receive` (with 0 timeout repeatedly until the queue is empty).

The UART peripheral also has "alerts", but calls them "events", and it has `UART_DATA`, which indicates data waiting in the input queue. The idea applies to multiple peripherals but it has different names and implementations.

I've considered adding such a TWAI alert into esp-idf myself but it would have to be upstreamed because I can't justify the effort of maintaining a patch beside esp-idf.

If this idea or an alternative is well received, I might look into making a pull request.

joncmaloney:
We also have this issue with TWAI interface. We have ended up using 2 threads one for TX and another for RX. It would be much better if there was an implementation that allowed for operations with only one thread. 

One solution could be that a tx_queue and rx_queue are added to twai_general_config_t structure and passed into the twai_driver_install function. This way the user could add both these queues into a queue set and then wait for either a tx frame or an rx frame by blocking on the queue set.

dev-techshlok:
Hello @crackwitz 

Amazing contribution. 

How can we use this in the main code, do we need to make a function and pass that to interrupt, can you give a small example, please.

Thank you.

crackwitz:
- Documentation of alerts: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html#alerts

- Docs for `twai_read_alerts`: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html#_CPPv416twai_read_alertsP8uint32_t10TickType_t

- Example for alerts: https://github.com/espressif/esp-idf/tree/master/examples/peripherals/twai/twai_alert_and_recovery

PR #7386, opened 2021-08-06
twai: TWAI_ALERT_RX_DATA indicates that frames were received (IDFGH-5664)

crackwitz:
related to #7374

the `#define` has to be inserted among the first few entries because `ALERT_LOG_LEVEL_WARNING`/`ALERT_LOG_LEVEL_ERROR` go by value. that necessarily requires all following `#define`s to be rewritten

this patch was made based on esp-idf v4.2.1, then applied to current master branch. I'll have a chance to verify next week.

### DIFF
--- a/components/driver/include/driver/twai.h
+++ b/components/driver/include/driver/twai.h
@@ -47,23 +47,24 @@ extern "C" {
  */
 #define TWAI_ALERT_TX_IDLE                  0x00000001  /**< Alert(1): No more messages to transmit */
 #define TWAI_ALERT_TX_SUCCESS               0x00000002  /**< Alert(2): The previous transmission was successful */
-#define TWAI_ALERT_BELOW_ERR_WARN           0x00000004  /**< Alert(4): Both error counters have dropped below error warning limit */
-#define TWAI_ALERT_ERR_ACTIVE               0x00000008  /**< Alert(8): TWAI controller has become error active */
-#define TWAI_ALERT_RECOVERY_IN_PROGRESS     0x00000010  /**< Alert(16): TWAI controller is undergoing bus recovery */
-#define TWAI_ALERT_BUS_RECOVERED            0x00000020  /**< Alert(32): TWAI controller has successfully completed bus recovery */
-#define TWAI_ALERT_ARB_LOST                 0x00000040  /**< Alert(64): The previous transmission lost arbitration */
-#define TWAI_ALERT_ABOVE_ERR_WARN           0x00000080  /**< Alert(128): One of the error counters have exceeded the error warning limit */
-#define TWAI_ALERT_BUS_ERROR                0x00000100  /**< Alert(256): A (Bit, Stuff, CRC, Form, ACK) error has occurred on the bus */
-#define TWAI_ALERT_TX_FAILED                0x00000200  /**< Alert(512): The previous transmission has failed (for single shot transmission) */
-#define TWAI_ALERT_RX_QUEUE_FULL            0x00000400  /**< Alert(1024): The RX queue is full causing a frame to be lost */
-#define TWAI_ALERT_ERR_PASS                 0x00000800  /**< Alert(2048): TWAI controller has become error passive */
-#define TWAI_ALERT_BUS_OFF                  0x00001000  /**< Alert(4096): Bus-off condition occurred. TWAI controller can no longer influence bus */
-#define TWAI_ALERT_RX_FIFO_OVERRUN          0x00002000  /**< Alert(8192): An RX FIFO overrun has occurred */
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

Dazza0:
@crackwitz The changes look good. It's just missing documentation and support for when `SOC_TWAI_SUPPORTS_RX_STATUS` is not supported. I'll push an merge request with those extra changes shortly

crackwitz:
Interesting. I was unaware of that value.

igrr:
I'm curious, is there any reason to add TWAI_ALERT_RX_DATA with mask 0x4, redefining all subsequent values? Would it be safer to add it at the end (0x00020000)? This way there is lower chance of introducing a breaking change for existing projects, especially if any pre-built libraries are involved.

crackwitz:
yes there's a reason. look at how `ALERT_LOG_LEVEL_WARNING` is used

igrr:
There seems to be only one usage of `alert_code >= ALERT_LOG_LEVEL_WARNING`, so how about we refactor it roughly as follows:

```c++
/* instead of "define ALERT_LOG_LEVEL_WARNING ..." */
static inline bool twai_alert_is_warning(uint32_t alert_code) {
  return alert_code >= TWAI_ALERT_ARB_LOST && alert_code != TWAI_ALERT_RX_DATA;
}

/* where ALERT_LOG_LEVEL_WARNING was used */
else if (twai_alert_is_warning(alert_code)) {
   /* ... */
}
```

crackwitz:
I find that to be an entirely unjustifiable special case.

continue that line of thought. you'll arrive at a *set* of constants of arbitrary order, and your function has to check *many* of them.

you should find that complexity unbearable.

an ordered list of values is clean and expected.

if you want to improve something about that code, convert the hex constants into `(1 << $bitvalue)` expressions:
```cpp
#define TWAI_ALERT_TX_IDLE                  (1 <<  0)  /**< Alert(1): No more messages to transmit */
#define TWAI_ALERT_TX_SUCCESS               (1 <<  1)  /**< Alert(2): The previous transmission was successful */
#define TWAI_ALERT_RX_DATA                  (1 <<  2)  /**< Alert(4): A frame has been received and added to the RX queue */
#define TWAI_ALERT_BELOW_ERR_WARN           (1 <<  3)  /**< Alert(8): Both error counters have dropped below error warning limit */
#define TWAI_ALERT_ERR_ACTIVE               (1 <<  4)  /**< Alert(16): TWAI controller has become error active */
#define TWAI_ALERT_RECOVERY_IN_PROGRESS     (1 <<  5)  /**< Alert(32): TWAI controller is undergoing bus recovery */
#define TWAI_ALERT_BUS_RECOVERED            (1 <<  6)  /**< Alert(64): TWAI controller has successfully completed bus recovery */
#define TWAI_ALERT_ARB_LOST                 (1 <<  7)  /**< Alert(128): The previous transmission lost arbitration */
#define TWAI_ALERT_ABOVE_ERR_WARN           (1 <<  8)  /**< Alert(256): One of the error counters have exceeded the error warning limit */
#define TWAI_ALERT_BUS_ERROR                (1 <<  9)  /**< Alert(512): A (Bit, Stuff, CRC, Form, ACK) error has occurred on the bus */
#define TWAI_ALERT_TX_FAILED                (1 << 10)  /**< Alert(1024): The previous transmission has failed (for single shot transmission) */
#define TWAI_ALERT_RX_QUEUE_FULL            (1 << 11)  /**< Alert(2048): The RX queue is full causing a frame to be lost */
#define TWAI_ALERT_ERR_PASS                 (1 << 12)  /**< Alert(4096): TWAI controller has become error passive */
#define TWAI_ALERT_BUS_OFF                  (1 << 13)  /**< Alert(8192): Bus-off condition occurred. TWAI controller can no longer influence bus */
#define TWAI_ALERT_RX_FIFO_OVERRUN          (1 << 14)  /**< Alert(16384): An RX FIFO overrun has occurred */
#define TWAI_ALERT_TX_RETRIED               (1 << 15)  /**< Alert(32768): An message transmission was cancelled and retried due to an errata workaround */
#define TWAI_ALERT_PERIPH_RESET             (1 << 16)  /**< Alert(65536): The TWAI controller was reset */
#define TWAI_ALERT_ALL                      ((1 << 17)-1)  /**< Bit mask to enable all alerts during configuration */
#define TWAI_ALERT_NONE                     0x00000000  /**< Bit mask to disable all alerts during configuration */
#define TWAI_ALERT_AND_LOG                  (1 << 17)  /**< Bit mask to enable alerts to also be logged when they occur. Note that logging from the ISR is disabled if CONFIG_TWAI_ISR_IN_IRAM is enabled (see docs). */
```

crackwitz:
This is great! Thanks!

Issue #7047, opened 2021-05-19
PPPoS unable to set authentication,will always fail. (IDFGH-5284)


persan666:
## Environment

- Development Kit:      [none]
- Module or chip used:  [ESP32-WROVER-B]
- IDF version (run ``git describe --tags`` to find it):
    // v4.3-beta3-195-g6be10fab0
- Build System:         [CMake]
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it):
    // xtensa-esp32-elf-gcc (crosstool-NG esp-2020r3) 8.4.0
- Operating System:     [Windows]
- (Windows only) environment type: [ESP Command Prompt].
- Using an IDE?: [Yes (Eclipse)]
- Power Supply:         [Battery]

## Problem Description

Unable to set authentication for PPPoS.
The call to 'esp_netif_ppp_set_auth' will always fail and return invalid parameter even though the parameters are correct. 
The evaluation of macro '_IS_NETIF_POINT2POINT_TYPE' evaluates to true if the parameters are correct.

This is the implementation which not work:
```cpp
esp_err_t esp_netif_ppp_set_auth(esp_netif_t *netif, esp_netif_auth_type_t authtype, const char *user, const char *passwd)
{
    if (_IS_NETIF_POINT2POINT_TYPE(netif, PPP_LWIP_NETIF)) {
        return ESP_ERR_ESP_NETIF_INVALID_PARAMS;
    }
#if PPP_AUTH_SUPPORT
    lwip_peer2peer_ctx_t *ppp_ctx = (lwip_peer2peer_ctx_t *)netif->related_data;
    assert(ppp_ctx->base.netif_type == PPP_LWIP_NETIF);
    pppapi_set_auth(ppp_ctx->ppp, authtype, user, passwd);
#else
    ESP_LOGE(TAG, "%s failed: No authorisation enabled in menuconfig", __func__);
    return ESP_ERR_ESP_NETIF_IF_NOT_READY;
#endif
    return ESP_OK;
}
```
And should be changed to something like this (Notice the '!' before the macro '_IS_NETIF_POINT2POINT_TYPE'):
```cpp
esp_err_t esp_netif_ppp_set_auth(esp_netif_t *netif, esp_netif_auth_type_t authtype, const char *user, const char *passwd)
{
    if (!_IS_NETIF_POINT2POINT_TYPE(netif, PPP_LWIP_NETIF)) {
[...]

PR #7397, opened 2021-08-10
Pinging specific iface does not work: Fix broken parentheses in setsockopt call (IDFGH-5676)

ivmarkov:
The call to `setsockopt` has the parentheses in a wrong way, which causes the whole functionality of pinging a specific netif instance to fail.

### DIFF
--- a/components/lwip/apps/ping/ping_sock.c
+++ b/components/lwip/apps/ping/ping_sock.c
@@ -274,7 +274,7 @@ esp_err_t esp_ping_new_session(const esp_ping_config_t *config, const esp_ping_c
         if(netif_index_to_name(config->interface, iface.ifr_name) == NULL) {
           goto err;
         }
-        if(setsockopt(ep->sock, SOL_SOCKET, SO_BINDTODEVICE, &iface, sizeof(iface) !=0)) {
+        if(setsockopt(ep->sock, SOL_SOCKET, SO_BINDTODEVICE, &iface, sizeof(iface)) != 0) {
           goto err;
         }
     }


Alvin1Zhang:
Thanks for your contribution.

suda-morris (approved):

AxelLin:
Just remind that v4.3 also needs the fix.

AxelLin:
Hi @igrr @Alvin1Zhang 

This one is a classic case of one-line-fix which fixes a real bug.
I'm wondering if espressif has some standard process to speed up merging such fix.

I asked this because it may take very long time to apply to impacted branches.

Below is another example of such one-line-fix which fixes a real bug.
https://github.com/espressif/esp-idf/issues/7047 (And people are waiting, e.g. https://github.com/espressif/esp-idf/issues/7047#issuecomment-862181856 )

19 May 2021  Issue reported
7 Jun  2021  Commit to master
30 Jun 2021  Commit to release/v4.3
And sync to github time is slightly longer than the commited time.

PS. I'm not asking to take special care of this PR. (@suda-morris  did very quick response on this PR)
I'm asking if you have some standard process to speed up merging such fix.

igrr:
@AxelLin Thank you for the suggestion. There is no existing special process for one-liner fixes compared to all other PRs. The basic process is: 1) the PR is assigned to a developer, 2) if the PR doesn't need additional work, the developer will create an internal merge request, adding release notes and any required tests, 3) code owners review the merge request, 4) once the merge request is approved it can be merged, 5) backport merge requests are created and handled in a similar way. Indeed PR handling gets stuck at different points, for example this PR is stuck at step 2, since the developer who is assigned happens to be on leave now. We are looking at options to automate more steps in this process to make the handling of PRs more efficient.

Regarding this specific bug, I don't necessarily agree with you that there is a need to speed up merging into release branches, unless the PR author specifically requests the bug to be fixed on a particular release. For bugs that result in random failures or unreliable behavior, we do need to backport them quickly, since users might encounter this issue without even changing much in their applications that are already deployed. On the other hand, for bugs that result in a certain use case being completely broken, we could argue that if noone has reported such issue for a release branch, the feature (or use case) is simply not used. Of course, we do need to fix the issue in release branches eventually. But due to lower impact, I would classify the urgency of backporting as "normal" rather than "high".

AxelLin:
well, this is indeed not a regression because it actually never work.
However the API is there and if you call ping with specific interface
it will run into Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Alvin1Zhang:
Changes have been merged with https://github.com/espressif/esp-idf/commit/ee531c748a83212312eba7390534a70fe6e88445, thanks for your contribution again.

AxelLin:
Would you apply this fix for v4.3? (Unfortunately, v4.3.1 does not include this trivial fix).

BaaBaaGoat:
just got fucked by that .V4.3 latest stable.

Issue #7418, opened 2021-08-14
Support client key passwords for HTTPS connections (IDFGH-5698)


joncmaloney:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

Please give as many details as you can. Include suggestions for useful APIs or interfaces if relevant.

**Additional context**

joncmaloney:
I don't know how to add a key into an esp32 bin image securely. We have a requirement that all secrets are encrypted at rest and in transit. If the key is not password protected it is in plain text and anyone with access to the bin image is able to view the key. Therefore the key is not encrypted at rest. If a build server was to secure the image by encrypting it needs to be decrypted to be transported via UART bootloader and hence not encrypted in transit. This is unless we use host generated key (not recommended). Further if host generated keys are use the keys need to be transmitted to ESP32 in plain text on first program. This is also doesn't meet our security requirements. 

We are currently using password protected keys then manually removing the password from these key using the mbedtls libraries.

negativekelvin:
How do you protect the passphrase?

joncmaloney:
It's not going to be cryptography secured. It can only be obfuscated.

PR #7420, opened 2021-08-14
Add support for client key password for HTTPS connections (IDFGH-5700)

joncmaloney:
I've tested password protected PEM file format keys that use RSA. I've also tested the inverse case using a non password protected key and this still works. Provided the element client_key_password is null or client_key_password_len is 0.


joncmaloney:
resolves #7418

Alvin1Zhang:
Thanks for your contribution.

components/esp_http_client/esp_http_client.c
```diff
@@ -632,6 +632,12 @@ esp_http_client_handle_t esp_http_client_init(const esp_http_client_config_t *co
         }
     }
 
+    if (config->client_key_password) {
+        if(config->client_key_password_len > 0){
```
AshUK, line 636: `config->client_key_password & config->client_key_password_len > 0`

Could be a single if statement.

components/esp_http_client/include/esp_http_client.h
```diff
@@ -113,6 +113,8 @@ typedef struct {
     size_t                      client_cert_len;     /*!< Length of the buffer pointed to by client_cert_pem. May be 0 for null-terminated pem */
     const char                  *client_key_pem;     /*!< SSL client key, PEM format as string, if the server requires to verify client */
     size_t                      client_key_len;      /*!< Length of the buffer pointed to by client_key_pem. May be 0 for null-terminated pem */
+    const char                  *client_key_password;      /*!< Client key decryption password string */
+    int                         client_key_password_len;   /*!< String length of the password pointed to by clientkey_password */
```
AshUK, line 117: size_t

AdityaHPatwardhan:
@AshUK Thank for the review. @joncmaloney can you please update the changes. I think that should be enough to merge this PR.

joncmaloney:
@AshUK Thanks for the feedback. I've updated the changes suggested and rebased to espressif/esp-idf:master

AxelLin:
@AdityaHPatwardhan 
Will this PR apply to v4.3? (It's a trivial change, the low level API already supports client key password)

AdityaHPatwardhan:
Hi @AxelLin, so if it is required  we may apply this change to `v4.3` as well.

PR #7436, opened 2021-08-18
LWIP CMake missing lowpan6 ble source files (IDFGH-5731)

marcusbirkin:
LWIP CMake missing lowpan6 ble source files


Alvin1Zhang:
Thanks for your contributions.

david-cermak:
@marcusbirkin Could you please elaborate on the usecase? Why do you need those file, what you're trying to achieve.
In general, we try to keep the lwip configuration simple in IDF and limited to the use-cases which are more or less tested. Would it make sense to add an example/test or some docs to this PR?

marcusbirkin:
There are two reasons for these files:
"lwip/src/netif/lowpan6.c" is already included, but will not link unless "lwip/src/netif/lowpan6_common.c" is included in the the makefile.

"lwip/src/netif/lowpan6_ble.c" is required for lowpan6 over bluetooth low engery (aka rfc7668), the ESP32 being the ideal target for this protocol. If lowpan6 support is included, which required additional hardware, then lowpan6_ble should also be included.

david-cermak:
Thanks for explaining.
Wouldn't it make sense then to remove the `lwip/src/netif/lowpan6.c` from the list

And add the related files only under a specific configuration:

```cmake
if(CONFIG_LWIP_LOWPAN6)
    list(APPEND srcs 
    "lwip/src/netif/lowpan6_common.c"
    "lwip/src/netif/lowpan6.c"
    "lwip/src/netif/lowpan6_ble.c"
     )
endif()
```

I can see some benefits in this approach as it
* would help with compilation time for most configs
* adds the files only if they are needed
* introduces the new config option, which would have to be briefly described in Kconfig and thus a very simple documentation be generated about this option.

marcusbirkin:
Good idea, it also allows for exposing the debug options.
I've updated the pull request

components/lwip/port/esp32/include/lwipopts.h
```diff
@@ -511,6 +511,49 @@
...
+*/
+
+#ifdef CONFIG_LWIP_LOWPAN6_SUPPORT
+
+/**
+ * LOWPAN6_DEBUG: Enable debugging for 6LoWPAN.
+ */
+#define LWIP_LOWPAN6_DEBUG_ON           CONFIG_LWIP_LOWPAN6_DEBUG_ON
+
+#if LWIP_LOWPAN6_DEBUG_ON
```
david-cermak, line 529:
```suggestion
#ifdef CONFIG_LWIP_LOWPAN6_DEBUG_ON
```

Nitpick: Just to shorten, simplify the header.
This works in IDF (we don't compile with `-Wundef`), but might cause trouble when build using lwip flags, where this could be treated as error. (Note that Kconfig macros are either defined as `1` (`yes`) or undefined (`'no`))

The same comment for the below config options.

david-cermak (approved):
Thanks for the quick update! LGTM.

Just one minor comment.

And a suggestion about the latter debug options of:
* `LWIP_LOWPAN6_DECOMPRESSION_DEBUG`
* `LWIP_RFC7668_IP_UNCOMPRESSED_DEBUG`
Do you think we can simply hard-code these two options to both enabled if `LWIP_LOWPAN6_DEBUG=y` and both disabled otherwise?
I'm asking to keep the Kconfig(s) compact, especially for experimental features, somewhat like what we did for [PPP options](https://github.com/espressif/esp-idf/blob/master/components/lwip/port/esp32/include/lwipopts.h#L740-L744) -- just enabled the most common debug options when `DEBUG_ON=y`

marcusbirkin:
> Thanks for the quick update! LGTM.
> 
> Just one minor comment.
> 
> And a suggestion about the latter debug options of:
> 
> * `LWIP_LOWPAN6_DECOMPRESSION_DEBUG`
> * `LWIP_RFC7668_IP_UNCOMPRESSED_DEBUG`
>   Do you think we can simply hard-code these two options to both enabled if `LWIP_LOWPAN6_DEBUG=y` and both disabled otherwise?
>   I'm asking to keep the Kconfig(s) compact, especially for experimental features, somewhat like what we did for [PPP options](https://github.com/espressif/esp-idf/blob/master/components/lwip/port/esp32/include/lwipopts.h#L740-L744) -- just enabled the most common debug options when `DEBUG_ON=y`

The LWIP_LOWPAN6_DECOMPRESSION_DEBUG & LWIP_RFC7668_IP_UNCOMPRESSED_DEBUG options are very useful. In fact I was using them last week.
But they are VERY verbose, having them enabled when you simply need the minimal output from LWIP_LOWPAN6_DEBUG would get distracting (and annoying!).

components/lwip/port/esp32/include/lwipopts.h
marcusbirkin: I was actually just directly coping the style of SLIPIF above, thinking it was needed in some unknown way.
However after looking over the rest of the header, doing it with ifdef is much neater.

david-cermak: Yes, we should cleanup the `lwipiopt.h` and use one common way of converting the Kconfig options to lwip defines. Thanks for the update.

> 
> The LWIP_LOWPAN6_DECOMPRESSION_DEBUG & LWIP_RFC7668_IP_UNCOMPRESSED_DEBUG options are very useful. In fact I was using them last week.
> But they are VERY verbose, having them enabled when you simply need the minimal output from LWIP_LOWPAN6_DEBUG would get distracting (and annoying!).

Okay, then it makes sense to keep them configurable separately.

---

Thanks again for your contribution. will move this PR to the internal round our reviews and testing, but please be patient, this will take some time.

david-cermak:
Hi @marcusbirkin,

just a little correction, we have cherry-picked only a portion of this PR (removal of unnecessary lowpan6 sources from the default build set).
We were having trouble creating a stable PAN IPv6 example with the internal BLE stack (had a sort of working example with an external BT stack only), so decided to accept this PR partially, at first.

Please feel free to reopen the PR, plus it would be helpful if you can add a simple example of using lwip's lowpan6-ble interface.

marcusbirkin:
@david-cermak There is no option for me to reopen...
As far as examples go, I have some very propriatory ones. Time permitting I try to adapt some for a more public audience.

david-cermak:
Have reopened, it makes sense to keep this open until it's 100% accepted.

> Time permitting I try to adapt some for a more public audience.

Thank you!

peturdainn:
Interested in seeing progress on this topic: I'm considering CoAP over 6LowPAN over BLE on an ESP32-S3 but by the looks of it this isn't mature/stable enough for use in a new development?

PR #7506, opened 2021-09-01
Fix bootloader size message in build log (IDFGH-5798)

boarchuz:
Fixes bootloader size message in the build log by using maximum size (instead of current binary size) to calculate free space percentage.

Before:
`Bootloader binary size 0x5b30 bytes. 0x64d0 bytes (111%) free.`
After:
`Bootloader binary size 0x5b30 bytes. 0x64d0 bytes (53%) free.`


Alvin1Zhang:
Thanks for your contribution.

KonstantinKondrashov (approved):

PR #7517, opened 2021-09-04
Fix compiler warnings with silent assertions

boarchuz:
Tidy up a couple unused variable warnings when compiling with CONFIG_OPTIMIZATION_ASSERTIONS_SILENT


0xjakob (approved):
Looks good! We'll mangle this through our internal CI and in a few weeks you should see your change showing up on master.

KaeLL:
>  in a few weeks

☹️

0xjakob:
> > in a few weeks
> 
> frowning_face

I personally would also really like it to be faster, but that's the current state of our internal CI. Everything needs to pass a review and all checks. Only once this is done, the code will be pushed. Sorry. We're working on it to make it faster (me personally as well, actually).

PR #7518, opened 2021-09-04
Fix bootloader output with silent assertions (IDFGH-5811)

boarchuz:
The bootloader assert function (__assert_func in bootloader_panic.c) is still linked despite CONFIG_OPTIMIZATION_ASSERTIONS_SILENT set. If invoked, the following output is printed:
`Assert failed in <null>, <null>:0 (<null>)`

I have changed this to `abort` for the bootloader, and confirmed that the function and string are no longer in the bootloader binary. 

The bootloader assert function would otherwise loop (until WDT reset) here:
https://github.com/espressif/esp-idf/blob/d5f58ab13551cd883e8d8478ba367b6e4543ffec/components/bootloader_support/src/bootloader_panic.c#L13-L18
Is abort ok or should that behaviour be preserved by simply omitting the esp_rom_printf?

### DIFF
--- a/components/newlib/platform_include/assert.h
+++ b/components/newlib/platform_include/assert.h
@@ -43,7 +43,11 @@
 
 #elif CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT
 
+#ifdef BOOTLOADER_BUILD
+#define assert(__e) (__builtin_expect(!!(__e), 1) ? (void)0 : abort())
+#else
 #define assert(__e) (__builtin_expect(!!(__e), 1) ? (void)0 : __assert_func(NULL, 0, NULL, NULL))
+#endif
 
 #else // !CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_SILENT
 


Alvin1Zhang:
Thanks for your contribution.

ESP-Marius:
Sorry for the late reply @boarchuz 

Good catch! I agree that we should do something to fix this. 

To me it might make more sense to keep the behavior the same, i.e. just loop. What do you think about simply ifdefing the print away in `bootloader_panic.c`? That way we dont need to introduce any checks for bootloader build in `assert.h` either.

@boarchuz if you dont feel like making changes to the PR, perfectly understandable with the time it took us to reply..., then please let me know and we can fix this internally ourselves.

CC @o-marshmallow  , what do you think?

boarchuz:
Hi @ESP-Marius,
Please feel free to make the necessary changes internally, thankyou.

o-marshmallow:
@ESP-Marius I would also go for #ifdef around the `esp_rom_printf` instruction

PR #7526, opened 2021-09-05
Added missing C++ guard/wrapper (IDFGH-5821)

0xFEEDC0DE64:
When the `dhcpserver.h` is included in a C++ source, it results in a linker error (as C symbols are named differently). To fix this, I added the same C++ guards as found in many other sources.


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/f2d32d5c0aaa17a5865dda10fae313bcba29f9e1.

PR #7530, opened 2021-09-07
Kconfig: Support M5Stack W5500 PLC base (IDFGH-5826)

ajlennon:
The W5500 pinout for the PLC base is:

MOSI	GPIO23
MISO	GPIO19
CLK	GPIO18
CS	GPIO26
RST	GPIO13
INTn	GPIO34

The Kconfig for the project had a max value of GPIO33
so I have changed this to GPIO34 and tested that with
this set the ethernet example works (picks up an IP
address)

@see: https://docs.m5stack.com/en/base/lan_base

Signed-off-by: Alex J Lennon <ajlennon@dynamicdevices.co.uk>

### DIFF
--- a/examples/ethernet/basic/main/Kconfig.projbuild
+++ b/examples/ethernet/basic/main/Kconfig.projbuild
@@ -6,7 +6,7 @@ menu "Example Configuration"
 
     config EXAMPLE_GPIO_RANGE_MAX
         int
-        default 33 if IDF_TARGET_ESP32
+        default 34 if IDF_TARGET_ESP32
         default 46 if IDF_TARGET_ESP32S2
         default 19 if IDF_TARGET_ESP32C3
 


kostaond:
@ajlennon thank you for you contribution. The reason of defining the ESP32 GPIO range in Ethernet related examples to 0-33 is the GPIOs 34-39 are inputs only, see section 4.10 IO_MUX Pad list in ESP32 TMR. It seemed to be safer approach since the same range definition is used for inputs and outputs pins for simplification in Ethernet examples. The idea was to prevent beginner user's headache if they used input only pins for outputs (some users sometimes tend to not to look properly into TMR). However, it causes troubles in cases as yours.

The correct full GPIO range should be then 0-39. Note that we are planning to expose soc_caps.h flags in kconfigs which will resolve this issue. Please resolve it locally for now and wait for the final solution.

ajlennon:
OK I see - maybe you can ping me here when you have the update done and I'll update my PR? Thanks

Alvin1Zhang:
Thanks for your contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/c2297669002622a3bdce4b8afe96c02af46d3744.

kostaond:
@ajlennon thanks for pointing us to the issue. The final solution has been chosen to be more complex but your inputs helped us to identify that there was a problem.

PR #7536, opened 2021-09-08
tcp_transport: Fix NULL pointer dereference in esp_transport_esp_tls_… (IDFGH-5834)

AxelLin:
…create

Add missing NULL checking to prevent NULL pointer dereference if calloc failed.


david-cermak (approved):
Thanks for the fix!

PR #7540, opened 2021-09-09
Add a sentry.io JSON mode to espcoredump.py (IDFGH-5838)

ryankurte:
Hi there,

Adds a `json_corefile` mode to `espcoredump.py` that generates a [sentry.io](https://sentry.io) compatible event object for those that like to collect [coredumps](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/core_dump.html). Invoked with something like:
```
espcoredump.py json_corefile -t raw -c COREDUMP.bin APP.elf > sentry-event.json
```
Also supports `--environment` to set the environment, and `--tags KEY:VALUE` for whatever else you may need.

Upload with:
```
curl -X POST -H 'Content-Type: application/json' -H "X-Sentry-Auth: Sentry sentry_version=7, sentry_key=$SENTRY_KEY, sentry_client=raven-native/1.0" https://sentry.io/api/$SENTRY_ID/store/ --data @./sentry-event.json
```
Where `SENTRY_KEY` and `SENTRY_ID` are the relevant portions of your [DSN](https://docs.sentry.io/product/sentry-basics/dsn-explainer/) (see `Settings -> Select Org -> Project -> Select Project -> Client Keys (DSN)`)

It's certainly not perfect, but, hopefully a useful start.

### Screenshots

image

image

### DIFF
--- a/components/espcoredump/espcoredump.py
+++ b/components/espcoredump/espcoredump.py
@@ -10,6 +10,8 @@ import sys
 from shutil import copyfile
 import json
 import re
+import uuid
+import datetime
 
 from construct import GreedyRange, Int32ul, Struct
 from corefile import RISCV_TARGETS, SUPPORTED_TARGETS, XTENSA_TARGETS, __version__, xtensa
@@ -125,6 +127,8 @@ def dbg_corefile():  # type: () -> Optional[list[str]]
     """
     exe_elf = ESPCoreDumpElfFile(args.prog)
     core_elf_path, target, temp_files = get_core_dump_elf(e_machine=exe_elf.e_machine)
+
+    print("Core: %s" % core_elf_path)
 
     rom_elf_path = get_rom_elf_path(target)
     rom_sym_cmd = load_aux_elf(rom_elf_path)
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

igrr:
@ryankurte thanks, this integration looks very neat! While there is some overlap between this and Espressif's new [esp-insights](https://github.com/espressif/esp-insights) service, i see no harm in adding this functionality. JSON output might actually be quite useful for other tools, e.g. for IDE integration (cc @kolipakakondal).

One concern i have though is about maintaining this feature. Since we likely won't be using sentry.io ourselves, this feature might get broken either due to some changes in espcoredump.py, or due to some changes in sentry.io. Do you have any recommendation for testing this feature? Currently we run some tests for `info_corefile` command in CI — mainly comparing its output to "expected" one — to ensure that this command doesn't get broken.
Perhaps there is some schema for this JSON format which we can validate the output against? We would prefer not to have to connect to real sentry.io service from our CI pipelines — such dependencies have proven to be flaky in the past.

Another question, how tolerant do you think sentry.io is to adding new fields into this JSON output? If we start using this JSON output for some other purpose (e.g. integration with IDEs), will it be a problem if we add more information to this JSON? Or would we have to maintain different kinds of JSON output for sentry.io and other consumers?

Finally, do you think we could do something to allow such features to be developed and maintained outside of ESP-IDF? E.g. converting espcoredump to a Python package available from PyPI, or something else?

ryankurte:
thanks for the response! esp-insights looks neat, will have to give it a try too.

> Finally, do you think we could do something to allow such features to be developed and maintained outside of ESP-IDF? E.g. converting espcoredump to a Python package available from PyPI, or something else?

addressing this first because i think it relates to all your questions/concerns. initially i tried to get this running as a standalone python script but, i wasn't able to get the dependencies to resolve outside of the ESP-IDF (i am seemingly incapable of python packaging 😂) , so, we ended up here.

i think the best way to address all your questions would be to split `espcoredump.py` into a CLI and a library and publish both to pypy, so that specific platform support (like this) could basically just be a wrapper that calls the library to retrieve the related objects, then restructures that to match whatever API.


> ... Do you have any recommendation for testing this feature? ...

for sentry specific testing, they publish their [types](https://crates.io/crates/sentry-types) as a crate. so, it should be possible to bash together a validator tool, but, i don't think there's anything right out-of-the-box. it seems like it'd be a bit substantial for inclusion here / would seem to make more sense as part of a standalone `espcoredump-sentry`.

> Another question, how tolerant do you think sentry.io is to adding new fields into this JSON output?

i'm not really sure, though the API seems to surface parsing errors if you do have things it doesn't like. another approach could be to genericise the json schema / output here then to have another tool that maps from that to this, but i think the libraryified one would be neater.

igrr:
Alright, we'll try to move the CLI/library into a package on PyPI then!

alekseiapa:
@ryankurte The corefile and coredump scripts from idf have been separated into their own package.
Please find the links below:
[GitHub](https://github.com/espressif/esp-coredump) 
[PyPi](https://pypi.org/project/esp-coredump/)

PR #7547, opened 2021-09-10
Add support for KSZ8091 ethernet PHY (IDFGH-5845)

ma-lalonde:
This is a preamble to adding Energy Efficient Ethernet (EEE) and Wake-on-LAN (WoL) support for the chip.

Also, what would be the best way to introduce these functions? At this time, only the KSZ8091 supports them (I think). However, energy efficiency is becoming a bigger concern and more chips will implements these functions in the future.

The most obvious way that I can see is adding functions `advertise_eee_ability(bool enable)` and `enable_wol(bool enable, int phy_pin)` to structure `esp_eth_phy_t` (and other parameters if I realise they are necessary).

### DIFF
--- a/components/esp_eth/include/esp_eth_phy.h
+++ b/components/esp_eth/include/esp_eth_phy.h
@@ -1,16 +1,9 @@
-// Copyright 2019-2021 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+/*
+ * SPDX-FileCopyrightText: 2019-2021 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 #pragma once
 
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

ma-lalonde:
I simplified the implementation since the initialization of the KSZ8091 is identical to that of the KSZ8081. Further simplification to KSZ80xx is not (easily) possible due to the registers of the KSZ8041 being differently arranged.

As for the implementation, due to WoL and EEE still being very hardware specific (and EEE still being full of bugs), I will leave those out of the PHY driver and to be implemented by the application programmers.

Note that the KSZ8091 implements MMD registers that the KSZ8081 doesn't, so they can be differentiated at this level. Writing to MMD registers for WoL and EEE configuration implies a specific sequence of PHY register read/writes detailed in the datasheet. I think this sequence is too specific to be implemented at the driver level either and should be left to application developer.

TL;DR: Simplified KSZ8091 implementation and WoL and EEE will be left out to application developer.

components/esp_eth/include/esp_eth_phy.h
```diff
@@ -304,6 +297,20 @@ esp_eth_phy_t *esp_eth_phy_new_ksz8041(const eth_phy_config_t *config);
...
+*
+* @param[in] config: configuration of PHY
+*
+* @return
+*      - instance: create PHY instance successfully
+*      - NULL: create PHY instance failed because some error occurred
+*/
+static inline esp_eth_phy_t *esp_eth_phy_new_ksz8091(const eth_phy_config_t *config)
+{
+	return esp_eth_phy_new_ksz8081(config);
```
kostaond, line 311: It was found the `tab` is used instead of spaces during the internal review. Could you please fix it?

ma-lalonde: Done.

PR #7554, opened 2021-09-13
Added an Arduino variant of the generic_gpio example

Reiikz:
Apart from the fact the original code is meant for another development environment.
I found a bunch of implicit castings issues I don't know if it's a compiler issue what but I hacked it by adding the following cast `(gpio_num_t)` to the uint32_t values holding the gpio value.
Tested on my ESP32 WROVER and it works fine.


Alvin1Zhang:
Thanks for your contribution.

negativekelvin:
This should be in arduino-esp32 https://github.com/espressif/arduino-esp32/tree/master/libraries/ESP32/examples/GPIO

Reiikz:
> This should be in arduino-esp32 https://github.com/espressif/arduino-esp32/tree/master/libraries/ESP32/examples/GPIO

oops, my bad, sorry for the inconvenience.

PR #7567, opened 2021-09-15
ksz80xx_negotiate now check for phy register auto negotiation enabled (IDFGH-5868)

ma-lalonde:
ksz80xx_negotiate is called by esp_eth_start, but if the user wants to force 10/100 mbps or half/full duplex, it is unable to. This commit has ksz80xx_negotiate check if auto negotiation is enabled in the phy registers first, then proceeds to auto negitiate the link or cancels according to the PHY register value.


ma-lalonde:
Note that I have no way to test this fix on other PHY peripherals, so this PR may create a discrepancy between how different PHY work. Most of the code seems to be copy-pasted, so the risks related to uniformizing this behavior with other PHYs is minimal.

kostaond:
@ma-lalonde hi, just note that we already have internal task to make Ethernet speed or half/full duplex user configurable via esp_eth_ioctl for all supported PHYs. It is planned to be worked soon, please stand by.

Issue #7570, opened 2021-09-16
Ethernet basic example broke with IDF version 4.3.1


dabbinavo:
## Environment

- Development Kit: **none**
- Module or chip used: **ESP32-WROOM-32U**
- IDF version (run ``git describe --tags`` to find it): **v4.3.1**
- Build System: **CMake**
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): **(crosstool-NG esp-2021r1) 8.4.0**
- Operating System: **Windows**
- (Windows only) environment type: **ESP Command Prompt**
- Using an IDE?: **Yes - VSCode with Espressif IDF Plugin**
- Power Supply: **external 3.3V**

## Problem Description

Starting with ESP-IDF V4.3.1, the dhcp client in ethernet mode is not able to start.
It seems that the root cause for the issue is [this change to components/esp_eth/src/esp_eth.c](https://github.com/espressif/esp-idf/commit/03a6c4975c7fd7d0267f26b93f43b5b391ad2de4#diff-f70b7e7eb060881f32460033b9c62d65cf93a019645367af6745d17e45c62f43). When undoing this change, everything works fine again.

Investigations showed, that the error message `esp_netif_lwip: dhcp client start failed` is generated, because the `dhcp_start()` function is called before the [`ethernet_low_level_init()`](https://github.com/espressif/esp-idf/blob/v4.3.1/components/lwip/port/esp32/netif/ethernetif.c#L77) function for some reason ([which sets netif->mtu to 1500](https://github.com/espressif/esp-idf/blob/v4.3.1/components/lwip/port/esp32/netif/ethernetif.c#L83)). Therefore in `dhcp_start()`, the [check if the MTU is large enough](https://github.com/espressif/esp-lwip/blob/2195f7416fb3136831babf3e96c027a73075bd4f/src/core/ipv4/dhcp.c#L822) fails. At this point, `netif->mtu` is still unitialized and therefore `0`.

@suda-morris Do you have any idea why the changes in the above mentioned commit lead to that behaviour?

### Expected Behavior
DHCP client does start.

### Actual Behavior
DHCP client is unable to start: `esp_netif_lwip: dhcp client start failed`

### Steps to reproduce
1. Prepare code (see [Code to reproduce this issue](#Code-to-reproduce-this-issue))
1. Open the basic ethernet example (`/esp-idf/examples/ethernet/basic`)
1. Change the sdkconfig to match your hardware
1. Build, run and monitor the example

[...]

PR #7573, opened 2021-09-16
esp_eth: adapt ksz8041.c to changes from commit 03a6c49

dabbinavo:
Local tests showed, that this should fix bug #7570.
As I just applied the changes analogous to commit [03a6c4975](https://github.com/espressif/esp-idf/commit/03a6c4975c7fd7d0267f26b93f43b5b391ad2de4#diff-3d91d958e6443c2d3c1ad88c35f4506cda5cc2b7ed83608bda21f36a67136ade), someone who understands these changes should review this. Maybe @suda-morris?

### DIFF
--- a/components/esp_eth/src/esp_eth_phy_ksz8041.c
+++ b/components/esp_eth/src/esp_eth_phy_ksz8041.c
@@ -193,6 +193,8 @@ static esp_err_t ksz8041_negotiate(esp_eth_phy_t *phy)
 {
     phy_ksz8041_t *ksz8041 = __containerof(phy, phy_ksz8041_t, parent);
     esp_eth_mediator_t *eth = ksz8041->eth;
+    /* in case any link status has changed, let's assume we're in link down status */
+    ksz8041->link_status = ETH_LINK_DOWN;
     /* Restart auto negotiation */
     bmcr_reg_t bmcr = {
         .speed_select = 1,     /* 100Mbps */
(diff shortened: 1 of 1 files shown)


suda-morris (approved):

suda-morris:
Thank you for your reporting and fixing! I believe this is left over when we backporting some fixes to other release branches. We will apply your fix and sync it to GitHub ASAP.

dabbinavo:
Thank you for approving the changes! Please, can you explain shortly why the fix solves the issue that the phy init is called after the dhcp client gets started? I just copied your changes but don't really understand why they resolve the issue.

suda-morris:
Yes, sure. In esp-idf Ethernet driver, the PHY link up/down status is fetched by a "timer". The timer is started in "esp_eth_start", **after** we send out the "ETHERNET_EVENT_START" signal.
There's a event handler for "ETHERNET_EVENT_START", which is responsible for setup up network interface (e.g. add Ethernet interface to LwIP).
The link timer will wake up periodically (default 2s), reading the PHY register and check if the link status got changed. Fo example, if it detects link up event, it will send out "ETHERNET_EVENT_CONNECT" signal. Likewise, there's an event handler for that, which is responsible for starting the DHCP client.

Without your fix, the buggy driver check the link status in the negotiation function (and then send out "ETHERNET_EVENT_CONNECT" event). But at that time, the network interface has been established, because we haven't call esp_eth_start yet.

dabbinavo:
Ok. Thank's for that explanation. So the significant part of the fix is the removal of `<phy>_update_link_duplex_speed()` inside the `<phy>_negotiate()` function...

dabbinavo:
@suda-morris, who takes care of merging this pull-request? Github tells me that I am not authorized to merge this pull request..

PR #7637, opened 2021-10-03
update clk_flags requirement for i2c docs (IDFGH-5949)

waheedi:
As it seems, the latest version is requiring the clk_flags to have a value for the i2c driver to initialize, not sure if its intended to be like this or not honestly, but the docs are not inclined with code apparently, either the docs or the code needs to be updated, this is the shortest path :)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=06a03d1dcbbd8ab334d5374670f2073d92f9d70c

mahavirj:
Merged with https://github.com/espressif/esp-idf/commit/d5bfbfc095df29fca20effc844d18c684474e290

PR #7680, opened 2021-10-12
undefined reference to `bootloader_flash_update_size' (IDFGH-5995)

ESP32DE:
undefined reference to `bootloader_flash_update_size'
ESP32-S3 was missing bootloader_flash_update_size


Alvin1Zhang:
Thanks for your contribution.

PR #7690, opened 2021-10-14
Added missing includes in esp_netif_types.h (IDFGH-6002)

0xFEEDC0DE64:
I had issues when including esp_netif_types.h in my sources as some required macros and types were missing.

Using these added includes all compiles fine again.

### DIFF
--- a/components/esp_netif/include/esp_netif_types.h
+++ b/components/esp_netif/include/esp_netif_types.h
@@ -15,6 +15,9 @@
 #ifndef _ESP_NETIF_TYPES_H_
 #define _ESP_NETIF_TYPES_H_
 
+#include "esp_event_base.h"
+#include "esp_err.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif


components/esp_netif/include/esp_netif_types.h
```diff
@@ -15,6 +15,9 @@
 #ifndef _ESP_NETIF_TYPES_H_
 #define _ESP_NETIF_TYPES_H_
 
+#include "esp_event_base.h"
+#include "esp_err.h"
```
david-cermak, line 19:
```suggestion
#include <stdbool.h>
#include <stdint.h>
#include "esp_event_base.h"
#include "esp_err.h"
#include "esp_netif_ip_addr.h"
```

david-cermak (approved):
Thanks for your help in fixing IDF public headers!
One optional suggestion, as we need to know standard boolean and integer types as well as the IP structs. These amends could be applied internally, updating your commit.

LGTM!

PR #7696, opened 2021-10-14
Fetch headers in POST example for native requests (IDFGH-6004)

svenfuchs:
In the POST section of `http_native_request` headers are not fetched. thus, just commenting out the GET request example would result in a `status` and `content_length` of `0` being reported (quite confusing).

(Plus, it should log `HTTP POST Status` here, instead of `HTTP GET Status`.)


Alvin1Zhang:
Thanks for contribution again, changes merged with https://github.com/espressif/esp-idf/commit/37da873da9777d852850d7d53a099fd637b7e3b2.

PR #7705, opened 2021-10-15
Increased LAN8720 ETH-PHY reset assertion time (IDFGH-6018)

sauttefk:
Increase reset assertion time from 100µs (specified as minimum in the datasheet) to 150µs.
Some specimen of the LAN8720 need the reset signal asserted longer than 100µs to initialise properly. Otherwise they are in a zombie state where they are establishing and loosing an Ethernet link once in a seconds interval.


Alvin1Zhang:
Thanks for your contribution.

Issue #6510, opened 2021-02-06
Sharing SPI bus between TFT and sdcard results in error (IDFGH-4699)


jkent:
## Environment

- Development Kit:      none
- Module or chip used:  ESP32-WROVER
- IDF version (run ``git describe --tags`` to find it):
    v4.3-dev-2940-g6e776946d
- Build System:         idf.py
- Compiler version
    xtensa-esp32-elf-gcc (crosstool-NG esp-2020r3) 8.4.0
- Operating System:     Linux
- Using an IDE?: Yes: vscode
- Power Supply:         USB / Battery

## Problem Description

Hardware is an ODROID-GO.  Sdcard is connected via SPI.  Sdcard initializes fine, able to read some files off the card.  After user event, reading same files again results in an error.

### Expected Behavior

No error.  Files read again as expected.

### Actual Behavior

```
E (65453) sdmmc_cmd: sdmmc_read_sectors_dma: sdmmc_send_cmd returned 0x107
E (65453) diskio_sdmmc: sdmmc_read_blocks failed (263)
```

### Code to reproduce this issue

Source can be found here, in the `wip` branch: https://gitlab.jkent.net/odroid-go/golauncher

```c
const gpio_num_t GPIO_NUM_MISO          = GPIO_NUM_19;
const gpio_num_t GPIO_NUM_MOSI          = GPIO_NUM_23;
const gpio_num_t GPIO_NUM_SCLK          = GPIO_NUM_18;
const gpio_num_t GPIO_NUM_DISPLAY_CS    = GPIO_NUM_5;
const gpio_num_t GPIO_NUM_DISPLAY_DC    = GPIO_NUM_21;
const gpio_num_t GPIO_NUM_DISPLAY_BL    = GPIO_NUM_14;
const gpio_num_t GPIO_NUM_SDCARD_CS     = GPIO_NUM_22;
```

```c
    spi_bus_config_t bus_config = { };
    bus_config.miso_io_num = GPIO_NUM_MISO;
    bus_config.mosi_io_num = GPIO_NUM_MOSI;
    bus_config.sclk_io_num = GPIO_NUM_SCLK;
    bus_config.quadwp_io_num = GPIO_NUM_NC;
    bus_config.quadhd_io_num = GPIO_NUM_NC;
    bus_config.max_transfer_sz = sizeof(lv_color_t) * LV_HOR_RES_MAX * 10;
    ret = spi_bus_initialize(VSPI_HOST, &bus_config, 1);
[...]

PR #7710, opened 2021-10-16
sdspi_host: bugfix The clock may be sent out before the bus that was used immediately before is released. (IDFGH-6024)

lovyan03:
If the SPI bus is shared between SDSPI and TFT, etc., and the SPI bus is used alternately, it may not work correctly in some cases. 
( For example, #6510 )

It makes sense to do `go_idle_clockout` after running `spi_device_acquire_bus`.
In the unmodified state, `go_idle_clockout` would be run with other devices CS-asserted, which would result in a loss of control.

### DIFF
--- a/components/driver/sdspi_host.c
+++ b/components/driver/sdspi_host.c
@@ -75,6 +75,8 @@ static esp_err_t start_command_default(slot_info_t *slot, int flags, sdspi_hw_cm
 
 static esp_err_t shift_cmd_response(sdspi_hw_cmd_t *cmd, int sent_bytes);
 
+static esp_err_t poll_busy(slot_info_t *slot, int timeout_ms, bool polling);
+
 /// A few helper functions
 
 /// Map handle to pointer of slot information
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

igrr (approved):

igrr:
sha=cd5d30b56a13b8f0933e8879be1f97724a88004a

Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/8bb95150bf2a4402e9bd91620cefbaefa8655882.

adokitkat:
Hi @lovyan03. Can I please ask you where did you get how did you decide on 40ms timeout? Thanks.

lovyan03:
Hello, @adokitkat
This pull request wasn't merged, so I thought there must have been some problem, but I just realized that it was incorporated in a different way.

I can't remember, but I think I set it to wait long enough for the MISO pin on the SD card I was using at the time to respond.

In other words, I have no clear numerical value. Therefore, if this caused any problems, I think you need to consider appropriate solutions other than poll_busy.

adokitkat:
Yes it was merged but as we use primarily our internal repository (this Github repo is just a mirror), we have to merge PRs in a special way (notice the "sha" comment).

I see. Thank you very much for the answer!

In the next IDF version we will make this constant 40ms value modifiable, so users can decide if they want to lower the timeout value or disable this behavior (on boards without/with weak pull-up resistors this fix slows down the communication considerably - but on properly designed boards it shouldn't matter as MISO should go high properly and pull_busy not wait).

lovyan03:
I just realized this, but depending on the content of the communication, there may be cases where only sending is performed. If the setting does not include receiving, it may be okay to skip the waiting process.

PR #7713, opened 2021-10-17
adds twai_get_bitrate_timings function (IDFGH-6027)

kdschlosser:
This function will calculate the brp, tseg1, tseg2 and sjw values that are needed to create the twai interface.

see notes for further information

This is my first PR with esp-idf  and as such I am sure there will be things that are not correct with it. I check out the contributing information but all of the links in it are not functional so I couldn't review past the single document. So that is going to need to be fixed.

All feedback is welcome good or bad and if I need to change something  please give an example and explain what the benefit of the change is. I know enough C to just about get my toes a little wet. I have a decent understanding of C so not a complete noob. LOL.

I am a firm believer of the easier it is to use the more people will use it. It took me 6 months to fully wrap my brain around the whole CAN packet specification and of that information can be transparent to the user then it should be. The user needs to be able to provide the things that they will know like the target bitrate, length of the BUS and the processing delay of the transceiver  which is gotten  from the documentation for the transceiver. If an exact match cannot be found for their target bitrate then then next closest bitrate  the interface is capable of  should be returned. That is what this code does. It makes using a CAN interface as easy as using a serial interface.

Thank you for taking the time to review this

### DIFF
--- a/components/driver/include/driver/twai.h
+++ b/components/driver/include/driver/twai.h
@@ -116,6 +116,20 @@ typedef struct {
     uint32_t bus_error_count;       /**< Number of instances a bus error has occurred */
 } twai_status_info_t;
 
+
+
+/**
+ * @brief   Structure to store bitrate matches from the twai_get_bitrate_timings function
+ */typedef struct _twai_bitrate_t {
+    uint32_t bitrate;  /**< Matched bitrate */
+    uint16_t brp;      /**< Bitrate prescalar. */
+    uint8_t tseg1;     /**< Time segment 1 */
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

kdschlosser:
I have found a much better way to do the calculation it should take about 0.57 microseconds to do the work.  It would return only a single match. I need to know what the APB oscillator tolerance is in order to have it work.

If someone knows what that tolerance is it would be really helpful. I would imagine it should be 0.1% or smaller

PR #7717, opened 2021-10-18
print correct chip model (IDFGH-6031)

davidzuhn:
This example does not accommodate the various ESP32 chip models (it's either ESP32 or Unknown).   

This change prints the correct ESP32 variant.


zikalino:
@davidzuhn this PR seems empty, have you recerted changes by mistake?

PR #7719, opened 2021-10-18
esp_modem: Fix report unknown line (IDFGH-6034)

AxelLin:
In esp_dte_handle_line(), it tokenize the data to call handlers separately
for each *line*. So it needs to post the tokenized data when report unknown
line instead of the first token (which could be a '\r' in my test).

Fixes: 336de2941350 ("Examples/pppos_client: Fix manual parsing to accept unexpected lines")


Alvin1Zhang:
Thanks for your contribution.

Issue #7716, opened 2021-10-17
assert failed: xQueueGenericSend queue.c:818 (!( ( pvItemToQueue == ((void *)0) ) && ( pxQueue->uxItemSize != ( UBaseType_t ) 0U ) )) (IDFGH-6030)


RichFalk:
## Environment

- Development Kit:      DollaTek ESP32 Type-C WiFi Bluetooth Module 4MB Development Board Based on ESP32-WROVER-B
- Kit version (for WroverKit/PicoKit/DevKitC): N/A
- Module or chip used:  ESP32-WROVER; chip revision 1
- IDF version (run ``git describe --tags`` to find it): v4.4-dev-3042-g220590d599 (but I checked latest code today and it still has the same problem seen in source code so isn't fixed).
- Build System:         idf.py (uses CMake)
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): 8.4.0
- Operating System:     macOS
- Using an IDE?: No
- Power Supply:         USB

## Problem Description

xRingbufferSend (and related function xRingbufferSendFromISR) in ringbuf.c incorrectly gives back the semaphore indicating the item was successfully sent (rbGET_RX_SEM_HANDLE) so ready for reading BEFORE giving back the semaphore so other tasks can send (rbGET_TX_SEM_HANDLE) so ready for writing.  If a separate task reads from the Ringbuffer, returns the read item to the buffer, and then deletes the Ringbuffer because it is no longer needed, then **an assertion crash can occur in xQueueGenericSend** in queue.c because the deleted Ringbuffer is freed memory that may be reused and the assertion check for uxItemSize being zero may fail.

The problem does not occur when using xRingbufferSendAcquire and xRingbufferSendComplete in sequence because xRingbufferSendAcquire gives back the semaphore to other tasks (rbGET_TX_SEM_HANDLE) before xRingbufferSendComplete gives back the semaphore indicating the item was successfully sent (rbGET_RX_SEM_HANDLE).  The sequence of semaphore release in xRingbufferSend and xRingbufferSendFromISR are reversed from what they should be.  The incorrect sequence copied from code is the following:

```cpp
    if (xReturn == pdTRUE) {
        //Indicate item was successfully sent
        xSemaphoreGive(rbGET_RX_SEM_HANDLE(pxRingbuffer));
    }
    if (xReturnSemaphore == pdTRUE) {
[...]

PR #7721, opened 2021-10-18
Fix assertion xQueueGenericSend queue.c:818 (IDFGH-6036)

RichFalk:
The release of the semaphore indicating the item was successfully sent must be the last semaphore released.  The receiver may be in another task and may delete the Ringbuffer (such as with a return code across tasks design pattern) if they are through with the Ringbuffer.

The function xRingbufferSendAcquire followed by xRingbufferSendComplete had the semaphores released in the proper order and that same pattern should have been used in xRingbufferSend and xRingbufferSendFromISR.  This commit fixes this order.

Issue (IDFGH-6030) #7716 describes the problem in more detail.

### DIFF
--- a/components/esp_ringbuf/ringbuf.c
+++ b/components/esp_ringbuf/ringbuf.c
@@ -1041,12 +1041,12 @@ BaseType_t xRingbufferSend(RingbufHandle_t xRingbuffer,
          */
     }
 
+    if (xReturnSemaphore == pdTRUE) {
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/1222f6dd3abe56f4bec985a01bfaa8ff19abb5f4.

PR #7769, opened 2021-10-25
if wpa peap username or password is not configured, build nak response message and send back to authenticator (IDFGH-6088)

feifranklin:
Fixed the issue that unable to use Wpa Tls to connect to windows radius server when the radius serer is configured to check PEAP and then TLS.
If radius server's rule is configured to authenticate PEAP and then TLS and esp-idf is configured as Tls, esp-idf needs to respond a nak message to radius server to trigger to it to retry with Tls.

### DIFF
--- a/components/wpa_supplicant/src/eap_peer/eap_peap.c
+++ b/components/wpa_supplicant/src/eap_peer/eap_peap.c
@@ -159,6 +159,12 @@ eap_peap_init(struct eap_sm *sm)
 		return NULL;
 	}
 
+	if (!sm->config.identity || !sm->config.password) {
+		wpa_printf(MSG_ERROR, "EAP-PEAP: failed to init with user name, password");
+		eap_peap_deinit(sm, data);
+		return NULL;
+	}
+
 	data->phase2_type.vendor = EAP_VENDOR_IETF;
 	data->phase2_type.method = EAP_TYPE_NONE;
 


feifranklin:
> @sagb2015 to validate config within eap_peer_get_eap_method(...)? the check should be done within 'peap'-specific function instead of in generic logic. I noticed eap_tls_init(...) checking config->private_key, that's why my proposed change is checking username and password in eap_peap_init(...).

@sagb2015 could you give more advise on this PR?

sagb2015:
Hi @feifranklin, can you help to test with [this](https://github.com/espressif/esp-idf/files/7512896/0001-esp_wifi-Skip-EAP-PEAP-mehtod-if-identity-and-passwo.txt) patch ? If it works okay, we will soon have it available in master.

sagb2015:
> Hi @feifranklin, can you help to test with [this](https://github.com/espressif/esp-idf/files/7512896/0001-esp_wifi-Skip-EAP-PEAP-mehtod-if-identity-and-passwo.txt) patch ? If it works okay, we will soon have it available in master.

@feifranklin There are certain problems with this patch as well. We are working on the better solution,

feifranklin:
> > Hi @feifranklin, can you help to test with [this](https://github.com/espressif/esp-idf/files/7512896/0001-esp_wifi-Skip-EAP-PEAP-mehtod-if-identity-and-passwo.txt) patch ? If it works okay, we will soon have it available in master.
> 
> @feifranklin There are certain problems with this patch as well. We are working on the better solution,

@sagb2015  keep me posted. I can help testing once the fix is available.

sagb2015:
@feifranklin Your method works, but ideally if PEAP is not configured, the corresponding init itself should not be called. In other words, the check should happen even before init is called IMO. The configuration should be enough to determine what the device can support. The problem stems from the fact traditionally ESP devices aimed to support all the methods, making all the config mandatory. However, later on, some part of config was made optional assuming not all the methods will be needed in the field. This now means config should derive supported methods and only proceed with those while handling EAP messages. Hope this is clear!

hpdhabe:
@feifranklin Can you please tryout the following patch and let us know how it behaves and if the NAK message goes through.
[eap_method_selection_patch.txt](https://github.com/espressif/esp-idf/files/7675770/eap_method_selection_patch.txt)

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/190b31bb1b9b1e6e21feb038797aed6e14cb765b.

PR #7770, opened 2021-10-25
client cert and key need to be configured; otherwise ESP_ERR_INVALID_STATE should be returned (IDFGH-6090)

feifranklin:
fixed set_client_config(...) to return ESP_ERR_INVALID_STATE if clientcert_buf or clientkey_buf is not set in cfg.


mahavirj:
Closing the PR based on discussion here. Please feel free to re-open if you are still facing any issues.

PR #7773, opened 2021-10-25
ftm example: clear FTM_FAILURE_BIT too (IDFGH-6094)

crackwitz:
Once some failure with FTM occurred, the FTM_FAILURE_BIT was set forever and caused `xEventGroupWaitBits` to "fall through" right away, dumping the previous session's report (if the success bit was set by the previous session), before the current session could even finish.

All bits need to be cleared (and "handled") if they were set. Instead of explicitly clearing the respective flag after the test, I chose having `xEventGroupWaitBits` clear the flags itself (`xClearOnExit`) atomically.


crackwitz:
Thanks for the merge!

PR #7811, opened 2021-11-01
Increased LAN8720 ETH-PHY reset assertion time (IDFGH-6018) (IDFGH-6132)

sauttefk:
Increase reset assertion time from 100µs (as specified minimum in the datasheet) to 150µs.
Some specimen of the LAN8720/42 need the reset signal asserted longer than 100µs to initialise properly. Otherwise they are in a zombie state where they are establishing and loosing an Ethernet link once in a seconds interval.

This is the second pull request for #7705 , as I messed up my last commit :-/


kostaond (approved):
LGTM!

PR #7819, opened 2021-11-03
Initialize `$envvars_array` to an empty list. (IDFGH-6140)

cmumford:
With strict debugging, i.e. `Set-PSDebug -strict`, execution of export.ps1 will
report the following error:

```
The variable '$envars_array' cannot be retrieved because it has not been set.
At C:\path\to\export.ps1:15 char:1
+ $envars_array # will be filled like:
+ ~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (envars_array:String) [], RuntimeException
    + FullyQualifiedErrorId : VariableIsUndefined
```


Alvin1Zhang:
Thanks for your contribution.

PR #7824, opened 2021-11-03
sntp example - updated ntp over dhcp code/comments (IDFGH-6147)

vortigont:
this is a followup for #7336 - a bit more detailed example on how to properly set NTPoDHCP along with statically configured servers

### DIFF
--- a/examples/protocols/sntp/main/sntp_example_main.c
+++ b/examples/protocols/sntp/main/sntp_example_main.c
@@ -172,13 +172,15 @@ static void initialize_sntp(void)
  * as a 'secondary' server. It will act as a fallback server in case that address
  * provided via NTP over DHCP is not accessible
  */
-#if defined LWIP_DHCP_GET_NTP_SRV && (SNTP_MAX_SERVERS > 1)
+#if  (LWIP_DHCP_GET_NTP_SRV == 1) && (SNTP_MAX_SERVERS > 1)
     sntp_setservername(1, "pool.ntp.org");
 
 #if LWIP_IPV6                   // statically assigned IPv6 address is also possible
     ip_addr_t ip6;
-    if (ipaddr_aton("2a01:3f7::1", &ip6))       // ipv6 ntp source "ntp.netnod.se"
+    if (ipaddr_aton("2a01:3f7::1", &ip6)) {
+         // ipv6 ntp source "ntp.netnod.se"
         sntp_setserver(2, &ip6);
+    }
 #endif  /* LWIP_IPV6 */
 
 #else   /* LWIP_DHCP_GET_NTP_SRV && (SNTP_MAX_SERVERS > 1) */
--- a/examples/protocols/sntp/main/sntp_example_main.c
+++ b/examples/protocols/sntp/main/sntp_example_main.c
@@ -23,6 +23,10 @@
 
 static const char *TAG = "example";
 
+#ifndef INET6_ADDRSTRLEN
+#define INET6_ADDRSTRLEN 48
+#endif
+
 /* Variable holding number of times ESP32 restarted since first boot.
  * It is placed into RTC memory using RTC_DATA_ATTR and
  * maintains its value when ESP32 wakes from deep sleep.
@@ -124,10 +128,15 @@ static void obtain_time(void)
 
     /**
      * NTP server address could be aquired via DHCP,
-     * see LWIP_DHCP_GET_NTP_SRV menuconfig option
+     * see following menuconfig options:
+     * 'LWIP_DHCP_GET_NTP_SRV' - enable STNP over DHCP
+     * 'LWIP_SNTP_DEBUG' - enable debugging messages
+     *
+     * NOTE: This call should be made BEFORE esp aquires IP address from DHCP,
+     * otherwise NTP option would be rejected by default.
      */
 #ifdef LWIP_DHCP_GET_NTP_SRV
-    sntp_servermode_dhcp(1);
+    sntp_servermode_dhcp(1);      // accept NTP offers from DHCP server, if any
 #endif
 
     /* This helper function configures Wi-Fi or Ethernet, as selected in menuconfig.
@@ -142,7 +151,7 @@ static void obtain_time(void)
     time_t now = 0;
     struct tm timeinfo = { 0 };
     int retry = 0;
-    const int retry_count = 10;
+    const int retry_count = 15;
     while (sntp_get_sync_status() == SNTP_SYNC_STATUS_RESET && ++retry < retry_count) {
         ESP_LOGI(TAG, "Waiting for system time to be set... (%d/%d)", retry, retry_count);
         vTaskDelay(2000 / portTICK_PERIOD_MS);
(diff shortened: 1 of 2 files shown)


KaeLL:
I'd also add a small anti-stupidity memo to remind people that option 42 doesn't work with hostnames, only IP addresses. I wasted a few hours last week on that particular treadmill.

vortigont:
@KaeLL agreed, might be a link to RFC... but it relates more to DHCP server configuration than client and there are plenty of those. Normally it should forbid defining ascii chars for opt 42. But that is why I added a piece of code to the example that prints out a list of active NTPs for lwip, both static domain names and received IP address from DHCP. It would be much easier this way to check what it's got actually.

Alvin1Zhang:
Thanks for your contribution.

examples/protocols/sntp/main/sntp_example_main.c
```diff
@@ -157,10 +162,33 @@ static void initialize_sntp(void)
...
 #endif
     sntp_init();
+
+    ESP_LOGI(TAG, "List of configured NTP servers:");
+
+    for (uint8_t i = 0; i < SNTP_MAX_SERVERS; ++i){
+        if (sntp_getservername(i)){
+            ESP_LOGI(TAG, "server %d: %s", i, sntp_getservername(i));
+        } else {
+            const ip_addr_t * _ip = sntp_getserver(i);
```
david-cermak, line 189: Nitpick: variables starting with underscore are reserved for compiler use
```suggestion
            const ip_addr_t * ip = sntp_getserver(i);
```

examples/protocols/sntp/main/sntp_example_main.c
```diff
@@ -157,10 +162,33 @@ static void initialize_sntp(void)
...
+            if (0 == _ip->type && _ip->u_addr.ip4.addr)    // non-empty IPv4 addresses
```
david-cermak, line 190:
```cpp
(0 == ip->type)
```
Would this compile if we disable `IPv6` ?

examples/protocols/sntp/main/sntp_example_main.c
```diff
@@ -157,10 +162,33 @@ static void initialize_sntp(void)
...
+                ESP_LOGI(TAG, "server %d: %u.%u.%u.%u", i, ip4_addr1_val(_ip->u_addr.ip4), ip4_addr2_val(_ip->u_addr.ip4), ip4_addr3_val(_ip->u_addr.ip4), ip4_addr4_val(_ip->u_addr.ip4) );
```
david-cermak, line 191: same here, would work only for dual stack `LWIP_IPV4 && LWIP_IPV6`

examples/protocols/sntp/main/sntp_example_main.c
vortigont: Ugh, you are right, dual stack only. Actually I'm not sure what is the best way to handle it here. Assume we have IPv4 only NTPs and just skip otherwise? I'm not sure if LWIP is capable to do NTPoDHCP via IPv6, need test in a lab setup.
I can fix it for sure, just want to hear your suggestion?

david-cermak: > I can fix it and limit to IPv4-only or dual stack

This sounds good. IDF supports only these two (dual stack and IPv4-only) settings, so that's perfectly okay to disregard IPv6-only scenario.

> LWIP is capable to do NTPoDHCP via IPv6,

lwip supports DHCPv6 in its stateless configuration and users can add options using `LWIP_HOOK_DHCP6_APPEND_OPTIONS` (so theoretically we can add `OPTION_NTP_SERVER (56)`) , but TBH I haven't tested that either, and even if we get this running it wouldn't be very practical to include it into this simple example.

This piece of code just prints addresses of NTP servers, correct? So not only those obtained by DHCP, but also the hardcoded ones, and maybe some of the names get resolved to IPv6 addresses (one day)? If you want to go the extra mile, you might look into how we can print the IP address in the IP-protocol agnostic way (using `ntoa()`-like API -- which again is not very readable if written in lwip)

Let's just assume we have IPv4 NTPs here. My only point was that if users disable `IPv6` in menuconfig this example won't compile.

vortigont: Thanks for your ideas, @david-cermak, appreciate it.

> This piece of code just prints addresses of NTP servers, correct?
it prints domain names for ntp's assigned statically and IP's for NTPs received via DHCP.

OK, I will concentrate on IPv4 and dual-stack case here and leave IPv6 opts for future improvements :)
Will update this PR and let you know than!
Cheers!

vortigont: Found a better approach. Now it works fine for IPv4-only and for dual-stack builds both for IPv4/IPv6 ntp addresses.

```
sntp_init: SNTP initialised
I (7400) example: List of configured NTP servers:
I (7410) example: server 0: 192.168.168.2
I (7410) example: server 1: pool.ntp.org
I (7420) example: server 2: 2A01:3F7::1
I (7420) example: Waiting for system time to be set... (1/15)
```

examples/protocols/sntp/main/sntp_example_main.c
```diff
@@ -157,10 +166,43 @@ static void initialize_sntp(void)
 {
     ESP_LOGI(TAG, "Initializing SNTP");
     sntp_setoperatingmode(SNTP_OPMODE_POLL);
+
+/*
+ * If 'NTP over DHCP' is enabled, we set dynamic pool address
+ * as a 'secondary' server. It will act as a fallback server in case that address
+ * provided via NTP over DHCP is not accessible
+ */
+#if defined LWIP_DHCP_GET_NTP_SRV && (SNTP_MAX_SERVERS > 1)
```
david-cermak, line 175:
```suggestion
#if  (LWIP_DHCP_GET_NTP_SRV == 1) && (SNTP_MAX_SERVERS > 1)
```
or
```cpp
#if defined CONFIG_LWIP_DHCP_GET_NTP_SRV
```

Otherwise, if the DHCP-NTP option is not set (default), this condition is true because after:

https://github.com/espressif/esp-idf/blob/4a011f318377ab717b9dc459019c1055602c93ff/components/lwip/port/esp32/include/lwipopts.h#L1065-L1067

`LWIP_DHCP_GET_NTP_SRV` is undefined, but

https://github.com/espressif/esp-lwip/blob/2749568fe15df2003f6c3f37f0dfd44f8f01fcd6/src/include/lwip/opt.h#L956-L958

sets this to `0`, so it's actually defined

examples/protocols/sntp/main/sntp_example_main.c
```diff
@@ -157,10 +166,43 @@ static void initialize_sntp(void)
...
+    sntp_setservername(1, "pool.ntp.org");
+
+#if LWIP_IPV6                   // statically assigned IPv6 address is also possible
+    ip_addr_t ip6;
+    if (ipaddr_aton("2a01:3f7::1", &ip6))       // ipv6 ntp source "ntp.netnod.se"
+        sntp_setserver(2, &ip6);
```
david-cermak, line 181: Formatting (see the contribution guide)
```suggestion
    if (ipaddr_aton("2a01:3f7::1", &ip6)) {
         // ipv6 ntp source "ntp.netnod.se"
        sntp_setserver(2, &ip6);
    }
```

david-cermak (approved):
Again LGTM, just two nitpicks

examples/protocols/sntp/main/sntp_example_main.c
david-cermak: Nice work, thanks for testing this!

vortigont:
> Again LGTM, just two nitpicks

thanks, David! indeed I missed that define could have a default 0, nice catch! Applied changes as you suggested.
Cheers!

Alvin1Zhang:
@vortigont Thanks for contribution again, changes merged with https://github.com/espressif/esp-idf/commit/696acf5e3f18925987f392f999782aadedfdc89e, feel free to reopen.

PR #7828, opened 2021-11-04
Modify the ENC28J60 MAC driver implementation to allow multicast packets (IDFGH-6151)

andycarle:
This patch modifies the example ENC28J60 MAC driver implementation (`esp_eth_mac_enc28j60.c`) to allow multicast packets to pass through the hardware's internal packet filter, rather than filtering them out. 

This change makes the example driver applicable to a wider range of projects and avoids confusion from new users of the driver who want to use multicast over Ethernet. Almost no projects need multicast packets filtered out, but some certainly may want them included.

### DIFF
--- a/examples/ethernet/enc28j60/components/eth_enc28j60/esp_eth_mac_enc28j60.c
+++ b/examples/ethernet/enc28j60/components/eth_enc28j60/esp_eth_mac_enc28j60.c
@@ -1,16 +1,8 @@
-// Copyright 2019 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

kostaond:
Hi @andycarle, thank you for the contribution! The update you performed makes perfect sense! However, we need to prepare the PR to pass formal checks too. Have you installed the pre-commit hook for esp-idf project as specified [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/index.html)? Please also check you conform with other bullets. Speaking of pre-commit hook, it should ensure the licence header update. Otherwise we wouldn't pass internal tests and the change could not be merged.

andycarle:
Hi @kostaond, thank you very much for the pointer. You are right, I had not installed the pre-commit hook.

I'm getting that set up now. Once I have that in place and fix the license header, is the best practice to simply open a new pull request with a fixed commit? Or should I force push/squash the corrected commit into this PR? 

Thank you!

kostaond:
I would prefer to use the same PR because of tracking the issue. Please "force push" the modified commit. Thanks! If you needed any assistance, do not hesitate to ask.

andycarle:
@kostaond Thank you for the recommendation. I have worked through the pre-commit fixes and force pushed the modified commit. Please take a look when you get a chance and let me know if I should make any additional changes.

(For what it's worth: I was unable to get the ESP-IDF `pre-commit` hooks working correctly on Windows. There seems to be a bug right now that causes false positives on the `tools` folder. This was also mentioned in #7547. But I got it working on macOS.) 

Thank you!

david-cermak (approved):

PR #7843, opened 2021-11-05
Check for failure in stopping http server (IDFGH-6166)

bugadani:
In some cases, sending the shutdown signal may fail, which causes `httpd_stop` to wait indefinitely. This PR adds a check to prevent this infinite loop.


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97 (approved):

Issue #7853, opened 2021-11-08
Lose of precision on pdTICKS_TO_MS convertions. (IDFGH-6176)


lisekt84:
## Environment

ESP-IDF:          v4.3.1-246-g37a736e419

## Problem Description

Lose of precision on pdTICKS_TO_MS convertions in **xTicks * 1000** multiply operation (output could exceed TickType_t max value).
Line 45: /components/freertos/include/freertos/projdefs.h 
```cpp
	#define pdTICKS_TO_MS( xTicks )   ( ( uint32_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ )
```

### Expected Behavior
```cpp
//configTICK_RATE_HZ = 1000
uint32_t output = pdTICKS_TO_MS(4,294,967,295);
```
`**output: 4,294,967,295**` 

### Actual Behavior 
```cpp
//configTICK_RATE_HZ = 1000
uint32_t output = pdTICKS_TO_MS(4,294,967,295);
```
`**output: 4,294,967**` 

### Code to resolve this issue

```cpp
	#define pdTICKS_TO_MS( xTicks )   ( ( TickType_t ) ( ( uint64_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ ) )
```

AxelLin:
v4.3 still does not include this fix.

AxelLin:
Just remind that v4.4-rc1 does not include this fix.

igrr:
Thanks for the reminder @AxelLin. We will merge this fix into release/v4.4 soon after v4.4 release is tagged. This fix will be part of v4.4.1.

Alvin1Zhang:
Fix on release/4.3 is available at https://github.com/espressif/esp-idf/commit/6dce875bf9ab3672508d0bb74cd360e60f622395, and will be part of IDF v4.3.3 release, thanks.

PR #7856, opened 2021-11-08
Update projdefs.h (IDFGH-6179)

lisekt84:
Lose of precision on pdTICKS_TO_MS convertions. (IDFGH-6176) #7853

This bug is on master and other branches too

### DIFF
--- a/components/freertos/include/freertos/projdefs.h
+++ b/components/freertos/include/freertos/projdefs.h
@@ -42,7 +42,7 @@ definition here is not suitable for your application. */
 #endif
 
 #ifndef pdTICKS_TO_MS
-	#define pdTICKS_TO_MS( xTicks )   ( ( uint32_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ )
+	#define pdTICKS_TO_MS( xTicks )   ( ( TickType_t ) ( ( uint64_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ ) )
 #endif
 
 


KaeLL:
PRs like this are supposed to be made against the master branch.

components/freertos/include/freertos/projdefs.h
```diff
@@ -42,7 +42,7 @@ definition here is not suitable for your application. */
 #endif
 
 #ifndef pdTICKS_TO_MS
-	#define pdTICKS_TO_MS( xTicks )   ( ( uint32_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ )
+	#define pdTICKS_TO_MS( xTicks )   ( ( TickType_t ) ( ( uint64_t ) ( xTicks ) * 1000 / configTICK_RATE_HZ ) )
```
Dazza0, line 45: We updated the macro in commits ff56848a0eca09865ec92f1c72d55e39b3984392 and 3f2f6d0cf5438037f891ce9e483af1709ad8a33d to match upstream FreeRTOS v10.4.3:

```c
/* Converts a time in milliseconds to a time in ticks.  This macro can be
 * overridden by a macro of the same name defined in FreeRTOSConfig.h in case the
 * definition here is not suitable for your application. */
#ifndef pdMS_TO_TICKS
    #define pdMS_TO_TICKS( xTimeInMs )    ( ( TickType_t ) ( ( ( TickType_t ) ( xTimeInMs ) * ( TickType_t ) configTICK_RATE_HZ ) / ( TickType_t ) 1000U ) )
#endif
```
As per upstream, I think it would be better to override this macro in [`FreeRTOSConfig.h`](https://github.com/espressif/esp-idf/blob/master/components/freertos/esp_additions/include/freertos/FreeRTOSConfig.h) so that we can keep `projdefs.h` clean

Dazza0 (changes requested):
-  Nitpick: Commit message format

```
freertos: Add higher precision pdTICKS_TO_MS macro

Closes https://github.com/espressif/esp-idf/issues/7853
Closes https://github.com/espressif/esp-idf/pull/7856

```
@lisekt84 Left a couple of nitpicks but the new macro LGTM.

components/freertos/include/freertos/projdefs.h
lisekt84: Sure, work on copy at FreeRTOSConfig.h is really better way and its workarund for this problem. Any way it still bug on code pdTICKS_TO_MS on projdefs.h and this is poor solution.

Dazza0: @lisekt84 sorry, my mistake. I mixed up `pdMS_TO_TICKS` and `pdTICKS_TO_MS`. `pdTICKS_TO_MS` doesn't exist in upstream FreeRTOS so it should be OK to update it directly in `projdefs.h`. I'll merge your commit internally asap.

Dazza0 (approved):

components/freertos/include/freertos/projdefs.h
lisekt84: I also checked pdMS_TO_TICKS and this define is wrong too.
I think the perfect way is set the comment about this problem.
Problem start at 71 minutes and 35 sec from start (UINT32_MAX / 1000).

Issue #7820, opened 2021-11-03
The `export.ps1` script used undefined variable: `$IsWindows` (IDFGH-6141)


cmumford:
## Environment

- Development Kit:      N/A
- Kit version (for WroverKit/PicoKit/DevKitC): N/A
- Module or chip used:  N/A
- IDF version (run ``git describe --tags`` to find it): v4.3.1
- Build System:         N/A
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): N/A
- Operating System:     Windows 10
- (Windows only) environment type: PowerShell
- Using an IDE?: no
- Power Supply:         N/A

## Problem Description

`$IsWindows` is used, but never defined.

### Steps to reproduce

1. Install IDE.
2. Enable strict debugging: `Set-PSDebug -strict`
3. `./export.ps1`

dobairoland:
Hi @cmumford. Thanks for the report. Do you know the proper solution for this? IsWindows should have been defined on other platform to $False: 

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7.1#iswindows

Shouldn't this be a PowerShell issue? It should not complain about his own internal variable.

dobairoland:
@cmumford I think this issue is caused by old version of PowerShell. The link given above is for 7.1. The variable is missing for version 5.1. 

I can reproduce this in PowerShell 5.1. Just execute `$IsWindows`. It won't print anything. If you repeat this after setting the debug flag then it will produce the same error as `export.ps1`.

cmumford:
Yes, here's my version:

```
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      19041  1320
```

I think I'll submit a PR to set `$IsWindows` to true if not defined. I think that PowerShell 6.x shipped on non-windows platforms so this won't be technically correct.

cmumford:
Actually it is defined in "PowerShell Core 6 and PowerShell 7", so maybe were OK here.

PR #7858, opened 2021-11-09
Define $IsWindows if not defined. (IDFGH-6181)

cmumford:
The `$IsWindows` PowerShell variable was added in PowerShell Core 6 and
PowerShell 7, and is not present in earlier PowerShell versions. Set to
true if undefined. This fixes https://github.com/espressif/esp-idf/issues/7820.

The first version to run on non-Windows platforms was PowerShell Core
6.0[^1] which means that IsWindows is guaranteed to be defined on all
non-Windows systems. So, if undefined this indicates a Windows platform.

[^1]: https://docs.microsoft.com/en-us/powershell/scripting/install/powershell-support-lifecycle?view=powershell-7.2#release-history

### DIFF
--- a/export.ps1
+++ b/export.ps1
@@ -23,6 +23,11 @@ foreach ($line  in $envars_raw) {
     $envars_array += (, ($var_name, $var_val))
 }
 
+if ($IsWindows -eq $null) {
+    # $IsWindows was added in PowerShell Core 6 and PowerShell 7.
+    $IsWindows = $true
+}
+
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

cmumford:
I verified this on PowerShell 5.1.

dobairoland:
Great fix @cmumford. Thanks! I move both of your PRs into our test queue. They will be closed when they get merged.

PR #7867, opened 2021-11-10
Ensurepip before installing virtualenv or use venv (IDFGH-6190)

ksmq:
I had a minor hiccup with the installation because my python doesn't have pip by default, which caused an error.

I changed the script to use venv, and that seems to work normally with the few examples I tried, but I understand some people may prefer to use virtualenv hence the warn... in either case, "ensurepip" should be ran before trying to use `python -m pip ...`. I was going to submit an issue but I thought it would be more useful to just provide the code, so apologies if there's stylistic inconsistencies.

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -1508,16 +1508,27 @@ def action_install_python_env(args):  # type: ignore
 
     if not os.path.exists(virtualenv_python):
         info('Creating a new Python environment in {}'.format(idf_python_env_path))
-
+      
         try:
-            import virtualenv  # noqa: F401
+            import virtualenv # noqa: F401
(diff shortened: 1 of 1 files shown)


dobairoland:
Hi @ksmq. Thank you for the contribution. I will consider addressing this in the near future in a different way because the Windows Tools Installer also deploys an environment with the one this script needs to be synchronized. We have already in place several fixes for virtualenv on Windows and switching just like this to venv would invoke a whole new set of possible issues. 

Pip for other platforms are required to be installed as a package prerequisite (https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-setup.html#install-prerequisites). 

Thank you again and sorry for rejecting this.

PR #7869, opened 2021-11-10
Fix: Ethernet: Add loopback for ksz8081 (IDFGH-6194)

dabbinavo:
I think there is a missing assignment of the loopback function for the ksz8081.
Can you verify this, @kostaond or @suda-morris?


kostaond:
Hi @dabbinavo, thank you very much for the contribution. You are right, it is really missing. I already have one internal merge request in review which resolves this issue (it is part of bigger update). Therefore it will be easier for me to process it that way. I hope you don't mind. Anyway, thank you again and if you find any other issue, don't hesitate to fix it :)

kostaond:
I just realized that my MR is intended for IDF v5.0 (introducing braking changes) but this issue should be backported and fixed in previous versions. So this PR could be utilized. I will let you know how we will proceed. Unfortunately, I am currently a little bit busy so please be patient...

Issue #7380, opened 2021-08-06
Bug with IDF SPI driver ?  Ethernet at same time as SPI LCD causes display corruption (IDFGH-5658)


jonshouse1:
This looks like a bug with the esp-idf SPI driver.

https://github.com/espressif/esp-iot-solution/issues/110

lexxvir:
Hi there,

It seems I encountered the same bug. We use IDF v.4.2 branch. We have a custom board with ST25R3911 connected by SPI to ESP32. For internet connectivity we are using WiFi **or** Ethernet. When we use WiFi all works fine. When we use Ethernet it most time works fine **but** it's possible to "break" ST25R3911 by sending many ICMP requests to the ESP32 (`ping -f` for couple of seconds is enough). Same `ping -f` command does not "break" ST25R3911 when we use WiFi.

Meantime, in another project we use the same HW and IDF v.3.3. In that case there is no such an issue, so I think it's an issue in the v.4.x branch.

By "break" ST25R3911 I mean that it falls to "odd" state in that it does not process the commands.

corecode:
Can confirm SPI transfer issues when ethernet is receiving packets.  For us, it seems that the SPI data read pointer is reversed by 32 bytes, then the read pointer jumps back to its original value and the rest of the data is transferred correctly.  Sometimes this reversal happens after just a few transmitted bytes (again seems 2^n length, e.g. 64 bytes); sometimes this offset exists from the beginning, for our whole SPI transfer (almost 4000 bytes).

elcojacobs:
We have been chasing this bug for 2 weeks and I think I finally identified the relationship between Ethernet and display glitches. Our display glitches could be explained by missed set x/y commands.

The ILI9488 samples the D/C pin on the first falling clock after CS goes low. This requires SPI mode 3 to have an idle high CLK.
But that's not all.

We use DMA for all display writes and set the DC pin in the pre-callback. When Ethernet is not connected, I think most display writes finish immediately and the dma queue stays empty. As a result, the CS pin is toggled by the DMA SPI handler for each transfer.
When Ethernet is connected, the ethernet task loads the CPU or dma handler with other tasks, which allows the display dma queue to fill with more than one transfer. When these dma transfers are handled back to back, the CS pin stays low. The second transfer in the queue does not have a CS pin high to low transition and the DC pin is not resampled.

We fixed it by adding a write high and write low to the CS pin immediately after setting the DC pin, in the pre-callback.

[...]

PR #7874, opened 2021-11-11
emac: reduce rx burst length to avoid SPI DMA collision bug (IDFGH-6199)

corecode:
See <https://github.com/espressif/esp-idf/issues/7380>.


corecode:
This PR applies to IDF 4.0 to 4.3; 4.4 has some source changes but presumably needs this change as well.

Alvin1Zhang:
Thanks for your contribution again, changes merged with https://github.com/espressif/esp-idf/commit/2553fb5845efbc8cb3ad2d6d4b72cf57fb86c4fc.

PR #7945, opened 2021-11-21
Added filename, line number and method name to log output (IDFGH-6276)

0xFEEDC0DE64:
The log output now also shows the source of the log statement, the filename, line number and funciton name.

Example from my firmware:
image

### DIFF
--- a/components/app_trace/include/esp_app_trace_util.h
+++ b/components/app_trace/include/esp_app_trace_util.h
@@ -174,7 +174,7 @@ void esp_apptrace_log_unlock(void);
 #define ESP_APPTRACE_LOG_LEV( _L_, level, format, ... )   \
     do { \
         if (LOG_LOCAL_LEVEL >= level) { \
-            ESP_APPTRACE_LOG(LOG_FORMAT(_L_, format), esp_log_early_timestamp(), TAG, ##__VA_ARGS__); \
+            ESP_APPTRACE_LOG(LOG_EARLY_FORMAT(_L_, format, TAG), ##__VA_ARGS__); \
         } \
     } while(0)
 
--- a/components/app_trace/test/test_trace.c
+++ b/components/app_trace/test/test_trace.c
@@ -54,7 +54,7 @@ const static char *TAG = "esp_apptrace_test";
 #define ESP_APPTRACE_TEST_LOG_LEVEL( _L_, level, format, ... )   \
     do { \
         if (LOG_LOCAL_LEVEL >= level) { \
-            ESP_APPTRACE_TEST_LOG(LOG_FORMAT(_L_, format), esp_log_early_timestamp(), TAG, ##__VA_ARGS__); \
+            ESP_APPTRACE_TEST_LOG(LOG_EARLY_FORMAT(_L_, format, TAG), ##__VA_ARGS__); \
         } \
     } while(0)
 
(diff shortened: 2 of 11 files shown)


Alvin1Zhang:
Thanks for your contribution.

0xFEEDC0DE64:
Any update on this?

0xFEEDC0DE64:
with every rebase I do something breaks around the logging, now someone changed something so that the format does not work anymore, should I give up on this and introduce a paralell logging system besides the already existing?

0xFEEDC0DE64:
anyone there?

0xFEEDC0DE64:
hello?

o-marshmallow:
Hi @0xFEEDC0DE64 ,

Sorry for the delay of response. Implementing this feature would result in a much bigger binary and so more flash/ram usage, which would not match other users' requirement. Thus, we currently don't have any plan to merge it.

However, we will keep in mind that custom messages may be necessary for some users in a potential future version of the Log component.

0xFEEDC0DE64:
is the increased size even be a problem when it's turned off by default? I dont think the binary size would be so much, as its just a few bytes more, did you actually measure and have some numbers for me?

igrr:
I have some numbers from when I tested this. For the examples/provisioning/wifi_prov_mgr example, this branch produces about 20 kB larger app binary and 3 kB larger bootloader. The bootloader size increase is quite problematic: the size goes over the default limit of 0x7000 (28672) bytes (for the ESP32), so with the default partition table offset most of the examples can no longer be built.

<details>
<summary>(details)</summary>

```
$ idf_size.py --diff build_master/wifi_prov_mgr.map build_7945/wifi_prov_mgr.map
<CURRENT> MAP file: build_7945/wifi_prov_mgr.map
<REFERENCE> MAP file: build_master/wifi_prov_mgr.map
Difference is counted as <CURRENT> - <REFERENCE>,           i.e. a positive number means that <CURRENT> is larger.
Total sizes of <CURRENT>:                                       <REFERENCE>     Difference
Used static DRAM:   49632 bytes (  74948 remain, 39.8% used)          49152           +480 (   -480 remain,      +0 total)
      .data size:   23760 bytes                                       23280           +480
      .bss  size:   25872 bytes                                       25872
Used static IRAM:  124502 bytes (   6570 remain, 95.0% used)         124126           +376 (   -376 remain,      +0 total)
      .text size:  123475 bytes                                      123099           +376
   .vectors size:    1027 bytes                                        1027
Used Flash size :  906937 bytes                                      888765         +18172
      .text     :  723899 bytes                                      717247          +6652
      .rodata   :  182782 bytes                                      171262         +11520
Total image size: 1055199 bytes (.bin may be padded larger)         1036171         +19028

$ idf_size.py --diff build_master/bootloader/bootloader.map build_7945/bootloader/bootloader.map                   
<CURRENT> MAP file: build_7945/bootloader/bootloader.map
<REFERENCE> MAP file: build_master/bootloader/bootloader.map
Difference is counted as <CURRENT> - <REFERENCE>,           i.e. a positive number means that <CURRENT> is larger.
Total sizes of <CURRENT>:                                       <REFERENCE>     Difference
Used static DRAM:    9332 bytes (  56204 remain, 14.2% used)           7064          +2268 (  -2268 remain,      +0 total)
      .data size:       4 bytes                                           4
      .bss  size:      48 bytes                                          48
   .rodata  size:    9280 bytes                                        7012          +2268
Used static IRAM:   20252 bytes (  77028 remain, 20.8% used)          19196          +1056 (  -1056 remain,      +0 total)
      .text size:   20252 bytes                                       19196          +1056
Total image size:   29536 bytes (.bin may be padded larger)           26212          +3324
```
</details>

KonstantinKondrashov:
Hi @0xFEEDC0DE64! Thanks for your contribution.
We are going to improve the esp_log component to make it more flexible. There is a list of features that we are implementing (including this - filename, line number, and method name to log output).
I close it for now because we can not take it in terms of the work that we have already started.

PR #7948, opened 2021-11-22
Clarify when the socket is closed if a custom session closing callbac… (IDFGH-6279)

FlorianR:
…k is used

When using a custom session closing callback, the IDF will not close the socket for the user. This might result in the system running out of fds. Without any note on that in the documentation, this is easy to miss.


Alvin1Zhang:
Thanks for your contribution.

mahavirj:
Merged with https://github.com/espressif/esp-idf/commit/4db3fd17d927b74f2f585227517fd9b7d7a2cee9. Thanks for your contribution!

PR #7958, opened 2021-11-24
fix: missing initializer waring in esp_eth (IDFGH-6293)

b1ackviking:
add auto_nego_en initializer to ETH_DEFAULT_CONFIG

fixes the following warning:
```
<source>/esp-idf/components/esp_eth/include/esp_eth.h:144:5: warning: missing initializer for member 'esp_eth_config_t::auto_nego_en' [-Wmissing-field-initializers]
     }
     ^
note: in expansion of macro 'ETH_DEFAULT_CONFIG'
   esp_eth_config_t config = ETH_DEFAULT_CONFIG(pMac, pPhy);
```

### DIFF
--- a/components/esp_eth/include/esp_eth.h
+++ b/components/esp_eth/include/esp_eth.h
@@ -43,12 +43,6 @@ typedef struct {
     uint32_t check_link_period_ms;
 
     /**
-     * @brief Configuration status of PHY autonegotiation feature
-     *
-     */
-    bool auto_nego_en;
-
-    /**
     * @brief Input frame buffer to user's stack
     *
     * @param[in] eth_handle: handle of Ethernet driver


Alvin1Zhang:
Thanks for your contribution.

components/esp_eth/include/esp_eth.h
```diff
@@ -136,6 +136,7 @@ typedef struct {
         .mac = emac,                     \
         .phy = ephy,                     \
         .check_link_period_ms = 2000,    \
+        .auto_nego_en = true,            \
```
kostaond, line 139: Thank you for pointing this issue out. This is really a mistake. However, proper fix should be removing `auto_nego_en` member from `esp_eth_config_t` structure.  `auto_nego_en` configuration member is not used anywhere and auto-negotiation should be only configurable via `esp_eth_ioctl` function from user's perspective. Could you please fix it in that way?

kostaond (commented):
Thanks for your contribution. You found a real issue. However, it should be fixed a little bit differently, please have a look below.

Just a reminder, please ensure you comply with all bullets stated [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/index.html). If you already do, please ignore this reminder.

components/esp_eth/include/esp_eth.h
b1ackviking: Sure, I will push that in the next commit. Thanks!

kostaond: Thank you for the fix. However, could you please squash the commits into one as stated in the link provide above?

b1ackviking: I squashed the commits now.

kostaond (approved):

kostaond:
sha=2b944ca056849a6c5c1fbd105a8b8936a9088e54

PR #7966, opened 2021-11-25
Changed argument pointer to correct data (IDFGH-6303)

tbravenec:
The queue packet function expects the header and the payload of the packet. recv_buf contains also the rx_ctrl structure resulting in writing incorrect data into the pcap file


Alvin1Zhang:
Thanks for your contribution.

kostaond (approved):
This is really a mistake. Thank you for the fix!

kostaond:
sha=25934167bd2c8d5c56fbcf19455b625b9763c3e8

PR #7971, opened 2021-11-26
bugfix: Conflicting declaration error when include<rom/secure_boot.h> (IDFGH-6308)

lovyan03:
Conflicting declaration error occurs between lines 28 and 31 of file rom/secure_boot.h.
This error can be reproduced by writing # include <esp_efuse.h> and building ESP32S2 or ESP32C3 as the target.
```
C:/esp-idf/components/esp_rom/include/esp32c3/rom/secure_boot.h:31:42: error: conflicting declaration 'typedef struct ets_secure_boot_signature ets_secure_boot_signature_t'
 typedef struct ets_secure_boot_signature ets_secure_boot_signature_t;
                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
C:/esp-idf/components/esp_rom/include/esp32c3/rom/secure_boot.h:28:8: note: previous declaration as 'struct ets_secure_boot_signature_t'
 struct ets_secure_boot_signature_t;
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This problem is caused by the extra "_t" at the end of the name, as shown below.
```
struct foo_t;
typedef struct foo foo_t;
```
This pull request removes the trailing _t from " struct ets_secure_boot_signature_t; " on line 28.

### DIFF
--- a/components/esp_rom/include/esp32c3/rom/secure_boot.h
+++ b/components/esp_rom/include/esp32c3/rom/secure_boot.h
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef _ROM_SECURE_BOOT_H_
-#define _ROM_SECURE_BOOT_H_
+#pragma once
 
 #include <stdint.h>
 #include <stdbool.h>
@@ -23,9 +22,6 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
-
-struct ets_secure_boot_sig_block;
-struct ets_secure_boot_signature_t;
 
 typedef struct ets_secure_boot_sig_block ets_secure_boot_sig_block_t;
 typedef struct ets_secure_boot_signature ets_secure_boot_signature_t;
@@ -127,5 +123,3 @@ struct ets_secure_boot_key_digests {
 #ifdef __cplusplus
 }
 #endif
-
-#endif /* _ROM_SECURE_BOOT_H_ */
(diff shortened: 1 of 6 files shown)


components/esp_rom/include/esp32c3/rom/secure_boot.h
```diff
@@ -25,7 +25,7 @@ extern "C" {
 #endif
 
 struct ets_secure_boot_sig_block;
-struct ets_secure_boot_signature_t;
+struct ets_secure_boot_signature;
```
mahavirj, line 28: I would suggest that we remove both these lines. Followup `typedef` should take care of appropriate forward declaration. Can you please check once and update this PR accordingly? @lovyan03

mahavirj (changes requested):

components/esp_rom/include/esp32c3/rom/secure_boot.h
lovyan03: @mahavirj 
thanks !
I have tested your suggestion.
You are indeed right, removing those two lines solves the error.

By the way, this has nothing to do with this directly, Only the esp32s3 file has the include guard set to " pragma once ".
For other processors, it is " ifndef _ROM_SECURE_BOOT_H_ ".
Is there any intention to do this?
If you don't mind, I'd like to unify this in this pull request as well.

mahavirj: @lovyan03 

Makes sense! Please update for `pragma once` in all headers. Thanks.

lovyan03: @mahavirj 
thanks!
I have completed the fix. Please confirm.

mahavirj (approved):

mahavirj:
sha=cb91b51c52c2dbdb3e063238293c41425b739d41

PR #8001, opened 2021-12-02
Added results count to MDNS (IDFGH-6340)

0xFEEDC0DE64:
I was missing the number of results in mdns_query_async_get_results()

### DIFF
--- a/components/mdns/include/mdns.h
+++ b/components/mdns/include/mdns.h
@@ -544,7 +544,7 @@ esp_err_t mdns_query_async_delete(mdns_search_once_t* search);
  *      True if search has finished before or at timeout
  *      False if search timeout is over
  */
-bool mdns_query_async_get_results(mdns_search_once_t* search, uint32_t timeout, mdns_result_t ** results);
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/mdns/include/mdns.h
```diff
@@ -544,7 +544,7 @@ esp_err_t mdns_query_async_delete(mdns_search_once_t* search);
  *      True if search has finished before or at timeout
  *      False if search timeout is over
  */
-bool mdns_query_async_get_results(mdns_search_once_t* search, uint32_t timeout, mdns_result_t ** results);
+bool mdns_query_async_get_results(mdns_search_once_t* search, uint32_t timeout, mdns_result_t ** results, uint8_t * num_results);
```
david-cermak, line 547: This is an API update, but since this function has been recently added, we could break the API now in `v5.0`

david-cermak (approved):
Thanks for posting this PR, LGTM. Special thanks for adding the NULL checks, which we've missed

PR #8024, opened 2021-12-06
Add esp_http_client_set_timeout_ms()

0xFEEDC0DE64:
I needed to change the timeout_ms of a http client.


0xFEEDC0DE64:
Seems like this has already been implemented in the esp-idf :)

PR #8035, opened 2021-12-08
Don't return a hard coded errno in vfs_fat_access (IDFGH-6376)

winzkigermany:
Why should the return value of vfs_fat_access be hard coded as ENOENT?
This hides errors like FR_DISK_ERR of the lower layers.


winzkigermany:
I'll reopen this pull request.

PR #8036, opened 2021-12-08
Don't overwrite errno by a hard coded ENOENT (IDFGH-6378)

winzkigermany:
Calling "open" in CHECK_AND_CALL sets a perfectly correct errno.
There is no need to overwrite that with a value of ENOENT, since doing so hides lower level errors like EIO

### DIFF
--- a/components/vfs/vfs.c
+++ b/components/vfs/vfs.c
@@ -415,7 +415,7 @@ int esp_vfs_open(struct _reent *r, const char * path, int flags, int mode)
         __errno_r(r) = ENOMEM;
         return -1;
     }
-    __errno_r(r) = ENOENT;
+    __errno_r(r) = errno;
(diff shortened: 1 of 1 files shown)


igrr:
@winzkigermany could you please describe the situation you are seeing this issue in? Just so that we can add a unit test to check this behavior...

winzkigermany:
The real world scenario is: The user pulls an SD card out (we're using sdmmc), and after that, you just try to fopen(...) a file.
As a result, errno contains ENOENT instead of EIO, because ENOENT is hard coded.

igrr:
Thank you for the PR @winzkigermany! It has been merged in https://github.com/espressif/esp-idf/commit/f01b9a583e475999fadff5fa106a2d98b595d1d1

PR #8068, opened 2021-12-12
Feature/psram malloc (IDFGH-6410)

Eschel:
Still learning


igrr:
@Eschel it looks like either you have picked the wrong source branch or the wrong target branch for this PR. I'm going to close it since it looks like it was opened by mistake. Please leave a comment if that's not the case.

PR #8076, opened 2021-12-13
example: Synchronize includes in BLE walkthrough (IDFGH-6417)

FrNecas:
While following one of the examples for BLE GATT server, I got stuck on an issue with includes where my code could not be compiled. After some struggling I found out that the walkthrough which I was partly taking code from had wrong includes for the BT headers (the source code for the example is correct).

I am completely new to ESP IDF so I am not sure whether the headers used to be correct and perhaps they were renamed later without reflecting the changes into the guide or whether the guide was wrong in the first place. I just found it a bit confusing, since this document is even linked from the  official docs :) The guide should now be the same as the source file in this regard.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=7277e6167234ffb975a57bb1f02ba89ce09eb2b6

PR #8093, opened 2021-12-16
Unified log print errno format (IDFGH-6434)

Caffreyfans:
```
Hi developer,
    I find there is some different log print about errno. Some places use errno and some places use Error. 
I'm not sure it makes sense.

Best Regards,
Caffreyfans
```


Alvin1Zhang:
Thanks for your contribution.

david-cermak (approved):
Yes, we should be more consistent in printing error messages.
Thanks for the fix!

Caffreyfans:
Sorry I forgot to switch to a new branch to submit a new pull request.

PR #8100, opened 2021-12-17
monitor: Print buffer when reading an escape character (IDFGH-6443)

bugadani:
The color converter script currently discards partially processed data when it encounters an escape character. This can result in missing characters when the user prints escape sequences that are not expected by the script. This PR fixes this issue.

PR #8101, opened 2021-12-17
[4.3 backport] monitor: Print buffer when reading an escape character (IDFGH-6442)

bugadani:
See #8100


dobairoland:
Also a duplicate of https://github.com/espressif/esp-idf/pull/8100. This will be backported internally.

PR #8107, opened 2021-12-17
Fix GDB backtraces in optimized builds (IDFGH-6444)

bugadani:
If `COMPILER_OPTIMIZATION` is not set to default, the `FREERTOS_TASK_FUNCTION_WRAPPER` option is ignored and wrapper functions are not generated. This makes backtraces in coredumps crash GDB, which makes `espcoredump.py` hang (see #8105).

### DIFF
--- a/components/freertos/Kconfig
+++ b/components/freertos/Kconfig
@@ -421,7 +421,7 @@ menu "FreeRTOS"
 
     config FREERTOS_TASK_FUNCTION_WRAPPER
         bool "Enclose all task functions in a wrapper function"
-        depends on COMPILER_OPTIMIZATION_DEFAULT
+        depends on COMPILER_OPTIMIZATION_DEFAULT || ESP_COREDUMP_ENABLE || ESP_GDBSTUB_ENABLED
         default y
         help
             If enabled, all FreeRTOS task functions will be enclosed in a wrapper function.


Alvin1Zhang:
Thanks for your contribution.

igrr:
@bugadani I think this was intentional — debugging functionality such as espcoredump is intended to be used in debug builds, and in release builds we didn't want to add extra overhead to the application.
Do you have some case when you need coredump functionality in a release build?

bugadani:
Yes, I have a remote crash reporting functionality that should work in release builds.

The task wrappers can still be disabled, it's just not enforced after this PR. It's not a huge deal if this can't be accepted, as I can patch esp-idf in my build system, but I am biased toward more flexibility instead of forcing my code to break (silently!) depending on the optimization level.

igrr:
Thanks, understood. What do you think about something along these lines?
```
    config FREERTOS_TASK_FUNCTION_WRAPPER
        bool "Enclose all task functions in a wrapper function"
        depends on COMPILER_OPTIMIZATION_DEFAULT || CONFIG_ESP_COREDUMP_ENABLE
        default y
        help
```

bugadani:
I feel like this option should be inverted (which I understand is a breaking change and most likely won't be done). Task wrapper functions are a necessity in some cases, and removing them is an optimization normally. This suggests to me that a "no task function wrappers" option is more natural, and that should be disabled if coredumps or related functionality is required.

What you propose is good enough for me, but I still feel it overly restrictive as it still forces a feature off in certain cases. For example, I might be developing my app using `COMPILER_OPTIMIZATION_PERF` or `SIZE` as I might have flash size restrictions and building with `COMPILER_OPTIMIZATION_DEFAULT` might not fit. In this case, if I don't need coredumps or gdb stub, I can't enable task wrappers (which don't really take up that much space I think), thus I might lose a feature that's useful for debugging.

igrr:
We may actually make it a breaking change (invert the option) since IDF is in v5.0-dev now.

But for the existing release branches (v4.4, v4.3) we would need to have some non-breaking fix.

bugadani:
> But for the existing release branches (v4.4, v4.3) we would need to have some non-breaking fix.

Sounds fair.

What are the downsides in removing the `depends` line entirely, besides a quiet increase in program size for some users? Is that size increase something we should avoid? In my project the difference is 150 bytes or so. I don't want to push my opinion, but I don't think I quite understand the reason this depends was originally added.

igrr:
I think the reason was that to avoid silently breaking user programs where the stack usage was very close to the total available stack size. I.e. it's one additional stack frame that was the issue, not the extra binary size.
In 5.0 we can make this change, but in release branches this is a bit more risky.

bugadani:
Alright, that makes sense, thanks!

components/freertos/Kconfig
```diff
@@ -421,7 +421,7 @@ menu "FreeRTOS"
 
     config FREERTOS_TASK_FUNCTION_WRAPPER
         bool "Enclose all task functions in a wrapper function"
-        depends on COMPILER_OPTIMIZATION_DEFAULT
+        depends on COMPILER_OPTIMIZATION_DEFAULT || CONFIG_ESP_COREDUMP_ENABLE || CONFIG_ESP_GDBSTUB_ENABLED
```
igrr, line 424: Sorry, this should be `|| ESP_COREDUMP_ENABLE || ESP_GDBSTUB_ENABLED` — inside Kconfig files all options are referenced without `CONFIG_` prefix.

igrr (changes requested):

igrr (approved):

igrr:
sha=a02d1e35ee0eccd6eaee5f5aa2832415bbd4415e

PR #8134, opened 2021-12-22
wpa_supplicant: Fix checking signal level in compare_scan_neighbor_re… (IDFGH-6477)

AxelLin:
…sults

Current logic is wrong because it makes the target acceptable
when "target->level < bss->level && target->level >= -80".


kapilkedawat (commented):
If our current bss's rssi is -90 and we find a bss with -80, we should consider it valid candidate.  Also this doesn't match with upstream supplicant and we don't want to deviate from upstream. Incase you still think its a valid change, Please submit to it wpa_supplicant/hostapd.

Issue #7528, opened 2021-09-06
How do I add non mesh connections to esp_mesh  (IDFGH-5823)


clementfumey:
## Environment

- Module or chip used:  ESP32-WROOM-32
- IDF version : v4.4-dev-2875-g5f38b766a8 (master)
- Build System:       CMake
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it):
    // 1.22.0-80-g6c4433a
- Operating System:    Linux
- Using an IDE?: Yes using Visual Studio Code with ESP-IDF extension
- Power Supply:         USB

## Problem Description

Based on recently  [merged commit](https://github.com/espressif/esp-idf/commit/42a1be10db3fdba1c7f2ba4b8a542bc72371ad6e) from @shenjun7 to add non mesh connections access to esp_mesh I am trying to give internet access to an external (non-mesh) device through an ESP-MESH network.

I am using ip_internal_network examples from mesh folder, where I deleted all mqtt stuff to have only a bare esp-mesh network.
I am only using two nodes in my mesh network, 1 that become root node and 1 I will use the AP to connect with my external device.
I set the configuration to `CONFIG_MESH_NON_MESH_AP_CONNECTIONS=2` to enable non-mesh connection. 
I found that mesh ap hidden ssid was ESPM_XXXXXX, with XXXXXX the last mac address bytes, so I used it to connect from my external device.

### Expected Behavior

I expect a root node to be chosen, and successful connection to router, and IPv4 addressed to be assigned to both root (10.0.0.1) and child node (10.0.0.2). When external device connect, I expect an IPv4 address to be assigned and IP packet forwarded to router.

### Actual Behavior
Root node is chosen, successful connection to router, IPv4 assigned to both root and child node. But no IPv4 assigned to non-mesh device. Very few logs about external devices trying to connect.

## Debug Logs

Logs from the child node when the external device connect and leave somes seconds later :
```
I (329574) wifi:new:<13,2>, old:<13,2>, ap:<13,2>, sta:<13,2>, prof:13
I (329574) wifi:station: 38:78:62:5a:81:ed join, AID=1, bgn, 20
W (329654) wifi:<ba-add>idx:4 (ifx:1, 38:78:62:5a:81:ed), tid:0, ssn:0, winSize:64
dhcp_coarse_tmr()
dhcp_coarse_tmr()
dhcp_coarse_tmr()
dhcp_coarse_tmr()
dhcp_coarse_tmr()
[...]

PR #8158, opened 2021-12-28
Non-mesh ip network example (IDFGH-6505)

wizche:
This is an attempt to "fix" #7528

EDIT2 (17.01.2021): Well I was completely wrong so here my findings after investigating this more.

Adapting the `internal_communication` wont work because even if we dont start the DHCP server/client the request are not relayed to the root node and therefore we cannot use the root node as the only DHCP server. As @shenjun7 pointed out _Root as the only dhcp server to allocate IP to non-mesh nodes requires a lot of changes, we will update in the future_.
Starting a DHCP server on every node is also not feasible since we will have duplicates IP across the (mesh) network. 

So basically the only solution I could find is based on my initial idea of adapting `ip_internal_network`. The idea is to have the entrypoint node (the one where non-mesh device will connect to) to reconfigure its soft-ap part to expose a different SSID.
Specifically I introduce the `ap` console command that will reconfigure the node in the following way:
- Reconfigure SoftAP to expose a new SSID that will be used by non-mesh clients
- Enable DHCP server on SoftAP
- Enable PNAT on the SoftAP interface (traffic from the non-mesh clients is NATed and forwarded to ST and therefore to the root node)

Even though this works, I experience major connectivity issues, not only from the non-mesh device but from the node themselves. I see packet drops, frame size error (when encapsulating into the mesh packet) and general instability. 
You can check the README in `nonmesh_ip_internal_network` for additional details.


Alvin1Zhang:
Thanks for your contribution.

wizche:
I kept the `nonmesh_ip_internal_network` example as lesson learned but if you think this branch will be merged without it I will just remove it (or maybe open another PR just for that?)

shenjun7:
Thanks for your contribution on ESP-MESH, we will check your code and give you feedback soon.

zhangyanjiaoesp:
@wizche Can you please rebase the PR and remove the `nonmesh_internal_communication` example ?

Sherry616:
Closing due to lack of reply. Feel free to submit a newer PR if you'd still prefer us to consider these changes. Thanks.

PR #8162, opened 2021-12-29
esp_wifi: Fix error message for rtc apb clk frequency (IDFGH-6509)

AxelLin:
Calling rtc_clk_apb_freq_get() in ESP_LOGE may return different freq value
from the value used in condition checking.
Avoid below message which is confusing because 80MHz == 80000000Hz.
E (37138888) wifi_apb80m_request: WiFi needs 80MHz APB frequency to work, but got 80000000Hz

### DIFF
--- a/components/esp_wifi/src/wifi_init.c
+++ b/components/esp_wifi/src/wifi_init.c
@@ -274,10 +274,12 @@ esp_err_t esp_wifi_init(const wifi_init_config_t *config)
 #ifdef CONFIG_PM_ENABLE
 void wifi_apb80m_request(void)
 {
+    uint32_t rtc_clk_apb_freq;
     assert(s_wifi_modem_sleep_lock);
     esp_pm_lock_acquire(s_wifi_modem_sleep_lock);
-    if (rtc_clk_apb_freq_get() != APB_CLK_FREQ) {
-        ESP_LOGE(__func__, "WiFi needs 80MHz APB frequency to work, but got %dHz", rtc_clk_apb_freq_get());
+    rtc_clk_apb_freq = rtc_clk_apb_freq_get();
+    if (rtc_clk_apb_freq != APB_CLK_FREQ) {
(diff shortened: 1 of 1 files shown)


esp-lis:
This issue should only appear on the release/v4.3 version, and it is a dual-core system. If it is, please modify the **$IDF_PATH/components/esp_pm/pm_impl.c** as shown in the figure below for testing.

Please feedback the test results to me, I will decide to add this fix on release/v4.3, Thank you!

AxelLin:
Hi @esp-lis 

1. With your fix, the error message no longer happen on v4.3. thanks.
   ps. I'd appreciate if you post a diff file rather than png file in the future support.
2. I think this PR is still valid.

esp-lis:
> Hi @esp-lis
> 
> 1. With your fix, the error message no longer happen on v4.3. thanks.
>    ps. I'd appreciate if you post a diff file rather than png file in the future support.
> 2. I think this PR is still valid.

In fact, the RF PHY has requirements for the system frequency. The design requires that after the `wifi_apb80m_request` is executed, the APB bus frequency must be 80 MHz, otherwise it will affect the RF.

AxelLin:
> In fact, the RF PHY has requirements for the system frequency. The design requires that after the `wifi_apb80m_request` is executed, the APB bus frequency must be 80 MHz, otherwise it will affect the RF.

But that is nothing to do with this PR.
The point of this PR is to avoid calling rtc_clk_apb_freq_get() twice which
may cause confusing message.

AxelLin:
> Hi @esp-lis
> 
> 1. With your fix, the error message no longer happen on v4.3. thanks.

@esp-lis  I hope the fix will be included in v4.3.3 release.

AxelLin:
> This issue should only appear on the release/v4.3 version

@esp-lis 
This fix is in v4.1 0f86d752e5ab56538349f80a9ce9941df18a02de , v4.4 d3b5f975739b7179b614d05a46d5147bfe430039 and master d3b5f975739b7179b614d05a46d5147bfe430039.
**However, the fix is NOT in v4.3.**

PR #8164, opened 2021-12-29
Replace lwIP byte order functions with built-in functions (IDFGH-6512)

josesimoes:
- This replaces the default software implementation (less efficient) with compiler builtin functions (much efficient and possibly assembler implemented).
- These byte order swap functions are used intensively and are on top of the list of the recommended improvements for performance. Please see [Detailed Description here](http://lwip.nongnu.org/2_1_x/def_8c.html) and [here](https://www.nongnu.org/lwip/2_1_x/optimization.html).


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=dd9fe9761ac714b1ea60f3b4ad4226c7ef1a288b

PR #8169, opened 2021-12-29
fix fatfs_create_spiflash_image() feature (IDFGH-6383) (IDFGH-6518)

### LINKED ISSUE #8042: 2 Bugs in new fatfs_create_spiflash_image() feature (IDFGH-6383)
## Environment

- Development Kit:      none
- Kit version (for WroverKit/PicoKit/DevKitC): none
- Module or chip used:  ESP32-WROOM-32
- IDF version: master at 2021-12-9
- Build System:         idf.py
- Compiler version : (crosstool-NG esp-2021r2) 8.4.0
- Operating System:     Windows
- (Windows only) environment type:ESP Command Prompt
- Using an IDE?: No
- Power Supply:  USB

## Problem Description

When I use the new fatfs_create_spiflash_image() feature in fatfs componenet, like following usage in CMakeLists.txt:
--------------------begining--------------------------------
fatfs_create_spiflash_image(resource resource FLASH_IN_PROJECT)
---------------------end-------------------------------

Then something got wrong like following logs at the building time:
-----------------------------begining------------------------------------------------
  File "D:/ESPIDF/components/fatfs/wl_fatfsgen.py", line 196, in <module>
    wl_fatfs.wl_generate(input_dir)
  File "D:/ESPIDF/components/fatfs/wl_fatfsgen.py", line 175, in wl_generate
    self.plain_fatfs.generate(input_directory=input_directory)
  File "D:\ESPIDF\components\fatfs\fatfsgen.py", line 190, in generate
    self._generate_partition_from_folder(folder_name, folder_path=path_to_folder, is_dir=True)
  File "D:\ESPIDF\components\fatfs\fatfsgen.py", line 183, in _generate_partition_from_folder
    self._generate_partition_from_folder(os.path.join(smaller_path, path), folder_path=folder_path)
  File "D:\ESPIDF\components\fatfs\fatfsgen.py", line 183, in _generate_partition_from_folder
    self._generate_partition_from_folder(os.path.join(smaller_path, path), folder_path=folder_path)
  File "D:\ESPIDF\components\fatfs\fatfsgen.py", line 171, in _generate_partition_from_folder
    content = file.read()
UnicodeDecodeError: 'gbk' codec can't decode byte 0x83 in position 1418: illegal multibyte sequence
----------------------------------end--------------------------------
it seens that fatfs partition generator wants to encode my raw file and failed.
I modified the file wl_fatfsgen.py, at line 168, added "wb" instruction as  "with open(output_path, 'wb') as output:", and modified the file fs_object.py, at line 50, removed "encode()" method  as "content_as_list = content_part", then it seens to work right.

But why does fatfs generator want to encode a file being put to a partition, is it enough to just treat the file as an unkown binary raw file.

Another bug is that the created partion burned into flash, will pollute the partition just next to it, cause mounting failure like following logs:
--------------------------begining-------------------------
W (260) vfs_fat_spiflash: f_mount failed (13) 
--------------------------end-------------------------

I reserved a gap between those two partition, the problem seens to be temporary solved.

tamisoft:
Open the files in binary mode, and remove the redundant encode.

Fixes #8042

### DIFF
--- a/components/fatfs/fatfsgen.py
+++ b/components/fatfs/fatfsgen.py
@@ -167,7 +167,7 @@ class FATFS:
         normal_path = os.path.normpath(folder_relative_path)
         split_path = normal_path.split(os.sep)
         if os.path.isfile(real_path):
-            with open(real_path) as file:
+            with open(real_path, "rb") as file:
                 content = file.read()
             file_name, extension = os.path.splitext(split_path[-1])
             extension = extension[1:]  # remove the dot from the extension
--- a/components/fatfs/fatfsgen_utils/fs_object.py
+++ b/components/fatfs/fatfsgen_utils/fs_object.py
@@ -47,7 +47,7 @@ class File:
         # we assume that the correct amount of clusters is allocated
         current_cluster = self._first_cluster
         for content_part in split_content_into_sectors(content, self.fatfs_state.sector_size):
-            content_as_list = content_part.encode()
+            content_as_list = content_part
             if current_cluster is None:
                 raise FatalError('No free space left!')
 


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
sha=df2ae4689cd1056bfda67c9dbe08c015da1202aa

PR #8172, opened 2021-12-30
Fix compile issue when TinyUSB task is not enabled. (IDFGH-6521)

ghent360:
Specifying CONFIG_TINYUSB_NO_DEFAULT_TASK in the project configuration produces errors in this code because
CONFIG_TINYUSB_TASK_STACK_SIZE and CONFIG_TINYUSB_TASK_PRIORITY
are not defined.

Conditionally compile this file only if CONFIG_TINYUSB_NO_DEFAULT_TASK is false.

### DIFF
--- a/components/tinyusb/additions/src/tusb_tasks.c
+++ b/components/tinyusb/additions/src/tusb_tasks.c
@@ -20,6 +20,8 @@
 #include "tinyusb.h"
 #include "tusb_tasks.h"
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=8e3ad708b2b8f1f7004fe0e23926672f6d5d118c

PR #8188, opened 2022-01-04
Add twai queue size observers (IDFGH-6537)

0xFEEDC0DE64:
I need to debug strange TWAI freezes on the esp32 and therefore I wanted to log the queue sizes of rx and tx queues,

I could observe that seconds before the esp32 is stuck the TX queue goes up to 5 or 6.

### DIFF
--- a/components/driver/include/driver/twai.h
+++ b/components/driver/include/driver/twai.h
@@ -223,6 +223,30 @@ esp_err_t twai_stop(void);
 esp_err_t twai_transmit(const twai_message_t *message, TickType_t ticks_to_wait);
 
 /**
+ * @brief   Get the number of pending messages in TX queue
+ *
+ * @param[out]   pending_tx         Number of messages
+ *
+ * @return
+ *      - ESP_OK: Transmission successfully queued/initiated
+ *      - ESP_ERR_INVALID_ARG: Arguments are invalid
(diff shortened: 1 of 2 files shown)


Dazza0:
@0xFEEDC0DE64 You can already get the current number of pending messages in `twai_status_info_t.msgs_to_tx` and `twai_status_info_t.msgs_to_rx`.

Did you check the TX error count to see if your messages are getting acknowledged?

0xFEEDC0DE64:
I just noticed the can status going to TWAI_STATE_BUS_OFF, when the can stops working.
This is until powered off and rebooted.
Also it seems that this problem happens since we rebased the esp-idf from a about a year old one to the latest master.
Every time this happens my little vehicle goes into full emergency mode and brakes down from 45kmh to 0kmh and almost kills me :)
for the next accident, should I try to call stop and start again or re initialize the complete can driver?
also going back to the older idf might be a fix for now, I would like to pinpoint the idf revision from which this behaviour starts

Dazza0:
@0xFEEDC0DE64 Once you enter a BUS OFF state, you can either uninstall the driver, or attempt to recover from the BUS OFF by calling `twai_initiate_recovery()`. See the [docs](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html) for more details

PR #8224, opened 2022-01-11
Reduce error log when calling esp_efuse_mac_get_custom() (IDFGH-6580)

0xFEEDC0DE64:
Our firmwares always try to find out, if a custom mac address has been stored in production, and the esp-idf itself logs an error if none is found.

I tried to reduce the log level to DEBUG when version equals to 0. It still logs an error if a newer (or invalid) version is found

### DIFF
--- a/components/esp_hw_support/mac_addr.c
+++ b/components/esp_hw_support/mac_addr.c
@@ -86,7 +86,12 @@ esp_err_t esp_efuse_mac_get_custom(uint8_t *mac)
     uint8_t version;
     esp_efuse_read_field_blob(ESP_EFUSE_MAC_CUSTOM_VER, &version, 8);
     if (version != 1) {
-        ESP_LOGE(TAG, "Base MAC address from BLK3 of EFUSE version error, version = %d", version);
+        // version 0 means has not been setup
+        if (version == 0) {
+            ESP_LOGD(TAG, "No base MAC address in eFuse (version=0)");
+        } else if (version != 1) {
+            ESP_LOGE(TAG, "Base MAC address version error, version = %d", version);
+        }
         return ESP_ERR_INVALID_VERSION;
     }
 


components/esp_hw_support/mac_addr.c
```diff
@@ -86,7 +86,8 @@ esp_err_t esp_efuse_mac_get_custom(uint8_t *mac)
     uint8_t version;
     esp_efuse_read_field_blob(ESP_EFUSE_MAC_CUSTOM_VER, &version, 8);
     if (version != 1) {
-        ESP_LOGE(TAG, "Base MAC address from BLK3 of EFUSE version error, version = %d", version);
+        // version 0 means has not been setup
+        ESP_LOG_LEVEL_LOCAL((version == 0 ? ESP_LOG_DEBUG : ESP_LOG_ERROR), TAG, "Base MAC address from BLK3 of EFUSE version error, version = %d", version);
```
igrr, line 90: Instead of a ternary operator I would suggest a (slightly more verbose but readable)
```c++
if (version == 0) {
    ESP_LOGD(TAG, "No base MAC address in eFuse (version=0)");
} else if (version != 1) {
    ESP_LOGE(TAG, "Base MAC address version error, version = %d", version);
}
```

(i have removed "from BLK3" since that is an ESP32-specific leftover; other chips may have custom MAC address written in another eFuse block)

igrr (changes requested):

igrr:
Thank you for addressing my comment @0xFEEDC0DE64! If you could squash your two commits into one, it would be excellent. If not, I'll squash them later before merging.

0xFEEDC0DE64:
Any update on this? @igrr ?

mahavirj:
sha=68875dfc82932a7c8617f20bd44abe80dc765c85

PR #8235, opened 2022-01-13
Added compiler optimization flag -Ofast (IDFGH-6591)

0xFEEDC0DE64:
Needed this for a cpu intensive firmware project, where I could Increase the calculation frequency from 24000 to 27100 (which is an increase by 12%)


igrr (commented):
Thanks for the PR @0xFEEDC0DE64! I have a few questions to better understand the impact.

What do you think about using -Ofast for `CONFIG_COMPILER_OPTIMIZATION_PERF`, if -Ofast it produces more performant code?

What is the approximate difference in binary size (in %) between -O2 and -Ofast for your application?

Do you think the calculation performance for your application could be similarly improved by using -Ofast for specific files or libraries (using target_compile_options, set_source_file_properties), without introducing a new user-facing option?

greenaddress:
@igrr I think -O2 and -Ofast should be separate options like this PR does rather than replace the performance option (afaik -Ofast is a super set of -03)

igrr:
@greenaddress Thanks for your comment. What I'm getting at is, how useful is -Ofast as a global option in the SDK? I understand you may want to compile certain performance-critical parts of the code with this option, but does it make sense to compile everything with -Ofast?
There are two things I would consider as part of "does it make sense" question:
1. What is the code size increase with that option, compared to -Os and -O2?
   (I suspect that it's not too bad — on simple MCUs like ESP32 is, code size is generally a pretty good proxy for execution performance. But I'd like to confirm that.
2. How much extra work is it to ensure that this option works for all our components? I see that an additional build is added in tools/ci/build_template_app.sh, but we might also need to run some tests on hardware with this option enabled.

As an alternative, instead of adding -Ofast as an option selectable in menuconfig, we could consider documenting it in the [guide about optimizing for performance](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/performance/speed.html), recommending the users to enable it selectively for the performance-critical components.

zikalino:
Closing this PR for now. Please feel free to continue discussion and reopen.

PR #8247, opened 2022-01-17
LEDC - expose ledc_get_max_duty() to be accessible in user code

vortigont:
Let user code to access ledc_get_max_duty() function and obtain current max duty value for the specific channel.
Otherwise it might be very tricky to implement it without accessing hal.
I'm not sure about the naming convention for internal functions, let me know if it needs to be changed.

### DIFF
--- a/components/driver/include/driver/ledc.h
+++ b/components/driver/include/driver/ledc.h
@@ -245,6 +245,20 @@ esp_err_t ledc_set_duty(ledc_mode_t speed_mode, ledc_channel_t channel, uint32_t
 uint32_t ledc_get_duty(ledc_mode_t speed_mode, ledc_channel_t channel);
 
 /**
+ * @brief LEDC get maximum channel duty
+ *
+ * Channel maximum duty depends on timer resolution where ledc channel is attached to, ((2^LEDC_TIMER_BIT) - 1)
+ *
+ * @param speed_mode Select the LEDC channel group with specified speed mode. Note that not all targets support high speed mode.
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=855e20b5a6ad8b655f528732c6d49978432dd8c6

songruo:
Hi @vortigont,

When you call `ledc_timer_config`, the `duty_resolution` parameter in `ledc_timer_config_t` is the one that determines the maximum duty value. 

Since `duty_resolution` is provided by the users, it is very simple to get the maximum duty value: 

Maximum duty value = 1 << duty_resolution

I don't think there is a need to use an API to read the duty_resolution from the register and then do the calculation.

vortigont:
Hi @songruo,
well, it depends on user's task and code. In such case user has to keep `duty_resolution` stashed somewhere else in a variable in his code just in case he'll need to calculate `max_duty`. Besides timer config could be changed from another thread or task being attached to another channel. So user's variable is one case and actual reg value is absolutely another.
More to this the code for `ledc_get_max_duty()` has been implemented already, this is just exposing it to user in a safe way. I think there is no harm, same way as we have a similar function `ledc_get_duty()`.
Currently I'm working on a lib code utilizing LEDC and calculationg `max_duty` in a way you suggested - as a stashed variable. It works fine, but it does not look as a generic robust solution.
Pls, consider it as a nice feature to expose for end user. :)
Tnx!

vortigont:
still hope we can get this anytime soon :) tnx!

PR #8248, opened 2022-01-17
Fix random crash with Plex mdns queries on my network (IDFGH-6606)

0xFEEDC0DE64:
While i was watching movies all my esp32 based projects crash at the same time every 30minutes. For now I only have the uploaded coredump stack traces the esp's.

This "fix" seems to avoid the crash, but I do not understand why it fixes it.

According sentry entry

image

### DIFF
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1552,6 +1552,8 @@ static bool _mdns_service_match_ptr_question(const mdns_service_t *service, cons
         }
         return false;
     }
+    if (question->host && !service->instance)
+        return false;
     if (question->host) {
         if (strcasecmp(service->instance, question->host) != 0) {
             return false;


peterpoetzi:
I managed to dump the packet that very likely crashed the ESP32
image

@espressif I can privately share the pcap file if you are interested

Alvin1Zhang:
Thanks for your contribution.

david-cermak:
@0xFEEDC0DE64 Thanks for posting the report with the potential fix!

@peterpoetzi would be much appreciated if you can send me the captured packet to my `git`-email:
```bash
git show 9fc9521 | grep Author
```
Thank you!

david-cermak:
Thanks for sharing the packet logs, I was able to reproduce the issue.

The problem is that mDNS service is registered without an instance (which is supported and the general instance name or hostname is used instead):
```cpp
    ESP_ERROR_CHECK( mdns_instance_name_set("general-instance") );
    ESP_ERROR_CHECK( mdns_service_add(NULL, "_http", "_tcp", 80, NULL, 0) );
```
Then, when posting a query for that specific host/instance name, the mdns library incorrectly tries to locate service's instance rather then the global instance. In that sense the proposed fix would be:
```patch
diff --git a/components/mdns/mdns.c b/components/mdns/mdns.c
index 0faa0b509d6..909250e7c72 100644
--- a/components/mdns/mdns.c
+++ b/components/mdns/mdns.c
@@ -1553,7 +1553,7 @@ static bool _mdns_service_match_ptr_question(const mdns_service_t *service, cons
         return false;
     }
     if (question->host) {
-        if (strcasecmp(service->instance, question->host) != 0) {
+        if (strcasecmp(_mdns_get_service_instance_name(service), question->host) != 0) {
             return false;
         }
     }
```
The fix proposed in this PR solve the crash, but wouldn't correctly respond (would never reply).

Is this the way you register the services, using default instance names?

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/f46dffca627e9578e49a510580f9754ec1e27e2e.

PR #8262, opened 2022-01-19
Websocket client example with Python Flask echo server (IDFGH-6619)

dizcza:
The `echo.websocket.org` service is no longer available. See https://www.lob.com/blog/websocket-org-is-down-here-is-an-alternative.

Changes:
1. `echo.websocket.org` -> `echo.websocket.events`
2. Added a minimalistic Python Flask socket echo server example to connect to a local websocket endpoint.


Alvin1Zhang:
Thanks for your contribution.

gabsuren (changes requested):
Thank you for your PR
Please remove Flask echo server, as we use WebSocketClient in our example. The other part looks good

dizcza:
> Please remove Flask echo server, as we use WebSocketClient in our example.

But that's really the contribution I want to share with others. Flask echo server works just fine with the WebSocketClient. See the readme.

Moreover, I wanted to ask a question on ESP32 forum with the very same question - how to write a Python Flask app that communicates with an ESP32 via a websocket.

david-cermak:
> Flask echo server works just fine with the WebSocketClient. See the readme.

Could you simply place the python code inside the README, then? Sorry, we don't document it anywhere, but the python scripts in example directory have special meanings and are executed in the CI to verify the examples.

We can replace the current websocket server in the `example_test.py`, but I'd suggest opening a separate PR for that.

dizcza:
Done.

Alvin1Zhang:
Thanks for your contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/23598dfcec0ae4788631f619fac9499fea5adfdb.

PR #8291, opened 2022-01-26
Ethernet basic example: Set default handlers for netifs (IDFGH-6655)

mkfrey:
Commit ef3038490275d02dee46319f9c596d807163213f removed the call to `esp_eth_set_default_handlers()`. Thich causes the example to not fully work (e.g. no DHCP).

This PR adds the required calls again.


Alvin1Zhang:
Thanks for your contribution.

mkfrey:
Thanks for your answer @KaeLL. You're right, I used the example from master with IDF 4.3.2 and didn't consider that a change like this happened. So the error was on my part.

kostaond:
@KaeLL thanks for helping to resolve this issue!

PR #8312, opened 2022-01-28
Enable I2C Slave mode with input-only pin as SCL. (IDFGH-6682)

jkearins:
If we use I2C in slave mode, we can use input-only pin for SCL line (GPI 34-39).
But there is a problem in i2c_set_pin() which is called by i2c_param_config().
In case of using GPI as SCL the following statements are failed:
	1. gpio_set_level(scl_io_num, I2C_IO_INIT_LEVEL);
Function fails because it cannot set level for input pin. As a result a ESP_LOGE() is called.
	2. gpio_set_direction(scl_io_num, GPIO_MODE_INPUT_OUTPUT_OD);
Function fails because it accepts only GPIO pins. ESP_LOGE() is called, input mode is not enabled for the pin and as a result - no interrupts from the pin.


ginkgm:
Thanks @jkearins ,
This is good proposal. We will try to get this into IDF.

L-KAYA:
We are planning to support stretch mode in the future, thus we better to remain SCL bidirectional.

Thanks for your contribution!

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/bcf7608cf812e9140e5ac2dcb4ad4afd53fdbca7.

Issue #8326, opened 2022-01-31
I2S wrong bclk for internal DAC (IDFGH-6697)


s-hadinger:
## Environment

- Development Kit:      [M5StickC]
- Module or chip used:  [ESP32-PICO-D4]
- IDF version [4.4]
- Build System:         [Pio]
- Compiler version toolchain-xtensa-esp32 8.4.0+2021r2-patch2
- Operating System:     [macOS]
- Using an IDE?: [PlatformIO]
- Power Supply:         [USB]

## Problem Description

When using I2S with internal DAC, 2 channels, 16000 frames/sec, the sound emitted is a short swaks and DMA buffers are emptied far too fast compared to regular I2S audio with external chip.

Problem is largely described in https://github.com/espressif/arduino-esp32/issues/5938

### Expected Behavior

The original PCM sound is the following and lasts over 660ms:

![151602386-8ca194e2-ea6a-466b-8a92-b9760f4d027e](https://user-images.githubusercontent.com/49731213/151852816-ced01491-8066-44df-9b1d-8770202435ea.jpg)

### Actual Behavior

Recording the output with a scope shows a correct waveform but played 11x faster than normal speed.
![151602460-e7997b68-d8c1-40fb-9442-6978bf8e68c0](https://user-images.githubusercontent.com/49731213/151852929-fb7d2c93-80f2-450b-a7c1-50df72e8282b.jpg)

I did a detour on `i2s_hal_tx_clock_config()` to check what clock values were calculated, and force-injected some different clock values until the sound was good again.

With the previous setting: 16KHz, 2 channels, internal DAC. IDF calculates the following clock:
`sclk=160000000 mclk=128000 bclk=64000 mclk_div=1250 bclk_div=2`

Although the number seem correct, `mclk_div` hardware is only 8 bits, so 1250 is overflowing.

Injecting my own calculated value, I made it work at correct speed with:
`sclk=160000000 mclk=1024000 bclk=32000 mclk_div=156 bclk_div=32`

Having a combined `mclk_div=156 bclk_div=32` achieves a divider of ~5000 over sclk.

I suppose that an easy patch would be to set `bclk_div=32` instead of `bclk_div=2` and it should allow for all audio frequencies.

## Other items if possible

- [ ] sdkconfig file (attach the sdkconfig file from your project folder)
- [ ] elf file in the ``build`` folder (**note this may contain all the code details and symbols of your project.**)
- [ ] coredump (This provides stacks of tasks.)

[...]

PR #8327, opened 2022-01-31
Fix for mclk/bclk divisors #8326 (IDFGH-6698)

s-hadinger:
This is a first tentative to #8326

`bclk_div` is set to 32 which is the highest power of 2 possible for this hardware register (6 bits).

`bclk` target is sample_rate * I2S_LL_AD_BCK_FACTOR. I removed the additional factor 2 which looks erroneous with audio output.

### DIFF
--- a/components/driver/i2s.c
+++ b/components/driver/i2s.c
@@ -983,9 +983,9 @@ static esp_err_t i2s_calculate_adc_dac_clock(int i2s_num, i2s_hal_clock_cfg_t *c
     ESP_RETURN_ON_FALSE(p_i2s[i2s_num]->hal_cfg.mode & (I2S_MODE_DAC_BUILT_IN | I2S_MODE_ADC_BUILT_IN), ESP_ERR_INVALID_ARG, TAG, "current mode is not built-in ADC/DAC");
 
     /* Set I2S bit clock */
-    clk_cfg->bclk = p_i2s[i2s_num]->hal_cfg.sample_rate * I2S_LL_AD_BCK_FACTOR * 2;
+    clk_cfg->bclk = p_i2s[i2s_num]->hal_cfg.sample_rate * I2S_LL_AD_BCK_FACTOR;
     /* Set I2S bit clock default division */
-    clk_cfg->bclk_div = I2S_LL_AD_BCK_FACTOR;
+    clk_cfg->bclk_div = I2S_LL_AD_BCK_FACTOR * 16;
     /* If fixed_mclk and use_apll are set, use fixed_mclk as mclk frequency, otherwise calculate by mclk = sample_rate * multiple */
     clk_cfg->mclk = (p_i2s[i2s_num]->use_apll && p_i2s[i2s_num]->fixed_mclk) ?
                     p_i2s[i2s_num]->fixed_mclk : clk_cfg->bclk * clk_cfg->bclk_div;


s-hadinger:
Applying this patch locally makes Audio DAC working again.

Here are the new computed values:
`sclk=160000000 mclk=1024000 bclk=32000 mclk_div=156 bclk_div=32`

L-KAYA:
Have you test this clock setting with ADC mode? It seems can't work for ADC mode

s-hadinger:
No I haven't tested ADC mode.

L-KAYA:
Oh sorry, don't mind ADC, it's my fault. 

So now the `bclk_div` is the real `bit_per_chan`, for `sclk=160000000 mclk=1024000 bclk=32000 mclk_div=156 bclk_div=32`, we got 32 bit * 2 channle = 64 mclk ticks in one sample period. Do we need to fixed it to 32? Or set `bclk_div` equal to `bit_per_chan`, so that users can control this field.

FYI, while we are not using APLL, and set a low sample rate for ADC/DAC, the mclk_div will exceed its maximum value 255 (it has only 8 bit in reg), which can lead the clock go wrong.

L-KAYA:
> Thanks for the feedback. I suppose that `mclk_div=156` is the maximum we can set to not override the 8 bits counter, then `bclk_div` can indeed be used to adjust.

Considering the sometimes DAC will work together with ADC, 8 bits is not the best solution, because ADC needs 16 bits channel width for receive 12 bits width data. I think we need to use `bit_per_chan` as `bclk_div` and add `mclk_div` check for it. 

This issue will be fixed recently, the commit in the PR will be involved, thanks for the contribution!

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/2f0e7e9b6c9c0c7bca6385f7a5d67abfcfcc6a7a

dzanis:
This fix broke the I2S ADC, had to revert the old code back

s-hadinger:
Can you give more details about the parameters you have been using?

dzanis:
My board is **ESP32-WROOM-32**.
Current version **IDF 4.4.1**

 I didn't use a DAC so I don't know how it behaves, but the ADC works with the wrong frequency to read samples. I recorded from a microphone and the sound turned out to be very slow.
 My parameters for one I2S ADC channel

```c
i2s_config_t i2s_config = {
    .mode = (i2s_mode_t)(I2S_MODE_MASTER | I2S_MODE_RX | I2S_MODE_ADC_BUILT_IN),
    .sample_rate = 16000,
    .bits_per_sample = I2S_BITS_PER_SAMPLE_16BIT,
    .channel_format = I2S_CHANNEL_FMT_ONLY_LEFT,
    .communication_format = I2S_COMM_FORMAT_STAND_I2S,
    .intr_alloc_flags = ESP_INTR_FLAG_LEVEL2,
    .dma_buf_count = 128,
    .dma_buf_len = 2,
    .use_apll = false
  };
```

I fixed it by returning the code with this condition to the compiler
```c
#if ESP_IDF_VERSION > ESP_IDF_VERSION_VAL(3, 3, 5)
  #define I2S_LL_AD_BCK_FACTOR           (2)
  int bclk = _sampleRate * I2S_LL_AD_BCK_FACTOR * 2;
  int bclk_div = I2S_LL_AD_BCK_FACTOR;
  i2s_config.use_apll = true;
  i2s_config.fixed_mclk = bclk * bclk_div;
#endif
```
And the problem was solved, the samples are normal.
But I hope that this will be fixed, such a crutch is not very beautiful

PR #8337, opened 2022-02-02
Set success when closing socket (IDFGH-6709)

bugadani:


mahavirj:
@bugadani Could you please elaborate on issue your were facing? Just asking to see if we can add simple test for this.

bugadani:
@mahavirj I have updated my project from esp-idf 4.3 to 4.4. I am making unsecure `http` requests and the `esp_http_client_close` function was returning `ESP_FAIL` (-1) even for requests that seemed successful. Might be that I forgot to update my config, but I tracked the source of that ESP_FAIL value to the code branch in this PR.

mahavirj:
sha=3aa41c438bccd92c1e381db335658193f5fdb7a1

PR #8338, opened 2022-02-02
unity: Better readline for interactive test menu (IDFGH-6708)

MacDue:
This commit replaces the use of `esp_rom_uart_rx_string()` in `unity_gets()` (used in `unity_run_menu()`) with new simple `readline()` function that supports echoing back the input, along with backspaces (the current version does neither of these making it a little annoying to use).

### DIFF
--- a/components/unity/unity_port_esp32.c
+++ b/components/unity/unity_port_esp32.c
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <string.h>
+#include <stdbool.h>
 #include "unity.h"
 #include "sdkconfig.h"
 #include "hal/cpu_hal.h"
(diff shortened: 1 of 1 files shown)


components/unity/unity_port_esp32.c
```diff
@@ -28,6 +29,37 @@ void unity_flush(void)
...
+                write_index--;
+                // Delete previously entered character
+                esp_rom_uart_tx_one_char('\b');
+                esp_rom_uart_tx_one_char(' ');
+                esp_rom_uart_tx_one_char('\b');
+            }
+        } else if (write_index < len - 1 && !iscontrol(c)) {
+            unity_putc(c);
+            dst[write_index++] = c;
+        }
```
atanisoft, line 59: It would likely be a good idea to add:
```
} else if (write_index >= len) {
  return;
}
```
That way it can exit from this loop without over running the buffer.

MacDue: I don't believe this code will overrun the buffer (though I may be mistaken). 
It only increments `write_index` and writes to `dst` if it's less than `len - 1` (to allow space for a null terminator in the return). I don't return when the buffer is full as it's still possible to backspace to delete the contents.

MacDue: It would overflow in one case, len = 0 (since len - 1 would underflow, which is now fixed)

igrr (approved):
Thank you for this very useful improvement!

igrr:
sha=ef25590277f831f763c175d57f5bc1563a3e8c50

components/unity/unity_port_esp32.c
```diff
@@ -28,6 +29,39 @@ void unity_flush(void)
...
+            unity_putc('\n');
+            dst[write_index] = '\0';
+            return;
+        } else if (c == '\b') {
+            if (write_index > 0) {
+                /* Delete previously entered character */
+                write_index--;
+                esp_rom_uart_tx_one_char('\b');
+                esp_rom_uart_tx_one_char(' ');
+                esp_rom_uart_tx_one_char('\b');
```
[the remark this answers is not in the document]
MacDue: Outputting this (`"\b \b"`) when `'\b'` is passed to `unity_putc()`?
If so, I'm not entirely sure if that would be correct as just outputting `\b` alone is valid (it just moves the cursor back one without clearing the character). However, `'\b'` is probably not used currently anyway, so it would probably be fine.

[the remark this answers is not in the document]
MaValki: Yes, that's what I meant. I thought it would make the function a bit cleaner. LGTM either way. Feel free to leave it as is.

PR #8350, opened 2022-02-04
Fix #8099 (IDFGH-6721)

### LINKED ISSUE #8099: IDF monitor fails to parse coredumps from uart (IDFGH-6439)
In the current implementation, it is possible for some bytes to be missing from coredump data received over UART. The branch that processed incomplete lines (referenced below) does not update the coredump processor at all:

https://github.com/espressif/esp-idf/blob/07bfc09d0c8837b4d149e1151c57635a0c02755a/tools/idf_monitor_base/serial_handler.py#L105-L120

The conditions where this may happen:
 - Data is received that contains an incomplete line
 - AND timer flushes the serial input buffer by calling `handle_serial_input` with `finalize_line=True`

Effect: monitor dumps a bunch of base64 encoded coredump data to the console, failing to process it due to missing bytes - either the base64 becomes invalid, or the data becomes shorter than expected.

I'm experiencing the issue with esp-idf 4.3.1, but looking at the implementation, master should still be affected.

bugadani:
This PR fixed processing coredumps that contain partially received lines.

### DIFF
--- a/tools/idf_monitor_base/serial_handler.py
+++ b/tools/idf_monitor_base/serial_handler.py
@@ -115,8 +115,18 @@ class SerialHandler:
         ))
         if self._last_line_part != b'' and force_print_or_matched:
             self._force_line_print = True
+            if self._decode_panic != PANIC_DECODE_DISABLE:
+                if self._reading_panic == PANIC_READING:
+                    self._panic_buffer += self._last_line_part.replace(b'\r', b'')
+
             self.logger.print(self._last_line_part)
             self.logger.handle_possible_pc_address_in_line(self._last_line_part)
+            if gdb_helper:
+                self.check_panic_decode_trigger(self._last_line_part, gdb_helper)
+            with coredump.check(line):
+                if self._force_line_print or line_matcher.match(line.decode(errors='ignore')):
+                    self.logger.print(line + b'\n')
+                    self.logger.handle_possible_pc_address_in_line(line)
             check_gdb_stub_and_run(self._last_line_part)
             # It is possible that the incomplete line cuts in half the PC
             # address. A small buffer is kept and will be used the next time


sio13:
Thanks for your contribution.

tools/idf_monitor_base/serial_handler.py
```diff
@@ -115,8 +115,18 @@ def handle_serial_input(self, data, console_parser, coredump, gdb_helper, line_m
...
             self._force_line_print = True
+            if self._decode_panic != PANIC_DECODE_DISABLE:
+                if self._reading_panic == PANIC_READING:
+                    self._panic_buffer += self._last_line_part.replace(b'\r', b'')
+
             self.logger.print(self._last_line_part)
             self.logger.handle_possible_pc_address_in_line(self._last_line_part)
+            if gdb_helper:
+                self.check_panic_decode_trigger(self._last_line_part, gdb_helper)
+            with coredump.check(line):
```
sio13, line 126: Hi @bugadani, is it meant to be `self._last_line_part` instead of `line` here? The variable `line` will be in many cases uninitialized.

bugadani: Hi @sio13

Thanks for catching it, it's definitely an error! On the other hand, I'm not sure if much of the patch is necessary. Let me check, and I'll update it accordingly.

bugadani:
Unfortunately, I don't this this patch is correct here. The script should, despite the forced processing, keep on to the partial line for coredump/panic trigger check, which it does not do. For some reason a similar change has worked in esp-idf v4.3, so I'm a bit confused why, but it no longer seems to. When I submitted the patch, my trial run must have been  a "lucky" run where the coredump trigger was not in a partial line.

PR #8388, opened 2022-02-11
esp_event: constify event_data pointer (IDFGH-6760)

amakuzaze:
The provided user data is immediately copied and never modified (as it should be), hence it should be marked const.
This change should be source compatible with existing code.

I tested this change by compiling and running both `esp_event/` sample projects.

### DIFF
--- a/components/esp_event/default_event_loop.c
+++ b/components/esp_event/default_event_loop.c
@@ -1,16 +1,8 @@
-// Copyright 2018 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+/*
+ * SPDX-FileCopyrightText: 2018-2022 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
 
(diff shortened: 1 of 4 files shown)


Alvin1Zhang:
Thanks for your contribution.

0xjakob (approved):
@amakuzaze Very nice improvement! I approved already, but if you don't mind, you could also apply the same change for the ISR-functions (`esp_event_isr_post`, `esp_event_isr_post_to`).

Otherwise we'll do the change ourselves after your PR has been merged.

0xjakob:
@amakuzaze Would you mind [installing the IDF pre-commit hooks](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html) and running them? We currently change our copyright headers, and the pre-commit will help you changing them correctly. Otherwise, we might need to rebase your commit. In that case your authorship will still be retained but the association with you is less direct (e.g. this PR won't be marked as merged but as closed instead with a reference to the commit ID).

amakuzaze:
@0xjakob Yes, apologies for not installing the hooks, I wrongly assumed they were about formatting. I will also constify the isr versions.

0xjakob (approved):

0xjakob:
> I commited the copyright header changes added by the pre-commit hook now (there was no other way to push anymore)

That's the correct thing to do. Thanks a loooot!

> I was wondering why you can't use the native rebase-and-merge feature github has?

Because for the time being, we use an internal mirror for our development. Only this internal one runs the full battery of unit tests and other CI-checks.

> Yes, apologies for not installing the hooks, I wrongly assumed they were about formatting. I will also constify the isr versions.

We have to apologize. We should maybe better document that feature and how to use it.

Thanks for the work, and sorry again for all the confusion. I myself know it's sometimes not easy to comply to all the checks and rules in IDF.

0xjakob:
sha=1858901241028bac6a6ff990af8c863e6a2094b6

amakuzaze:
@0xjakob thanks a lot for your feedback, and taking the time to explain the process behind the pre-commit hook :)

PR #8418, opened 2022-02-16
Added return type 'None' to serial_write() (IDFGH-6791)

nlantau:
Added return type of serial_write() for clarification of function prototypes.

### DIFF
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -178,7 +178,7 @@ class Monitor:
                 pass
             normal_print('\n')
 
-    def serial_write(self, *args, **kwargs):  # type: ignore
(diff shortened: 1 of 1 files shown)


dobairoland:
Hi @nlantau. Thank you for the contribution but I don't think that this changes anything in the current state. You've defined the return values but not the arguments.

And the type ignore comment remains for the lines so the typing is not checked. If you would like to improve the typing then please remove those `#type: ignore` comments, and install the [pre-commit](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html) hooks to check the commit in order to make sure it passes the tests.

PR #8430, opened 2022-02-18
ADC Fix compiling when included from C++ (IDFGH-6804)

0xFEEDC0DE64:
c++ is stricter with enum types and needs a external "C" for correct linkage


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=c86fba7919fed086ae3dfd07911d589391cf7e89

PR #8455, opened 2022-02-23
Implement websocket client connect error (IDFGH-6830)

0xFEEDC0DE64:
I dont know why, but the current websocket client implementation keeps emitting a DISCONNECTED event when the host address is simply wrong.

To me it seems more clear that this is an error case and usually you have not been connected before, so emitting a DISCONNECTED doesnt seem right?

Let me know what you think

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -161,6 +161,21 @@ static esp_err_t esp_websocket_client_abort_connection(esp_websocket_client_hand
     return ESP_OK;
 }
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

gabsuren:
@0xFEEDC0DE64 thanks for your contribution!
For now we are migrating **_esp_websocket_client_** from [esp-idf](https://github.com/espressif/esp-idf) to [esp-protocols](https://github.com/espressif/esp-protocols) and it will no longer be available in esp-idf starting from v5.0. Could you please migrate the PR to [esp-protocols](https://github.com/espressif/esp-protocols)  repository so we can merge it accordingly?

PR #8476, opened 2022-02-27
Add Release Support Schedule to Readme (IDFGH-6852)

negativekelvin:


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
I think this is a good idea but I'm wondering that maybe we could add https://dl.espressif.com/dl/esp-idf/support-periods.svg instead. The image is already shown on the https://docs.espressif.com/projects/esp-idf/en/latest/esp32/versions.html page.

The advantage is that that image is automatically re-generated so we wouldn't have to keep the readme up-to-date. We also plan to improve it in the near future by making the dates more visible.

negativekelvin:
I would say table format is preferred because I have never liked Gantt charts but automatically generated would be good.

dobairoland:
Regenerating a table would be harder because the script would have to open a PR as well. We will see how the image would be updated finally. It could contain a table as well!

In any way, if you are interested to update this PR so we could give a credit to your idea, then please add the linked image instead of the table. Otherwise, we will make the changes and close this PR with our commit.

PR #8496, opened 2022-03-04
Bugfix: LCD component support for large SPI color transfers (IDFGH-6874)

rrforte:
Currently, attempting to transmit over SPI a single color buffer that is too large using the LCD component will cause an integer overflow which will cause only some of the data being sent. This is due to SPI transfers being limited by the size that could be written into the SPI register `SPI_USR_MOSI_DBITLEN`.
This PR attempts to fix that behavior by splitting the transfer to multiple transactions no larger than the maximal allowable size.

Maximum size adressable by `SPI_USR_MOSI_DBITLEN` is actually `SPI_USR_MOSI_DBITLEN+1` bytes since the register holds the transfer size minus one.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=ffe19e0c8d033a5e2e80dc1b376cf7ef3f3fe63f

PR #8497, opened 2022-03-04
ledc: Do not drive output during init if the output is inverted (IDFGH-6875)

nonoo:
If the output of LEDC GPIO is inverted, then the correct level must be set before initializing the GPIO, otherwise it will be driven until the next call to stop LEDC output.

### DIFF
--- a/components/driver/ledc.c
+++ b/components/driver/ledc.c
@@ -638,6 +638,7 @@ esp_err_t ledc_channel_config(const ledc_channel_config_t *ledc_conf)
             );
     /*set LEDC signal in gpio matrix*/
     gpio_hal_iomux_func_sel(GPIO_PIN_MUX_REG[gpio_num], PIN_FUNC_GPIO);
+    gpio_set_level(gpio_num, output_invert);
     gpio_set_direction(gpio_num, GPIO_MODE_OUTPUT);
     esp_rom_gpio_connect_out_signal(gpio_num, ledc_periph_signal[speed_mode].sig_out0_idx + ledc_channel, output_invert, 0);
 


KaeLL (commented):
Shouldn't `gpio_set_direction` be called before `gpio_set_level`?

nonoo:
No.

If the output is inverted, and gpio_set_level() is called after
gpio_set_direction(), then the output will have an invalid value for a
short period of time.

Setting the level before the direction avoids the invalid output.

components/driver/ledc.c
```diff
@@ -638,6 +638,7 @@ esp_err_t ledc_channel_config(const ledc_channel_config_t *ledc_conf)
             );
     /*set LEDC signal in gpio matrix*/
     gpio_hal_iomux_func_sel(GPIO_PIN_MUX_REG[gpio_num], PIN_FUNC_GPIO);
+    gpio_set_level(gpio_num, output_invert);
```
mrengineer7777, line 641: This makes sense.  You should set the pin H/L value before driving the pin as an output.  I'm guessing esp_rom_gpio_connect_out_signal() also sets pin H/L, but by that point you would see a glitch on the pin without this fix.  I'm done the same thing in other hardware drivers I've written.

mrengineer7777: I should also mention esp_rom_gpio_connect_out_signal() is in the ESP32 ROM file, so I didn't find a source file.

mrengineer7777 (approved):

songruo:
sha=062f6164ae89d54df3f1a89a67f01e82ae56e0d8

KaeLL:
@nonoo Unrelated to the PR, but not completely, I found out I know nothing about GPIOs. Can you tell me why driving a GPIO before setting the direction works? And since it works and it's set to output, why set the direction at all?

PR #8544, opened 2022-03-10
esp_https_ota: add user_ctx to decrypt_cb (IDFGH-6924)

MacDue:
This makes the decryption layer easier to use in C++ wrappers, or whenever you want to avoid a global.

Without this you're forced to make a global variable for the `esp_decrypt_handle_t` for pre-encrypted OTA (instead of a member variable in the case of a C++ wrapper).


Alvin1Zhang:
Thanks for your contribution.

mahavirj (approved):

mahavirj:
sha=e47419374c8b5bdd5d4d0245e84233bb99bf6068

PR #8564, opened 2022-03-12
Fix nested redirect (IDFGH-6944)

### LINKED ISSUE #8563: Redirect location is not reset after updating URL (IDFGH-6943)
## Environment

- Module or chip used:  
  [WT32-ETH01](https://www.seeedstudio.com/Ethernet-module-based-on-ESP32-series-WT32-ETH01-p-4736.html)
- IDF version (run ``git describe --tags`` to find it):  
  `v5.0-dev-1730-g229ed08484`
- Build System:  
  `idf.py`
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it):  
  `xtensa-esp32-elf-gcc (crosstool-NG esp-2021r2-patch3) 8.4.0`
- Operating System:  
  Linux (Ubuntu 20.04)
- Using an IDE?:  
  Yes, VS Code
- Power Supply:
  USB

## Problem Description

If you request an HTTP endpoint, which will redirect you twice, the URL for the second redirect is set incorrectly.

I assume that the `client->location` is not reset after updating the URL during the first redirect.

After further research, it seems that [the function below](https://github.com/espressif/esp-idf/blob/master/components/esp_http_client/esp_http_client.c#L237) might need an additional if statement, which resets the content of `client‑>location` if it already contains data.

```cpp
static int http_on_header_value(http_parser *parser, const char *at, size_t length)
{
    esp_http_client_handle_t client = parser->data;
    if (client->current_header_key == NULL) {
        return 0;
    }
    if (strcasecmp(client->current_header_key, "Location") == 0) {
        // TODO: If location has a value, clear it before extracting new location.
        http_utils_append_string(&client->location, at, length);
    } else if (strcasecmp(client->current_header_key, "Transfer-Encoding") == 0
               && memcmp(at, "chunked", length) == 0) {
        client->response->is_chunked = true;
    } else if (strcasecmp(client->current_header_key, "WWW-Authenticate") == 0) {
        http_utils_append_string(&client->auth_header, at, length);
    }
    http_utils_append_string(&client->current_header_value, at, length);
    return 0;
}
```

### Expected Behavior

1. https://example.com/start
1. https://example.com/intermediate
1. https://example.com/end

### Actual Behavior

1. https://example.com/start
1. https://example.com/intermediate
1. https://example.com/intermediatehttps://example.com/end

```cpp
    .url = firmware_url,
    .cert_pem = (char*)ca_cert_pem_start,
    .timeout_ms = 10 * 1000,
    .keep_alive_enable = true,
    .user_agent = user_agent,
    .buffer_size = BUFFER_SIZE,
};

esp_http_client_handle_t client = esp_http_client_init(&config);
if (client == NULL) {
  ESP_LOGE(TAG, "Failed to configure HTTP client");
  return ESP_FAIL;
}

// Follow redirects until resource is found.
int32_t status = 0;
do {
  char url[512];
  esp_http_client_get_url(client, url, 512);
  ESP_LOGW(TAG, "Request: %s", url);

  err = esp_http_client_open(client, 0);
  if (err != ESP_OK) {
    ESP_LOGE(TAG, "Failed to open HTTP connection: %s", esp_err_to_name(err));
    esp_http_client_cleanup(client);
    return err;
  }
  esp_http_client_fetch_headers(client);

  status = esp_http_client_get_status_code(client);
  ESP_LOGW(TAG, "Status: %d", status);

  if (http_is_redirect(status)) {
    status = 0;
    esp_http_client_set_redirection(client);
  }
} while (status == 0);
```

## Debug Logs

```
W (12023) update: Request: https://github.com/nicklasfrahm/zeus/releases/latest/download/zeus-esp32.bin
W (17473) update: Status: 302
W (17473) update: Request: https://github.com/nicklasfrahm/zeus/releases/download/v1.2.2/zeus-esp32.bin
W (17673) update: Status: 302
W (17673) update: Request: https://github.com/nicklasfrahm/zeus/releases/download/v1.2.2/zeus-esp32.binhttps://objects.githubusercontent.com/github-production-release-asset-2e65be/462908779/b959e13d-bf22-4819-893d-c1ae01ff0047
```

nicklasfrahm:
Closes #8563.

Signed-off-by: Nicklas Frahm <nicklas.frahm@gmail.com>

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -234,6 +234,10 @@ static int http_on_header_value(http_parser *parser, const char *at, size_t leng
         return 0;
     }
     if (strcasecmp(client->current_header_key, "Location") == 0) {
+        if (client->location != NULL) {
+            free(client->location);
+            client->location = NULL;
+        }
         http_utils_append_string(&client->location, at, length);
     } else if (strcasecmp(client->current_header_key, "Transfer-Encoding") == 0
                && memcmp(at, "chunked", length) == 0) {
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -234,10 +234,6 @@ static int http_on_header_value(http_parser *parser, const char *at, size_t leng
         return 0;
     }
     if (strcasecmp(client->current_header_key, "Location") == 0) {
-        if (client->location != NULL) {
-            free(client->location);
-            client->location = NULL;
-        }
         http_utils_append_string(&client->location, at, length);
     } else if (strcasecmp(client->current_header_key, "Transfer-Encoding") == 0
                && memcmp(at, "chunked", length) == 0) {
@@ -516,6 +512,10 @@ static esp_err_t esp_http_client_prepare(esp_http_client_handle_t client)
     client->process_again = 0;
     client->response->data_process = 0;
     client->first_line_prepared = false;
+    if (client->location != NULL) {
+        free(client->location);
+        client->location = NULL;
+    }
     http_parser_init(client->parser, HTTP_RESPONSE);
     if (client->connection_info.username) {
         char *auth_response = NULL;


Alvin1Zhang:
Thanks for your contribution.

shubhamkulkarni97 (approved):

shubhamkulkarni97:
sha=e90f0f647cbff69721c55c5b1432a7f297d8eb76

nicklasfrahm:
@shubhamkulkarni97 I am still experiencing issues with the redirect, so my suspicion is that this is not the entire fix unfortunately. I am still investigating.

nicklasfrahm:
Okay, my research shows that resetting the `client->location` in this part of the code throws off the parser. Investigating further.

nicklasfrahm:
It can now be merged. All works as intended.

shubhamkulkarni97:
@nicklasfrahm Could you share a test URL with which you observed issue with first version of this PR?

nicklasfrahm:
https://github.com/nicklasfrahm/zeus/releases/latest/download/zeus-esp32.bin

The issue will show as a mismatch between the content length header and the length reported in the header callback. This will cause the header parser to use a wrong offset and it will truncate the location url. This in turn will then cause the HTTP URL parser to fail because and invalid URL is passed to the `esp_http_client_set_url` method.

nicklasfrahm:
@shubhamkulkarni97 How soon can this be merged? Can I help?

shubhamkulkarni97:
@nicklasfrahm This PR is under internal review and will be merged once it is approved

shubhamkulkarni97:
sha=b1fa44893b80686244046a7a220bbe543b103d52

shubhamkulkarni97:
> The issue will show as a mismatch between the content length header and the length reported in the header callback. This will cause the header parser to use a wrong offset and it will truncate the location url. This in turn will then cause the HTTP URL parser to fail because and invalid URL is passed to the `esp_http_client_set_url` method.

Hi @nicklasfrahm, could you please elaborate the issue you observed with the first approach?

Ideally the place where we free location field in the client handle should be independent of content length header and the length reported in the header callback.

It would be great if you can share an URL which will help me to reproduce the issue at my end.

nicklasfrahm:
The URL I am using to test this is https://github.com/nicklasfrahm/zeus/releases/latest/download/zeus-esp32.bin. `curl` has no issues with the URL and follows the redirects as intended.

I happen to **also** run this firmware to test it, but you can of course also use a different setup. My suspicion is that the parser keeps some internal state, because the `http_utils_append_string()` function takses a pointer as second argument and may get confused if `free()` is called after the parser initialization.

I am not sure if I agree that ideally it should not matter. I did not deep-dive into the parser implementation, but _if_ you work with pointer arithmetics for performance reasons, this may happen.

Thinking about this more, and again I didn't deep-dive, it may also be that the `client->location` is part of a larger internal buffer, which is kept between requests. In this case I agree and ideally, the `client->location` should be it's own buffer with its own lifecycle.

mahavirj:
Fix has been merged with https://github.com/espressif/esp-idf/commit/16ebc608dbfbbebad61f65ce3bbf75f630765533. Closing the PR. Thanks.

PR #8629, opened 2022-03-22
Update simple_ota_example.c

jfbrunner:
The original code passes the argument &ota_config, which is of type esp_https_ota_config_t *. The called function is esp_https_ota(const esp_http_client_config_t *config). The passed argument does not match the expected argument type.
ota_config.http_config is of type esp_http_client_config_t *, which matches the expected type of the called function.

### DIFF
--- a/examples/system/ota/simple_ota_example/main/simple_ota_example.c
+++ b/examples/system/ota/simple_ota_example/main/simple_ota_example.c
@@ -125,7 +125,7 @@ void simple_ota_example_task(void *pvParameter)
         .http_config = &config,
     };
     ESP_LOGI(TAG, "Attempting to download update from %s", config.url);
-    esp_err_t ret = esp_https_ota(&ota_config);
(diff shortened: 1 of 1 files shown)


igrr:
@jfbrunner any chance that you are using an example from `master` branch with a different version of IDF? In `master`, `esp_https_ota` function does indeed take `const esp_https_ota_config_t*`:
https://github.com/espressif/esp-idf/blob/c1bcb8756b89435d1ef6b964ac9ce953b18b6c90/components/esp_https_ota/include/esp_https_ota.h#L75
This is a breaking change in the upcoming ESP-IDF v5.0, described in the migration guide: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/migration-guides/protocols.html#esp-https-ota

If you are using one of the previous releases of ESP-IDF, please use the examples from the same branch or tag, they should be compatible.

Alvin1Zhang:
Thanks for your contribution.

PR #8689, opened 2022-03-30
uart: expose uart context for custom uart interrupt handlers

nonoo:
Exposing the UART context is needed to be able to easily access the HAL dev and the context's spinlock in user-defined UART interrupt handlers.


Alvin1Zhang:
Thanks for your contribution.

PR #8755, opened 2022-04-09
Fix nvs_flash_init_partition_ptr (IDFGH-7152)

io1o:


Alvin1Zhang:
Thanks for your contribution.

0xjakob:
@io1o Any explanation why we should change this? It looks as if all NVS pages will [load their data from a wrong address](https://github.com/espressif/esp-idf/blob/master/components/nvs_flash/src/nvs_pagemanager.cpp#L33) with this change.

io1o:
partition->address is not the start  address of  partition, https://github.com/espressif/esp-idf/blob/dd27d351ee0f50a47baa3b229a9803c2ea3a56ce/components/spi_flash/include/esp_partition.h#L111

PR #8779, opened 2022-04-14
esp_wifi: fix clearing default wifi netif procedure (IDFGH-7179)

klew:
Prior to this change "esp_wifi_clear_default_wifi_driver_and_handlers"
will not remove netif pointer from table when both AP and STA interfaces were
created and destroying default wifi interfaces is done in unfortunate
order. As a result there is dangling pointer left and it may cause crash in
later code (i.e. when esp_wifi_stop() is called).


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/43f769dcb4b15e1564805e7f9dc6580ddfc4c741.

Issue #8361, opened 2022-02-08
ability to route lwip debugs through ESP_LOGx macros (IDFGH-6731)


someburner:
When enabling LWIP debugs with `idf.py menuconifg`, these appear to (mostly?) get routed to `printf`. On the other hand pretty much everything else in the IDF is logged through ESP_LOGx macros. I would like to have a unified manner of logging where everything goes through the same macros, in order to support redirecting logs over UDP or elsewhere.

Doing something like this in `components/lwip/port/esp32/include/arch/cc.h` works, but could probably be made better with some macro magic to expand `x` and pass to the ESP_LOGx macros.

```c
#include "esp_log.h"

static const char * _LWIP_TAG = "lwip";
static int _ind = 0;
static char _pbuf[256];
static void _lwip_dbg_fn(const char * fmt, ...)
{
    va_list arglist;
    va_start(arglist, fmt);
    _ind += vsnprintf(_pbuf+_ind, sizeof(_pbuf) - _ind, fmt, arglist);
    va_end(arglist);
    if(_ind > 0) {
        if(_pbuf[_ind-1] == '\n') {
            _pbuf[_ind - 1] = 0;
            ESP_LOGI(_LWIP_TAG, "%s", _pbuf);
            _ind = 0;
        }
    }
}

// #define LWIP_PLATFORM_DIAG(x)   do {printf x;} while(0)
#define LWIP_PLATFORM_DIAG(x)   do {_lwip_dbg_fn x;} while(0)
```

I'm sure there are some pitfalls in doing this, particularly if routing logs over UDP, but I think such a feature would be easy enough to make optional and would result in consistent log output that could be `grep`d or redirected more easily. Thoughts?

negativekelvin:
>some macro magic

something like ...

```
#define LOG_LOCAL_LEVEL ESP_LOG_DEBUG
#include "esp_log.h"

#define LOG_FUNC(format, ...) ESP_LOGD("lwip", format, ##__VA_ARGS__ ) 

#define LWIP_PLATFORM_DIAG(x) LOG_FUNC x
```

xueyunfei998:
hi @negativekelvin @someburner 

This is a very good suggestion, and this change will be added to idf later.

xueyunfei998:
hi @someburner 
You can submit a PR, thanks

someburner:
@negativekelvin

The only issue with doing it this way:

```
[...]

PR #8785, opened 2022-04-17
add option to route LWIP logs through ESP_LOG interface (IDFGH-7186)

someburner:
Add an option to route LWIP debugs through ESP_LOGD instead of `printf`. See discussion here: https://github.com/espressif/esp-idf/issues/8361

Only downside with this approach is each LWIP debug has its own newline and ESP_LOG adds its own. In the future a new macro could be added that doesnt append a newline specifically for these logs or other 3rd party components where modifying source is not desired.

### DIFF
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -998,6 +998,13 @@ menu "LWIP"
 
             All lwIP debug features increase the size of the final binary.
 
+    config LWIP_DEBUG_ESP_LOG
+        bool "Route LWIP debugs through ESP_LOG interface"
+        depends on LWIP_DEBUG
+        default n
+        help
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

chipweinberger:
It would be nice If there was a drop down to choose between ESP_LOGD and ESP_LOGI. ESP_LOGD is pretty cluttered, and I want the ability to not have to enable all of LOGD.

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/5ebd86ab6aa8640a92fc70ee6a2e0ef054b1b0e0.

PR #8788, opened 2022-04-17
Add idf.py monitor argument --no-reset (-R) (IDFGH-7189)

nonoo:
Add idf.py monitor argument --no-reset (-R) to be able to optionally prevent resetting the CPU on monitor startup.

### DIFF
--- a/tools/idf_monitor.py
+++ b/tools/idf_monitor.py
@@ -75,6 +75,7 @@ class Monitor:
         print_filter,  # type: str
         make='make',  # type: str
         encrypted=False,  # type: bool
+        reset=True,  # type: bool
         toolchain_prefix=DEFAULT_TOOLCHAIN_PREFIX,  # type: str
         eol='CRLF',  # type: str
         decode_coredumps=COREDUMP_DECODE_INFO,  # type: str
@@ -109,7 +110,7 @@ class Monitor:
         if isinstance(self, SerialMonitor):
             socket_mode = serial_instance.port.startswith('socket://')
             self.serial = serial_instance
-            self.serial_reader = SerialReader(self.serial, self.event_queue)
+            self.serial_reader = SerialReader(self.serial, self.event_queue, reset)
 
             self.gdb_helper = GDBHelper(toolchain_prefix, websocket_client, self.elf_file, self.serial.port,
                                         self.serial.baudrate) if self.elf_exists else None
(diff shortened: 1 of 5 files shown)


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Hi @nonoo. https://github.com/espressif/esp-idf/runs/6052952029?check_suite_focus=true is failing. Can you please set up the pre-commit hooks locally (https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html) and fix the failures.

tools/idf_monitor.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         print_filter,  # type: str
         make='make',  # type: str
         encrypted=False,  # type: bool
+        no_reset=False,  # type: bool
```
dobairoland, line 78: This is just an optional suggestion.

If we would have 
```python
reset = True
```
instead then it would be easier to read and understand it than a reverse logic variable.

After that, negation has to be used only at one place:
```python
self.reset = not args.no_reset

dobairoland (changes requested):
Thanks for your contribution. Looks good to me in general.

nonoo:
There's one caveat, if no --port argument is given, tools/idf_py_actions/serial_ext.py calls esptool.get_default_connected_device() which opens the serial port by setting the chip into reset state. This needs to be fixed in esptool.get_default_connected_device() (both DTR and RTS needs to be set high in the esptool Python module).

A temporary workaround for this could be that we require the user to define the --port argument when using --no-reset.

dobairoland:
@nonoo 

> There's one caveat, if no --port argument is given, tools/idf_py_actions/serial_ext.py calls esptool.get_default_connected_device() which opens the serial port by setting the chip into reset state. This needs to be fixed in esptool.get_default_connected_device() (both DTR and RTS needs to be set high in the esptool Python module).

I don't think we can avoid that. 

> A temporary workaround for this could be that we require the user to define the --port argument when using --no-reset.

Yes, this looks good to me. Maybe don't need to "require" it. We could just print a warning that "`--no-reset` is ignored. Please specify the port with the --port argument in order to use this option".

dobairoland:
One more thing I forgot to mention.

Please squash the commits into one. (https://www.git-tower.com/learn/git/faq/git-squash/)

dobairoland:
Thanks for the corrections!

dobairoland:
sha=9266a7c52e078389e36fb90f1fba905683db5eae

PR #8827, opened 2022-04-24
led_strip component : fix timing by adding the reset signal (IDFGH-7233)

supcik:
The current `led_strip` component does not implement the required reset signal (280µs) at the end of the colors sequence. The constant `WS2812_RESET_US` is indeed defined, but it is not used.

A simple example such as the following one demonstrates the problem:

```c
led_strip_t *pStrip_a;
pStrip_a = led_strip_init(0, 25, 3);
ESP_ERROR_CHECK(pStrip_a->clear(pStrip_a, 500));
ESP_ERROR_CHECK(pStrip_a->set_pixel(pStrip_a, 0, 16, 16, 16));
ESP_ERROR_CHECK(pStrip_a->refresh(pStrip_a, 500));
```

The LED does not light up.

We can see on the logical analyzer that the delay between the bursts is too short (45µs instead of 280µs) :

Screenshot 2022-04-24 at 23 15 10

I resolved this issue by adding another byte to the buffer and the `ws2812_rmt_adapter` uses this additional byte to emit the proper reset signal.

With this fix, the signal looks much better :

Screenshot 2022-04-25 at 00 20 14

... and the LED lights up.

### DIFF
--- a/examples/common_components/led_strip/led_strip_rmt_ws2812.c
+++ b/examples/common_components/led_strip/led_strip_rmt_ws2812.c
@@ -1,16 +1,9 @@
-// Copyright 2019 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=c7fe38926a712b283134dc791697f5b2888f2cad

Alvin1Zhang:
Thanks for your contribution, changes merged with https://github.com/espressif/esp-idf/commit/517c0b54d83d42fb4f6598c946c80b2669d9c6c2.

PR #8835, opened 2022-04-26
Support for external 32.768kHz oscillator as Bluetooth clock (IDFGH-7241)

johnboiles:
Currently, only a 32.768kHz crystal can be used as the Bluetooth low-power clock since `BTDM_CTRL_LPCLK_SEL_EXT_32K_XTAL` depends on `RTC_CLK_SRC_EXT_CRYS` (and not also `RTC_CLK_SRC_EXT_OSC`).

This change makes it so that you can set `BTDM_CTRL_LPCLK_SEL_EXT_32K_XTAL` when using the external oscillator.

Tested using a similar change on ESP-IDF v4.4.1.

### DIFF
--- a/components/bt/controller/esp32/Kconfig.in
+++ b/components/bt/controller/esp32/Kconfig.in
@@ -257,10 +257,10 @@ menu "MODEM SLEEP Options"
                 the bluetooth low power clock source.
 
         config BTDM_CTRL_LPCLK_SEL_EXT_32K_XTAL
-            bool "External 32kHz crystal"
-            depends on RTC_CLK_SRC_EXT_CRYS
+            bool "External 32kHz crystal/oscillator"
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

xiongweichao:
Hi @johnboiles 

We will verify if the External 32kHz oscillator can be used as a bluetooth low power clock and will reply to you as soon as possible.

Thanks

johnboiles:
FWIW it's been working reliably for me for me on the ESP32-PICO-D4 for the past month or so.

I just started working on doing the same for the ESP32-S3. I'm getting hard lockups where the the USB device ceases to respond if the chip spends too long asleep. Still investigating that.

xiongweichao:
Hi @johnboiles ， Sorry for the late reply. This issue has been fixed in the [master](https://github.com/espressif/esp-idf/commit/f5673411686f1dff1924e7c8bbe23fdd1c9c1c12) branch.

PR #8851, opened 2022-04-28
Fixed the commented out coap_new_optlist() in the coap_client example. (IDFGH-7262)

nosuz:
coap_new_optlist() requires 3 arguments but the example had 2 arguments.


Alvin1Zhang:
Thanks for your contribution.

PR #8900, opened 2022-05-04
hello_world_main : fix incorrect printf format (IDFGH-7312)

huardti:
spi_flash_get_chip_size return a size_t to printf this type it's better to use %zu than %d.
This commit change that.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=5b53a073d1b8a74492c77c24aa82d18b0c0d7126

PR #8934, opened 2022-05-10
Speed up reads from encrypted partitions

rojer:
Mapping and unmapping is expensive so we map the entire partition and keep it mapped
anticipating that it will be reused. We only keep one partition mapped so alternating
reads between encrypted partitions will incur a penalty but hopefully that is uncommon.

### DIFF
--- a/components/spi_flash/partition.c
+++ b/components/spi_flash/partition.c
@@ -430,19 +430,32 @@ esp_err_t esp_partition_read(const esp_partition_t* partition,
         if (partition->flash_chip != esp_flash_default_chip) {
             return ESP_ERR_NOT_SUPPORTED;
         }
-
-        /* Encrypted partitions need to be read via a cache mapping */
-        const void *buf;
-        spi_flash_mmap_handle_t handle;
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/spi_flash/partition.c
```diff
@@ -430,19 +430,32 @@ esp_err_t esp_partition_read(const esp_partition_t* partition,
...
-                                 SPI_FLASH_MMAP_DATA, &buf, &handle);
-        if (err != ESP_OK) {
-            return err;
+        /*
+         * Encrypted partitions need to be read via a cache mapping.
+         * Mapping and unmapping is expensive so we map the entire partition and keep it mapped
+         * anticipating that it will be reused. We only keep one partition mapped so alternating
+         * reads between encrypted partitions will incur a penalty but hopefully that is uncommon.
+         */
+        static uint32_t s_mapped_partition_end = 0;
```
greenaddress, line 439: this is not thread safe, would require locking to work with multiple tasks

PR #8955, opened 2022-05-16
Tools: fix idf.py monitor reset with hotkey with --no-reset arg (IDFGH-7375)

nonoo:
Previous --no-reset PR (https://github.com/espressif/esp-idf/pull/8788) was not merged completely. Without this patch, the hotkey reset in idf.py monitor (ctrl+t ctrl+r) does not work when --no-reset argument is used.

### DIFF
--- a/tools/idf_monitor_base/serial_reader.py
+++ b/tools/idf_monitor_base/serial_reader.py
@@ -57,6 +57,8 @@ class SerialReader(Reader):
 
                 # Add a delay to meet the requirements of minimal EN low time (2ms for ESP32-C3)
                 time.sleep(MINIMAL_EN_LOW_DELAY)
+            elif not self.reset:
+                self.serial.setDTR(high)  # IO0=HIGH, default state
             self.gdb_exit = False
             self.serial.rts = high             # Set rts/dtr to the working state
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Hi @nonoo. Oh I see it now. I assumed this was a mistake. I'm afraid the same mistake could happen in the future.

Could you add this into the `run()` method in `tools/idf_monitor_base/serial_reader.py` instead? That would make it clearer.

Something like the following:

```python
if not self.gdb_exit and self.reset:
...
elif not self.reset:
...
```

nonoo:
`serial_reader.py`'s `run()` is not executed when a hotkey reset is requested. A hotkey reset request causes a `CMD_RESET` which is handled in `serial_handler.py`.

dobairoland:
But the hotkey reset is not working for you because the DTR/RTS signals were not left at the right state after the port has been opened. Or am I missing something?

dobairoland:
Thanks for the update!

dobairoland:
sha=c063cbd7f7d0996fbaab3d97c8011f216576c8fc

PR #8957, opened 2022-05-16
components/lwip: errno.h should not be included if LWIP provides errno. (IDFGH-7377)

iabdalkader:
The standard `<errno.h>` should not be included directly in `cc.h`, to allow lwip to provide error numbers (via `LWIP_PROVIDE_ERRNO`) or a custom errno.h (via `LWIP_ERRNO_INCLUDE`).

### DIFF
--- a/components/lwip/port/esp32/include/arch/cc.h
+++ b/components/lwip/port/esp32/include/arch/cc.h
@@ -35,7 +35,6 @@
 #define __ARCH_CC_H__
 
 #include <stdint.h>
-#include <errno.h>
 #include <assert.h>
 #include <stdio.h>
 
--- a/components/lwip/port/esp32/include/lwipopts.h
+++ b/components/lwip/port/esp32/include/lwipopts.h
@@ -61,6 +61,11 @@
 
 #define LWIP_RAND       esp_random
 
+/**
+ * Include <errno.h> from the standard libraries.
+ */
+#define LWIP_ERRNO_STDINCLUDE   1
+
 /*
    ------------------------------------
    ---------- Memory options ----------


components/lwip/port/esp32/include/arch/cc.h
```diff
@@ -35,7 +35,9 @@
 #define __ARCH_CC_H__
 
 #include <stdint.h>
+#ifndef LWIP_PROVIDE_ERRNO
 #include <errno.h>
+#endif
```
igrr, line 40: Perhaps this include can probably be removed altogether, and instead we could define `LWIP_ERRNO_STDINCLUDE` in components/lwip/port/esp32/include/lwipopts.h? Then `<errno.h>` will be included automatically from components/lwip/lwip/src/include/lwip/errno.h.

iabdalkader: Yes, that should also fix it because `LWIP_PROVIDE_ERRNO` is checked first before including `<errno.h>`, so I'll amend this PR. That said, I noticed that some files include `<errno.h>` directly, I think they should include `lwip/errno.h`. For example `components/esp_http_server/src/httpd_main.c`

iabdalkader: It seems that I have to revert changes in my projects and use `<errno.h>` anyway, but I'll keep this PR open if you say it is useful. Btw would this `components/lwip/lwip/test/unit/cc_esp_platform.h` cause any unit tests to fail after this change ?

MaxwellAlan:
Hi @iabdalkader 

Thanks for the PR, we will not merge this.

PR #8981, opened 2022-05-19
Tools: Add missing DTR/RTS set before serial port open on reconnect (IDFGH-7403)

nonoo:
This makes sure --no-reset works correctly when the monitor reconnects the serial port.


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Thanks @nonoo!

dobairoland:
sha=f6e414017122f4de1c227d36769ed6233661da57

PR #8987, opened 2022-05-19
pthread: avoid heap alloc in pthread_cond_timedwait if possible (IDFGH-7409)

MacDue:
This creates the wait sem with `xSemaphoreCreateCountingStatic` if possible taking around 80 bytes off the stack.
This semaphore is only used for the life of this function, which can be quite heavily called, so this avoids any possible heap/performance issues.

### DIFF
--- a/components/pthread/pthread_cond_var.c
+++ b/components/pthread/pthread_cond_var.c
@@ -1,21 +1,8 @@
-// Copyright 2017 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

igrr (changes requested):
This looks like a useful improvement, thank you! Have left one minor comment.

By the way, have you done any benchmarks to see how much this change improves performance?

MacDue:
I don't think this changes the runtime performance really (since you normally will be waiting at least a few milliseconds in the cond_timedwait). 

My main thought with this change was to avoid the possibility of heap fragmentation when waiting for longer timeouts, 
though I've not got anything quantifiable for that (it was just something I thought about when looking over the code due to an unrelated issue).

0xjakob:
sha=c104662192c35ddc01050310ec96705894a75889

0xjakob:
@MacDue Your code is [available on master ](https://github.com/espressif/esp-idf/commit/253b952f4db8768b0b123037cd1c1e050b8e66c0) already. Because it is merged already, we are closing this PR.

PR #8990, opened 2022-05-19
esp_netif_leave_ip6_multicast_group_api: Fixup inversion error causing error prints (IDFGH-7412)

rretanubun:
ESP_RETURN_ON_FALSE(a, err_code, ...) macro

Will print and return err_code
if (unlikely(!(a)))

so "a" needs to state what we expect to be TRUE
which in this case means:

mld6_leavegroup_netif(...) == ERR_OK


Alvin1Zhang:
Thanks for your contribution.

david-cermak (approved):
Good catch! Thanks for the fix

PR #9014, opened 2022-05-24
[WIP] Implement EAP-NOOB (RFC9140) (IDFGH-7438)

Janfred:
This is an implementation of RFC9140 (EAP-NOOB)

The implementation is not completely standard compliant. The JSON library will remove any unnecessary spaces in objects (which the RFC forbids, the encoding has to be exactly the same for the HMAC calculation) and I'm not sure about UTF-8 support, since I use functions like `strlen`.

Since EAP-NOOB is a non-conventional EAP-Method in the meaning that it requires an out-of-band-message before the actual login is successful, I modified some of the logic in the esp_wpa stack.

This Pull Request is still work in progress, for now I would like to get feedback if the way I implemented this is ok or if I should adjust the architecture, especially regarding the public facing methods for OOB-Message generation/retrieval/offloading and initializing the EAP method.

### DIFF
--- a/components/wpa_supplicant/esp_supplicant/src/esp_wpa2.c
+++ b/components/wpa_supplicant/esp_supplicant/src/esp_wpa2.c
@@ -28,6 +28,7 @@
 #ifdef EAP_PEER_METHOD
 #include "eap_peer/eap_methods.h"
 #endif
+#include "eap_peer/eap_noob.h"
 
 #include "esp_wifi_driver.h"
 #include "esp_private/wifi.h"
@@ -381,6 +382,7 @@ int eap_sm_process_request(struct eap_sm *sm, struct wpabuf *reqData)
     } else {
         m = eap_peer_get_eap_method(reqVendor, reqVendorMethod);
         if (m == NULL) {
+            wpa_printf(MSG_DEBUG, "EAP: no EAP method associated with vendor %u method %u.", reqVendor, reqVendorMethod);
             goto build_nak;
         }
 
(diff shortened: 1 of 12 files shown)


Janfred:
For the JSON encoding compliance this is what the RFC says in section 3.3.2:
> The inputs to the hash function for computing the fingerprint Hoob and to the HMAC for computing MACs, MACp, MACs2, and MACp2 are JSON arrays containing a fixed number (17) of elements. The array elements MUST be copied to the array verbatim from the sent and received in-band messages. When the element is a JSON object, its members MUST NOT be reordered or reencoded. White space MUST NOT be added anywhere in the JSON structure. Implementers should check that their JSON library copies the elements as UTF-8 strings, does not modify them in any way, and does not add white space to the HMAC input.

Alvin1Zhang:
Thanks for your contribution.

components/wpa_supplicant/esp_supplicant/src/esp_wpa2.c
```diff
@@ -1193,3 +1195,63 @@ esp_err_t esp_wifi_sta_wpa2_use_default_cert_bundle(bool use_default_bundle)
...
+
+eap_noob_oob_msg_t *esp_wifi_sta_wpa2_ent_eap_noob_generate_oob_message(void)
+{
+    return eap_noob_generate_oob_msg();
+}
+
+esp_err_t esp_wifi_sta_wpa2_ent_eap_noob_receive_oob_message(u8 *noob, u8 *hoob)
+{
+    eap_noob_oob_msg_t *oobmsg = os_zalloc(sizeof(eap_noob_oob_msg_t));
+    memcpy(oobmsg->noob, noob, 16);
```
obgm, line 1225: I suggest using `sizeof(oobmsg->noob)` here to avoid the magic number.
(Same for hoob in the next line).

components/wpa_supplicant/src/eap_peer/eap.c
```diff
@@ -612,6 +617,9 @@ int eap_peer_config_init(
 			//set EAP-TTLS
 			config_methods[allowed_method_count].vendor = EAP_VENDOR_IETF;
 			config_methods[allowed_method_count++].method = EAP_TYPE_TTLS;
+            //set EAP-NOOB
+            config_methods[allowed_method_count].vendor = EAP_VENDOR_IETF;
```
obgm, line 621: The leading whitespace looks off.

components/wpa_supplicant/src/eap_peer/eap_noob.c
```diff
@@ -0,0 +1,1686 @@
...
+    g_wpa_eap_noob_state.ephemeral_state = os_zalloc(sizeof(struct eap_noob_ephemeral_state_info));
+    struct eap_noob_ephemeral_state_info *eph = g_wpa_eap_noob_state.ephemeral_state;
+
+    if (data->shared_key != NULL) {
+        eph->shared_secret = os_zalloc(data->shared_key_length);
+        eph->shared_secret_length = data->shared_key_length;
+        memcpy(eph->shared_secret, data->shared_key, eph->shared_secret_length);
+    }
+
+    memcpy(eph->np, data->np, 32);
```
obgm, line 477: Here, I also advise against magic numbers.

kapilkedawat:
Hi @Janfred , From architectural point of view, changes looks ok (some cleanup is required for public functions and code though). Let us know once you complete it, we will review this in depth.

components/wpa_supplicant/src/eap_peer/eap_noob.c
```diff
@@ -0,0 +1,1686 @@
...
+
+    mbedtls_entropy_context entropy;
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_ctr_drbg_init( &ctr_drbg );
+    mbedtls_entropy_init( &entropy );
+    int ret_val;
+    const char *mbed_seed = "EAP-NOOB-Nonce";
+    ret_val = mbedtls_ctr_drbg_seed( &ctr_drbg, mbedtls_entropy_func, &entropy, (unsigned char *)mbed_seed, 14);
+    if (ret_val != 0) {
+        return NULL;
```
Janfred, line 342: Note to myself: Here a `free` for the oobMsg and oobMsgNode is missing

Alvin1Zhang:
@Janfred Thanks for contribution, any further updates for the PR? Thanks.

Janfred:
Unfortunately, this was part of a study program and my efforts in the IETF to specify a revised EAP protocol that uses CBOR instead of JSON did not receive feedback, so I kind of abandoned this.
If there is renewed interest with a use case, I would be happy to pick the ball up again.

kapilkedawat:
Hi @Janfred, thanks for the update. We will close this request for now. Feel free to reopen if you start working on this again.

Issue #4162, opened 2019-10-04
Add color to build output log (more readable errors and warnings) (IDFGH-1965)


AloyseTech:
## Environment

- Development Kit:      ESP32-PICO-Kit
- Kit version v4
- Module or chip used:  ESP32-PICO-D4
- IDF version v4.1-dev-256-g9f145ff16
- Build System:        CMake
- Compiler version xtensa-esp32-elf-gcc (crosstool-NG esp32-2019r1) 8.2.0
- Operating System:     Windows
- Power Supply:         USB

## Problem Description

The build process output a lot of text. It is sometimes difficult to spot errors and warnings.

### Expected Behavior

It could be great if the output could be colorized (red for errors, orange/yellow for warning etc), to make the build log more readable.

AloyseTech:
I've added `add_compile_options (-fdiagnostics-color=always)` to the main CMakeList.txt file of my project and set an environment variable `GCC_COLORS="error=01;31:warning=01;33:note=01;36:caret=01;32:locus=01:quote=01"` which give me some colored output. Is there any better way to do this?

renzbagaporo:
There is no official support for this in the sense that you can't pass an argument to `idf.py` to enable this. You can, however, skip adding `add_compile_options (-fdiagnostics-color=always)` to every project you work with by adding the -`fdiagnostics-color=always` to the environment variable EXTRA_CFLAGS and EXTRA_CXXFLAGS.

You still need to set the GCC_COLORS however for the colors you want the output to be.

esanai:
You can colorize the log with following code.
```
static const char *TAG = "TEST_TASK";
ESP_LOG_LEVEL(ESP_LOG_ERROR, TAG, "%d - Red log", 1);
ESP_LOG_LEVEL(ESP_LOG_WARN, TAG, "%d - Yello log", 2);
ESP_LOG_LEVEL(ESP_LOG_INFO, TAG, "%d - Green log", 3);
```
I found the way from the source code of [esp_log.h](https://github.com/espressif/esp-idf/blob/master/components/log/include/esp_log.h#L319)

renzbagaporo:
@esanai I think @AloyseTech was referring to colorizing the actual output build log, not the runtime logs.

esanai:
@renzbagaporo You are right. I'm sorry for my misunderstanding.

KaeLL:
[...]

PR #9024, opened 2022-05-25
cmake: enable coloured compiler output (IDFGH-7449)

MacDue:
This adds -fdiagnostics-color=always to the default compile_options,
without this GCC won't enabled colored output when called from
ninja (via idf.py), making errors much harder to decipher.

### DIFF
--- a/tools/cmake/build.cmake
+++ b/tools/cmake/build.cmake
@@ -105,7 +105,10 @@ function(__build_set_default_build_specifications)
                                     "-Wno-sign-compare"
                                     # always generate debug symbols (even in release mode, these don't
                                     # go into the final binary so have no impact on size
-                                    "-ggdb")
+                                    "-ggdb"
+                                    # without this GCC may not enable ANSI colouring, making errors
+                                    # harder to decipher
+                                    "-fdiagnostics-color=always")
 
     list(APPEND c_compile_options   "-std=gnu99")
 


MacDue:
This is something that has bugged me quite a few times when building with `idf.py`

igrr:
Hi @MacDue, thank you for the PR.
Unfortunately this approach will enable color output also when the build system output is not a terminal which understands ANSI escape sequences.

To solve this issue, we plan to enable color diagnostic output in the following manner (ref. https://github.com/espressif/esp-idf/issues/4162#issuecomment-1063323074):
- upgrade CMake to the version which supports CMAKE_COLOR_DIAGNOSTICS environment variable — done in https://github.com/espressif/esp-idf/commit/2ee43cf6c4e282f343bf657a6a8c8c6af0f1e422
- in idf.py, check if stdout is a TTY and set CMAKE_COLOR_DIAGNOSTICS=1 if so
- IDEs which call CMake directly without idf.py (such as CLion or Eclipse) can set CMAKE_COLOR_DIAGNOSTICS if they support ANSI escape codes. For example, CLion has recently shipped this feature: https://blog.jetbrains.com/clion/2022/05/clion-starts-the-2022-2-eap/
- in headless builds (CI, or when output is redirected to another program or a file) color output will stay disabled.

If you have any suggestions, please leave a comment in the linked issue about color diagnostic output. Thanks again!

MacDue:
While I appreciate a better fix is coming, I still think it's silly this has been disabled for so long. It's trivial to remove ANSI sequences from the output after the fact with something like `ansi2txt`, but impossible to add them & makes for unpleasant error messages.

igrr:
Support for color diagnostics is now available in master, please see https://github.com/espressif/esp-idf/issues/4162#issuecomment-1235155746. I'm sorry we did not merge your PR as is.

PR #9025, opened 2022-05-25
fix glb_clk may be uninitialized build error in ledc driver (IDFGH-7450)

tamisoft:
I am not sure if this is a complete solution to the problem, but at least it seems to fix it on esp32.
The issue was introduced in: d64c6f581d4391146c91ccc6c75d5630944c9c26 ledc: cleanup the clock selection code

The offending line is a print: line 570:         ESP_LOGD(LEDC_TAG, "In slow speed mode, global clk set: %d", glb_clk);

Am I missing something else here?

build error:

In file included from components/driver/ledc.c:10:
components/driver/ledc.c: In function 'ledc_set_timer_div':
components/log/include/esp_log.h:419:47: error: 'glb_clk' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         else if (level==ESP_LOG_DEBUG )     { esp_log_write(ESP_LOG_DEBUG,      tag, LOG_SYSTEM_TIME_FORMAT(D, format), esp_log_system_timestamp(), tag, ##__VA_ARGS__); } \
                                               ^~~~~~~~~~~~~
components/driver/ledc.c:497:25: note: 'glb_clk' was declared here
     ledc_slow_clk_sel_t glb_clk;
                         ^~~~~~~

### DIFF
--- a/components/driver/ledc.c
+++ b/components/driver/ledc.c
@@ -162,6 +162,10 @@ static uint32_t ledc_get_glb_clk_freq(ledc_slow_clk_sel_t clk_cfg)
             src_clk_freq = esp_clk_xtal_freq();
             break;
 #endif
+        case LEDC_CLK_NOT_INIT:
+        default:
+            ESP_LOGE(LEDC_TAG, "Clock not initialized");
+            break;
     }
 
     return src_clk_freq;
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

dabbinavo:
@suda-morris, @songruo
Any news here? Still an issue in v5.0-rc1.

PR #9033, opened 2022-05-26
add address field to read_rmt_name_param in legacy bluetooth (IDFGH-7459)

key-moon:
This PR adds the field `bda` to `esp_bt_gap_cb_param_t::read_rmt_name_param` and the code to copy the value into the field. This structure is used for the callback function's argument when the name inquiry is resolved. Knowing who responded is mandatory for the callback function to process the information.


Alvin1Zhang:
Thanks for your contribution.

PR #9042, opened 2022-05-27
add user_panic_handler() hook in esp_panic (IDFGH-7472)

mmoskal:
Use cases: 
* we have custom logging infrastructure and want to print out these logs upon crash.
* we want to blink some LED pattern upon crash and then reboot


Alvin1Zhang:
Thanks for your contribution.

mmoskal:
thank you @igrr

PR #9062, opened 2022-05-30
Secure boot v2 image format: CRC size (IDFGH-7493)

wallento:
The CRC size is documented as 1095 bytes, but in the code I find 1196 bytes:
https://github.com/espressif/esp-idf/blob/master/components/bootloader_support/src/secure_boot_v2/secure_boot.c#L35


Alvin1Zhang:
Thanks for your contribution.

mahavirj (approved):

mahavirj:
sha=dd7aece4bf46819b0e8f8ac2c0f2cacc3099144c

PR #9073, opened 2022-06-01
initialize all elements of esp_bt_controller_config_t (IDFGH-7506)

davidzuhn:
Fixes warnings when using bt in C++ on ESP32-S3 by explicitly initializing all elements of the esp_bt_controller_config_t struct (even those that are deprecated).


Alvin1Zhang:
Thanks for your contribution.

Issue #9109, opened 2022-06-06
ESP32C2 tjpgd example run will crash (IDFGH-7549)


dreamcmi:
I try to run the example of tjpgd on the ESP8684-DevKitM-1 development board, but it will crash, the following is the error log, how can I solve this problem?
```shell
ESP-ROM:esp8684-api2-20220127
Build:Jan 27 2022
rst:0xc (RTC_SW_CPU_RST),boot:0xc (SPI_FAST_FLASH_BOOT)
Saved PC:0x403805be
0x403805be: esp_restart_noos at /home/darren/esp/esp-idf/components/esp_system/port/soc/esp32c2/system_internal.c:112 (discriminator 1)

SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd6108,len:0x18b4
load:0x403ae000,len:0x864
load:0x403b0000,len:0x2598
entry 0x403ae000
I (23) boot: ESP-IDF v5.0-dev-3290-g01d014c42d 2nd stage bootloader
I (23) boot: compile time 22:51:47
I (23) boot: chip revision: 0
I (26) boot.esp32c2: MMU Page Size  : 64K
I (31) boot.esp32c2: SPI Speed      : 60MHz
I (36) boot.esp32c2: SPI Mode       : DIO
I (40) boot.esp32c2: SPI Flash Size : 2MB
I (45) boot: Enabling RNG early entropy source...
I (51) boot: Partition Table:
I (54) boot: ## Label            Usage          Type ST Offset   Length
I (61) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (69) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (76) boot:  2 factory          factory app      00 00 00010000 00100000
I (84) boot: End of partition table
I (88) esp_image: segment 0: paddr=00010020 vaddr=3c020020 size=18f38h (102200) map
I (128) esp_image: segment 1: paddr=00028f60 vaddr=3fca9d70 size=00ce8h (  3304) load
I (130) esp_image: segment 2: paddr=00029c50 vaddr=40380000 size=063c8h ( 25544) load
I (143) esp_image: segment 3: paddr=00030020 vaddr=42000020 size=10c90h ( 68752) map
I (165) esp_image: segment 4: paddr=00040cb8 vaddr=403863c8 size=0399ch ( 14748) load
I (176) boot: Loaded app from partition at offset 0x10000
I (177) boot: Disabling RNG early entropy source...
[...]

PR #9119, opened 2022-06-08
Fix the PARALLEL_LINES size of the tjpgd example to ensure that esp32c2 can also be used normally (IDFGH-7559)

dreamcmi:
https://github.com/espressif/esp-idf/issues/9109

Try changing the default PARALLEL LINES size to be compatible with ESP32C2 running the example


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=2024ae943466a967f48607c3bd9c258526f08b10

PR #9133, opened 2022-06-09
driver(spi): fix flag check in bus initialization

yannishuber:
The current flag check in the SPI bus initialization is wrong and causes exceptions when using certain SPI hosts in quad transfert mode. 

If the SPI host only supports quad mode and not octal mode (which is the case on the SPI3 host of the ESP32-S3) then we unnecessarily initialize the data4-7 pins which causes an exception.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=3397cf465d60ddb9806838ebeba57b826fa20fdc

PR #9147, opened 2022-06-13
Add esp_tls option to skip server verification (IDFGH-7592)

0xFEEDC0DE64:
I needed to debug a faulty tls websocket connection and did not want to compromise the whole firmware by setting the compile time flag `CONFIG_ESP_TLS_SKIP_SERVER_CERT_VERIFY`.

Instead I implemented a way to disable server verification for one single connection and I put warnings everywhere before you can turn on this option.

I implemented this new option additionaly in other clients mqtt (separate repo), websocket_client (separate repo), http_client

### DIFF
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -151,6 +151,10 @@ typedef struct esp_tls_cfg {
     bool use_global_ca_store;               /*!< Use a global ca_store for all the connections in which
                                                  this bool is set. */
 
+    bool skip_server_verification;          /*!< Don't try to verificate the server's certificate in any
+                                                 way. Should never be used in production and only help to debug
+                                                 a faulty connection. */
+
     const char *common_name;                /*!< If non-NULL, server certificate CN must match this name.
                                                  If NULL, server certificate CN must match hostname. */
 
--- a/components/esp-tls/esp_tls_mbedtls.c
+++ b/components/esp-tls/esp_tls_mbedtls.c
@@ -664,6 +664,9 @@ esp_err_t set_client_config(const char *hostname, size_t hostlen, esp_tls_cfg_t 
         if (esp_ret != ESP_OK) {
             return esp_ret;
         }
+    } else if (cfg->skip_server_verification == true) {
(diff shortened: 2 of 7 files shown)


Alvin1Zhang:
Thanks for your contribution.

negativekelvin:
>I put warnings everywhere

Should be at warning level then not error

>Should never be used in production

Then it should be conditional compilation

0xFEEDC0DE64:
so we enable debugging at certain customers without changing their binaries, just setting a bool to allow all server certificates for one component of the firmware, we have 10 connections in paralell and I need verification in 9 out of 10 cases.

Just let me specify if that one instance of tls connection should also verify the server cert or not.

negativekelvin:
Ok well it should still be optional to add this feature to a firmware via conditional compilation

ESP-YJM:
Hi @0xFEEDC0DE64 , i think we no need this PR also can achieve your requirement. Add a compile flag(Enable `CONFIG_ESP_TLS_SKIP_SERVER_CERT_VERIFY`) or add a configuration like your PR both can make verify optional. When you enable `CONFIG_ESP_TLS_SKIP_SERVER_CERT_VERIFY` and you not set `cacert` in your configuration, we will not verify the server. And if you set `cacert` in your configuration, we will verify the server. So `cacert` configuration has similar effect to your PR.

ESP-YJM:
Do you mean global ca store?

0xFEEDC0DE64:
yes

ESP-YJM:
Yes, i think it can work with global ca store, because global ca store is controled by flag https://github.com/espressif/esp-idf/blob/master/components/esp_http_client/include/esp_http_client.h#L131

mahavirj:
@0xFEEDC0DE64 Ideally we would not like to expose a run time knob to skip server verification, in-fact compile time configuration also comes with sufficient warnings upfront. At the same time I understand that this patch is useful in case you described above. Would it possible for you to maintain this patch on your side?

0xFEEDC0DE64:
ok i keep the patch on my side, with all the other patches :)

0xFEEDC0DE64:
One thing that makes maintaining this patch really bad for me is that I also have patches on mqtt and http and websocket clients. Those are submodules of esp-idf which makes it really a pain to keep up-to-date

PR #9209, opened 2022-06-22
fixed misaligned data transfer for ssd1306 (IDFGH-7661)

Slamy:
I got hold of a https://heltec.org/project/wifi-kit-32/ board.
Then I went for the example peripherals/lcd/i2c_oled which makes use of the ssd1306 oled on the board.
Sadly the scrolling text has shown artifacts. The problem was the sending of one additional 0x00 byte at the start of the color data transfer.

This PR might not be exactly what you are aiming for. I'm pretty new to the esp-idf. So maybe your team has a better solution for this.

Color data is not translated with an additional command as control_phase_data is already what is needed here. Then the color data has to follow immediately.

### DIFF
--- a/components/esp_lcd/src/esp_lcd_panel_io_i2c.c
+++ b/components/esp_lcd/src/esp_lcd_panel_io_i2c.c
@@ -141,10 +141,15 @@ static esp_err_t panel_io_i2c_tx_buffer(esp_lcd_panel_io_t *io, int lcd_cmd, con
         ESP_GOTO_ON_ERROR(i2c_master_write_byte(cmd_link, is_param ? i2c_panel_io->control_phase_cmd : i2c_panel_io->control_phase_data, true),
                         err, TAG, "write control phase failed"); // control phase
     }
-    uint8_t cmds[4] = {BYTESHIFT(lcd_cmd, 3), BYTESHIFT(lcd_cmd, 2), BYTESHIFT(lcd_cmd, 1), BYTESHIFT(lcd_cmd, 0)};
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

tore-espressif:
@Slamy Could you please share relevant pieces of code, so we can reproduce the issue? Or photos of display artifacts?

I can get a basic scrolling text working even without the fix

Slamy:
I'm terribly sorry for forgetting this pull request. It has now quite aged. Let me wrap up things with all requests from you.
First of all, thank you for integration. The current master state fde4afc67a2035a1f461dc511cd85dde1ef90368 does no longer suffer from the problem. As you wish to have some evidence, I've reverted the commits a7bd917c427abe5062a33df0e37e31f8792a8899 and 256945997d0edcb41f0ed2a0e311ec42d3e9b42f and then executed i2c_oled_example_main.c (as is) on my board. I've recorded a video of the artefacts and uploaded it as non listed on YouTube. It is better seen in motion: https://youtube.com/shorts/PFjmo4dvGb4?feature=share

jonathanfoster:
Any idea when this fix will make it into a release? I'm on v4.4.2 and just ran into this issue. The fix is already in the `release/v4.4` branch.

espzav:
Hi @jonathanfoster,
It will be in next release (v4.4.3). The release is waiting for the fix of the last task. It will be released in the next weeks.

espzav:
Hi @jonathanfoster,
the version 4.4.3 was released.

PR #9213, opened 2022-06-23
Fix out of range in mcpwm example (IDFGH-7666)

PieterVandaeleDekimo:
hall_sensor_value 7 must not be allowed (s_hall_actions  array has length 7, index 0 to 6)


suda-morris (approved):

suda-morris:
sha=856ab38681f1881c428ee3faf9971ff41b792296

Issue #9198, opened 2022-06-21
Problem with UART, if UART is placed in IRAM [problem found] (IDFGH-7646)


vavroa:
Problem found in IDF 4.4.1

If the SPI function that accesses the external flash is being executed and the ISR is called from the UART at this point, the entire program will crash.

The problem is that the ISR from UART uses RINGBUFFER, which has some functions located in external memory.

I am attaching a patch that I applied to this problem.

[0001-Ringbuffer-to-iram.zip](https://github.com/espressif/esp-idf/files/8949117/0001-Ringbuffer-to-iram.zip)
.

meltdown03:
You should make a Pull Request instead of attaching the patch here. That way it can be tested and edited by the Espressif team

igrr:
Hi @vavroa,

Could you please attach your sdkconfig file?

In v4.4.1 release, FreeRTOS and ringbuffer functions are already placed into IRAM, unless CONFIG_FREERTOS_PLACE_FUNCTIONS_INTO_FLASH configuration option is enabled. I'd like to check if you are seeing this issue with this option is enabled or not.

vavroa:
Indeed, this option was enabled, probably for this reason, the ringbuffer functions were placed in FLASH, which caused problems in their access from ISR.

Now I've tested it with option CONFIG_FREERTOS_PLACE_FUNCTIONS_INTO_FLASH turned off and it looks like it works.

igrr:
Thanks for confirming!

Looks like the action on our side would be to either prevent `CONFIG_FREERTOS_PLACE_FUNCTIONS_INTO_FLASH` from being used together with UART-ISR-in-IRAM, or somehow automatically move ringbuffer functions back to IRAM in that case. Another option would be to create a new option just for ringbuffers, `CONFIG_ESP_RINGBUFFER_PLACE_FUNCTIONS_INTO_FLASH`, and make sure it is disabled when UART IRAM interrupts are enabled.

vavroa:
Both options are a good idea.

In our case, we are at the limit of IRAM capacity, so we tried these different options.

PR #9239, opened 2022-06-27
Ringbuffer to iram (IDFGH-7696)

vavroa:
#9198

### DIFF
--- a/components/esp_ringbuf/ringbuf.c
+++ b/components/esp_ringbuf/ringbuf.c
@@ -37,6 +37,12 @@
 #define rbGET_RX_SEM_HANDLE( pxRingbuffer ) ( pxRingbuffer->xRecvSemHandle )
 #endif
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

bugadani:
Is it really necessary to place `prvInitializeNewRingbuffer` into IRAM?

ESP-Marius:
@vavroa Thanks for reporting this.

Seems like the problem might be bigger than just the UART though, as other drivers might also experience similar issues.

So we decided to take a different approach and have an internal MR in-review now where we add separate kconfig option for placing the ringbuffers in flash.

Lets keep this PR open until the fix is merged though, just incase we decide to with your solution instead :)

ESP-Marius:
Resolved by a940064748fcf9da26592516258fe65d2e52aa9a

PR #9264, opened 2022-06-30
Fix tinyusb_driver_install (dangling pointer) (IDFGH-7723)

Lsitar:
Pointer to `gpio_conf` variable was used out of its scope. This caused dereference to some trash values in the following call `usb_new_phy()`. I moved the variable definition to include usb_new_phy() call in its scope and now it's working ok.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=db90b512a11479a11275b4f08d5ea35127d9c4c0

PR #9266, opened 2022-06-30
esp_lcd: add ILI9341 panel support

PaintYourDragon:
Adapted from the st7789 code; mostly the initialization is different. Tested on Adafruit 2.8" ILI9341 breakout, ESP32-S3-DevKitC-1 N8R8.

### DIFF
--- a/components/esp_lcd/include/esp_lcd_panel_vendor.h
+++ b/components/esp_lcd/include/esp_lcd_panel_vendor.h
@@ -65,6 +65,19 @@ esp_err_t esp_lcd_new_panel_nt35510(const esp_lcd_panel_io_handle_t io, const es
  */
 esp_err_t esp_lcd_new_panel_ssd1306(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config, esp_lcd_panel_handle_t *ret_panel);
 
+/**
+ * @brief Create LCD panel for model ILI9341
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
@PaintYourDragon Thank you for sharing the driver code. The code looks nice to me. But AFAIK, the ILI9341 shares a similar register layout to the ST7789, that's why esp-idf doesn't provide another ILI9341 API out of the box. 

We can't hardcode the panel-specific configurations (e.g. the power control, gama settings) into the driver, because they can be different for LCD manufactured by different OEMs. So we'd prefer to put those configurations to the user side (i.e. in the bsp). For example, 

```c
// create a LCD panel
esp_lcd_new_panel_st7789(io_handle, &panel_config, &panel_handle);

// basic and common initialization
esp_lcd_panel_init(panel_handle);

// LCD module specific configurations, like Gama, Power
esp_lcd_panel_io_tx_param(io_handle, 0xEF, (uint8_t[]) { 0x03, 0x80, 0x02 }, 3);
...

// turns on the screen
esp_lcd_panel_disp_on_off(panel_handle, true);
```

PaintYourDragon:
Fair points! The ST7789 driver does in fact work well enough, once the gamma is set “externally” on the user side. Will close this PR, thanks for looking and for the tips.

Issue #8336, opened 2022-02-02
Crash in xEventGroupSetBits after a few hours of running (IDFGH-6707)


maBarabas:
## Environment

- Development Kit:     none
- Module or chip used:  ESP32-WROOM-32D
- IDF version (run ``git describe --tags`` to find it): v4.3.1-dirty (we have updated nimble to https://github.com/espressif/esp-nimble/commit/94afe2711752ba08bef93780be1b9e5143a87159, there are no other changes)
- Build System:         idf.py
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): xtensa-esp32-elf-gcc (crosstool-NG esp-2021r1) 8.4.0
- Operating System:     Linux
- Using an IDE?: No
- Power Supply:        external 5V

## Problem Description

After a few hours of running, a crash happens due to null pointer dereference in FreeRTOS code. We have reproduced this on multiple hardware units and multiple times on each unit.

We are currently running the test with latest esp-idf, and trying to create a minimal project for reproduction.

Our firmware is setting event bits every second to update time information on a screen. Before the crash, everything works as expected.

It looks like there might be a race condition where the task is already unblocked from a timeout when `xEventGroupSetBits` is trying to unblock it?

This looks similar to https://github.com/espressif/esp-idf/issues/242. The [commit message](https://github.com/espressif/esp-idf/commit/01c4a980ee05b5eddefd9d01a2bbdecde9944ce7) for the fix says the race condition is less likely to happen. Does this mean that the issue is still present?

### Expected Behaviour

No crashes.

### Actual Behaviour

See description above.

### Steps to reproduce

1. Start firmware
2. After a while, the crash happens

### Code to reproduce this issue

```cpp
// This is how we set the bits:
// The DISPLAY_INFO bits and change value are provided by a UART peripheral connected to the system
    EventBits_t bits = 0;
    if(change & DISPLAY_INFO_MASK_NEW_INDEX)
    {
        bits |= CD_NEW_INDEX;
    }
    if(change & DISPLAY_INFO_MASK_NEW_TRACK)
    {
        bits |= CD_NEW_TRACK;
    }
    if(change & DISPLAY_INFO_MASK_TIME_CHANGED)
    {
        bits |= CD_NEW_TIME;
    }

    xEventGroupSetBits(xCdEventGroup, bits);


// This is how we are waiting for them in a different task:
    xEventGroupValue = xEventGroupWaitBits( xCdEventGroup,
[...]

PR #9271, opened 2022-07-02
Fix SMP race in xEventGroupSetBits() -> xEventGroupWaitBits() timeout  (IDFGH-7731)

SpComb:
Fixes https://github.com/espressif/esp-idf/issues/8336

This is a draft PR for my best attempt to fix the SMP race condition described in https://github.com/espressif/esp-idf/issues/8336, where a task calling `xEventGroupSetBits()` on one core can race with `xTaskIncrementTick()` handling a `xEventGroupWaitBits()` timeout on a different core, leading to a crash in `vTaskRemoveFromUnorderedEventList()` -> `uxListRemove()` or lockup.

Tested against the `release/v4.4` branch using the repro case in https://gist.github.com/SpComb/909d20847645aa22361cf4d3e8bc3700 to verify that this fixes the crash/lockup and does not lead to lost event bits.

I'll be happy to clean this up and rebase against master based on feedback if this approach is acceptable - but I'm skeptical about the modifications to the FreeRTOS interfaces, and it looks like the FreeRTOS SMP implementation in the master branch might be changing for 5.0?

### DIFF
--- a/components/freertos/event_groups.c
+++ b/components/freertos/event_groups.c
@@ -589,7 +589,7 @@ EventBits_t xEventGroupSetBits( EventGroupHandle_t xEventGroup,
     List_t const * pxList;
     EventBits_t uxBitsToClear = 0, uxBitsWaitedFor, uxControlBits;
     EventGroup_t * pxEventBits = xEventGroup;
-    BaseType_t xMatchFound = pdFALSE;
+    BaseType_t xMatchFound = pdFALSE, xTaskUnblocked = pdFALSE;
 
     /* Check the user is not attempting to set the bits used by the kernel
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

Dazza0:
@SpComb Thanks for the PR. We've noticed this issue as well and have fixed this internally. However we went with a slightly different approach. The fix has already been merged on our internal master and is awaiting sync with Github. I'll Keep #8336 open until the fix is backported to v4.4

SpComb:
I'm happy to hear that a proper fix is on the way!

PR #9276, opened 2022-07-03
Change placeholder in ESP_LOGD conditionally depending on FF_FS_EXFAT (IDFGH-7735)

jkearins:
In case of using EXFAT by setting in ffconf.h:
`#define FF_FS_EXFAT    1`
the type FSIZE_t is changing from 4 to 8 bytes.

As a result,  ESP_LOGD() in vfs_fat_lseek() does not compile:
```
error: format '%d' expects argument of type 'int', but argument 8 has type 'FSIZE_t' {aka 'long long unsigned int'} [-Werror=format=]
     ESP_LOGD(TAG, "%s: offset=%ld, filesize:=%d", __func__, new_pos, f_size(file));
```

To solve the problem we need to change %d with %lld conditionally, depending on FF_FS_EXFAT.


Alvin1Zhang:
Thanks for your contribution.

rrtandler:
sha=7e023b7bce179679f318850757768cbaa5c2e863

jkearins:
I cleaned the commit history and recreated PR.

PR #9302, opened 2022-07-06
ESP32S3 USB external PHY pinout

Lsitar:
Change USB pinout for ESP32-S3 as in Reference Manual Figure 29-3.
Now the pinout causes crash and uses pins non existing in ESP32-S3. The connection for external PHY is not well described, pinout seems not to be user configurable and the only place where it is documented is Reference Manual chapter 29.
I use internal PHY for JTAG and external for USB CDC simultaneously, after this pinout change it works ok.

### DIFF
--- a/components/soc/esp32s3/include/soc/usb_pins.h
+++ b/components/soc/esp32s3/include/soc/usb_pins.h
@@ -15,11 +15,11 @@
 #pragma once
 
 /* GPIOs used to connect an external USB PHY */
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=9ceff23c6d50b37a9fba4c370cffa1654fedeecd

suda-morris (approved):

PR #9311, opened 2022-07-07
Expand lcd/i80_controller example to support ILI9341 display (IDFGH-7775)

PaintYourDragon:
This replaces #9266 (closed yesterday as the wrong approach). i80 LCD example is extended to show ILI9341 usage; nearly but not quite the same as ST7789, requires no driver changes, just user code. If nothing else, the addition of .flags.swap_color_bytes being the reverse of the corresponding LvGL setting might be worth keeping.
Tested on ESP32-S3-DevKitC-1 N8R8 with Adafruit ILI9341 breakout (8-bit parallel interface).

### DIFF
--- a/examples/peripherals/lcd/i80_controller/main/i80_controller_example_main.c
+++ b/examples/peripherals/lcd/i80_controller/main/i80_controller_example_main.c
@@ -55,20 +55,16 @@ static const char *TAG = "example";
 #define EXAMPLE_PIN_NUM_BK_LIGHT       1
 
 // The pixel number in horizontal and vertical
-// Bit number used to represent command and parameter
-#if CONFIG_EXAMPLE_LCD_I80_CONTROLLER_ST7789
 #define EXAMPLE_LCD_H_RES              240
 #define EXAMPLE_LCD_V_RES              280
+// Bit number used to represent command and parameter
+#if CONFIG_EXAMPLE_LCD_I80_CONTROLLER_ST7789
 #define EXAMPLE_LCD_CMD_BITS           8
 #define EXAMPLE_LCD_PARAM_BITS         8
 #elif CONFIG_EXAMPLE_LCD_I80_CONTROLLER_NT35510
-#define EXAMPLE_LCD_H_RES              240
-#define EXAMPLE_LCD_V_RES              280
 #define EXAMPLE_LCD_CMD_BITS           16
 #define EXAMPLE_LCD_PARAM_BITS         16
 #elif CONFIG_EXAMPLE_LCD_I80_CONTROLLER_ILI9341
-#define EXAMPLE_LCD_H_RES              320
-#define EXAMPLE_LCD_V_RES              240
 #define EXAMPLE_LCD_CMD_BITS           8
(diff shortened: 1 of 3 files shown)


examples/peripherals/lcd/i80_controller/main/i80_controller_example_main.c
```diff
@@ -55,15 +55,22 @@ static const char *TAG = "example";
...
 #define EXAMPLE_LCD_CMD_BITS           8
 #define EXAMPLE_LCD_PARAM_BITS         8
 #elif CONFIG_EXAMPLE_LCD_I80_CONTROLLER_NT35510
+#define EXAMPLE_LCD_H_RES              240
+#define EXAMPLE_LCD_V_RES              280
 #define EXAMPLE_LCD_CMD_BITS           16
 #define EXAMPLE_LCD_PARAM_BITS         16
+#elif CONFIG_EXAMPLE_LCD_I80_CONTROLLER_ILI9341
+#define EXAMPLE_LCD_H_RES              320
+#define EXAMPLE_LCD_V_RES              240
```
suda-morris, line 71: Even for the same LCD controller, we can't guarantee the resolution will be the same for all LCMs, maybe your ILI9341's resolution is 320*240, but mine is 280*240. It's impossible to make the example work out of the box for all LCD break-out boards, without changing some of these default values :thinking: That's why I didn't make the resolution configuration strong associated with the LCD model.

examples/peripherals/lcd/i80_controller/main/i80_controller_example_main.c
```diff
@@ -177,19 +187,44 @@ void app_main(void)
...
+    // the gap is LCD panel specific, even panels with the same driver IC, can have different gap value
 #if CONFIG_EXAMPLE_LCD_I80_CONTROLLER_ST7789
     esp_lcd_panel_invert_color(panel_handle, true);
+    esp_lcd_panel_set_gap(panel_handle, 0, 20);
 #elif CONFIG_EXAMPLE_LCD_I80_CONTROLLER_NT35510
     esp_lcd_panel_swap_xy(panel_handle, true);
     esp_lcd_panel_mirror(panel_handle, true, false);
-#endif
-    // the gap is LCD panel specific, even panels with the same driver IC, can have different gap value
     esp_lcd_panel_set_gap(panel_handle, 0, 20);
```
suda-morris, line 213: it's was a mistake made by me, in fact, I didn't see an NT35510 screen that needs an extra gap setting. So would you please help me remove this line or just set it to zero gap?

suda-morris (changes requested):
Nice job @PaintYourDragon I like this solution! Thank you for your contribution!

suda-morris (approved):

PaintYourDragon:
Changes made as requested. Thank you for looking!
Deleted both set_gap lines that were 0,0 rather than leave them in; ONLY the ST7789 sets a gap. Confirmed the gap defaults are always 0; this is implicit as the panel_t structure is calloc’d (which sets allocated memory to 0).

suda-morris:
sha=148ce09d97a46602f58791d67db37bfd0f20b751

Issue #9327, opened 2022-07-11
Version 5.0 doesn't work system-wide


stkw0:
## Environment

- Development Kit:      none
- Module or chip used: ESP32-WROOM-32E
- IDF version: release/5.0
- Build System:  idf.py
- Compiler version: xtensa-esp32-elf-gcc (crosstool-NG esp-2022r1-RC1) 11.2.0
- Operating System:  Linux
- Using an IDE?: Kdevelop
- Power Supply:  USB

## Problem Description

I have a package that installs esp-idf system-wide because it have numerous advantages (doesn't need a manual set-up, better integration with IDEs and other tools, easier to upgrade idf version, etc). With version 4.4.1 it worked correctly but if I try to use release/5.0 this approach doesn't work anymore. When running the same command, `idf build` It only shows the next error message and doesn't build the image.

```
ERROR: /home/droman/.espressif/python_env/idf5.0_py3.9_env/bin/python doesn't exist! Please run the install script or "idf_tools.py install-python-env" in order to create it

fatal: not a git repository: '/usr/share/esp-idf/.git'
WARNING: Git version unavailable, reading from source
ESP-IDF v5.0.0
``` 
I am aware that version 5.0 isn't released yet but we use a ksz8863 Ethernet chip and the official idf driver on [eth-drivers](https://github.com/espressif/esp-eth-drivers) requires version 5.0

### Expected Behavior

IDF is capable of building the image

### Actual Behavior

An error is shown

dobairoland:
Have you run the suggested command? I assume that you've had and used your environment from `idf4.4_py3.9_env` until now and after you've updated ESP-IDF, it tries to find and use `idf5.0_py3.9_env` but it cannot find because the installer hasn't been run yet for the new ESP-IDF version.

stkw0:
No, I never run the installer for version 4.4 (and it wasn't a problem for 4.x versions). Starting with 5.0 this breaks as can be seen in the error message, so I guess this is a new check that wasn't in place in previous versions

dobairoland:
All v4.x and v5.0 versions require to run `install.sh` at least once and consequently `. export.sh` in all new shells in order to set up the tools.

The above message is new for v5.0, but the user-side setup requirements have not changed.

Please see https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/linux-macos-setup.html for more information.

dobairoland:
[...]

PR #9328, opened 2022-07-11
add IDF_SKIP_PYTHON_CHECK env var (IDFGH-7790)

stkw0:
In some environments IDF is installed system-wide. With v5.0 some checks are done against user directories which obviously doesn't contain the required dependencies (they are installed system-wide) and as a result IDF fails to execute. A workaround for this is to simply ignore those checks if the user knows those dependencies are really provided, this way IDF can be run successfully. 

Another approach would be to correctly check if dependencies are met wherever they are located.

See https://github.com/espressif/esp-idf/issues/9327

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -2027,6 +2027,11 @@ def action_install_python_env(args):  # type: ignore
 
 
 def action_check_python_dependencies(args):  # type: ignore
+    # In some enviroments python deps are installed system-wide in which case this checks
+    # should be ignored to be able to run idf
+    if os.getenv('IDF_SKIP_PYTHON_CHECK') == '0':
+        return
+
     use_constraints = not args.no_constraints
     req_paths = get_requirements('')  # no new features -> just detect the existing ones
 


igrr:
@stkw0 If you create a global (not user-specific) virtual environment and add the environment's interpreter to PATH, do you still get this error? I think it should work without bypassing the check.

igrr:
Do you mean that you install all the Python package dependencies of ESP-IDF globally (as in, `sudo pip install ...`)?

If that is the case, I'm afraid this won't work in general, since IDF's dependencies may conflict that with dependencies of other software in the system. There will also be a problem if multiple versions of IDF are installed side by side on the same system. With our current installation approach, each version will use its own virtual environment, and as a result can use different versions of packages.

> To avoid breaking this behavior no venv can be used as this would require user set-up (which is unfeasible because the package manager runs in a different user)

I'm afraid I don't understand the reason behind this. It is possible to create a virtual environment as one user and then use it as another user. (For example, you can do `python -m venv /opt/my-venv && /opt/my-venv/bin/python -m pip install ...` as one user and then use this virtual environment as another user.)

stkw0:
I mean more like `sudo apt install idf`. The package contains the dependencies of IDF and "apt" installs all dependencies. The problem of conflict between dependencies will then relay on the package manager itself, in some cases it would be no problem to have different versions of IDF at the same time.

stkw0:
> ESP-IDF install scripts also support an environment variable, `IDF_TOOLS_PATH`. If not set, it is assumed to be `$HOME/.espressif`. However you can set it to another location, outside of user's home directory.
> 
> For example:
> 
> ```
> export IDF_TOOLS_PATH=/opt/esp
> $IDF_PATH/install.sh
> ```
> 
> This will install all the tools _and the virtual environment_ to /opt/esp.
> 
> Later, as a regular user, you can run $IDF_PATH/export.sh and it will pick up the tools installed in /opt/esp. The only requirement is that IDF_TOOLS_PATH should be set to the same value in the user's environment.

Thanks, I would try if modifying the package to put IDF_TOOLS_PATH=/usr/share/esp-idf/tools in /etc/env.d fixes the issue

stkw0:
Okay, thank you for your time

HiFiPhile:
It's really useful to make locally reproducible workflow. I presume if I checked out a revision then all rebuild should be the same (except hash which has a date change) .

What happened actually is 2 weeks later after my vacation my build workflow failed, as it downloaded new espidf.constraints.v5.0.txt and complains esptool 4.2 should be installed instead of 4.1

PR #9350, opened 2022-07-13
Add a helpful comment to sdkconfig.defaults files in examples (IDFGH-7820)

egnor:
ESP-IDF's behavior for `sdkconfig.defaults` files can be confusing; if you edit the file after your first build (when `sdkconfig` is automatically generated) the edits have no effect.

This adds a short comment header to the `sdkconfig.defaults` file in example projects explaining the file and pointing to documentation (which seems helpful regardless) and calling out this particular gotcha. These examples get widely copied and adapted, so having a pointer comment seems helpful?

(Relevant to https://github.com/platformio/platform-espressif32/issues/847)

### DIFF
--- a/components/cxx/test_apps/exception/sdkconfig.defaults
+++ b/components/cxx/test_apps/exception/sdkconfig.defaults
@@ -1,3 +1,9 @@
+# Defaults for generating a new "sdkconfig" file (project settings).
+# https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/build-system.html#custom-sdkconfig-defaults
+#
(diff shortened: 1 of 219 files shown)


Alvin1Zhang:
Thanks for your contribution.

Dazza0:
@egnor Thank you for your contribution and sorry for the late review. However I don't think the proposed change are feasible as it would require the comment to be added to every `sdkconfig.default` in ESP-IDF.

I think adding the help comment to the `idf.py menuconfig --help` prompt would be an easier change instead. Please feel free to reopen this PR if you have other considerations.

PR #9365, opened 2022-07-15
fix write to null pointer if malloc failed (IDFGH-7836)

devosavchenko:


KaeLL:
Why not just use `calloc`?

devosavchenko:
> Why not just use `calloc`?

This example uses the malloc function throughout to dynamically allocate one piece of memory. Also, this Espressif documentation [Heap Memory Allocation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/mem_alloc.html ) says next
"For most purposes, the standard libc malloc() and free() functions can be used for heap allocation without any special consideration."

There is no such mention of the calloc function. Judging by the search for calloc calls in the project, I think there should be no problems using calloc. Also in this example, only one block of data is allocated, given all of the above, I don’t see much point in replacing malloc with calloc. Please correct me if I'm wrong.

examples/wifi/espnow/main/espnow_example_main.c
```diff
@@ -321,13 +321,13 @@ static esp_err_t example_espnow_init(void)
 
     /* Initialize sending parameters. */
     send_param = malloc(sizeof(example_espnow_send_param_t));
-    memset(send_param, 0, sizeof(example_espnow_send_param_t));
```
KaeLL, line 324: Alright, so the reason to use calloc is to avoid the redundancy. calloc = malloc + memset.

Alvin1Zhang:
Thanks for your contribution.

PR #9375, opened 2022-07-17
Tools: resolve symlinks in IDF_PATH when installing (IDFGH-7846)

simpkins:
Use `pwd -P` to resolve any symlinks in the current directory path.
This makes the behavior in the shell script similar to the idf_tools.py
code, which calls `os.path.realpath()`.  Without this, multiple entries
can get created in the `idfInstalled` dictionary in idf-env.json, and
the installed targets and features are not fully present in all entries.
This results in a broken installation, where `export.sh` cannot set up
the environment correctly afterwards.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@ set -e
 set -u
 
 basedir=$(dirname "$0")
-IDF_PATH=$(cd "${basedir}"; pwd)
+IDF_PATH=$(cd "${basedir}"; pwd -P)
 export IDF_PATH
 
 echo "Detecting the Python interpreter"


Alvin1Zhang:
Thanks for your contribution.

simpkins:
> Thank you for the PR @simpkins. Could you please extend this functionality to the `install.fish` script as well, so we can accept the suggestion under your name and keep the ESP-IDF consistent?

It looks like install.fish already calls realpath to resolve symlinks, so I believe it already handles this case correctly, and doesn't need any updates:
https://github.com/espressif/esp-idf/blob/9b8c558e63d95b491355a22e69f247834e899b47/install.fish#L3

dobairoland:
Closed by ff75c4f3b5e03cb7b066ad14c3d02c09a3636fe8

PR #9408, opened 2022-07-22
nvs_partition_gen: Allow up to 4000 byte strings with NVS V2 (IDFGH-7884)

projectgus:
Hi folks! :wave: 

[As documented](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html#keys-and-values), the limit for NVS format v2 is 4000 bytes for a string.

nvs_partition_gen.py had what looks like a copy-paste bug, and was enforcing the NVS V1 format blob size limit for V2 strings also.

### DIFF
--- a/components/nvs_flash/nvs_partition_generator/nvs_partition_gen.py
+++ b/components/nvs_flash/nvs_partition_generator/nvs_partition_gen.py
@@ -51,13 +51,6 @@ def reverse_hexbytes(addr_tmp):
 
 
 class Page(object):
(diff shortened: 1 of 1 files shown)


mahavirj:
@projectgus Thanks for sending across this fix. LGTM!

mahavirj:
sha=bb141956dfbac7d263862fda6f6e6865846e3c03

Alvin1Zhang:
@projectgus Thanks for the contribution.

PR #9498, opened 2022-08-04
bt member typo (IDFGH-7987)

tgotic:
Terminate sec_event.cfm_req.bd_name string, not sec_event.pin_req.bd_name


Alvin1Zhang:
Thanks for your contribution.

PR #9563, opened 2022-08-16
Add git-lfs to docker container (IDFGH-8060)

tpambor:
I propose to add git-lfs to the docker container to support pulling git repositories and submodules which use [Git LFS](https://git-lfs.github.com/).

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -14,6 +14,7 @@ RUN : \
     curl \
     flex \
     git \
+    git-lfs \
     gperf \
     lcov \
     libbsd-dev \


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
Hi @tpambor. Could you please explain the use case where this could be useful? There are many users of this image who are very unhappy by adding anything to these images.

Git is there for cloning and later describing tags. However, I don't see why `git-lfs` could be useful there.
1. If you adding anything else into the image then the proper way could be to have a custom one for your need deduced from the esp-idf one.
2. The image is used with projects outside the image. Therefore, you can set it up there everything locally using your local installation of Git LFS without the need to adding it into the esp-idf image.

tpambor:
Hi @dobairoland,

I got a component which is stored in a repository which uses Git LFS. The `IDF Component Manger` clones it as part of the build process but due to the lack of git-lfs only the placeholder files are checked out and not the files stored with LFS.

dobairoland:
Oh, thank. That makes sense.

dobairoland:
sha=dba5ff8b3abbdb77def2caa106f91403b76b05a2

tpambor:
@dobairoland 

Could this also be backported to v4.4 and v5.0 branch?

dobairoland:
> Could this also be backported to v4.4 and v5.0 branch?

Yes, I'll attempt to do so. It is blocked by another issue but we are working on it....

PR #9582, opened 2022-08-18
[flash] Implement generic read register (IDFGH-8082)

KonssnoK:
As per title,
The current read_status register needs to be expanded to a more generic "Read register" in order to support latest winbond flash.

(And probably also other brands?)

If you find a better way of updating the status register address depending on the flash family please let me know.


Alvin1Zhang:
Thanks for your contribution.

KonssnoK:
we found other issues, this PR will be updated with more changes

PR #9591, opened 2022-08-19
[flash] Winbond W25N support (IDFGH-8091)

KonssnoK:
As per title, support for winbond W25N family.
Since there are multiple changes from normal winbond devices, it was easier to generate a new driver.

We are still testing the code to see if we missed some implementation changes.

Replace 
https://github.com/espressif/esp-idf/pull/9582

### DIFF
--- a/components/spi_flash/CMakeLists.txt
+++ b/components/spi_flash/CMakeLists.txt
@@ -51,6 +51,7 @@ else()
         "spi_flash_chip_boya.c"
         "spi_flash_chip_mxic_opi.c"
         "spi_flash_chip_th.c"
+        "spi_flash_chip_winbond_w25n.c"
         "memspi_host_driver.c")
 
     list(APPEND cache_srcs
--- a/components/spi_flash/include/spi_flash_chip_driver.h
+++ b/components/spi_flash/include/spi_flash_chip_driver.h
@@ -1,16 +1,8 @@
-// Copyright 2015-2019 Espressif Systems (Shanghai) PTE LTD
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


igrr:
Hi @KonssnoK,

Thank you for the PR! Have you considered packaging the support for the new flash chip as a standalone component, so that it could be included into various projects as shown in this example: https://github.com/espressif/esp-idf/tree/master/examples/storage/custom_flash_driver?
You could then host this component under your Github account, so that the development and releases wouldn't be tied to IDF release cycle. Other users interested in such a chip could then add this component to their project, either using a submodule or via the IDF component manager.

KonssnoK:
Hi @igrr , no we didn't consider it, but we'll take a look!

Consider that some changes to idf will still be necessary: 
in read_reg the status register is defined as an enum but this is not generic.

ginkgm:
Hi @KonssnoK ,
I would like to know what's the benefits of changing the type of the register ID from uint32_t to uint8_t?

Chainging the argument type of interface may break the ROM's behavior.

KonssnoK:
@ginkgm uint32_t is fine (although status addresses are usually 8 bit)
I think we misunderstood each other.
I'm talking about changing

`esp_err_t spi_flash_chip_generic_read_reg(esp_flash_t* chip, spi_flash_register_t reg_id, uint32_t* out_reg);`

into

`esp_err_t spi_flash_chip_w25n_read_reg(esp_flash_t* chip, uint8_t reg_id, uint32_t* out_reg)`

where 

spi_flash_register_t  is a useless enum

```
typedef enum {
    SPI_FLASH_REG_STATUS = 1,
} spi_flash_register_t;
```

which is never used (ignored in every call, i checked).

KonssnoK:
closing while we work on a separated driver (which might be trickier than previously thought)

PR #9620, opened 2022-08-24
Allow global CA to be written directly via an API (IDFGH-8122)

jrepp:
In a very low memory environment with a non-trivial number of certificates, it is much more memory efficient to construct the global CA chain directly. This global chain of trust can then be used from both the HTTPS and MQTTS transports.

### DIFF
--- a/components/esp-tls/esp_tls_mbedtls.c
+++ b/components/esp-tls/esp_tls_mbedtls.c
@@ -829,6 +829,16 @@ esp_err_t esp_mbedtls_init_global_ca_store(void)
     return ESP_OK;
 }
 
+esp_err_t esp_mbedtls_set_global_ca(mbedtls_x509_crt *trust_ca)
+{
+    if (global_cacert != NULL) {
+        free(global_cacert);
+    }
+
+    global_cacert = trust_ca;
+    return ESP_OK;
+}
+
 esp_err_t esp_mbedtls_set_global_ca_store(const unsigned char *cacert_pem_buf, const unsigned int cacert_pem_bytes)
 {
 #ifdef CONFIG_MBEDTLS_DYNAMIC_FREE_CA_CERT
--- a/components/esp-tls/private_include/esp_tls_mbedtls.h
+++ b/components/esp-tls/private_include/esp_tls_mbedtls.h
@@ -8,6 +8,10 @@
 #include "esp_tls.h"
 #include "esp_tls_private.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /**
  * Internal Callback API for mbedtls_ssl_read
  */
@@ -123,6 +127,11 @@ void esp_mbedtls_free_client_session(esp_tls_client_session_t *client_session);
 esp_err_t esp_mbedtls_init_global_ca_store(void);
 
 /**
+ * Set the global trust CA directly
+ */
+esp_err_t esp_mbedtls_set_global_ca(mbedtls_x509_crt *trust_ca);
+
+/**
  * Callback function for setting global CA store data for TLS/SSL using mbedtls
  */
 esp_err_t esp_mbedtls_set_global_ca_store(const unsigned char *cacert_pem_buf, const unsigned int cacert_pem_bytes);
@@ -136,3 +145,8 @@ mbedtls_x509_crt *esp_mbedtls_get_global_ca_store(void);
  * Callback function for freeing global ca store for TLS/SSL using mbedtls
  */
 void esp_mbedtls_free_global_ca_store(void);
+
+
+#ifdef __cplusplus
+}
+#endif


Alvin1Zhang:
Thanks for your contribution.

AdityaHPatwardhan:
Hi, @jrepp Thanks for your contribution.
IIUC, I see that you want to set the global ca chain directly instead of calling the `esp_tls_set_global_ca_store` API to parse and add each certificate to the global ca chain.  Do you plan to provide the cert chain directly from the flash?

Can you please mention the numbers about heap saved with this alternate approach for your case?

1) With this approach I see the following things.
* The cert chain shall not be parsed till the time when it is actually used for the TLS handshake.

In your PR, I see that you have added the respective API only in `esp_tls_mbedtls.h` and not in `esp_tls.h`. The API needs to be in `esp_tls.h` as well to be able to be used by the application.

jrepp:
Thank you for looking at this change. I appreciate your time.

Yes, we provision the .der format certificates to the device flash, at startup the certificates are built into a chain and set as the global CA chain. This is mostly a memory savings trick as we have very little memory at runtime for application functionality. 

The drawback of the existing esp_crt_bundle functionality is that you lose the ability to side-load certificates and they bloat the executable flash which is a problem when doing multi-partition OTA (cert bundle size x #OTA partitions)

It might be interesting to look at improving the esp_crt_bundle mechanism to initialize a cert partition but that would be a much more invasive change.

> In your PR, I see that you have added the respective API only in esp_tls_mbedtls.h and not in esp_tls.h. The API needs to be in esp_tls.h as well to be able to be used by the application.

I will make this change.

jrepp:
> > In your PR, I see that you have added the respective API only in esp_tls_mbedtls.h and not in esp_tls.h. The API needs to be in esp_tls.h as well to be able to be used by the application.
> 
> I will make this change.

In investigating this change it appears you are requesting this functionality be made to work across other TLS implementations (adding the API to esp_tls.h). However, it is too low-level and specific to mbedtls for that to be practical. The esp_crt functionality also has this limitation.

jrepp:
Regarding memory utilization - there are two main savings:
- Storage memory of PEM
- Parse memory of PEM

This approach allows the user to choose their own format for certificates, for example, DER format certificates. 
At rest, this is a savings of >2x (from 2021 bytes to 889 bytes in an example GSR certificate). At runtime, there is a non-trivial amount of memory that is used to parse the PEM certificates which may not be available to the application.

AdityaHPatwardhan:
> Yes, we provision the .der format certificates to the device flash, at startup the certificates are built into a chain and set as the global CA chain. This is mostly a memory savings trick as we have very little memory at runtime for application functionality.

Do you mean that the certificates are parsed at the start of the code, if they are done so then I think the API `esp_tls_set_global_ca_store` does the same. It parses the certificate and loads it into the global ca chain. So if it is called on startup then the memory required for parsing would be one time and once it is freed internally, only the heap consumed by CA chain shall be remaining. What difference does your PR add in this case?
I had thought that you have parsed the certificate before startup and generated `mbedtls_x509_crt` kind of structure and have stored it directly on flash.


> The drawback of the existing esp_crt_bundle functionality is that you lose the ability to side-load certificates and they bloat the executable flash which is a problem when doing multi-partition OTA (cert bundle size x #OTA partitions)

I think if you use default certificate bundle that includes a large number of CA certificates, then it increases the size by a larger extent.
However, if you use only the most common certs then the size of the bundle is reduced. Moreover, if you give you a custom list of certificates to be made into a bundle then the size depends on the number of certificates that you are providing. Have you tried that ?

> In investigating this change it appears you are requesting this functionality be made to work across other TLS implementations (adding the API to esp_tls.h). However, it is too low-level and specific to mbedtls for that to be practical. The esp_crt functionality also has this limitation.

What I had meant was , `esp_tls_mbedtls.h` is a private header. Even if you add any API's in that file, the application won't be able to use it. You need to have respective entry in `esp_tls.h` as well. It is okay if it is not supported for `wolfSSL` stack, it can just be marked as `not supported`.


> This approach allows the user to choose their own format for certificates, for example, DER format certificates.
At rest, this is a savings of >2x (from 2021 bytes to 889 bytes in an example GSR certificate). At runtime, there is a non-trivial amount of memory that is used to parse the PEM certificates which may not be available to the application.

Ohh, okay I think this is a different question. I dont know why you are not using the same API for `DER` encoded certificates as well? I think the API `mbedtls_x509_crt_parse` which has been internally used in the `esp_tls_set_global_ca_store` can be used to parse both PEM and DER encoded certificates. ([reference](https://github.com/Mbed-TLS/mbedtls/blob/fac3ea5656baa78ef5aa7ae84e79797d330f4215/include/mbedtls/x509_crt.h#L537)

jrepp:
I think based on your feedback we should abandon this PR. Thanks for your time and the pointers.

Perhaps there are changes in the latest stable that need to be investigated or some misuse of APIs. This change would just be a workaround for some other underlying problems. I'll keep the change on the local branch to keep the firmware of the devices working until there is time to investigate further or migrate off of these APIs to something else.

AdityaHPatwardhan:
> I think based on your feedback we should abandon this PR. Thanks for your time and the pointers.

@jrepp  Please feel free reopen the PR if you still think that this change fixes your issue and current IDF behaviour does not utilise your resources efficiently. Actually, I thought a discussion about requirements and possible solutions should happen so that we would reach to a best-case solution that solves all your problems and without adding any additional complexity to IDF.

jrepp:
Sure I appreciate that. After discussing it I don't think it's a good change. It's a kludge. It's working around issues in the underlying system that perhaps there are other (better?) workarounds.

Storing certificates on a global and hiding the parse and memory allocation is not a clean design. Adding an override on that (this PR) and giving access to that global is not making it better, it's making it worse. It's fine on a local build as a workaround but not a good PR.

IMO a better design would be to have a container abstraction for certs that the user controls and is passed into the communication librararies as a dependent api. This way you can avoid the hidden global state management and interlocking inconsistent behavior and cert chain control of the https and mqtts libraries. 

An inversion of control would allow deferred loading of elements in the cert chain instead of commiting heap. In fact mbedtls supports just this style of working internally but it's not exposed in the IDF. 

There are probably plenty of reasons it's not implemented this way as it's complicated to support all these libraries and optional things glued together. E.g I have no idea how wolfSSL works.

Requirements:
- one cert chain for all TLS connections
- user controlled memory, ideally a cert checking callback to avoid commiting heap to all certs
- container for certs that supports storage in its own partition to avoid bloating the application partitions 
- tool to prepare and load certs to partition

Ideally the cert validation is implemented with callbacks and the reference implementation reads from the cert partition without requiring all of the certs to reside in dram.

PR #9628, opened 2022-08-24
essl_sdio: should be available even if platform can only be sdio master

franck-exceenis:
When using a ESP32S3 to control a ESP32 over ESSL/SDIO, compilation doesn't work:
```
/build/esp-idf/examples/sdio/host/main/app_main.c:202: undefined reference to `essl_sdio_init_dev'
```
but applying this patch let the magic happen.

Tested with examples from `examples/peripherals/sdio`

image

image


Alvin1Zhang:
Thanks for your contribution.

PR #9637, opened 2022-08-25
export.fish: set IDF_PATH without changing current working directory (IDFGH-8140)

imcaizheng:


dobairoland:
Thank you for the contribution. We'll take a look at this soon.

imcaizheng:
I just found that my initial commit didn't take care of the case when the source command is executed under the root directory of esp-idf. The last commit I uploaded should fix it.

export.fish
```diff
@@ -7,13 +7,16 @@ end
 
 function __main
     if not set -q IDF_PATH
-        set -gx IDF_PATH (cd (dirname (status -f)); and pwd)
+        set -gx IDF_PATH (dirname (status -f))
+        if test "$IDF_PATH" = "."
+            set -gx IDF_PATH $PWD
+        end
```
mfialaf, line 13: Hi @imcaizheng, thanks for finding this bug and your suggested change. 

I would improve it a little, so that ``IDF_PATH`` is extracted from absolute path of this script and then there is no need to test the ``$IDF_PATH`` for ``.`` since it always is a valid path. 

The same applies to setting and testing ``script_dir``.

Please edit these lines in your changes and squash or delete the second commit so we can merge it as one change.

Thank you
```suggestion
        set -gx IDF_PATH (dirname (readlink -m (status -f)))
```

dobairoland:
sha=84ec01c3301c33ec9a61ff03ab0703c653d39180

Issue #8364, opened 2022-02-09
esp-2021r2-patch2-8.4.0 not working if user folder has a space in it (IDFGH-6735)


apiesse:
## Environment
- Development Kit:  none
- Module or chip used: ESP32
- IDF version: v4.4
- Build System: CMake
- Compiler version : xtensa-esp32-elf-gcc (crosstool-NG esp-2021r2-patch2) 8.4.0
- Operating System: Windows
- (Windows only) environment type: idf.py
- Using an IDE?: Yes, Eclipse 2021-06 with esp-idf plugin

## Problem Description
I've just updated a project from esp-idf-v4.3.2 to esp-idf-v4.4 and I have the following error message:
```
xtensa-esp32-elf-gcc.exe: error: C:\Users\Marco: No such file or directory
xtensa-esp32-elf-gcc.exe: error: Mariotti\AppData\Roaming/.ccache/tmp/cmd_args.bHD3G8: No such file or directory
xtensa-esp32-elf-gcc.exe: fatal error: no input files
compilation terminated.
```
As you can guess, the user directory is "C:\Users\Marco Mariotti\", but somehow the space in it is no longer escaped.
The error was not present in esp-idf-v4.3.2 (using esp-2021r2-8.4.0 toolchain).

igrr:
Hi @apiesse, thanks for reporting this. It's interesting to hear that having a directory with space in its path worked for you in older versions of ESP-IDF. This is explicitly not supported and there were several places in the build system where spaces were not handled correctly.
https://docs.espressif.com/projects/esp-idf/en/v4.4/esp32/get-started/index.html#step-1-install-prerequisites

We have fixed these issues in `master` recently and the future v5.0 release will support spaces in paths.

Could you please check the following:
* when you used v4.3.2, did ESP-IDF directory have path with spaces?
* did your project have path with spaces?

apiesse:
> 

Thanks for the reply @igrr .
That surprised me as well give the IDE Tools are installed in C:\espressif and the project is located in C:\svn\project\firmware\trunk\app (same with esp-idf-v4.3.2) as the documentation suggested.

igrr:
I see, looks like it might be a ccache issue. Are you building using `idf.py build` from command line or from an IDE?
If it's the former, could you do `set IDF_CCACHE_ENABLE=0` in the command prompt, then `idf.py fullclean` and finally `idf.py build` again.

apiesse:
[...]

PR #9656, opened 2022-08-27
upgrade ccache installed version to 4.6.2 (IDFGH-8160)

binary1230:
- primary reason: 4.6.2 fixes errors if the user's home directory contains a space
- alternative fix: the CCACHE_DIR env var can be set to something without a space in it, or set to TMP which properly escapes the home dir name
- there may be other issues with spaces in filenames on builds in other parts of ESP-IDF, see https://github.com/espressif/esp-idf/issues/8364 for more info.  Espressif is saying v5.0 is required for those other fixes, so I suggest only merging this with that version. I am testing with 4.4.2 and it works great over here

tested locally on windows only, use at your own risk

### DIFF
--- a/tools/tools.json
+++ b/tools/tools.json
@@ -947,7 +947,7 @@
       "description": "Ccache (compiler cache)",
       "export_paths": [
         [
-          "ccache-4.3-windows-64"
+          "ccache-4.6.2-windows-x86_64"
         ]
       ],
       "export_vars": {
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

DNedic:
There should be no issues with this, i've been using ccache 4.6.1 on Arch Linux for a while now and we have upped the CMake version for windows so it satisfies the minimum requirement as well.

Let's merge this.

dobairoland:
sha=823ae5cf4550f6455ed761f496616361980fa4b1

PR #9667, opened 2022-08-29
reinit provisioning manager fixed (IDFGH-8171)

Pofke:
after wifi_prov_mgr_deinit() calling wifi_prov_mgr_init() function throws error. Not initializing prov_ctx_lock second time. Fixed it by writing prov_ctx_lock = NULL where it should be deleted.


Alvin1Zhang:
Thanks for your contribution.

IshaESP:
We have already lined up the fix internally. Rebasing the master branch in after it merges will work.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/b8fc63d6582d9306c794a7faa36f1b038471fcad.

Issue #7446, opened 2021-08-20
Stop TX when 485 collision detects (IDFGH-5726)


Dandjinh:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when 485 collision detects, it only clears the rx fifo while tx is still working.
Now it's only possible to know the collision after TX is done because no events will be posted and uart_wait_tx_done() doesn't handle such issue.
Also there is no api to stop TX according to _uart.h_.
When the packet is large it's waste of time.

**Describe the solution you'd like**

When 485 collision detects, stops TX in the ISR handler and posts an event.

**Describe alternatives you've considered**

1. clear txfifo in uart_rx_intr_handler_default() when collision detects.
2. add an event UART_485_BUS_COLLISION and post it when collision detects.
3. add an api that stops UART TX.

alisitsyn:
@Dandjinh,

Thank you for your issue. The UART driver currently supports collision detection using hardware feature and corresponded flag `UART_INTR_RS485_CLASH`. When the transmission is complete and collision is detected UART prepares for receiving data by flushing the receive buffer. This approach is ok for J1708 networks while the `Message Identification character (MID)` is used. The user application can get status of collision `uart_get_collision_flag()` and start collision-retry procedure. I agree the API are very limited now and allows checking if the collision as described above.
Your proposals will be taken into account and it is planned to check and update collision support feature.
For now if you need customized handling of collision detection in your project I would propose you to realize it using customized UART interrupt handler and UART HAL layer.

Alvin1Zhang:
Thanks for reporting, seems this duplicates https://github.com/espressif/esp-idf/issues/3569, will close this issue. Let us track progress in https://github.com/espressif/esp-idf/issues/3569. Thanks.

alisitsyn:
@Dandjinh, @AshUK,

The example which uses HAL layer and adopted RS485 software driver to detect collisions in RS485 bus based networks is located [here](https://github.com/alisitsyn/modbus_support/tree/feature/driver_collision_detect/rs485_collision_detector) .

![collisions_with_modbus_master](https://user-images.githubusercontent.com/37483886/188855396-a44dd095-e5f9-4af2-9325-ad80e7e4b90b.png)

[collision_detection_log.txt](https://github.com/espressif/esp-idf/files/9505293/collision_detection_log.txt)

It is not free of some minor issues and still in WIP state but can describe what configuration of internal UART hardware can be used to detect collisions in RS485 networks.
[...]

PR #9695, opened 2022-09-01
Enable missing Interrupts for half duplex mode (IDFGH-8202)

plasger-ebv:
The mode UART_MODE_RS485_HALF_DUPLEX is capable of collision detection (given the required circuit), but the interrupts were not enabled. (See e.g. at function `uart_get_collision_flag()` about required modes for detection.

### DIFF
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -1690,7 +1690,7 @@ esp_err_t uart_set_mode(uart_port_t uart_num, uart_mode_t mode)
     }
     UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
     uart_hal_set_mode(&(uart_context[uart_num].hal), mode);
-    if (mode ==  UART_MODE_RS485_COLLISION_DETECT) {
+    if ((mode == UART_MODE_RS485_COLLISION_DETECT) || (mode == UART_MODE_RS485_HALF_DUPLEX)) {
         // This mode allows read while transmitting that allows collision detection
         p_uart_obj[uart_num]->coll_det_flg = false;
         // Enable collision detection interrupts


alisitsyn:
Hello @plasger-ebv ,

Thank you for the issue. You are right that the collision detection can be very useful and would be able to work when the [variant A RS485 connection ](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html#circuit-a-collision-detection-circuit) is used. However it is not possible with the ESP32 in  `UART_MODE_RS485_HALF_DUPLEX` mode. This is issue related to internal HW and will prevent to detect collisions correctly. So, in current driver your changes will not allow to detect the collisions correctly and the `UART_INTR_RS485_CLASH` will be triggered in spite of the correct signal on the RO pin. This also can cause additional side affects with UART receiver. This is why the collision detection interrupts are not enabled in the driver for `UART_MODE_RS485_HALF_DUPLEX` mode and the [Variant B of RS485 schematic](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html#circuit-b-manual-switching-transmitter-receiver-without-collision-detection) is recommended.

plasger-ebv:
Hi @alisitsyn ,

Thank you for the hints. I would like to use the Variant A circuit for RS485 with collision detection and so I would like to implement said workaround. Can you give me some hints what the problem with the internal HW is why the UART_INTR_RS485_CLASH interrupt is triggered at the wrong time?

Best regards

alisitsyn:
Hi @plasger-ebv ,

As per my information:
1. Due to hardware problems, the collision detection function of half-duplex mode (rs485_conf.tx_rx_en = 0) does not work (in fact, it can be used to detect collisions, but the RX idle level is required to be low when transmitting data).

2. The collision detection function of full-duplex mode (rs485_conf.tx_rx_en = 1) can work. If the data sent (on TX line) is different from the received data (on the RX line), a collision interrupt will occur. 

Let me know if you have any questions. I will try to check this again later.

alisitsyn:
The related issue:  https://github.com/espressif/esp-idf/issues/7446

alisitsyn:
This issue will be closed for now. The details of implementation can be found [here](https://github.com/alisitsyn/modbus_support/tree/mb_support/add_uart_isr_hal_example).
Feel free to reopen issue if there are any comments.

PR #9700, opened 2022-09-02
fix build error in uart driver (IDFGH-8207)

ValterMinute:
when building on Windows the compiler reports an error about
the sclk_freq variable being potentially uninitialized.
Setting the var to 0 keeps compiler happy.
I think this isn't an actual issue since variable is passed by
reference to uart_get_sclk_freq and this sets its value or returns
an error that should be trapped by the assert, at least on debug
builds.

### DIFF
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -300,7 +300,7 @@ esp_err_t uart_set_baudrate(uart_port_t uart_num, uint32_t baud_rate)
     ESP_RETURN_ON_FALSE((uart_num < UART_NUM_MAX), ESP_FAIL, UART_TAG, "uart_num error");
 
     uart_sclk_t src_clk;
-    uint32_t sclk_freq;
+    uint32_t sclk_freq=0;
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

ESP-Marius:
@ValterMinute Thanks for the fix, but I think this issue has already been fixed on master now (b844df8c852e2b147279e0dcd185f80247288741).

 Could you help double check by pulling the latest master and see if the error still occurs?

ValterMinute:
@ESP-Marius you are right, I am sorry for the useless MR.
I did the last pull before your commit was merged into master and forgot to check again before opening the MR.

ESP-Marius:
@ValterMinute Not useless, it was just bad luck with the timing.

Thanks for contributing anyway, nothing to be sorry about!

ESP-Marius:
Closing this one as the issue has been resolved.

PR #9728, opened 2022-09-06
[provisioning] Allow multiple WiFi scan (IDFGH-8240)

KonssnoK:
As per title,
this change allow clients to scan multiple times during provisioning, either to test different networks or because one network goes offline or whatever.

It cannot be verified with the current wifi provisioning app because after a successful provision the app will go on.

This is instead built to be used in a maintenance menu.

In any case, the code doesn't change behavior of the original provisioning flow.

image

### DIFF
--- a/components/wifi_provisioning/src/manager.c
+++ b/components/wifi_provisioning/src/manager.c
@@ -907,6 +907,8 @@ static void wifi_prov_mgr_event_handler_internal(
 esp_err_t wifi_prov_mgr_wifi_scan_start(bool blocking, bool passive,
                                         uint8_t group_channels, uint32_t period_ms)
 {
+    esp_err_t err;
+
     if (!prov_ctx_lock) {
         ESP_LOGE(TAG, "Provisioning manager not initialized");
         return ESP_ERR_INVALID_STATE;
@@ -923,6 +925,14 @@ esp_err_t wifi_prov_mgr_wifi_scan_start(bool blocking, bool passive,
         ESP_LOGD(TAG, "Scan already running");
         RELEASE_LOCK(prov_ctx_lock);
         return ESP_OK;
+    }
+
+    // Ensure we are disconnected
+    err = esp_wifi_disconnect();
+    if (err) {
+        ESP_LOGE(TAG, "Failed to disconnect %d", err);
+        RELEASE_LOCK(prov_ctx_lock);
+        return ESP_FAIL;
     }
 
     /* Clear sorted list for new entries */
@@ -961,6 +971,7 @@ esp_err_t wifi_prov_mgr_wifi_scan_start(bool blocking, bool passive,
     }
 
     ESP_LOGD(TAG, "Scan started");
+    prov_ctx->prov_state = WIFI_PROV_STATE_STARTED;
     prov_ctx->scanning = true;
     prov_ctx->curr_channel = prov_ctx->scan_cfg.channel;
     RELEASE_LOCK(prov_ctx_lock);


laukik-hase:
Hello, @KonssnoK!

Thank you for your contribution. I had a few questions regarding the pull request.

1. What does the PR intend to achieve? Is it to check whether the Android/iOS device is able to connect to different networks shown in the scan results?
2. Is there a way we can verify the PR using the [`esp_prov`](https://github.com/espressif/esp-idf/tree/master/tools/esp_prov) tool?

KonssnoK:
> Hello, @KonssnoK!
> 
> Thank you for your contribution. I had a few questions regarding the pull request.
> 
>     1. What does the PR intend to achieve? Is it to check whether the Android/iOS device is able to connect to different networks shown in the scan results? (Example flow: Scan → Provision with SSID 1 → Success/Failure → Scan and repeat again with a different SSID)
> 
>     2. Is there a way we can verify the PR using the [`esp_prov`](https://github.com/espressif/esp-idf/tree/master/tools/esp_prov) tool?

Hi @laukik-hase 
1. it basically decouples the wifi scan functionality from the provisioning flow. Which means you can do what you put in your example (Connect to SSID1, do something, connect to SSID2, all in the same provisioning session)
2. I don't know esp_prov tool, i use a modified version coming from https://github.com/espressif/esp-idf-provisioning-android
    I see the ESPDevice java class is probably a copy of the python class you have in esp_prov. I didn't have to do any change to ESPDevice in order to make my change work.

components/wifi_provisioning/src/manager.c
```diff
@@ -961,6 +971,7 @@ esp_err_t wifi_prov_mgr_wifi_scan_start(bool blocking, bool passive,
     }
 
     ESP_LOGD(TAG, "Scan started");
+    prov_ctx->prov_state = WIFI_PROV_STATE_STARTED;
```
mahavirj, line 974: I would not recommend changing the provisioning SM state as part of scan request here. We can dedicated handler for this, which can be invoked from provisioning entity side. We will discuss this internally and provide update here.

KonssnoK: well actually this line is a bugfix, otherwise you are able to enter the same function a second time but nobody will change the state of the provisioning to "started".
This will cause the "Scan started" print to be shown when the provisioning SM is actually in a different state.

Harshal5:
Hello, @KonssnoK !

We now support reprovisioning (`ctrl_reprov`) command under a new provisioning endpoint `wifi_ctrl`. [Reference](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/provisioning/wifi_provisioning.html#id12)
The endpoint can be used in case you want to re-provision the device after a previous successful provisioning.
Could you try using this once?

KonssnoK:
Hi @Harshal5 , we work on v4.4, so that change is not available to us. Do you plan on backporting f1dcb1967cca4fdb0b40136781d29b1c71539589 from master to 4.4?

Harshal5:
We don't plan to backport this feature into `release/v4.4`. You can try cherry-picking the following commit [9bc1cc](https://github.com/espressif/esp-idf/commit/9bc1cc706e394466dcd202b6c4a4586d8742bab2) for adding support of re-provisioning.

Thanks!

Harshal5:
We won't be backporting this feature in v4.4. I tried cherry-picking some commits over v4.4 and sharing a patch with you, but that will be lots of changes and refactoring of too many commits. For example, `esp_prov` earlier used BlueZ and now we have moved to Bleak.
You can try the feature using `release/v5.0` by cherry-picking two commits using the following commands:
1. `git cherry-pick 9bc1cc7` 
2. `git cherry-pick f1dcb19 -m 1`

Thanks!

KonssnoK:
ok, so we'll continue using our own patch until we'll move to 5.0.
Thanks.

PR #9749, opened 2022-09-08
Fix export.sh failure if python 2.x is unavailable (IDFGH-8261)

gormster:
In the latest release of macOS (and probable some other recent *nixes as well?) `python` is no longer available by default, only `python3`. This causes `export.sh` to fail as it still had a reference to plain `python`. This now works as expected.

### DIFF
--- a/export.fish
+++ b/export.fish
@@ -47,7 +47,7 @@ function __main
     set -x PATH "$IDF_ADD_PATHS_EXTRAS":"$PATH"
 
     echo "Checking if Python packages are up to date..."
-    python "$IDF_PATH"/tools/idf_tools.py check-python-dependencies || return 1
+    "$ESP_PYTHON" "$IDF_PATH"/tools/idf_tools.py check-python-dependencies || return 1
 
     set added_path_variables
     for entry in $PATH;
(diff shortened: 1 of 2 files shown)


dobairoland:
Hi @gormster. The change looks good to me. Thanks! Could you please apply the same patch to `export.fish` as well? Please apply it to the same commit (`git commit --amend`).

gormster:
@dobairoland ah yes, of course! can't forget the fish people. you can't amend commits once they've been pushed, though - I'll add a new commit to the same branch.

dobairoland:
@gormster Commit can be amended even after that. One has to do forced push after. Please try to squash the commits into one so we could accept your commit as is and merge this. We prefer to not have multiple commits for simple things belonging together. Thanks for understanding!

You can do this by interactive rebase, i.e. `git rebase -i origin/master` and by marking the second commit for squashing or fixup. You will also need to do `git push --force` after that.

dobairoland:
I'm squashing the commits by myself internally.

dobairoland:
20487e7a06135db993886129d6316f973f82f7b8

PR #9803, opened 2022-09-17
Improve Udev Rule procedure (IDFGH-8326)

ajsb85:
Dearest.

- Some product IDs are 1001 and not ??00.
- Running just `sudo udevadm trigger` is not enough to apply the rules.

Cheers.

### DIFF
--- a/docs/en/api-guides/dfu.rst
+++ b/docs/en/api-guides/dfu.rst
@@ -97,12 +97,43 @@ Udev is a device manager for the Linux kernel. It allows us to run ``dfu-util`` 
 
 Create file ``/etc/udev/rules.d/40-dfuse.rules`` with the following content::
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

dobairoland:
> Some product IDs are 1001 and not ??00.

Thanks @ajsb85 for pointing this out. At the time of writing this was true. I'll ask a colleague to check this who is more familiar with this.

As regarding the update of the reloading the rules, I would really advise against it. The original description is intentionally very short. I like your improvement but these kind of descriptions are hard to maintain. Others might want to add here instructions for their favorite distribution and it is not practical try to keep here descriptions for all maintained distributions (and versions). Instead one should find the manual in the distribution's documentation.

PR #9805, opened 2022-09-18
Speed up FATFS aligned read & writes by merging adjacent read/write calls (IDFGH-8328)

bloodmage:
A simple approach to merge sequential disk_reads and disk_writes Will vastly accelerate aligned file syscall (read/write)
On SD card, when you use buffer larger than 4kb (I use 16kb), it will improve read/write speed by n-fold (such as 4 fold).

A little catch: data may lost when dirty-data flushing failed, or any ABORT is called. This behavior is inconsistent with unpatched one.
If this is important, one can only use merged reads, leaving writes unmerged.

### DIFF
--- a/components/fatfs/src/ff.c
+++ b/components/fatfs/src/ff.c
@@ -3843,6 +3843,45 @@ FRESULT f_open (
 
 
 
+struct last_read_op_t {
+    BYTE valid;
+    BYTE pdrv;
+    BYTE* buff;
+    LBA_t sector;
+    UINT ss;
+    UINT count;
+};
+
+DRESULT disk_read_flush(struct last_read_op_t* last_read_op) {
+    if (last_read_op->valid==0) return RES_OK;
+    DRESULT ret = disk_read(last_read_op->pdrv, last_read_op->buff, last_read_op->sector, last_read_op->count);
+    last_read_op->valid = 0;
+    return ret;
+}
+DRESULT disk_read_joined(struct last_read_op_t* last_read_op, BYTE pdrv, BYTE* buff, LBA_t sector, UINT ss, UINT count) {
+    DRESULT ret = RES_OK;
+    if (last_read_op->valid==0) {
+initop:
+        last_read_op->valid = 1;
+        last_read_op->pdrv = pdrv;
+        last_read_op->buff = buff;
+        last_read_op->sector = sector;
+        last_read_op->ss = ss;
+        last_read_op->count = count;
+        return ret;
+    }
+    //Try to join read op
+    if (last_read_op->sector + last_read_op->count == sector && last_read_op->buff + (last_read_op->count*ss) == buff) {
+        //Sector id is continuous and buffer is continuous
+        last_read_op->count += count;
+        return RES_OK;
+    } else {
+        //Segment, flush and replace
+        ret = disk_read_flush(last_read_op);
+        goto initop;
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

igrr:
Hi @bloodmage,
As far as I understand, FATFS already supports or reading writing multiple sectors at a time. For example, if you allocate a large enough buffer in your application and [call `write`](https://github.com/bloodmage/esp32-remoteota-captive-softap-sdcard-httpd/blob/e00c69aa215557101b5b0eff8919a66cf1ab7719/main/sd_webserver.cpp#L60) once that buffer is full, FATFS will write it in one go. Alternatively, if you don't want to implement buffering yourself, you can use the C standard library APIs `fopen`, `fread`, `fwrite`. After opening the file, call `setvbuf` function to set a larger buffer for the file stream.

I would suggest solving the problem this way, rather than by patching FATFS library. We would prefer to keep the source code of FATFS as close to upstream version as possible, to facilitate future upgrades to the new version. If we diverge from the upstream version now, someone will have to figure out how to apply these changes to the new upstream release each time we upgrade FATFS.

igrr:
> FATFS would do block reading / writing sectorwise. It will split read/write commands by sectors regardless the size of buffer.

May I ask how you have tested this? I'm reaching the opposite conclusion.

Here's the test you can perform:
```c++
    FILE* f = fopen(name, "wb");
    const size_t write_size = 8192;
    char* buf = calloc(write_size, 1);
    TEST_ASSERT_NOT_NULL(buf);

    ESP_LOGI(__func__, "writing %d bytes to file", write_size);
    ssize_t res = write(fileno(f), buf, write_size);
    TEST_ASSERT_EQUAL(write_size, res);
    ESP_LOGI(__func__, "write done");

    free(buf);
    TEST_ASSERT_EQUAL(0, fclose(f));
```

and add a log [here](https://github.com/espressif/esp-idf/blob/29830f609d96ed2cfb0639908e0a24ae40d2936c/components/fatfs/diskio/diskio_sdmmc.c#L65-L66):
```c++
    ESP_LOGI(TAG, "%s: sector=%d count=%d", __func__, sector, count);
```

which results in:
```c++
I (3881) test_fatfs_large_write: writing 8192 bytes to file
I (3883) diskio_sdmmc: ff_sdmmc_write: sector=564 count=16
I (3890) test_fatfs_large_write: write done
```

So 16 sectors (8192 bytes) have been written in one call to `sdmmc_write_sectors`.

bloodmage:
There are 2 sector size. One is the physical sector of sdcard or other medium. The other is the logic sector size of FAT32 or other format (which is call "Allocation Unit Size" during formatting). FATFS r/w operation handles each logic sector separately, this behavior is in ff.c line 3896-3901. 

So maybe your SDCard has formatted using sector size of 8192. The sdcard I test uses 4096.

Try to enlarge read buf to 16k or reformat sdcard using smaller Allocation Unit Size. You will observe the difference.

igrr:
For cluster size (not sector size) you are right, of course. In this particular case my card was formatted with 16kB cluster size.

Coalescing writes of multiple clusters indeed is a nice idea, please do submit it as an upstream patch! We will upgrade to the new fatfs version once it is released.

bloodmage:
😊

Issue #9753, opened 2022-09-08
BLE Mesh: Dynamic Configuration of node property (IDFGH-8266)


Shane555:
### Is your feature request related to a problem?

Currently, the node properties like LPN, friend, relay cannot be set at run time. We have to go into menuconfig and enable or disable these functions. I would like the capability to do so at run time e.g. change this relay node to a low power node while the program is running. I noticed there is a function `esp_err_t esp_ble_mesh_config_client_set_state(esp_ble_mesh_client_common_param_t *params, esp_ble_mesh_cfg_client_set_state_t *set_state)` that could use a configuration client model to modify the configuration server parameters, which seems like it might be possible to do so at run time. However, `esp_ble_mesh_cfg_client_set_state_t` only contains `esp_ble_mesh_cfg_relay_set_t` and `esp_ble_mesh_cfg_friend_set_t` but does not have something for the LPN equivalent.

 Another thing is i am not sure if simply setting say the esp_ble_mesh_cfg_friend_set_t to TRUE would allow the node to be a friend node instantly. Do clarify regarding this issue   

### Describe the solution you'd like.

I would like to be able to flash the firmware once and have the node be able to do friendship, LPN or relay as and when i would want to without reflashing.

Shane555:
Regarding the APIs provided for a configuration client model node to send a message to configure another configuration server model node, it seems like there are things missing.  For the client side to send a message i will need to use `sp_err_t esp_ble_mesh_config_client_set_state(esp_ble_mesh_client_common_param_t *params,
                                               esp_ble_mesh_cfg_client_set_state_t *set_state)`

Shane555:
I am not sure if you can simply modify `esp_ble_mesh_cfg_server_state_change_t` to account for the friend and relay component but i believe it should work without a hiccup 
```cpp
typedef union {
    /**
     * The recv_op in ctx can be used to decide which state is changed.
     */
    esp_ble_mesh_state_change_cfg_mod_pub_set_t         mod_pub_set;        /*!< Config Model Publication Set */
    esp_ble_mesh_state_change_cfg_model_sub_add_t       mod_sub_add;        /*!< Config Model Subscription Add */
    esp_ble_mesh_state_change_cfg_model_sub_delete_t    mod_sub_delete;     /*!< Config Model Subscription Delete */
    esp_ble_mesh_state_change_cfg_netkey_add_t          netkey_add;         /*!< Config NetKey Add */
[...]

PR #9811, opened 2022-09-19
BLE_MESH: Added missing state changes event in cfg srv model

Shane555:
Added missing events on cfg server model side to match existing events in cfg client model side


Shane555:
@JiaLWang As discussed in https://github.com/espressif/esp-idf/issues/9753

Alvin1Zhang:
Thanks for your contribution.

JiaLWang:
Hi @Shane555
Thanks for your contribution.
If you want to add state change event, not only need to add structs at the application layer, but also need to add related structs at the BTC layer. and the use of callback functions. You can refer to the existing events.
Contact me if you have any difficulties.

Shane555:
> Hi @Shane555 Thanks for your contribution. If you want to add state change event, not only need to add structs at the application layer, but also need to add related structs at the BTC layer. and the use of callback functions. You can refer to the existing events. Contact me if you have any difficulties.

Hi @JiaLWang, I am not quite sure what needs to be changed. As you can see below, esp_ble_mesh_config_client_set_state has already accounted for the relevant bluetooth layer msg to an even lower layer.
```
esp_err_t esp_ble_mesh_config_client_set_state(esp_ble_mesh_client_common_param_t *params,
                                               esp_ble_mesh_cfg_client_set_state_t *set_state)
{
    btc_ble_mesh_config_client_args_t arg = {0};
    btc_msg_t msg = {0};

    if (params == NULL || params->model == NULL ||
        params->ctx.net_idx == ESP_BLE_MESH_KEY_UNUSED ||
        !ESP_BLE_MESH_ADDR_IS_UNICAST(params->ctx.addr) ||
        (params->opcode != ESP_BLE_MESH_MODEL_OP_NODE_RESET && set_state == NULL)) {
        return ESP_ERR_INVALID_ARG;
    }

    ESP_BLE_HOST_STATUS_CHECK(ESP_BLE_HOST_STATUS_ENABLED);

    msg.sig = BTC_SIG_API_CALL;
    msg.pid = BTC_PID_CONFIG_CLIENT;
    msg.act = BTC_BLE_MESH_ACT_CONFIG_CLIENT_SET_STATE;
    arg.cfg_client_set_state.params = params;
    arg.cfg_client_set_state.set_state = set_state;

    return (btc_transfer_context(&msg, &arg, sizeof(btc_ble_mesh_config_client_args_t), btc_ble_mesh_config_client_arg_deep_copy)
            == BT_STATUS_SUCCESS ? ESP_OK : ESP_FAIL);
}
```
The configuration client code for setting states has already been accounted for [here](https://github.com/espressif/esp-idf/blob/29830f609d96ed2cfb0639908e0a24ae40d2936c/components/bt/esp_ble_mesh/api/models/include/esp_ble_mesh_config_model_api.h#L381). All i am doing is ensuring the callback event on server side has more options to be handled as it has already been integrated in the client side. Not quite sure what you mean exactly, would appreciate if you could clarify, thanks! :)

JiaLWang:
Hi @Shane555 
There is nothing wrong with what you're currently adding, it's just missing something. 
Take exist cfg_appkey_update as an example, when the server receive the config appkey update message and update the appkey successfully, it will[ throw state change ](https://github.com/espressif/esp-idf/blob/master/components/bt/esp_ble_mesh/mesh_core/cfg_srv.c#L520)events to the BTC layer. The BTC  task will handle the message and throw the message to the APP layer. The APP layer using the structs that defined to store the parameters.

But for cfg_friend_set, when the server receive the config friend set message, it [doesn't throw](https://github.com/espressif/esp-idf/blob/master/components/bt/esp_ble_mesh/mesh_core/cfg_srv.c#L2809) state change events to BTC layer. Therefore, even if the APP layer defines the relevant structures, the underlying parameters cannot be obtained. (The [structs of BTC layer](https://github.com/espressif/esp-idf/blob/master/components/bt/esp_ble_mesh/mesh_core/include/cfg_srv.h#L213) has defined.)
So we also need to add some `bt_mesh_config_server_cb_evt_to_btc`  in the cfg_srv.c.

Shane555:
@JiaLWang i have added `bt_mesh_config_server_cb_evt_to_btc`, take a look

Shane555:
Hi @JiaLWang  any news on this? I am hoping either you would update the issue or if i am doing it in the right direction, you could merge the code because i think its quite important to add the missing events to allow users to handle the control of application in the application layer.

JiaLWang:
Hi @Shane555 ,

Sorry for the late reply.

We will first merge it in the internal repository. Then it will be periodically synchronized to the github repository.

Thanks for your contribution.

PR #9813, opened 2022-09-19
Fix wrong error reporting when https server tries to malloc(0) (IDFGH-8338)

axos88:
Fix issue of reporting allocation failure when length is zero

### DIFF
--- a/components/esp_https_server/src/https_server.c
+++ b/components/esp_https_server/src/https_server.c
@@ -217,7 +217,7 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
     ssl_ctx->user_cb = config->user_cb;
 
 /* cacert = CA which signs client cert, or client cert itself */
-    if(config->cacert_pem != NULL) {
+    if(config->cacert_pem != NULL && config->cacert_len > 0) {
         cfg->cacert_buf = (unsigned char *)malloc(config->cacert_len);
         if (!cfg->cacert_buf) {
             ESP_LOGE(TAG, "Could not allocate memory");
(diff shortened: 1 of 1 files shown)


components/esp_https_server/src/https_server.c
```diff
@@ -217,7 +217,7 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
     ssl_ctx->user_cb = config->user_cb;
 
 /* cacert = CA which signs client cert, or client cert itself */
-    if(config->cacert_pem != NULL) {
+    if(config->cacert_pem != NULL && config->cacert_len > 0) {
```
hmalpani, line 220:
```suggestion
    if(config->cacert_pem != NULL) {
        if (config->cacert_len <= 0) {
            ESP_LOGE(TAG, "Invalid certificate length");
            free(cfg);
            free(ssl_ctx);
            return NULL;
        }
```

components/esp_https_server/src/https_server.c
```diff
@@ -231,7 +231,7 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
 
 /* servercert = cert of server itself */
     cfg->servercert_buf = (unsigned char *)malloc(config->servercert_len);
-    if (!cfg->servercert_buf) {
+    if (!cfg->servercert_buf && config->servercert_len > 0) {
```
hmalpani, line 234: Same as the above suggestion.

hmalpani (commented):
Hello @axos88,
Thanks for the PR. Left a few comments otherwise LGTM.

Also please sign the CLA.

axos88:
@hmalpani , please see https://github.com/espressif/esp-idf/pull/9833, which includes this change as well. I think this should be closed in favour of that one.

hmalpani:
> @hmalpani , please see #9833, which includes this change as well. I think this should be closed in favour of that one.

Sure. I will close this PR. Thanks!

PR #9833, opened 2022-09-22
add sni support (IDFGH-8363)

axos88:
This PR adds full server side SNI support.

- Server side SNI is only supported by mbedTLS.
- A new KConfig value has been introduced, it depends on mbedTLS being selected
- the sni callback and the supplied data can be used to select a certificate based on the incoming SNI header containing the host the client is looking for.
- pkey, cert, client verification ca bundle, authmode can also be selected: https://mbed-tls.readthedocs.io/en/latest/kb/how-to/use-sni/#server-side
- a TLS server can now be initialized WITHOUT a pkey and a certificate as long as SNI is enabled, and an SNI callback is set up. A warning will be issued that handshakes will fail unless the SNI callback sets up the pkey and cert to use for each handshake.
- Using this mechanism the server certificate and private key can be changed without restarting the server, allowing the certificate to be updated without reboots (certificate renewal for example)

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -594,6 +594,24 @@ esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
 #endif
 }
 
+#ifdef CONFIG_ESP_TLS_SERVER
+esp_err_t esp_tls_cfg_server_sni_init(esp_tls_cfg_server_t *cfg, esp_tls_server_sni_callback *cb, void *data)
+{
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+    if (!cfg) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    cfg->sni_callback = cb;
+    cfg->sni_callback_p_info = data;
+
+    return ESP_OK;
+#else
+    return ESP_ERR_NOT_SUPPORTED;
+#endif
+}
+#endif
+
+#ifdef CONFIG_ESP_TLS_SERVER
 void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
 {
 #if defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
@@ -602,6 +620,7 @@ void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
     }
 #endif
 }
+#endif
 
 /**
  * @brief      Create a server side TLS/SSL connection
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -594,24 +594,6 @@ esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
 #endif
 }
 
-#ifdef CONFIG_ESP_TLS_SERVER
-esp_err_t esp_tls_cfg_server_sni_init(esp_tls_cfg_server_t *cfg, esp_tls_server_sni_callback *cb, void *data)
-{
-#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
-    if (!cfg) {
-        return ESP_ERR_INVALID_ARG;
-    }
-    cfg->sni_callback = cb;
-    cfg->sni_callback_p_info = data;
-
-    return ESP_OK;
-#else
-    return ESP_ERR_NOT_SUPPORTED;
-#endif
-}
-#endif
-
-#ifdef CONFIG_ESP_TLS_SERVER
 void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
 {
 #if defined(CONFIG_ESP_TLS_SERVER_SESSION_TICKETS)
@@ -620,7 +602,6 @@ void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
     }
 #endif
 }
-#endif
 
 /**
  * @brief      Create a server side TLS/SSL connection
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -297,25 +297,6 @@ typedef struct esp_tls_cfg_server {
 esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg);
 
 /**
- * @brief Initialize the server side TLS server name identification hook
- *
- * This function initializes the server side tls server name identification hook
- * which is called during a client connect handshake, and has the ability to check
- * the advertised hostname and supply a specific certificate based on that for example.
- * Check https://mbed-tls.readthedocs.io/en/latest/kb/how-to/use-sni/#server-side for details.
- *
- * @param[in]  cfg server configuration as esp_tls_cfg_server_t
- * @param[in]  cb callback function pointer
- * @param[in]  data pointer to implementation specific data supplied as the first parameter of the callback
- * @return
- *             ESP_OK if setup succeeded
- *             ESP_ERR_INVALID_ARG if context is NULL
- *             ESP_ERR_NOT_SUPPORTED if server side sni is not available due to build configuration
- *             ESP_FAIL if setup failed
- */
-esp_err_t esp_tls_cfg_server_sni_init(esp_tls_cfg_server_t *cfg, esp_tls_server_sni_callback *cb, void *data);
-
-/**
  * @brief Free the server side TLS session ticket context
  *
  * @param cfg server configuration as esp_tls_cfg_server_t
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -197,6 +197,18 @@ typedef struct esp_tls_server_session_ticket_ctx {
 } esp_tls_server_session_ticket_ctx_t;
 #endif
 
+/**
+ * @brief SNI callback function prototype
+ * Can be used to configure per-handshake attributes for the TLS connection.
+ * E.g. Client certificate / Key, Authmode, Client CA verification
+ *
+ * @param p_info Input data provided when registering the callback
+ * @param ssl mbedtls_ssl_context that can be used for changing settings
+ * @param name advertised server name by the client
+ * @param len length of the name buffer
+ */
+typedef int esp_tls_server_sni_callback(void *p_info, mbedtls_ssl_context *ssl, const unsigned char *name, size_t name_len);
+
 typedef struct esp_tls_cfg_server {
     const char **alpn_protos;                   /*!< Application protocols required for HTTP2.
                                                      If HTTP2/ALPN support is required, a list
@@ -259,6 +271,11 @@ typedef struct esp_tls_cfg_server {
                                                     Call esp_tls_cfg_server_session_tickets_free
                                                     to free the data associated with this context. */
 #endif
+
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+    esp_tls_server_sni_callback *sni_callback; /*!< Server Name Identification callback to use */
+    void *sni_callback_p_info;                 /*!< Data to pass to the SNI callback. */
+#endif
 } esp_tls_cfg_server_t;
 
 /**
@@ -278,6 +295,25 @@ typedef struct esp_tls_cfg_server {
  *             ESP_FAIL if setup failed
  */
 esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg);
+
+/**
+ * @brief Initialize the server side TLS server name identification hook
+ *
+ * This function initializes the server side tls server name identification hook
+ * which is called during a client connect handshake, and has the ability to check
+ * the advertised hostname and supply a specific certificate based on that for example.
+ * Check https://mbed-tls.readthedocs.io/en/latest/kb/how-to/use-sni/#server-side for details.
+ *
+ * @param[in]  cfg server configuration as esp_tls_cfg_server_t
+ * @param[in]  cb callback function pointer
+ * @param[in]  data pointer to implementation specific data supplied as the first parameter of the callback
+ * @return
+ *             ESP_OK if setup succeeded
+ *             ESP_ERR_INVALID_ARG if context is NULL
+ *             ESP_ERR_NOT_SUPPORTED if server side sni is not available due to build configuration
+ *             ESP_FAIL if setup failed
+ */
+esp_err_t esp_tls_cfg_server_sni_init(esp_tls_cfg_server_t *cfg, esp_tls_server_sni_callback *cb, void *data);
 
 /**
  * @brief Free the server side TLS session ticket context
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -197,17 +197,19 @@ typedef struct esp_tls_server_session_ticket_ctx {
 } esp_tls_server_session_ticket_ctx_t;
 #endif
 
+
+#if defined(CONFIG_ESP_TLS_SERVER_CERT_SELECT_HOOK)
 /**
- * @brief SNI callback function prototype
+ * @brief tls handshake callback
  * Can be used to configure per-handshake attributes for the TLS connection.
- * E.g. Client certificate / Key, Authmode, Client CA verification
+ * E.g. Client certificate / Key, Authmode, Client CA verification, etc.
  *
- * @param p_info Input data provided when registering the callback
  * @param ssl mbedtls_ssl_context that can be used for changing settings
- * @param name advertised server name by the client
- * @param len length of the name buffer
+ * @return The reutn value of the callback must be 0 if successful,
+ *         or a specific MBEDTLS_ERR_XXX code, which will cause the handhsake to abort
  */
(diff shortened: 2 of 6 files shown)


KaeLL:
I believe PRs are supposed to be made against the master branch.

axos88:
For some odd reason I cannot get current master to compile, so I can't port this to the current master, and there are merge conflicts, so I can't confidently resolve them.

AdityaHPatwardhan:
@axos88 Thanks for adding the support of the feature in ESP-IDF.
But as mentioned by @KaeLL, We only accept pull requests against the master branch. Please raise a PR against the master branch.

Just to note here, We are using `mbedtls-3.x`, Please make sure you have followed, [migration guide](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/3.0-migration-guide.md) for using mbedtls-3.x on the master branch.

axos88:
@AdityaHPatwardhan, I am using esp-rs/esp-idf-sys to build esp-idf, but I get the following error on master (works on 4.4 though):


```

running: "cmake" "/home/akos/projects/rust/esp-rs/esp-doorlock/target/riscv32imc-esp-espidf/release/build/esp-idf-sys-740a94ad1cf0d2c8/out" "-G" "Ninja" "-DCMAKE_TOOLCHAIN_FILE=/home/akos/projects/rust/esp-rs/esp-idf/tools/cmake/toolchain-esp32c3.cmake" "-DCMAKE_BUILD_TYPE=" "-DPYTHON=/home/akos/.espressif/python_env/idf5.1_py3.10_env/bin/python" "-DCMAKE_INSTALL_PREFIX=/home/akos/projects/rust/esp-rs/esp-doorlock/target/riscv32imc-esp-espidf/release/build/esp-idf-sys-740a94ad1cf0d2c8/out" "-DCMAKE_C_FLAGS= -march=rv32imc -ffunction-sections -fdata-sections -march=rv32imc -mabi=ilp32 -mcmodel=medany" "-DCMAKE_CXX_FLAGS= -march=rv32imc -ffunction-sections -fdata-sections -march=rv32imc -mabi=ilp32 -mcmodel=medany" "-DCMAKE_ASM_FLAGS=  -ffunction-sections -fdata-sections -march=rv32imc -mabi=ilp32 -mcmodel=medany"
  -- Checking Python dependencies...
  pkg_resources cannot be imported. The most common cause is a missing pip or setuptools package. If you've installed a custom Python then these packages are provided separately and have to be installed as well. Please refer to the Get Started section of the ESP-IDF Programming Guide for setting up the required packages.
  Skipping the download of /home/akos/projects/rust/esp-rs/esp-doorlock/.embuild/espressif/espidf.constraints.v5.1.txt because it was downloaded recently.
  Constraint file: /home/akos/projects/rust/esp-rs/esp-doorlock/.embuild/espressif/espidf.constraints.v5.1.txt
  Requirement files:
   - /home/akos/projects/rust/esp-rs/esp-idf/tools/requirements/requirements.core.txt
  Python being checked: /home/akos/projects/rust/esp-rs/esp-doorlock/.embuild/espressif/python_env/idf5.1_py3.10_env/bin/python
  -- Configuring incomplete, errors occurred!
  See also "/home/akos/projects/rust/esp-rs/esp-doorlock/target/riscv32imc-esp-espidf/release/build/esp-idf-sys-740a94ad1cf0d2c8/out/build/CMakeFiles/CMakeOutput.log".


 --- stderr
  Build configuration: BuildConfig {
      esp_idf_tools_install_dir: None,
      esp_idf_sdkconfig: None,
      esp_idf_sdkconfig_defaults: None,
      mcu: None,
      native: NativeConfig {
          esp_idf_version: Some(
              Tag(
                  "v5.1",
              ),
          ),
          esp_idf_repository: None,
          esp_idf_cmake_generator: None,
          idf_path: Some(
              "/home/akos/projects/rust/esp-rs/esp-idf",
          ),
          extra_components: [],
          esp_idf_components: None,
      },
      esp_idf_sys_root_crate: None,
  }
  Using activated esp-idf v5.1.0 environment at '/home/akos/projects/rust/esp-rs/esp-idf'
  CMake Error at /home/akos/projects/rust/esp-rs/esp-idf/tools/cmake/build.cmake:345 (message):
    Failed to run Python dependency check.  Python:
    /home/akos/.espressif/python_env/idf5.1_py3.10_env/bin/python, Error: 255
  Call Stack (most recent call first):
    /home/akos/projects/rust/esp-rs/esp-idf/tools/cmake/build.cmake:480 (__build_check_python)
    CMakeLists.txt:14 (idf_build_process)
```

```

**Comment by axos88** [comment]
fixed build issue, rebased onto master.

File: components/esp-tls/esp_tls.c [hunk 1]
```diff
@@ -594,6 +594,24 @@ esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
 #endif
 }
 
+#ifdef CONFIG_ESP_TLS_SERVER
```
AdityaHPatwardhan, line 597: This `#if` is not necessary. I think the code is already under the macro of the same name.

components/esp-tls/esp_tls.c
```diff
@@ -594,6 +594,24 @@ esp_err_t esp_tls_cfg_server_session_tickets_init(esp_tls_cfg_server_t *cfg)
...
+    cfg->sni_callback_p_info = data;
+
+    return ESP_OK;
+#else
+    return ESP_ERR_NOT_SUPPORTED;
+#endif
+}
+#endif
+
+#ifdef CONFIG_ESP_TLS_SERVER
```
AdityaHPatwardhan, line 614: Same here, not necessary!

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -566,7 +566,16 @@ esp_err_t set_server_config(esp_tls_cfg_server_t *cfg, esp_tls_t *tls)
             return esp_ret;
         }
     } else {
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+        if ((cfg->sni_callback == NULL)) {
```
AdityaHPatwardhan, line 570:
```suggestion
        if (cfg->sni_callback == NULL) {
```

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -787,6 +796,14 @@ int esp_mbedtls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp
         tls->conn_state = ESP_TLS_FAIL;
         return -1;
     }
+
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+    if (cfg->sni_callback != NULL) {
+        ESP_LOGI(TAG, "Initializing server side SNI callback");
+        mbedtls_ssl_conf_sni(&tls->conf, cfg->sni_callback, cfg->sni_callback_p_info);
+    }
+#endif
```
AdityaHPatwardhan, line 805: This part should come under `set_server_config` API  since this part also does server configuration.
I think we can keep this `conf_sni` part in the `else` condition added above.

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,100 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
     ssl_ctx->user_cb = config->user_cb;
 
 /* cacert = CA which signs client cert, or client cert itself */
-    if(config->cacert_pem != NULL) {
+    if(config->cacert_pem != NULL && config->cacert_len > 0) {
         cfg->cacert_buf = (unsigned char *)malloc(config->cacert_len);
-        if (!cfg->cacert_buf) {
-            ESP_LOGE(TAG, "Could not allocate memory");
+
+        if(cfg->cacert_buf) {
```
AdityaHPatwardhan, line 234:
```suggestion
        if (cfg->cacert_buf) {
```

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,100 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
+        };
+    }
+#endif
+
     ssl_ctx->tls_cfg = cfg;
     ssl_ctx->user_cb = config->user_cb;
 
 /* cacert = CA which signs client cert, or client cert itself */
-    if(config->cacert_pem != NULL) {
+    if(config->cacert_pem != NULL && config->cacert_len > 0) {
```
AdityaHPatwardhan, line 231:
```suggestion
    if (config->cacert_pem != NULL && config->cacert_len > 0) {
```

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,100 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
-        memcpy((char *)cfg->cacert_buf, config->cacert_pem, config->cacert_len);
-        cfg->cacert_bytes = config->cacert_len;
     }
 
 /* servercert = cert of server itself */
-    cfg->servercert_buf = (unsigned char *)malloc(config->servercert_len);
-    if (!cfg->servercert_buf) {
-        ESP_LOGE(TAG, "Could not allocate memory");
-        free((void *)cfg->cacert_buf);
+    if(config->servercert != NULL && config->servercert_len > 0) {
```
AdityaHPatwardhan, line 246:
```suggestion
    if (config->servercert != NULL && config->servercert_len > 0) {
```

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,100 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
+        cfg->servercert_buf = (unsigned char *)malloc(config->servercert_len);
+
+        if(cfg->servercert_buf) {
```
AdityaHPatwardhan, line 249:
```suggestion
        if (cfg->servercert_buf) {
```

components/esp-tls/esp_tls.h
```diff
@@ -279,6 +296,25 @@ typedef struct esp_tls_cfg_server {
...
+ * @param[in]  cfg server configuration as esp_tls_cfg_server_t
+ * @param[in]  cb callback function pointer
+ * @param[in]  data pointer to implementation specific data supplied as the first parameter of the callback
+ * @return
+ *             ESP_OK if setup succeeded
+ *             ESP_ERR_INVALID_ARG if context is NULL
+ *             ESP_ERR_NOT_SUPPORTED if server side sni is not available due to build configuration
+ *             ESP_FAIL if setup failed
+ */
+esp_err_t esp_tls_cfg_server_sni_init(esp_tls_cfg_server_t *cfg, esp_tls_server_sni_callback *cb, void *data);
```
AdityaHPatwardhan, line 316: What is the need of this init function ?
IIUC, this function only sets the function pointer and the pointer to the additional data in the server configuration structure.
Instead of using this API, the user can just set the resp. pointers in the cfg structure itself.

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,100 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
+#endif
+            ESP_LOGE(TAG, "No Server key supplied");
+            free((void *) cfg->servercert_buf);
+            free((void *) cfg->cacert_buf);
             free(cfg);
             free(ssl_ctx);
             return NULL;
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+            } else {
+            ESP_LOGW(TAG, "Server key not supplied, make sure to supply it in the SNI hook!");
```
AdityaHPatwardhan, line 304:
```suggestion
                ESP_LOGW(TAG, "Server key not supplied, make sure to supply it in the SNI hook!");
```

AdityaHPatwardhan, line 304: I think instead of adding multiple `#if def` in the code.
We could merge the code when `SNI hook` is enabled into one and then keep the other code in another part.
I think the current format is affecting code readability.
The same is applicable to the above case.

AdityaHPatwardhan (commented):
@axos88 Thanks for updating the changes.

I have reviewed them in detail. I have left some comments and suggestions.
I also wanted to request you, if possible could you keep a separate commit for the fix that you have added? I think we may have to backport it.

components/esp-tls/esp_tls.h
axos88: Yeah, that's what it does. Just though it was more elegant than meddling with the cfg structure itself at the call site. I can remove it though.

components/esp-tls/esp_tls_mbedtls.c
axos88: I'm not sure this is the case. The set_server_config is called when a new server side socket is created, while server_session_create is called each time a client connection is accepted. It is at that  point that SNI has to be initialized, per incoming socket.

Adding it to set_server_config did not actually activate SNI as far as I remember - that's what I tried first.

axos88:
@AdityaHPatwardhan let me know if you have outstanding concerns.

components/esp-tls/esp_tls_mbedtls.c
AdityaHPatwardhan: @axos88 Sorry for the delayed reply, I wanted to test this first at my end before commenting.
I will do that now and give my comments in some time.
Do you have a MWE which I could use to test out this feature ?
Thanks!

AdityaHPatwardhan: Hi @axos88 I was able to test the SNI feature locally.
I can confirm that with current approach if `servercert` and `server key` are not supplied and we configure the sni_callback using `mbedtls_ssl_conf_sni` in the `ser_server_config` function. Then the sni callback is set at once and is used for every connection formed by that specific context. There is no need to set the sni callback each time in the `esp_mbedtls_server_session_create`. I will update my comments based on that.

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -566,7 +566,16 @@ esp_err_t set_server_config(esp_tls_cfg_server_t *cfg, esp_tls_t *tls)
             return esp_ret;
         }
     } else {
+#if defined(CONFIG_ESP_TLS_SERVER_SNI_HOOK)
+        if (cfg->sni_callback == NULL) {
```
AdityaHPatwardhan, line 570: I think the SNI callback needs to be configured here.
I think only the check if sni_callback is present needs to be in the `else` part.
We should allow users to set the sni_callback along with providing server cert and keys. I think the sni callback would only be triggered when the ClientHello message would contain the `ServerName` extension. Otherwise, the normal certs would be used.

components/esp_https_server/src/https_server.c
```diff
@@ -213,51 +213,99 @@ static httpd_ssl_ctx_t *create_secure_context(const struct httpd_ssl_config *con
...
+            if (cfg->serverkey_buf) {
+                memcpy((char *) cfg->serverkey_buf, config->prvtkey_pem, config->prvtkey_len);
+                cfg->serverkey_bytes = config->prvtkey_len;
+            } else {
+                ESP_LOGE(TAG, "Could not allocate memory for server key");
+                free((void *) cfg->servercert_buf);
+                free((void *) cfg->cacert_buf);
+                free(cfg);
+                free(ssl_ctx);
+                return NULL;
```
AdityaHPatwardhan, line 284: I think this part is duplicated 3 times, can we have a `goto:` statement for this so that the code would jump to this in case of error.

AdityaHPatwardhan:
@axos88 Sorry for the delayed review, Thanks again for your contribution for adding the sni feature in esp-tls.
I have left some small suggestions overall the MR looks good to me.
Thanks,
Aditya

axos88:
Thanks, I'll update the pr next week, when I'm back

axos88:
Updated per your comments. Also discovered the certificate selection hook in mbedtls https://github.com/Mbed-TLS/mbedtls/issues/5430, so moved to use that one instead of the SNI hook, since it has guaranteed access to all TLS extensions received in the ClientHello.

This means that the application can make a better decision on which certificate to send, based on the ALPN, or the server certificate type, or any other TLS extension received.

This callback does not take a void* unfortunately, so it is a bit more circumstancial to use IMHO, so I have added support for setting the user_data of the ssl_conf as well, so one can supply their inputs there.

axos88:
@AdityaHPatwardhan  ^

axos88:
No, if the cert_callback does not set the private key and the ca_cert, the ones provided will be used by default, so it does make sense to allocate memory if the user wants a default.

The cert_selection callback can be used for setitng other parameters, such as verification mode, client ca bundle, etc.

axos88:
Yeah I'm curious too, but as far as I have seen this is a strict superset. Just made that comment to have an official confirmation on this.

As far as https://mbed-tls.readthedocs.io/en/latest/kb/how-to/use-sni/ is concerned, the certificate_selection callback can call all of these functions described there, but has access to more information.

AdityaHPatwardhan (approved):

AdityaHPatwardhan:
@axos88 Thank you very much for the PR. I will see to it that these changes are merged in our internal code-base, after that they should be available on GitHub in a few days.

axos88:
So do I understand correctly that GitHub is a mirror of some interior repository you use? Saw some strange comments in some PRS with shasums, that would explain them

axos88:
@AdityaHPatwardhan, do you have an estimate of how long until this can be merged?

AdityaHPatwardhan:
Hi @axos88 The MR should be merged and should be available on GitHub by next week.

Issue #9726, opened 2022-09-06
Information obtained by WPS cannot distinguish between 2.4GHz AP and 5GHz AP (IDFGH-8238)


sakamoto330:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

## Environment

- Chip : esp32S3
- IDF version : v4.4.2-1b16ef6cfc2479a08136782f9dc57effefa86f66

## Problem Description

Information obtained by WPS cannot distinguish between 2.4GHz AP and 5GHz AP.

### Issue
The following information can be obtained by executing WPS with reference to the sample code.

```c
/** Argument structure for WIFI_EVENT_STA_WPS_ER_SUCCESS event */
typedef struct {
    uint8_t ap_cred_cnt;                        /**< Number of AP credentials received */
    struct {
        uint8_t ssid[MAX_SSID_LEN];             /**< SSID of AP */
        uint8_t passphrase[MAX_PASSPHRASE_LEN]; /**< Passphrase for the AP */
    } ap_cred[MAX_WPS_AP_CRED];                 /**< All AP credentials received from WPS handshake */
} wifi_event_sta_wps_er_success_t;
```

esp32S3 doesn't support WiFi 5GHz, but some routers can get 5GHz AP information as well as 2.4GHz AP.
Is there a way to distinguish this 2.4GHz AP from a 5GHz AP?

kapilkedawat:
Hi @sakamoto330 , we take care of this by saving all the credentials and then try to connect to them one by one. 

This is already handled in wps example code and you can take reference from that.
For credentials save: https://github.com/espressif/esp-idf/blob/master/examples/wifi/wps/main/wps.c#L91 , connection retry to next network: https://github.com/espressif/esp-idf/blob/master/examples/wifi/wps/main/wps.c#L70

sakamoto330:
Thanks! @kapilkedawat 

> For credentials save: https://github.com/espressif/esp-idf/blob/master/examples/wifi/wps/main/wps.c#L91 

Above we can get two information of 5GHz AP and 2.4GHz AP, but we can get only SSID and passphrase.
Is there a way to determine which is the 5GHz AP and which is the 2.4GHz AP?

[...]

PR #9840, opened 2022-09-23
Fillter out the wps credential with unsupported band. (IDFGH-8370)

kneko715:
#9726 
Some access point passes the 5GHz band credential in 2.4GHz wps procedure.
ESP32 can filter out the unnecessary credentials by the attribute of rf_bands or ap_channel.
But this way is not perfect, because rf_bands and ap_channel is optional attribute.
Although, this is helpful for application .
ESP32 won't have to try worthless 5GHz band credential.


Alvin1Zhang:
Thanks for your contribution.

kapilkedawat:
As explained in https://github.com/espressif/esp-idf/issues/9726, we will not be able to take this patch. However we do appreciate your contribution in making the SDK better :) .

Issue #9825, opened 2022-09-21
In AP+STA mode starting to WPS causes disconnection of STA joined to AP


sakamoto330:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

## Environment

- Chip : esp32S3
- IDF version : v4.4.2-1b16ef6cfc2479a08136782f9dc57effefa86f66

## Problem Description

In AP+STA mode starting to WPS causes disconnection of STA (PC etc.) joined to AP (ESP32).

### Issue
Connection of STA (PC etc.) joined to AP (ESP32) will lost.
It is the part of "/// STA (PC etc.) disconnects from AP (ESP32)  ///" described in the Debug Logs.

### Expected Behavior
STAs (PC etc.) joined to AP (ESP32) keep connection.

### Steps to reproduce

1. Boot in AP+STA mode.
2. Join to AP (ESP32) from STA (PC etc.).
3. Start to WPS with STA (ESP32).

### Code to reproduce this issue

```cpp
	printf("################ start"NL);
	ESP_ERROR_CHECK(esp_netif_init());
	ESP_ERROR_CHECK(esp_event_loop_create_default());
	esp_netif_create_default_wifi_ap();
	esp_netif_create_default_wifi_sta();
	wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
	ESP_ERROR_CHECK(esp_wifi_init(&cfg));

	wifi_config_t wifi_config = {
			.sta = {
					.bssid_set = false,
					.ssid = "logitec62",
					.password = "abcd1234",
					.threshold.authmode = WIFI_AUTH_WPA2_PSK,
					.channel = 1
					}
			};
	wifi_config_t wifi_config1 = {
			.ap = {
					.ssid = "123",
					.password = "abcd1234",
					.max_connection = 10,
					.authmode = WIFI_AUTH_WPA_WPA2_PSK,
					.channel = 1
					}
			};

	ESP_ERROR_CHECK(esp_wifi_set_mode(WIFI_MODE_APSTA));
	ESP_ERROR_CHECK(esp_wifi_set_config(WIFI_IF_STA, &wifi_config));
[...]

PR #9873, opened 2022-09-28
Fix AP stop sending beacon during WPS on STA in wifi concurrent mode.

sakamoto330:
https://github.com/espressif/esp-idf/issues/9825

Connection of STA (PC etc.) joined to AP (ESP32) may be lost when the following steps.

1. Boot in AP+STA mode.
2. Join some STA (PC etc.) to AP (ESP32). 
3. Start to WPS on STA (ESP32).

Since the channel is switched when scanning with WPS, AP mode beacons may not be sent.
As a result, STAs (PC etc.) connected in AP (ESP32) are disconnected.

Instead of scanning all channels at once, to take interval on each channel scan and gives AP mode the time to send beacon.

### DIFF
--- a/components/wpa_supplicant/esp_supplicant/src/esp_wps.c
+++ b/components/wpa_supplicant/esp_supplicant/src/esp_wps.c
@@ -1864,7 +1864,7 @@ wifi_wps_scan_done(void *arg, STATUS status)
             sm->ignore_sel_reg = true;
             sm->wps_pin_war = false;
         }
-        ets_timer_arm(&sm->wps_scan_timer, 100, 0);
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

kapilkedawat:
Hi @sakamoto330 ,

Thanks for the contribution however this case should be handled in much more cleaner way by the driver.

We will consider implementing bucket scan/split scan to take care of similar scenario in future versions.

Also incase you are using this patch in your current IDF, suggest following changes :

.show_hidden = 0, // WPS AP cannot set hidden ssid(as per specs)

#define WPS_IGNORE_SEL_REG_MAX_CNT 53

PR #9900, opened 2022-10-02
CACHEERR on second core (IDFGH-8434)

svenbieg:
1) esp_cache_err_int_init() is called from both cores (call_start_cpu0 and call_start_cpu1), so ETS_CACHE_IA_INTR_SOURCE gets routed twice.
2) ETS_CACHE_CORE1_ACS_INTR_SOURCE is routed to core 1, this should be added to the documentation in soc.h.


mahavirj:
@svenbieg What problem you faced without this fix? Can you please confirm that the test case mentioned at https://github.com/espressif/esp-idf/blob/02605f1a31be4233b97e83a0460bb458f277dfea/components/spi_flash/test/test_cache_disabled.c#L101 still works as expected?

svenbieg:
You are right, this problem is mentioned in the comments. But why should this interrupt get routed twice?

svenbieg:
I'm sorry, then only the documentation in soc.h is missing. CACHEERR on core 1 can be raised by CACHE_CORE1_ACS_INTR_SOURCE.
Thank You very much for Your explenation!

PR #9917, opened 2022-10-05
Fix vfs uart_end_select() for concurrent calls to select() (IDFGH-8455)

hadrava:
There seems to be a bug in the function uart_end_select(). Let's assume that there is a task A, which waits for data on UART1 using select, and task B similarly waits for data on (same or other) UART. In theory, those selects should not interfere much with each other. But they unfortunately do. When one of the select finishes, the function uart_end_select() unregisters notification callback for all UARTs. This is done despite the fact, that there is still the other select() running and waiting for different UART.

The first two commits add a new sub-test to "concurrent selects work" test case. This checks for the issue mentioned above. This test will fail with current versions of esp-idf, because second task will not be notified.

The third commit fixes the issue, so the modified test will pass again. To be honest, I am not very satisfied with the look of my fix, because it loops through all combinations of UART and registered select. So any ideas on how to clean it up are very welcomed.

### DIFF
--- a/components/vfs/test/test_vfs_select.c
+++ b/components/vfs/test/test_vfs_select.c
@@ -544,6 +544,69 @@ TEST_CASE("concurrent selects work", "[vfs]")
         vSemaphoreDelete(send_param.sem);
     }
 
+    {
+        // First task waits for UART writing second for reading with longer timeout.
+        // Data is written to UART after the first task time-out.
+        // Second task should be notified about them immediately.
+
+        struct timeval tv1 = {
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

RathiSonika:
Hello @hadrava. 
This issue has already been resolved. Please review the commit [e48ca243](https://github.com/espressif/esp-idf/commit/e48ca243) for more details. Thank you so much for your efforts!

PR #9950, opened 2022-10-10
ESP32-S3 page-table initialized twice (IDFGH-8492)

svenbieg:
mmu_ll_unmap_all() is called twice in mmu_hal_init(), but mmu_id is ignored in mmu_ll.h.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
Hi @svenbieg would you please squash the commits in this PR? **four** commits are too much for the small fix.

svenbieg:
I'm sorry, i don't know anything about "squashing".


---- suda-morris schrieb ----


Hi @svenbieg<https://github.com/svenbieg> would you please squash the commits in this PR? four commits are too much for the small fix.

svenbieg:
You should have gotten a compiler-warning already! ;-)

svenbieg:
The question is if there are any systems on a chip with two page-tables? In a multi-tasking environment any core is able to take over the same task. How should this be possible with different page-tables?

svenbieg:
I'm satisfied with one page-table already.

svenbieg:
I think i know what You mean: I'd like to undo changes i don't like anymore (squashing?). I've tried, but in case i have to commit Your changes too. You can try to contact Microsoft, i didn't have any success with it. I'm really sorry about this, the only thing i'm trying to do is taking over Your kernel. I have to make a compliment at this point because i couldn't find any bugs by now.
But two page-tables? What is this?

Issue #9462, opened 2022-07-29
[Feature Request] sdkconfig: "Print registers and reboot after delay" would be useful (IDFGH-7947)


chipweinberger:
A `CONFIG_ESP_SYSTEM_PANIC_PRINT_DELAY_REBOOT` option would be usable for both development & production, and would be most ideal for Beta testing phase.

1. "print & reboot" is annoying for development. A Boot Loop will log to serial too much, forever.
2. "print registers and halt" is not good for users.

We need a middle approach,  "print & delay & reboot, with a configurable delay. e.g. 6 seconds.

Thanks for your consideration.

chipweinberger:
Bump. 

A `CONFIG_ESP_SYSTEM_PANIC_PRINT_DELAY_REBOOT` option would let me keep it the same between development and production, and it just generally more useful IMO. 20htz boot loops are annoying =)

chipweinberger:
@ESP-Marius, @igrr, @Alvin1Zhang 

Just want to make sure 1 person sees this.

igrr:
Hi @chipweinberger,
All issues posted on Github are synced to our internal issue tracker, categorized and land on of of the teams' backlog. We've seen this feature request but currently consider it relatively low priority. Most efforts right now are focused on v5.0 release and various bug fixes.

Have you checked this example demonstrating how to set up builds for multiple configurations (development/production)? https://github.com/espressif/esp-idf/blob/master/examples/build_system/cmake/multi_config/README.md
I hope it can satisfy your requirement for now.

chipweinberger:
Thanks for your input Ivan, as always. 

That is a useful page, for now =)

**Use case:** For Beta testing, reboot after delay is most ideal. A beta device may need to be debugged "live", but must still reboot automatically.

chipweinberger:
Pull Request: https://github.com/espressif/esp-idf/pull/9962

@igrr, hope you consider it!

chipweinberger:
merged into master.

PR #9962, opened 2022-10-12
[Panic] add user configurable reboot delay in seconds (IDFGH-8506)

chipweinberger:
Issue: https://github.com/espressif/esp-idf/issues/9462

Adds a **CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS** option to the panic handler.

Very simple change & tested on a ESP32-S3, and ESP32-S2.

Why?
 - **"print & reboot"** is annoying. Boot loops log way too much, forever & ever. There's no time to look at the stack trace. I always need to unplug the device, and then scroll back and look for the point it crashed.
 - **"print & halt"** is not good for production or beta testing. A device that does not reboot is not useable.

This middle approach is the best of both, especially for beta testing, and is sorely needed! =)

### DIFF
--- a/components/esp_system/panic.c
+++ b/components/esp_system/panic.c
@@ -379,6 +379,27 @@ void esp_panic_handler(panic_info_t *info)
     wdt_hal_write_protect_enable(&rtc_wdt_ctx);
 #if CONFIG_ESP_SYSTEM_PANIC_PRINT_REBOOT || CONFIG_ESP_SYSTEM_PANIC_SILENT_REBOOT
 
+#if CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS
+
+    disable_all_wdts();
+
+    panic_print_str("Waiting to reboot...\r\n");
+
+    for(int i = 0; i < CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS; i++) {
+
+        // Display a progress countdown.
+        // Only print every 5th second so users terminal is still "calm".
+        if (i % 5 == 0) {
+            panic_print_dec(CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS - i);
+            panic_print_str(" seconds...\r\n");
+        }
+
+        esp_rom_delay_us(1000000);
+    }
+
+    esp_panic_handler_reconfigure_wdts();
+#endif /* CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS */
+
     if (esp_reset_reason_get_hint() == ESP_RST_UNKNOWN) {
         switch (info->exception) {
         case PANIC_EXCEPTION_IWDT:
--- a/components/esp_system/Kconfig
+++ b/components/esp_system/Kconfig
@@ -48,6 +48,14 @@ menu "ESP System Settings"
                 Invoke gdbstub on the serial port, allowing for gdb to attach to it and to do a debug on runtime.
     endchoice
 
+    config ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS
+        int "Panic reboot delay (Seconds)"
+        default 0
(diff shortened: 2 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

chipweinberger:
I have been testing this change for a ~week without issue. Very nice quality of life change =)

chipweinberger:
@igrr Also, hoping to get this merged. Super useful!

chipweinberger:
@0xjakob bump.

I've been using this myself for weeks now. Users will really appreciate the option.

chipweinberger:
@igrr Just hoping for an update.

components/esp_system/panic.c
```diff
@@ -373,6 +373,27 @@ void esp_panic_handler(panic_info_t *info)
     wdt_hal_write_protect_enable(&rtc_wdt_ctx);
 #if CONFIG_ESP_SYSTEM_PANIC_PRINT_REBOOT || CONFIG_ESP_SYSTEM_PANIC_SILENT_REBOOT
 
+#if CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS
+
+    disable_all_wdts();
+
+    panic_print_str("Waiting to reboot...\r\n");
+
+    for(int i = 0; i < CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS; i++){
```
0xjakob, line 382: Please note our [style guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/style-guide.html#horizontal-space):
`for (int i = 0; i < CONFIG_ESP_SYSTEM_PANIC_REBOOT_DELAY_SECONDS; i++) {`

0xjakob (commented):
Code in general looks good to me. Just one very minor note about coding style.

chipweinberger:
Thanks for the update!

>  What tool are you using to monitor output from devices?

VSCode. AFAIK, there is no way to "pause" the VSCode console after a crash. I always need to pull the cable. 

> you can always press Ctrl+] to stop the output

Even still, by the time I've noticed the crash my device will have logged 1000 lines immediately on boot, so it's quite a bit of annoying scrolling backwards...

> Another minor risk is that something goes wrong inside the code where the WDT is disabled.

`disable_all_wdts();` is already called in the panic handler code, so this should not add additional risk.

I don't see the reason not to add it, given that it is just an "option", with no obvious downside AFAICT. I've really enjoyed the past month of using it. But of course, it is your decision ultimately.

components/esp_system/panic.c
chipweinberger: updated.

chipweinberger:
Not sure how to fix precommit. All I did was fix a bracket.

`fatal: could not read Username for 'https://gitlab.com/': No such device or address`

0xjakob:
> Not sure how to fix precommit. All I did was fix a bracket.
> 
> `fatal: could not read Username for 'https://gitlab.com/': No such device or address`

Sorry for the inconvenience, this is an issue with a mirror of a third party project. The fix has been merged already on our internal gitlab server, but needs to run the pipeline on master to sync to github. With the next sync, it should be available here, please rebase then.

chipweinberger:
updated.

Please decide to merge or deny! thanks @0xjakob !

0xjakob (commented):
Hi @chipweinberger,

we finally discussed this PR internally and found a couple of unsolved questions that we would like to addresse before we merge it. Feel free to resolve them yourself, otherwise we would like to add another commit addressing these.

First of all, the concern is that disabling the watchdog timers for such a long period is quite risky. It would be better to adjust the timeout of the watchdog according to the reboot delay. We furthermore would like to test somehow that the configuration works, probably inside the test applications. Also, would you be OK to adjust the code so that there is no continuous reboot print anymore but just a one-time print, like "Waiting %d seconds before reboot"?

chipweinberger:
That sounds OK to me. 

It is great if you make the adjustments you like.

0xjakob:
@chipweinberger We merged the feature. It's available on master now.

chipweinberger:
I noticed! looks great. thanks for you work getting it in.

PR #10082, opened 2022-10-31
esp_wifi: Fix deprecated CONFIGs still being used, causing WiFi to not work when SPIRAM is enabled (IDFGH-8637)

zRedShift:
These configs were removed 6 months ago and unified into `CONFIG_SPIRAM`, this is the only place they are left at. KConfig replaces them with `CONFIG_SPIRAM`, so they're always unset. `WIFI_CACHE_TX_BUFFER_NUM` is then always set to 0, while `CONFIG_SPIRAM` sets the `CONFIG_FEATURE_CACHE_TX_BUF_BIT` capability of `esp_wifi, which then causes the WiFi to never be able to connect to any AP.

Good candidate to patching in `v5.0` as well.


ESP-Marius:
Hi @zRedShift, good catch. We should update the config name.

Are you sure that it was causing issues though? Because we still add old config name using the `sdkconfig.rename` files to keep backwards compatibility (e.g. see `components/esp_hw_support/sdkconfig.rename.esp32s3`)

When I tested this seems to work fine, `CONFIG_ESP32S3_SPIRAM_SUPPORT=y` gets set and  `WIFI_CACHE_TX_BUFFER_NUM` is correctly redefined.

zRedShift:
@ESP-Marius Yes I saw the `sdkconfig.rename`, and I assumed at first that it does the opposite of what it does, renames all the deprecated configs to the new configs (see below). It does look like `menuconfig` sets it, and it's set in `sdkconfig.h` automatically during `idf.py build` even without `menuconfig`.

In C-based `esp-idf` projects this works fine, but I ran it in the context of esp-rs/esp-idf-sys#148, which doesn't seem to handle `sdkconfig.rename` files correctly. Instead of adding the old deprecated config names to the `sdkconfig.h`, it removes them even when they're set in `sdkconfig.defaults`, and they were set. And it doesn't add `CONFIG_SPIRAM=y` either, if the deprecated config is set. I'll let them know so we can investigate this.

ESP-Marius:
@zRedShift I also asked the rust team about this as I suspected it might be related to how the renames are handled in other projects.

We still wanna update the names to the new ones in IDF anyways though! 

Do you mind just updating the commit message to clarify that ` causing WiFi to not work when SPIRAM is enabled`  probably just occurs in the context of certain non esp-idf projects? Then i'll sync this commit to our internal branch and get it merged. :smile:

zRedShift:
@ESP-Marius done, amended the commit message :+1:

ESP-Marius:
sha=d556c2d7df42224f3365227f3dca4c881665e3ab

ESP-Marius:
@zRedShift Thanks! 

The topic will be automatically updated when this is merged and synced to github

zRedShift:
@ESP-Marius FYI, I think I found the reason why the renaming didn't work with `esp-rs`: the said config was target specific (`sdkconfig.rename.esp32s3`) but `__target_init()` wasn't called by CMake and the `IDF_TARGET` env var wasn't converted to a property so it looked for `sdkconfig.rename.` in the various components and didn't pick this one up. Adding it [here](https://github.com/esp-rs/esp-idf-sys/pull/148/files#diff-30506ba71549c7b4d9b2bf6ab8eeac2707c6e9f3d23e8f962a46917b07a44a55) seemed to fix this, but `__target_init()`  is a private function...

jack0c:
merged: ccd5a5a866dcb9f814c1a71d7a4ab98f7ae5ac7d

Issue #9366, opened 2022-07-15
[ESP32-S3] How to detect if USB CDC Console is connected to a USB Host? (USB OTG Peripheral) (IDFGH-7837)


chipweinberger:
**ESP32-S3**

I'd like to know when a USB Host computer is connected to the ESP32-S3's USB CDC Console (which I've read is implemented with the USB OTG Peripheral).

**Why? I want to prevent my esp32-s3 application from enabling Usb Host mode if a CDC Host computer is already connected within the first 3 seconds of boot.** 

Note: my device only has a standard USB-A Female port on it. Connecting a CDC Host will happen with a USB-A-to-USB-A cable, for firmware updates / console access.

Specifically:
- (A) Under normal operation, my ESP32-S3 uses Usb Host to communicate with external devices
- (B) However, for the first 3 seconds after boot I want the ESP32-S3 to listen in USB CDC Console mode, and if a computer connects I skip step (A) and remain in CDC Console mode

igrr:
In case your device isn't connected to the host PC (and your program initializes USB host itself), where do you need the console output (logs) to go? Do you still want to be able to access them over the UART?

If you plan to retarget the console output to UART when going into the USB host mode, it might actually be easier to start with UART console in the first place, and only initialize the USB CDC console if the device is connected to the host PC. If the host is not connected, then initialize USB host and carry on with the UART console.

chipweinberger:
Thanks for the suggestions =)

> In case your device isn't connected to the host PC (and your program initializes USB host itself), where do you need the console output (logs) to go? Do you still want to be able to access them over the UART?

Ideally yes. Or possibly USB Serial/JTAG. Is there any example code for "retargeting" log output?

>  might actually be easier to start with UART console in the first place, and only initialize the USB CDC console if the device is connected to the host PC

Can you elaborate? AFAICT:

Boot w/ Uart logs Advantages:
- more reliable for early logs & panic handler
- no need to "switch" log outputs when turning on USB Host mode, so an attached uart monitor will have access to all logs

Boot w/ CDC logs Advantages:
- my device has an external USB port, so more logs will be available at boot time without opening the device

I think you are right that booting w/ uart advantages outweighs the cons.

chipweinberger:
[...]

PR #10097, opened 2022-11-02
[USB Serial/JTAG] [VFS] add flag if driver has ever received bytes (IDFGH-8651)

chipweinberger:
Related issue: https://github.com/espressif/esp-idf/issues/9366

This PR was spun out from https://github.com/espressif/esp-idf/pull/9983

introduces: `usb_serial_jtag_has_driver_received_bytes();`

I need a way to detect if a USB terminal is attached, so that I can decide which port to open my terminal on. If the user types a character over USB during boot, I open USB Console, otherwise I open UART Console.

(I also send a 'terminal probe request' during boot, so USB Console detection is automatic for most terminals 😊)

I like that this solution is less complex (i.e. more reliable) than calling `read()` or initializing `linenoise` or `esp_console`.

### DIFF
--- a/components/driver/include/driver/usb_serial_jtag.h
+++ b/components/driver/include/driver/usb_serial_jtag.h
@@ -72,6 +72,16 @@ int usb_serial_jtag_read_bytes(void* buf, uint32_t length, TickType_t ticks_to_w
  */
 int usb_serial_jtag_write_bytes(const void* src, size_t size, TickType_t ticks_to_wait);
 
+
+/**
+ * @brief Returns true if the USB Serial/JTAG Controller driver has ever
+ * received any bytes over its serial port since the driver was installed.
+ *
+ * This can be used as a simple way to determine that a serial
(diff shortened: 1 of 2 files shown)


chipweinberger:
Looking around for alternatives, maybe I could instead use this function?

`static inline int usb_serial_jtag_ll_rxfifo_data_available(void)`

I'd just need to make sure nothing else reads the data, but I think thats already the case unless I initialize the usb_serial_jtag_driver.

chipweinberger:
I'll test `usb_serial_jtag_ll_rxfifo_data_available`, and close this PR if I find it to be sufficient.

chipweinberger:
Closing this PR. 

This functionality can now be better implemented with `bool usb_serial_jtag_is_connected(void)` which was added to master on Februrary 26, 2023. https://github.com/espressif/esp-idf/commit/1a66459b4454f0b45b1910b2bf260c77aef119f4

PR #10099, opened 2022-11-02
[Console Repl] USB Serial/JTAG default to non-blocking for TX (IDFGH-8653)

chipweinberger:
This is the 2nd PR spun out from: https://github.com/espressif/esp-idf/pull/9983

- add `esp_vfs_usb_serial_jtag_use_driver_for_rx(); ` to keep TX in non-blocking mode
- **esp_console_repl.c** - use `esp_vfs_usb_serial_jtag_use_driver_for_rx();` by default. 

Users dont want to block the entire program when a USB terminal is not connected. This PR makes it the same behavior as UART Console, which is what the vast majority of users want, IMO. 

Note: even in non-blocking mode TX still blocks for 50ms to see if theres a chance the buffer will clear up, but it does not block again until there is space again. This is precisely the behavior most users want IMO.

### DIFF
--- a/components/console/esp_console_repl.c
+++ b/components/console/esp_console_repl.c
@@ -162,8 +162,8 @@ esp_err_t esp_console_new_repl_usb_serial_jtag(const esp_console_dev_usb_serial_
         goto _exit;
     }
 
-    /* Tell vfs to use usb-serial-jtag driver */
-    esp_vfs_usb_serial_jtag_use_driver();
+    /* Tell vfs to use usb-serial-jtag driver for reads*/
+    esp_vfs_usb_serial_jtag_use_driver_for_rx();
 
     // setup history
     ret = esp_console_setup_history(repl_config->history_save_path, repl_config->max_history_len, &usb_serial_jtag_repl->repl_com);
--- a/components/vfs/include/esp_vfs_dev.h
+++ b/components/vfs/include/esp_vfs_dev.h
@@ -110,8 +110,30 @@ void esp_vfs_dev_uart_use_driver(int uart_num);
  * @brief set VFS to use USB-SERIAL-JTAG driver for reading and writing
  * @note application must configure USB-SERIAL-JTAG driver before calling these functions
  * With these functions, read and write are blocking and interrupt-driven.
+ * 
+ * @warning If there is no terminal *consuming* the bytes we are sending,
+ * sending data will block indefinitely, until a terminal attaches and consumes them.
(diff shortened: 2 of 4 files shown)


chipweinberger:
From @igrr 

```
Instead of introducing this new function, would it instead make more sense to fix the TX-blocking-indefinitely behavior?

This is already done in usb_serial_jtag_tx_char with the TX_FLUSH_TIMEOUT_US check, we exit and drop the character if the flush doesn't happen for too long.

Seems like the same behavior could be achieved in usbjtag_tx_char_via_driver, by passing pdMS_TO_TICKS(TX_FLUSH_TIMEOUT_US / 1000) instead of portMAX_DELAY?
```

So we could change it to this? lmk what you think Ivan.

```
int usb_serial_jtag_write_bytes(const void* src, size_t size, TickType_t ticks_to_wait)
{
    ESP_RETURN_ON_FALSE(size != 0, ESP_ERR_INVALID_ARG, USB_SERIAL_JTAG_TAG, "size should be larger than 0");
    ESP_RETURN_ON_FALSE(src != NULL, ESP_ERR_INVALID_ARG, USB_SERIAL_JTAG_TAG, "Invalid buffer pointer.");
    ESP_RETURN_ON_FALSE(p_usb_serial_jtag_obj != NULL, ESP_ERR_INVALID_ARG, USB_SERIAL_JTAG_TAG, "The driver hasn't been initialized");

    usb_serial_jtag_obj_t* sjtag = p_usb_serial_jtag_obj;

    // try to send without blocking
    if (xRingbufferSend(sjtag->tx_ring_buf, (void*) (buff), size, 0)){
       goto send_success;
    }

    
    // If the ring buffer is full, try to send again with a short
    // blocking delay, after which we'll never block again 
    // until the buffer has space again.
    if (esp_timer_get_time() - sjtag.last_tx_ts < TX_FLUSH_TIMEOUT_US) {

      // try to send, with up to a short blocking delay
      if (xRingbufferSend(sjtag->tx_ring_buf, (void*) (buff), size, TX_FLUSH_TIMEOUT_US);){
         goto send_success;
      }
    }

   // send failure
   return 0;

sent_success:
   // Now trigger the ISR to read more data from the ring buffer.
   usb_serial_jtag_ll_ena_intr_mask(USB_SERIAL_JTAG_INTR_SERIAL_IN_EMPTY);
   sjtag->last_tx_ts = esp_timer_get_time();
   return size;
}
```

chipweinberger:
Closing in favor of https://github.com/espressif/esp-idf/pull/10142

PR #10106, opened 2022-11-04
hal/uart_ll.h: Fix compile with C++ (IDFGH-8661)

nomis:
I'm including `<hal/uart_ll.h>` in my C++ application because I need to bypass the uart driver. The inline functions in the header file fail to compile as C++.

With xtensa-esp32s2-elf-g++ (crosstool-NG esp-2021r2-patch3) 8.4.0 it doesn't allow copying the union like this:
```
hal/uart_ll.h:131:39: error: binding reference of type 'const uart_dev_s::<unnamed union>&' to 'volatile uart_dev_s::<unnamed union>' discards qualifiers
     typeof(hw->clk_div) div_reg = hw->clk_div;
hal/uart_ll.h:131:39: error: cannot bind rvalue reference of type 'uart_dev_s::<unnamed union>&&' to lvalue of type 'volatile uart_dev_s::<unnamed union>'
     typeof(hw->clk_div) div_reg = hw->clk_div;
```
All of the enums need explicit casts for conversion to/from integers.

### DIFF
--- a/components/hal/esp32/include/hal/uart_ll.h
+++ b/components/hal/esp32/include/hal/uart_ll.h
@@ -111,7 +111,8 @@ FORCE_INLINE_ATTR void uart_ll_set_baudrate(uart_dev_t *hw, uint32_t baud, uint3
  */
 FORCE_INLINE_ATTR uint32_t uart_ll_get_baudrate(uart_dev_t *hw, uint32_t sclk_freq)
 {
-    typeof(hw->clk_div) div_reg = hw->clk_div;
+    typeof(hw->clk_div) div_reg;
+    div_reg.val = hw->clk_div.val;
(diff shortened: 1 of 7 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=5d46d1ada01a34ba3bd930578c5b841967ff11c4

0xjakob (commented):
@nomis Thanks a lot for the fix! 

Unfortunately, C++ compatibility isn't best in IDF right now. But we're doing our best to improve it. If you encounter further problems with C++, please let us know as well.

0xjakob (approved):

suda-morris:
PR is merged in the commit: https://github.com/espressif/esp-idf/commit/d1993c01f26a59c0b0d85402b61563beba5f74fe

PR #10127, opened 2022-11-07
[Heap Trace Standalone] Support storing records in SPI RAM (IDFGH-8683)

chipweinberger:
Add support for storing records in SPI RAM. Internal RAM is just used as a temporary buffer.

**With this change, amazingly a small internal ram record buffer size of 13 was enough to handle all records!** And this is for a very complex app with bluetooth, wifi, console. etc.

**With this change, I am able to store *all* allocations since boot.**  During development one can leave this enabled, & If they unexpectedly hit a _Low Ram_ condition, they can immediately print all allocations and analyze the problem without needing to reproduce with host mode enabled.

This makes standalone mode 10x more useful, IMO.

Also adds `heap_trace_dump_internal_ram_only()`, which is often all we care about.

And adds `heap_trace_summary()` so that users can access more useful heap tracing information.

## Code: 
In this PR, **Internal RAM** is just used as a temporary buffer. We stream records to SPI ram whenever we get a chance (i.e. at the next heap_trace_X() invocation from a non-IRAM function).

### DIFF
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -23,16 +23,39 @@ static portMUX_TYPE trace_mux = portMUX_INITIALIZER_UNLOCKED;
 static bool tracing;
 static heap_trace_mode_t mode;
 
-/* Buffer used for records, starting at offset 0
-*/
-static heap_trace_record_t *buffer;
-static size_t total_records;
+struct records_t {
 
-/* Count of entries logged in the buffer.
+    /* Buffer used for records, starting at offset 0 */
+    heap_trace_record_t *buffer;
 
-   Maximum total_records
-*/
-static size_t count;
+    /* capacity of the buffer */
+    size_t capacity;
+
+    /* Count of entries logged in the buffer.*/
+    size_t count;
+
+    /* During execution, we remember the minimum 
+       value of (capacity - count). This can help you
+       choose the right size for your buffer capacity.*/
+    size_t high_water_mark;
+
+    /* Has the buffer overflowed and lost trace entries? */
+    bool has_overflowed;
+};
+
+typedef struct records_t records_t;
+
+// Forward Defines
+static void drain_internal_cache();
+static void remove_record(records_t* r, int index);
+static void heap_trace_dump_base(bool internalRam, bool spiRam);
+
+/* The actual records. Only used as a 
+   temporary cache if SPI RAM is used */
+static records_t internal_records;
+
+/* Long term storage of records */
+static records_t spi_records;
 
 /* Actual number of allocations logged */
 static size_t total_allocations;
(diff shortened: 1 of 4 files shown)


components/heap/heap_trace_standalone.c
```diff
@@ -115,20 +115,55 @@ esp_err_t heap_trace_get(size_t index, heap_trace_record_t *record)
...
+
+            const char* label = "";
+            if (esp_ptr_internal(rec->address)) {
+                label = ", Internal-RAM";
+            }
+            if (esp_ptr_external_ram(rec->address)) {
+                label = ", SPI-RAM";
+            }
+
+            printf("%d bytes (@ %p%s) allocated CPU %d ccount 0x%08x caller ",
```
atanisoft, line 164: It looks like the `%p` and `%s` may be reversed in this print statement.

components/heap/include/esp_heap_trace.h
```diff
@@ -114,10 +138,10 @@ esp_err_t heap_trace_stop(void);
  * - ESP_ERR_INVALID_STATE Heap tracing was already started.
  * - ESP_OK Heap tracing resumed.
  */
-esp_err_t heap_trace_resume(void);
+esp_err_t heap_trace_resume(void)
```
atanisoft, line 141: this looks like an unintentional removal of the `;`. I'm not certain it would compile without it.

chipweinberger: thanks.

 I may of removed that in my last force-push

components/heap/heap_trace_standalone.c
chipweinberger: good catch.

components/heap/include/esp_heap_trace.h
chipweinberger: updated.

components/heap/heap_trace_standalone.c
chipweinberger: updated.

chipweinberger:
Fixed some odds and ends, and re-tested both standalone modes.

This PR should be good to merge.

chipweinberger:
Okay, changes are done. I  re-tested `heap_trace_init_standalone` and `heap_trace_init_standalone_with_spi_ram` and they work as expected.

This should be good to merge.

Alvin1Zhang:
Thanks for your contribution.

chipweinberger:
@igrr 

From `esp_heap_trace.h`
```
* @param record_buffer Provide a buffer to use for heap trace data. Must remain valid any time heap tracing is enabled, meaning
 * it must be allocated from internal memory not in PSRAM.
```

I'd think avoiding the "cache disabled" issue is important.

igrr:
In ESP-IDF internal components we don't perform heap allocations from interrupt handlers, so there should be no cases of accessing heap functions while cache is disabled. This might only be the case in user code, if a heap related function is called from an ISR. However I don't think there is a need to support these two things simultaneously:
1. heap functions called from ISRs (hence, possibly while cache is disabled)
2. heap tracing into a buffer in external RAM

since 1) is a bad practice and is not recommended anyway.

Adding an assert somewhere in heap tracing code that `!(trace buffer pointer is in external RAM && cache is disabled)` seems to be sufficient to catch this potential issue.

chipweinberger:
Ok. Can you give more details on how to do this? `!(trace buffer pointer is in external RAM && cache is disabled)`

chipweinberger:
I'll close this PR and open a new one without the dual Internal / SPI ram buffers.

igrr:
Sorry, I should have mentioned that. Probably something along the lines of: `spi_flash_cache_enabled() || esp_ptr_internal(ptr)`

PR #10142, opened 2022-11-09
[USB Serial/JTAG Driver] use time-limited blocking for TX (IDFGH-8700)

chipweinberger:
Older PR: https://github.com/espressif/esp-idf/pull/10099

Tested working on a ESP32-S3.

Uses Ivan's suggestions for Usb Serial JTAG TX with the driver.

> Instead of introducing this new function, would it instead make more sense to fix the TX-blocking-indefinitely behavior?
> 
> This is already done in usb_serial_jtag_tx_char with the TX_FLUSH_TIMEOUT_US check, we exit and drop the character if the flush doesn't happen for too long.
> 
> Seems like the same behavior could be achieved in usbjtag_tx_char_via_driver, by passing pdMS_TO_TICKS(TX_FLUSH_TIMEOUT_US / 1000) instead of portMAX_DELAY?

### DIFF
--- a/components/driver/usb_serial_jtag.c
+++ b/components/driver/usb_serial_jtag.c
@@ -17,6 +17,8 @@
 #include "soc/periph_defs.h"
 #include "esp_check.h"
 
+#define TX_DRIVER_FLUSH_TIMEOUT_MS 50
+
 // The hardware buffer max size is 64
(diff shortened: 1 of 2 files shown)


chipweinberger:
new PR here, due to renaming the branch: https://github.com/espressif/esp-idf/pull/10208

Issue #9618, opened 2022-08-23
ASN1 tag error when downloading components (IDFGH-8120)


josesimoes:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.2

### Operating System used.

Windows

### How did you build your project?

Command line with CMake

### If you are using Windows, please specify command line type.

PowerShell

### What is the expected behavior?

Being able to use install scripts

### What is the actual behavior?

Error in download step

### Steps to reproduce.

run install.bat or install.ps1

### Build or installation Logs.

```plain
Downloading tools for win64: xtensa-esp32-elf, xtensa-esp32s2-elf, xtensa-esp32s3-elf, riscv32-esp-elf, esp32ulp-elf, esp32s2ulp-elf, cmake, openocd-esp32, ninja, idf-exe, ccache, dfu-util
Downloading xtensa-esp32-elf@esp-2021r2-patch3-8.4.0
Downloading xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip to C:\Users\jjss\.espressif\dist\xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip.tmp
WARNING: Download failure [WinError -2146881269] ASN1 valor de tag inválido encontrado
WARNING: Failed to download https://github.com/espressif/crosstool-NG/releases/download/esp-2021r2-patch3/xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip to C:\Users\jjss\.espressif\dist\xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip.tmp
Downloading xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip to C:\Users\jjss\.espressif\dist\xtensa-esp32-elf-gcc8_4_0-esp-2021r2-patch3-win64.zip.tmp
WARNING: Download failure [WinError -2146881269] ASN1 valor de tag inválido encontrado
[...]

PR #10147, opened 2022-11-09
Improve creating SSL context for download (IDFGH-8705)

josesimoes:
- This improves the SLL context creation for downloading files. It states the purpose for the context and loads the "internal" DigiCert certificate.
- Definitive fix for #9618.
- With this change I've managed to successfully install v4.4.2.

### DIFF
--- a/tools/idf_tools.py
+++ b/tools/idf_tools.py
@@ -48,7 +48,7 @@ import tempfile
 import time
 from collections import OrderedDict, namedtuple
 from json import JSONEncoder
-from ssl import SSLContext  # noqa: F401
+from ssl import Purpose, SSLContext  # noqa: F401
 from tarfile import TarFile  # noqa: F401
 from zipfile import ZipFile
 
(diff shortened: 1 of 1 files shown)


tools/idf_tools.py
```diff
@@ -432,8 +432,7 @@ def download(url, destination):  # type: (str, str) -> Optional[Exception]
         # This works around the issue with outdated certificate stores in some installations.
         if 'dl.espressif.com' in url or 'github.com' in url:
             try:
-                ctx = ssl.create_default_context()
-                ctx.load_verify_locations(cadata=DIGICERT_ROOT_CERT)
+                ctx = ssl.create_default_context(purpose=Purpose.SERVER_AUTH, cadata=DIGICERT_ROOT_CERT)
```
igrr, line 435: I think we should also call `ssl.load_default_certs` in this case, otherwise we will only load the DIGICERT_ROOT_CERT.

This is necessary since IDF_MIRROR_PREFIX_MAP and IDF_GITHUB_ASSETS environment variables can be used to redirect the download URLs to a mirror, which may be using a different certificate.

josesimoes: Please check

josesimoes: I'm not a Python expert so I can't comment further... 😅 
I do know that this change made the ANSI TAG error go away.
Maybe that's because of the cdata being passed here instead of `load_verify_locations`...
Feel free to test and change!

dobairoland:
Please also use https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html in order to solve all the issues and comply with our coding style.

PR #10165, opened 2022-11-11
Clarify GPIO Mask Comment (IDFGH-8724)

mrengineer7777:
I was confused by the comment in esp32/include/soc/soc_caps.h on SOC_GPIO_VALID_GPIO_MASK.

After verifying the mask bits are correct, I have updated the comments for esp32, esp32s2, esp32s3.
I did not check the mask base 0x...FF size.

Not touched because they use a different mask format: esp32c2, esp32c3, esp32c6, exp32h4.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=11a0aa548e14904bfef7caba6161bd80d5086480

suda-morris:
commits are squashed in https://github.com/espressif/esp-idf/commit/0c89ab3b582988eb561a177d946e4f45bda400b1

PR #10225, opened 2022-11-22
[bt] osi thread - use task notifications instead of semaphores

tgotic:
Use lighter task notifications instead of semaphores. Proposed code uses task notifications to replace semaphores. With existing task handle and additional "signal" task handle, three semaphores have been replaced. work_queue has been trimmed down to QueueHandle_t because queues already have capacity. Number of malloc calls has been reduced from three to one. In osi_event, is_queued is handled atomically, so event->lock mutex is not needed anymore.

Regarding priority of work_queues in osi_thread_run, work_queues[1] will be executed only when work_queues[0] is empty (idx = 0 in second while loop). Is this correct?

### DIFF
--- a/components/bt/common/osi/thread.c
+++ b/components/bt/common/osi/thread.c
@@ -23,308 +23,209 @@
 #include "freertos/queue.h"
 #include "osi/semaphore.h"
 #include "osi/thread.h"
-#include "osi/mutex.h"
+#include "stdatomic.h"
 
-struct work_item {
+typedef struct work_item_s {
     osi_thread_func_t func;
     void *context;
-};
+} work_item_t;
 
-struct work_queue {
-    QueueHandle_t queue;
-    size_t capacity;
-};
-
-struct osi_thread {
-  TaskHandle_t thread_handle;           /*!< Store the thread object */
-  int  thread_id;                       /*!< May for some OS, such as Linux */
-  bool stop;
-  uint8_t work_queue_num;               /*!< Work queue number */
-  struct work_queue **work_queues;      /*!< Point to queue array, and the priority inverse array index */
-  osi_sem_t work_sem;
-  osi_sem_t stop_sem;
-};
-
-struct osi_thread_start_arg {
-  osi_thread_t *thread;
-  osi_sem_t start_sem;
-  int error;
+struct osi_thread_s {
+    TaskHandle_t task;              /*!< Store the task object */
+    atomic_uintptr_t signal;        /*!< task thread signal */
+    uint8_t work_queue_num;         /*!< Number of work queues */
+    QueueHandle_t work_queues[];    /*!< variable length queue array */
 };
 
 struct osi_event {
-    struct work_item item;
-    osi_mutex_t lock;
-    uint16_t is_queued;
-    uint16_t queue_idx;
     osi_thread_t *thread;
+    work_item_t item;
+    atomic_bool is_queued;
+    uint8_t queue_idx;
 };
 
 static const size_t DEFAULT_WORK_QUEUE_CAPACITY = 100;
+#define DEFAULT_JOIN_TIMEOUT_MS         1000U
 
-static struct work_queue *osi_work_queue_create(size_t capacity)
+static QueueHandle_t osi_work_queue_create(size_t capacity)
 {
-    if (capacity == 0) {
-        return NULL;
-    }
-
-    struct work_queue *wq = (struct work_queue *)osi_malloc(sizeof(struct work_queue));
-    if (wq != NULL) {
-        wq->queue = xQueueCreate(capacity, sizeof(struct work_item));
-        if (wq->queue != 0) {
-            wq->capacity = capacity;
-            return wq;
-        } else {
-            osi_free(wq);
-        }
-    }
-
-    return NULL;
+    return xQueueCreate(capacity, sizeof(work_item_t));
 }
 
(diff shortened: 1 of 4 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/bt/common/osi/thread.c
```diff
@@ -23,389 +23,289 @@
...
+    work_item_t item;
+    uint8_t idx;
+
+    OSI_TRACE_DEBUG("task %p started", thread->task);
+    // signal task that we are running and clear value
+    osi_thread_task_give((TaskHandle_t)atomic_exchange(&thread->signal, 0));
+
+    do {
+        // wait to be notified through thread->task
+        osi_thread_task_take(OSI_SEM_MAX_TIMEOUT);
```
wmy-espressif, line 115: `osi_thread_task_take` immediately follows `osi_thread_task_give`, it is probable that the task calling `osi_thread_create` cannot get notified and thus getting blocked waiting for successful creation and starting of the task, in case that the task calling osi_thread create has a lower priority.

wmy-espressif (changes requested):

wmy-espressif:
> 
> Regarding priority of work_queues in osi_thread_run, work_queues[1] will be executed only when work_queues[0] is empty (idx = 0 in second while loop). Is this correct?

Yes, work_queue[0] has higher priority than work_queue[1].

wmy-espressif (commented):
Hi @tgotic , thanks for your continuous effort for improving the esp-idf bluetooth stack.

I would like to explain about our strategy and difficulties in accepting some of the changes in these days.

- In espressif we are concerntrating on an internal project and there is a lack of engineers to review and test the PRs for Classic Bluetooth part. We will accept PRs that are important bugfixes or major performance improvements. Some PRs that include changes that needs comprehensive tests, or evaluated with a certain level of risks, may not be accepted, or can take a relatively long review, evaluation and test period. Other PRs with some code refactorings, minor improvements, may or may not be accepted according to the available enginner resources.

Regarding this PR, I have the following comments:
1. Code cleaning and changes on the APIs in thread.h is OK and will be accepted

2. The use of atomic operations for is_queued is clever, and will be accepted

3. The changes made on the internal APIs for work_queue is not favorable, according to the design we would like to make osi_work_queue an indenpendent components with some abstraction that can be implemented on multiple Operating Systems. We don't want this and similar changes: "struct work_queue * osi_thread_work_queue_create(size_t capacity)` --> "QueueHandle_t osi_work_queue_create(size_t capacity)". 
However, I agree with the changes in implementing `osi_thread_work_queue_len` that uses `uxQueueMessageWaiting`.

4. The replacement in using taskNotification of binary semaphores,  even if can reduce the use of memory cost and use of "malloc", and may introduce a level of performance improvement, needs to be carefully verified and comprehensive tests. I also think that the currently modified code has lose some readability by using the `signal` with multiple resposibilities. So this part of change may not be accepted currently.

components/bt/common/osi/thread.c
tgotic: You are correct. To my understanding of original code, parent task (one that calls `osi_thread_create`) should wait for child task (new one, running `osi_thread_run` ) to start and child task should give `start_sem`/`signal` to parent when ready.

Before calling `osi_thread_create`, parent task sets its task handle to signal.

If parent has higher priority than child, it should create child task, reach `osi_sem_take`/`osi_thread_task_take` before child task starts and block. Scheduler switches to child task, it starts running, gives start_sem/notifies parent task and blocks in `osi_sem_take`/`osi_thread_task_take`. Scheduler switches back to parent task, semaphore/notification is given and it continues.

If parent has lower priority, creates child task and scheduler switches to child task. Child task starts, gives start_sem/signal and blocks. Scheduler switches back to parent task, it calls `osi_sem_take`/`osi_thread_task_take`, doesn't block because semaphore/task is already given and continues.

Is there some other reason for not calling `osi_thread_task_take `immediately after `osi_thread_task_give`? Maybe scheduling problems, priority problems, ...? If yes, how to avoid them?

Maybe I didn't figure out original code correctly or missed something.

tgotic:
Hi @wmy-espressif . Thank you for your explanation. 
Please find new patch regarding comment 3. `work_queue_s` has been restored and I hope that it is now more in line with the design.

components/bt/common/osi/thread.c
wmy-espressif: Hi @tgotic I agree with your analysis and explanation on this issue, I don't think of other reasons, the original design is to make the parent task synchronized with the operations of child task.

wmy-espressif:
> 

Hi @tgotic  thanks for the effort, I will provide my feedback later this week.

wmy-espressif:
Hi @tgotic , your changes on reverting work_queue_s is good to me. Thanks very much. 
What do you think of my previous comment 4?

tgotic:
Hi @wmy-espressif , I agree that replacement of binary semaphores by task notifications requires lot of tests and that takes time which you might not have. 
`signal` role could be split into two task handles to be more in line with the current code. One task handle would be `start`, and it would be replacement for `start_sem`. Other task handle would be `signal` renamed to `stop` and it would take role of `bool stop` and `stop_sem`. Would that be acceptable?

wmy-espressif:
> 

Hi @tgotic , I guess you mean `semaphore handle`, instead of task handle? If yes, I think it is OK, since we do not change much from the orignal implementation and please go ahead.

PR #10226, opened 2022-11-22
[IPC] use task notifications instead of semaphores (IDFGH-8795)

tgotic:
Proposed code uses task notifications to replace semaphores. Semaphore s_ipc_sem is replaced with existing s_ipc_task_handle and s_ipc_ack is converted from semaphore to task handle. s_func, s_func_arg and s_ipc_wait are declared as atomic.

### DIFF
--- a/components/esp_system/esp_ipc.c
+++ b/components/esp_system/esp_ipc.c
@@ -144,8 +144,8 @@ static esp_err_t esp_ipc_call_and_wait(uint32_t cpu_id, esp_ipc_func_t func, voi
         return ESP_ERR_INVALID_STATE;
     }
 
-#ifdef CONFIG_ESP_IPC_USES_CALLERS_PRIORITY
     TaskHandle_t task_handler = xTaskGetCurrentTaskHandle();
+#ifdef CONFIG_ESP_IPC_USES_CALLERS_PRIORITY
     UBaseType_t priority_of_current_task = uxTaskPriorityGet(task_handler);
     UBaseType_t priority_of_running_ipc_task = uxTaskPriorityGet(s_ipc_task_handle[cpu_id]);
     if (priority_of_running_ipc_task < priority_of_current_task) {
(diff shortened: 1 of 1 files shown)


Dazza0:
@tgotic Thanks for the contribution. I just wanted to check if there was any other reasons we're missing regarding this change (other than the increased speed and reduced memory usage from using task notifications).

tgotic:
Only speed and memory. No other reasons.

components/esp_system/esp_ipc.c
```diff
@@ -150,11 +158,12 @@ static esp_err_t esp_ipc_call_and_wait(uint32_t cpu_id, esp_ipc_func_t func, voi
...
 
-    s_func[cpu_id] = func;
-    s_func_arg[cpu_id] = arg;
-    s_ipc_wait[cpu_id] = wait_for;
-    xSemaphoreGive(s_ipc_sem[cpu_id]);
-    xSemaphoreTake(s_ipc_ack[cpu_id], portMAX_DELAY);
+    atomic_store(&s_func[cpu_id], (uintptr_t)func);
+    atomic_store(&s_func_arg[cpu_id], (uintptr_t)arg);
+    atomic_store(&s_ipc_wait[cpu_id], (int)wait_for);
+    atomic_store(&s_ipc_ack[cpu_id], (uintptr_t)task_handler);
```
bugadani, line 164: This `task_handler` may not exist if `CONFIG_ESP_IPC_USES_CALLERS_PRIORITY` is not set

tgotic: Thank you for reporting. Fixed by moving `task_handler` above `CONFIG_ESP_IPC_USES_CALLERS_PRIORITY `.

KonstantinKondrashov:
Hi @tgotic!
Thanks for your contribution to the IPC component. It was accepted with some modifications due to some test failures. Using a notification functionality for an IPC caller seems not a good idea because it can iterfiar with other notifications that the caller can have, so we're still using a semaphore to notify back.
This PR will be closed once it will be published on GitHub.
Thanks.

Issue #9092, opened 2022-06-03
ENC28J60 Driver does not recover after EMI-Burst (IDFGH-7529)


nx518:
## Environment

- Module or chip used:  ESP32-WROOM-32 and ESP32S3WROOM1
- IDF version: master(v5.0-dev-3290-g0d014c42d) and v4.4
- Build System: idf.py
- Compiler version crosstool-NG esp-2021r2 8.4.0
- Operating System:   Linux
- Using an IDE?: No
- Power Supply: external 3.3V

## Problem Description

While doing the conformity tests of our esp32 based product we noticed that the [ENC28J60-Driver](https://github.com/espressif/esp-idf/tree/master/examples/ethernet/enc28j60/components/eth_enc28j60) does not recover from doing the EMI-Burst EN 61000-6-1 tests. 
While the ESP32 overall keeps running fine during the tests, the enc28j60 driver seems to get stuck without notification. Only a warning "tx_ready_sem expired" is being sent once. The device is not pingable over ethernet anymore, only reboot of ESP32 resolves the issue.

### Expected Behavior

Driver is working fine after the EMI-burst is over. Device is still pingable.

### Actual Behavior

Driver never recovers. Device is not pingable. Rebooting the ESP32 resolves the issue.

### Steps to reproduce

1. run [ENC28J60 Example](https://github.com/espressif/esp-idf/tree/master/examples/ethernet/enc28j60) example
2. sudo ping -s 1450 -i 0.01 <ESP-IP>
3. start EMI burst generator
4. stop EMI burst generator
5. ESP32 is still running but enc28j60 driver is stuck

### Code to reproduce this issue
[ENC28J60 Example](https://github.com/espressif/esp-idf/tree/master/examples/ethernet/enc28j60)

## Debug Logs

I (3172) enc28j60: working in 10Mbps
I (3172) enc28j60: working in half duplex
I (3177) enc28j60: Ethernet Link Up
I (3177) enc28j60: Ethernet HW Addr 02:00:00:12:34:56
I (4147) esp_netif_handlers: eth ip: 192.168.1.186, mask: 255.255.255.0, gw: 192.168.1.1
I (4147) enc28j60: Ethernet Got IP Address
I (4150) enc28j60: ~~~~~~~~~~~
I (4153) enc28j60: ETHIP:192.168.1.186
I (4158) enc28j60: ETHMASK:255.255.255.0
I (4162) enc28j60: ETHGW:192.168.1.1
[...]

PR #10237, opened 2022-11-23
[ENC28J60] clear EIE INTIE instead of non-existent EIR INTIE (IDFGH-8806)

nx518:
This fix should improve the enc28j60 driver a bit. #9092


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=61686221762e906f432113975fb1a1cf072aeed1

PR #10292, opened 2022-12-01
fix small bug when logging eth error (IDFGH-8869)

teamplayer3:
I think these two arguments `free_size` and `length` must be switched.


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
sha=93a8510dc5cffe9efd3619d113b7d788a040012e

PR #10304, opened 2022-12-03
Fixed mismatch in printf types (IDFGH-8883)

gregjesl:
Closes #10303

PR #10317, opened 2022-12-06
Update hello_world_main.c to adopt new Format Identifiers (IDFGH-8898)

franz-ms-muc:
use the new Format identifiers, so this Code can be copied into new (ESP-IDF 5.0) and old (prior 5.0) projects.


igrr:
@franz-ms-muc this was recently submitted in https://github.com/espressif/esp-idf/pull/10304, it is in the process of being merged.

PR #10329, opened 2022-12-08
Fix type of cast (IDFGH-8910)

NathanJPhillips:
It's not important to the code but in terms of documentation it's worth getting things like this right.


Alvin1Zhang:
Thanks for your contribution.

mahavirj:
sha=2d047f54399f8e4eb50cd3f50fa9ff4cc70f529f

PR #10331, opened 2022-12-08
Fix bad initiailisation in RMT (IDFGH-8912)

NathanJPhillips:
loop_count field was added conditional on SOC_RMT_SUPPORT_TX_LOOP_COUNT but was never initialised in RMT_DEFAULT_CONFIG_TX. This fixes that.

### DIFF
--- a/components/driver/deprecated/driver/rmt_types_legacy.h
+++ b/components/driver/deprecated/driver/rmt_types_legacy.h
@@ -192,6 +192,27 @@ typedef struct {
  * @brief Default configuration for Tx channel
  *
  */
+#if SOC_RMT_SUPPORT_TX_LOOP_COUNT
+#define RMT_DEFAULT_CONFIG_TX(gpio, channel_id)      \
+    {                                                \
+        .rmt_mode = RMT_MODE_TX,                     \
+        .channel = channel_id,                       \
+        .gpio_num = gpio,                            \
+        .clk_div = 80,                               \
+        .mem_block_num = 1,                          \
+        .flags = 0,                                  \
+        .tx_config = {                               \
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

suda-morris:
@NathanJPhillips Hi, I think the `loop_count` will be initiated to zero by default even if we don't explicitly set it in the initializer. Did you see the `rmt_tx_config_t::loop_count` was set to a garbage value?

chegewara:
> @NathanJPhillips Hi, I think the `loop_count` will be initiated to zero by default even if we don't explicitly set it in the initializer. Did you see the `rmt_tx_config_t::loop_count` was set to a garbage value?

It may happen to initialize with garbage. We got similar situation with I2C once, when new parameter has been added to structure and it caused many issues in arduino. There was simple solution to init like this:
```
struct i2c_t variable = {};
```
which initializes all variables in struct with 0.
Im not saying here is the same situation, but it may happen.

suda-morris:
Correct. If you initialize the `rmt_config_t` in this way:

```c
rmt_config_t tx_config;
tx_config.xxx = yyy;
...
```
Then the uninitialized member can be filled with garbage.
But if you initialize it with the initializer `rmt_config_t tx_config = RMT_DEFAULT_CONFIG_TX();`, there shouldn't be such an issue.

suda-morris:
sha=2a6e8f81fd1adb5fdee588e60b1b6fe650452a1e

suda-morris:
merged in https://github.com/espressif/esp-idf/commit/6a7e8581e9816dfe42d8414a944e47ba0c921e67

PR #10337, opened 2022-12-09
[BT] list fix (IDFGH-8921)

tgotic:
Fix use after free in `list_remove()` and `list_delete()`. 
`list->head` / `node` are freed in `list_free_node()` / `list_delete_node()` and used after that to check if they are equal to `list->tail`.

Just a notice that hypothetical change in `fixed_queue_new()` form `list_new(NULL)` to `list_new(osi_free)` would lead to use after free of data returned by `fixed_queue_dequeue()` and `fixed_queue_try_remove_from_queue()`. In that case `list_remove(`) should be replaced with `list_delete()` in `fixed_queue_dequeue()` and `fixed_queue_try_remove_from_queue()`.

### DIFF
--- a/components/bt/common/osi/list.c
+++ b/components/bt/common/osi/list.c
@@ -181,20 +181,19 @@ bool list_remove(list_t *list, void *data)
     }
 
     if (list->head->data == data) {
-        list_node_t *next = list_free_node(list, list->head);
         if (list->tail == list->head) {
-            list->tail = next;
+            list->tail = NULL;
         }
-        list->head = next;
+        list->head = list_free_node(list, list->head);
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

lennyJL:
Hi @tgotic, we reviewed your PR, and we think this PR is not very necessary. Reason is:

Although we free the node pointed to by `list->head`, we don't **dereference** `list->head`. We only access `list->head` itself. Therefore it cannot be called `use after free`.

If you still think this PR is necessary, feel free to correct me. Or, we will close this PR.

Finally, thanks for your contribution!

tgotic:
Hi @lennyJL. I agree that this PR is not very necessary.

Alvin1Zhang:
Thanks for contribution again, feel free to reopen.

PR #10338, opened 2022-12-09
[BT] btc_spp_write mem leak

tgotic:
Fix potential memory leak in `btc_spp_write()`.
`btc_spp_arg_deep_free()` doesn't free `write.p_data` allocated in `btc_spp_arg_deep_copy()`. 
Cases of memory leak:
- btc_spp_write() fails
- fixed_queue_enqueue() fails 

Memory will be freed if `BTA_JvRfcommWrite()` fails when connection is closed (BTA_JV_RFCOMM_CLOSE_EVT).
If `BTA_JvRfcommWrite()` fails, it would be better to pull data from queue and free it in `btc_spp_arg_deep_free()` when `btc_spp_write()` returns.

### DIFF
--- a/components/bt/host/bluedroid/bta/jv/bta_jv_api.c
+++ b/components/bt/host/bluedroid/bta/jv/bta_jv_api.c
@@ -1156,8 +1156,9 @@ tBTA_JV_STATUS BTA_JvRfcommWrite(UINT32 handle, UINT32 req_id, int len, UINT8 *p
         p_msg->p_data = p_data;
         p_msg->len = len;
         APPL_TRACE_API( "write ok");
-        bta_sys_sendmsg(p_msg);
-        status = BTA_JV_SUCCESS;
+        if (bta_sys_sendmsg(p_msg)) {
+            status = BTA_JV_SUCCESS;
+        }
     }
     return (status);
 }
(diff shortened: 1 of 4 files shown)


Alvin1Zhang:
Thanks for your contribution.

components/bt/host/bluedroid/btc/profile/std/spp/btc_spp.c
```diff
@@ -895,7 +895,12 @@ static void btc_spp_write(btc_spp_args_t *arg)
             }
         } else {
             if (fixed_queue_enqueue(slot->tx.queue, arg->write.p_data, 0)) {
-                BTA_JvRfcommWrite(arg->write.handle, slot->id, arg->write.len, arg->write.p_data);
+                if (BTA_JvRfcommWrite(arg->write.handle, slot->id, arg->write.len, arg->write.p_data) == BTA_JV_SUCCESS) {
```
xiongweichao, line 898: I think other places that call `BTA_JvRfcommWrite` should also be dealt with accordingly.

components/bt/host/bluedroid/bta/jv/bta_jv_api.c
```diff
@@ -1156,8 +1156,9 @@ tBTA_JV_STATUS BTA_JvRfcommWrite(UINT32 handle, UINT32 req_id, int len, UINT8 *p
         p_msg->p_data = p_data;
         p_msg->len = len;
         APPL_TRACE_API( "write ok");
```
xiongweichao, line 1158: Maybe it can be moved to line 1160.

xiongweichao:
Hi @tgotic , thanks for your contribution, are you still following up this PR?

Issue #10335, opened 2022-12-09
[Flash Encryption][v4.4.3][ESP32-S3] Encryption gets stuck when virtual fuses are enabled? (IDFGH-8917)


chipweinberger:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.3-215-g9f3f82f54b

### Operating System used.

Windows

### How did you build your project?

Command line with Make

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32-S3 Dev Kit C

### Power Supply used.

USB

### What is the expected behavior?

Flash gets encrypted on first boot.

### What is the actual behavior?

We get stuck after this log:

`I (50828) boot: Disabling RNG early entropy source...`

### Steps to reproduce.

1. I think it might have to do with virtual fuses being enabled
2. burn host generated AES-256 (512) key onto device with espsecure.py
3. I enable flash encryption (Release Mode) in KConfig
4. hit 'build / flash / monitor' from Visual Studio Code

[sdkconfig.txt](https://github.com/espressif/esp-idf/files/10191884/sdkconfig.txt)

### Debug Logs.

```plain
export IDF_PATH=/Volumes/User/MBP-Google-Drive/jamcorder/firmware/esp-idf
/Volumes/User/MBP-Google-Drive/jamcorder/firmware/.espressif/python_env/idf4.4_py3.9_env/bin/python /Volumes/User/MBP-Google-Drive/jamcorder/firmware/esp-idf/tools/idf_monitor.py -p /dev/cu.SLAB_USBto%                                                                               
charlesweinberger@Chips-Macbook-Pro-2 jamcorder_app % export IDF_PATH=/Volumes/User/MBP-Google-Drive/jamcorder/firmware/esp-idf
charlesweinberger@Chips-Macbook-Pro-2 jamcorder_app % /Volumes/User/MBP-Google-Drive/jamcorder/firmware/.espressif/python_env/idf4.4_py3.9_e
nv/bin/python /Volumes/User/MBP-Google-Drive/jamcorder/firmware/esp-idf/tools/idf_monitor.py -p /dev/cu.SLAB_USBtoUART -b 460800 --toolchain
[...]

PR #10383, opened 2022-12-16
[Flash Enc] better logs for virtual efuses

chipweinberger:
related issue: https://github.com/espressif/esp-idf/issues/10335

### DIFF
--- a/components/bootloader_support/src/flash_encryption/flash_encrypt.c
+++ b/components/bootloader_support/src/flash_encryption/flash_encrypt.c
@@ -111,7 +111,7 @@ esp_err_t esp_flash_encrypt_check_and_update(void)
         err = esp_flash_encrypt_enable();
         if (err != ESP_OK) {
             ESP_LOGE(TAG, "Enabling of Flash encryption failed (%d)", err);
-            return err;f
+            return err;
         }
 #else
         ESP_LOGE(TAG, "flash encryption is not enabled, and SECURE_FLASH_REQUIRE_ALREADY_ENABLED "
@@ -311,10 +311,10 @@ esp_err_t esp_flash_encrypt_enable(void)
 #endif
     }
 
+    ESP_LOGI(TAG, "Flash encryption completed");
+
 #if CONFIG_EFUSE_VIRTUAL
     ESP_LOGW(TAG, "Flash encryption not really completed. Must disable virtual efuses");
-#else
-    ESP_LOGI(TAG, "Flash encryption completed");
 #endif
 
     return err;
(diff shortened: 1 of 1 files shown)


chipweinberger:
@mahavirj

components/bootloader_support/src/flash_encryption/flash_encrypt.c
```diff
@@ -311,7 +311,11 @@ esp_err_t esp_flash_encrypt_enable(void)
 #endif
     }
 
+#if CONFIG_EFUSE_VIRTUAL
+    ESP_LOGW(TAG, "Flash encryption not really completed. Must disable virtual efuses");
```
mahavirj, line 315: This will probably fail the tests cases from efuse example:

https://github.com/espressif/esp-idf/blob/edd815af2e7d541b25ff3a74c59a7fe0612df42d/examples/system/efuse/pytest_system_efuse_example.py#L243

components/bootloader_support/src/flash_encryption/flash_encrypt.c
```diff
@@ -111,7 +111,7 @@ esp_err_t esp_flash_encrypt_check_and_update(void)
         err = esp_flash_encrypt_enable();
         if (err != ESP_OK) {
             ESP_LOGE(TAG, "Enabling of Flash encryption failed (%d)", err);
-            return err;
+            return err;f
```
mahavirj, line 114: This can be removed?

chipweinberger: oops. thanks

components/bootloader_support/src/flash_encryption/flash_encrypt.c
chipweinberger: Solution: just log both messages?
```
    ESP_LOGI(TAG, "Flash encryption completed');

#if CONFIG_EFUSE_VIRTUAL
    ESP_LOGW(TAG, "Flash encryption not really completed. Must disable virtual efuses");
#endif
```

mahavirj: I think the clean solution would be to now allow `CONFIG_EFUSE_VIRTUAL` for release mode of the security features through Kconfig dependency itself. But we have some test configs that depend on this scenario, we may have to refactor those. I will check it further.

chipweinberger: I think the both logs solution is actually better. It shows that the normal operations have "completed"

PR #10469, opened 2022-12-31
Reference code fixed. (IDFGH-9066)

gasaimaple:
Fixed a small code bug in reference document.

PS: I have no idea why this goes wrong when the english version is correct.

### DIFF
--- a/docs/zh_CN/api-reference/protocols/esp_http_server.rst
+++ b/docs/zh_CN/api-reference/protocols/esp_http_server.rst
@@ -22,7 +22,7 @@ HTTP Server 组件提供了在 ESP32 上运行轻量级 Web 服务器的功能�
         esp_err_t get_handler(httpd_req_t *req)
         {
             /* 发送回简单的响应数据包 */
-            const char[] resp = "URI GET Response";
+            const char resp[] = "URI GET Response";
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

docs/zh_CN/api-reference/protocols/esp_http_server.rst
```diff
@@ -22,7 +22,7 @@ HTTP Server 组件提供了在 ESP32 上运行轻量级 Web 服务器的功能
         esp_err_t get_handler(httpd_req_t *req)
         {
             /* 发送回简单的响应数据包 */
-            const char[] resp = "URI GET Response";
+            const char resp[] = "URI GET Response";
```
suda-morris, line 25: Would you please fix Line58 as well, it has the same issue.

mahavirj:
> PS: I have no idea why this goes wrong when the english version is correct.

It was fixed in https://github.com/espressif/esp-idf/pull/3940, but seems that we missed out on propagating the fix to Chinese version of the docs.

hmalpani:
SHA=27fc115b98127094790067c82abb0a3237b9f97a

mahavirj:
Merged with https://github.com/espressif/esp-idf/commit/ed1b2ba60e033c58131d4d2e2f554533a5e1527b

PR #10133, opened 2022-11-08
[Heap Trace Standalone] fix terrible Leaks perf on large records by using doubly linked list (IDFGH-8690)

chipweinberger:
This augments: https://github.com/espressif/esp-idf/pull/10127 (**Edit:** this PR will need to be rebased onto https://github.com/espressif/esp-idf/pull/10141)

**Leaks:** The original implementation `memmove()`'s the entire record buffer every time a record is freed. This is obviously terrible for perf.

**My ESP32-S3 was not useable after ~300 records in SPI ram.** 

With this change, I can keep `Heap Trace Standalone` enabled without a noticeable impact on perf, tested up to 2540 records!

PR #10478, opened 2023-01-03
[Heap Trace Standalone] fix terrible Leaks perf on large records by using doubly linked list (rebased) (IDFGH-9075)

chipweinberger:
❌ This PR has been rejected in favor of a new PR: https://github.com/espressif/esp-idf/pull/10521 ❌❌

**Original PR:** https://github.com/espressif/esp-idf/pull/10133 (rejected) 

**Leaks:** The original implementation `memmove()`'s the entire record buffer every time a record is freed. This is obviously terrible for perf.

My ESP32-S3 was not useable after ~300 records in SPI ram.

Using a linked lists improves _O(N)_ to _O(1)_ for leaks mode record removal, and makes it reasonable to record all _still-alive_ allocations *since boot* using standalone mode and SPI ram.

With this change, I can keep Heap Trace Standalone enabled without a noticeable impact on perf, tested up to 2540 records!

**My thoughts:** yes, this is more complicated code than before. But it is self contained, reasonably simple, and enables profoundly more comprehensive memory debugging. For example, now it is reasonable to keep this mode enabled during debug or release, and automatically print all current allocations if a low memory scenario is encountered. This is particularly useful for slow growing hard to reproduce leaks where keeping a PC attached for days or weeks is unreasonable.

**Future work:** It would be great to combine this with a simple hash lookup table (address -> record), so leaks mode is completely O(1), near zero cost.

### DIFF
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -449,7 +449,7 @@ static IRAM_ATTR void linked_list_remove(records_t* rs, heap_trace_record_t* rRe
 // pop record from unused list
 static IRAM_ATTR heap_trace_record_t* linked_list_pop_unused(const records_t* rs)
 {
-    if (rs->count >= rs->capacity){
+    if (rs->count >= rs->capacity) {
         return NULL;
     }
 
@@ -461,14 +461,18 @@ static IRAM_ATTR heap_trace_record_t* linked_list_pop_unused(const records_t* rs
     assert(pop->address == NULL);
     assert(pop->size == NULL);
 
+    // update next unused record
     heap_trace_record_t* next = pop->next;
-
-    // update unused list
     if (next) {
         next->prev = NULL;
-        rs->unused = next;
-    } else {
-        rs->unused = NULL;
+    }
+
+    // update unused list
+    rs->unused = next;
+
+    // assert we popped the last unused record
+    if (rs->count == rs->capacity - 1){
+        assert(next == NULL)
     }
 
     return pop;
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -452,15 +452,20 @@ static IRAM_ATTR heap_trace_record_t* linked_list_pop_unused(const records_t* rs
     if (rs->count == rs->capacity){
         return NULL;
     }
+
     heap_trace_record_t* pop = rs->unused;
     assert(pop != NULL);
     assert(pop->address == NULL);
     assert(pop->size == NULL);
 
-    // update linked list
-    rs->unused = pop->next;
-    if (rs->unused != NULL) {
-        rs->unused->prev = NULL;
+    heap_trace_record_t* next = pop->next;
+
+    // update unused list
+    if (next) {
+        next->prev = NULL;
+        rs->unused = next;
+    } else {
+        rs->unused = NULL;
     }
 
     return pop;
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -79,6 +79,14 @@ esp_err_t heap_trace_init_standalone(heap_trace_record_t *record_buffer, size_t 
         return ESP_ERR_INVALID_STATE;
     }
 
+    if (num_records == 0) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    if (record_buffer == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
     records.buffer = record_buffer;
     records.capacity = num_records;
     memset(records.buffer, 0, num_records * sizeof(heap_trace_record_t));
@@ -91,7 +99,6 @@ esp_err_t heap_trace_start(heap_trace_mode_t mode_param)
     if (records.buffer == NULL || records.capacity == 0) {
         return ESP_ERR_INVALID_STATE;
     }
-
 
     portENTER_CRITICAL(&trace_mux);
 
@@ -465,17 +472,20 @@ static IRAM_ATTR heap_trace_record_t* linked_list_pop_unused(const records_t* rs
 
     // update next unused record
     heap_trace_record_t* next = pop->next;
-    if (next) {
+
+    // popped last item, next should be NULL
+    if (rs->count == rs->capacity - 1) {
+        assert(next == NULL);
+    }
+
+    // next should be non-null
+    if (rs->count < rs->capacity - 1) {
+        assert(next != NULL);
         next->prev = NULL;
     }
 
     // update unused list
     rs->unused = next;
-
-    // assert we popped the last unused record
-    if (rs->count == rs->capacity - 1) {
-        assert(next == NULL)
-    }
 
     return pop;
 }
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -145,7 +145,9 @@ esp_err_t heap_trace_get(size_t index, heap_trace_record_t *rOut)
     portENTER_CRITICAL(&trace_mux);
 
     if (index >= records.count) {
+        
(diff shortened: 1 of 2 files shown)


chipweinberger:
@SoucheSouche

This is just the initial refactor. It *should* work if I did the refactor correct =), but I have yet to test it.

components/heap/heap_trace_standalone.c
```diff
@@ -167,52 +199,71 @@ void heap_trace_dump_caps(const uint32_t caps) {
 
 static void heap_trace_dump_base(bool internal_ram, bool psram)
 {
+    // In order to keep the linked list valid,
+    // we must stop tracing
+    bool was_tracing = false;
+    if (tracing) {
```
atanisoft, line 205: perhaps:
```
bool was_tracing = tracing;
if (was_tracing) {
  tracing = false;
}
```
later the code can simply assign `tracing = was_tracing` and avoid another if check.

chipweinberger: ya thats cleaner!

chipweinberger: On second look, since I already grab the mutex, `portENTER_CRITICAL(&trace_mux);`, I don't need to stop tracing. And stopping tracing means some records will be skipped.

It will cause all other code to stop executing for a long time though. Hard to say it that's "good". But I think it's preferable since the user can always stop tracing themself.

components/heap/heap_trace_standalone.c
```diff
@@ -304,46 +345,187 @@ static IRAM_ATTR void record_free(void *p, void **callers)
...
+// Append a record to the end of the linked list.
+// This deep copies rAppend into the linked list. 
+static IRAM_ATTR bool linked_list_append_copy(records_t* rs, const heap_trace_record_t *rAppend)
+{
+    if (rs->count < rs->capacity) {
+
+        // copy record into the next available slot
+        heap_trace_record_t* rAvailable = linked_list_next_available(r);
+        linked_list_copy(rAvailable, rAppend);
+
```
chipweinberger, line 489: I should add a comment here why we don't need to update the linked list connectivity.

reason: the linked list is already fully connected. It's just that some of its nodes are "empty". So we just copy the data into the last node.

SoucheSouche:
Hi @chipweinberger,
Thanks for this PR. I will start looking into it today.

components/heap/heap_trace_standalone.c
chipweinberger: updated.

chipweinberger:
@SoucheSouche 

thanks!

I did just think of a small improvement for code clarity. (I don't plan to implement it for a ~week or more, so feel free to implement it)

Currently I just have these fields:

```
typedef struct {
    heap_trace_record_t *buffer;
    heap_trace_record_t* first;
    heap_trace_record_t* last;
    .....
} records_t;
```

It might be more clear to add another field: `heap_trace_record_t* unused;`

All the records would start in the `unused` linked list, and then moved to the `first/last` linked list when allocated.

It would be nearly the same code, but just easier to follow.

chipweinberger:
I've just noticed ESP-IDF already has a doubly linked list implementation: 

https://github.com/husarnet/esp-idf/blob/master/components/esp32/include/rom/queue.h

With some small changes to the API (namely, we maybe should allocate the buffer ourself, and not have it passed by the user), we could change this code to use `LIST_HEAD`, etc.

I think a custom linked list implementation is fine here. But just something to consider.

components/heap/heap_trace_standalone.c
```diff
@@ -239,47 +291,32 @@ static void heap_trace_dump_base(bool internal_ram, bool psram)
...
+static IRAM_ATTR void record_allocation(const heap_trace_record_t *rAllocation)
 {
-    if (!tracing || record->address == NULL) {
+    if (!tracing || rAllocation->address == NULL) {
         return;
     }
 
     portENTER_CRITICAL(&trace_mux);
 
     if (tracing) {
```
SoucheSouche, line 307: Maybe this is not necessary since it is checked  on line 334. I missed it on the last PR.

components/heap/heap_trace_standalone.c
```diff
@@ -239,47 +291,32 @@ static void heap_trace_dump_base(bool internal_ram, bool psram)
...
-            However, ringbuffer code gets tricky when we remove elements
-            in mid-buffer (for leak trace mode) while trying to keep
-            track of an item count that may overflow.
-            */
-            memmove(&records.buffer[0], &records.buffer[1],
-                sizeof(heap_trace_record_t) * (records.capacity -1));
-
-            records.count--;
+            linked_list_remove(&records, records.first);
         }
```
SoucheSouche, line 316: This check is done again in `linked_list_add`. Maybe you could move this block at the beginning of `linked_list_add` instead of only adding the element if capacity is not reached. In that way `record_allocation` gets more readable and `linked_list_add` never fails and can therefore return void which makes sense since the return value is not checked line 319.

components/heap/heap_trace_standalone.c
```diff
@@ -304,46 +341,228 @@ static IRAM_ATTR void record_free(void *p, void **callers)
...
+
+        // we checked that there is capacity, so this
+        // should never be null.
+        assert(rDest != NULL);
+        
+        // copy allocation data
+        linked_list_copy(rDest, rAppend);
+
+        // update linked list connectivity
+        rDest->next = NULL;
```
SoucheSouche, line 521: It would be better to detach rDest from the unused list in `linked_list_pop_unused` so the function can return an element no longer attached to anything.

components/heap/heap_trace_standalone.c
```diff
@@ -304,46 +341,228 @@ static IRAM_ATTR void record_free(void *p, void **callers)
...
 
-        /* search backwards for the allocation record matching this free */
-        int i = -1;
-        for (i = records.count - 1; i >= 0; i--) {
-            if (records.buffer[i].address == p) {
-                break;
-            }
-        }
+        // search backwards for the allocation record matching this free 
+        heap_trace_record_t* rFound = linked_list_find_address_reverse(&records, p);
```
SoucheSouche, line 345: Is there a reason for walking the list from the last record to the first record? Because if not, then I think the whole concept could be reduced to a singly linked list which would result in a lighter implementation.

chipweinberger: walking backwards is super important for perf

most allocations are quick and short lived.

but when we print, we want to dump print in allocation order, not reverse order.

components/heap/heap_trace_standalone.c
```diff
@@ -46,8 +56,13 @@ typedef struct {
 
 
 // Forward Defines
-static void remove_record(records_t *r, int index);
 static void heap_trace_dump_base(bool internal_ram, bool psram);
+static void linked_list_setup(records_t* rs);
```
igrr, line 60: If possible, please try to avoid adding custom linked list implementation and use the one provided by sys/queue.h, instead. ([Docs](https://www.freebsd.org/cgi/man.cgi?query=queue&sektion=3))

components/heap/heap_trace_standalone.c
SoucheSouche: Yes, it makes sense, thanks for clarifying.

components/heap/heap_trace_standalone.c
chipweinberger: yes I did find that implementation afterwards

I'll give it a shot

components/heap/heap_trace_standalone.c
chipweinberger: we must re-check while holding the mutex. the first check is a fast path.

chipweinberger:
Rejecting this PR in favor of a new PR: https://github.com/espressif/esp-idf/pull/10521

PR #10518, opened 2023-01-10
BT fixes (IDFGH-9117)

tgotic:
Fixes for potential NULL pointer dereferences (wait_entry) and memory leaks.


Alvin1Zhang:
Thanks for your contribution.

PR #10521, opened 2023-01-11
[Heap Trace Standalone] fix terrible Leaks perf on large records by using doubly linked list (try 3) (IDFGH-9121)

chipweinberger:
**Previous PR: https://github.com/espressif/esp-idf/pull/10478 (rejected)**

**Try 3:** This new PR uses sys/queue.h

### Leaks Mode
The original implementation `memmove()`'s the record buffer on every `free()`. This is obviously terrible for perf. My ESP32-S3 was not useable after ~300 records in SPI ram.

With linked-lists, I can keep Heap Trace Standalone enabled without a noticeable impact on perf, tested up to 2540 records!

**My thoughts:** yes, this is more complicated code than before. But it is self contained, reasonably simple, and enables profoundly more comprehensive memory debugging. For example, now it is reasonable perf-wise to keep this mode enabled during debug or release, and automatically print all living allocations if a low memory scenario is encountered. This is particularly useful for slow growing hard to reproduce leaks where keeping a PC attached for days or weeks is unreasonable.

**Future work:** It would be great to combine this with a simple hash lookup table (address -> record), so leaks mode is completely O(1), near zero cost.

### DIFF
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -41,10 +41,10 @@ typedef struct {
     /* Linked list of available record objects */
     heap_trace_record_list_t unused;
 
-    /* capacity of the buffer */
+    /* capacity of 'buffer' */
     size_t capacity;
 
-    /* Count of entries logged in the buffer. */
+    /* Count of entries in 'list' */
     size_t count;
 
     /* During execution, we remember the maximum
@@ -73,6 +73,10 @@ static size_t total_allocations;
 
 /* Actual number of frees logged */
 static size_t total_frees;
+
+/* Used to speed up heap_trace_get */
+static heap_trace_record_t* rGet;
+static size_t rGetIdx;
 
 esp_err_t heap_trace_init_standalone(heap_trace_record_t *record_buffer, size_t num_records)
 {
@@ -160,24 +164,39 @@ esp_err_t heap_trace_get(size_t index, heap_trace_record_t *rOut)
 
     } else {
 
-        // Iterate through through the linked list
+        // Perf: speed up sequential access
+        if (rGet && rGetIdx == index - 1) {
 
-        heap_trace_record_t* rCur = TAILQ_FIRST(&records.list);
+            rGet = TAILQ_NEXT(rGet, tailq);
+            rGetIdx = index;
 
-        for (int i = 0; i < index; i++) {
+        } else {
 
-            // this should not happen, since we already
-            // checked that index >= records.count == false
-            if (rCur == NULL) {
-                result = ESP_ERR_INVALID_STATE;
-                break;
+            // Iterate through through the linked list
+
+            rGet = TAILQ_FIRST(&records.list);
+
+            for (int i = 0; i < index; i++) {
+
+                if (rGet == NULL) {
+                    break;
+                }
+
+                rGet = TAILQ_NEXT(rGet, tailq);
+                rGetIdx = i + 1;
             }
-
-            rCur = TAILQ_NEXT(rCur, tailq);
         }
 
         // copy to destination
-        memcpy(rOut, rCur, sizeof(heap_trace_record_t));
+        if (rGet) {
+            memcpy(rOut, rGet, sizeof(heap_trace_record_t));
+        } else {
+            // this should not happen since we already
+            // checked that index < records.count,
+            // but could be indicative of memory corruption
+            result = ESP_ERR_INVALID_STATE;
+            memset(rOut, 0, sizeof(heap_trace_record_t));
+        }
     }
 
     portEXIT_CRITICAL(&trace_mux);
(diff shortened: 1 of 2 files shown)


chipweinberger:
I've given this a quick test, and it works.

chipweinberger:
Pushed a commit - just some function renames.

SoucheSouche:
@chipweinberger I incorporated your latest change to the MR.
Also, I noticed that in `list_setup()` you are adding records to unused list from the head and not from the tail. Is there a reason behind this choice or is it arbitrary?

chipweinberger:
it's arbitrary. all the unused records are identical.

chipweinberger:
@SoucheSouche pushed one more commit.  Speeding up `heap_trace_get` had been on my mind awhile. 

 That should be the last thing I change. Apologies for the multiple changes!

chipweinberger:
@SoucheSouche, can you give overview of the internal merging process in Espressif? Anything I can help with?

What does PR-Sync-Merge mean for instance?

SoucheSouche:
Hi @chipweinberger,
I synced your PR on the GitLab repository of ESP-IDF. When running tests on your feature I realized that some of them were failing so I had to figure out why they failed, and fixed the cause. Now all tests are passing. I am waiting for internal review (for your PR I think 2 other employees familiar with the heap component are expected to review your feature). I will let you know how the review process is going but I am not expecting any major comments on their side.

chipweinberger:
Thanks for your update and contributing to my PR! 

Hopefully the issue was easy to fix!

SoucheSouche:
Yes, I had to modify the way `heap_trace_dump_base()` was implemented because using `printf()` under lock to print fields of `records` was leading to some load access error. I solved the issue by making local copies of the information to be printed under lock and printing outside. It means that the trace mutex is taken and released for every record.

chipweinberger:
Hmm. I believe you must lock the entire time because you can't iterate on a changing list. 

For example, what happens if we:
1. lock a record, copy the record, 'G', then unlock it
2. **separate thread:** 'G->next' is freed & removed from the list
3. we print 'G' using printf
4. we then go to 'G->next', only to find it no longer exists. dump will fail.

SoucheSouche:
Actually we can keep your implementation with locks if we use `esp_rom_printf()` instead of `printf()`. `printf()` relies on interrupt and can't be used under lock, however `esp_rom_printf()` doesn't so we can use it instead and keep `heap_trace_dump_base()`  as you implemented it.

chipweinberger:
quick question - does `esp_rom_printf()` support USB Serial JTAG?

igrr:
I would recommend checking the implementation of `esp_pm_dump_locks(FILE* stream)` as it solves basically the same problem.

Keeping `printf`, and in the future allowing to pass the `FILE* stream` argument might be a good idea, since that allows for more flexibility. For example, in combination of `funopen`, this allows for redirection of the heap dump to an arbitrary location.

igrr:
> and I guess more generally, why does printf not work this mutex?

This is because printf internally will attempt to acquire a mutex of `stdout` stream, which in general might block the task. Blocking operations are not possible when the interrupts are disabled, which happens when you acquire a spinlock.

chipweinberger:
```
#include <sys/lock.h>
static _lock_t s_list_lock;
_lock_acquire(&s_list_lock);
_lock_release(&s_list_lock);
```

So we can use `_lock_t` instead of `portENTER_CRITICAL`.

chipweinberger:
@SoucheSouche No rush, but curious an ETA of when this would be on github master branch.

I have 1 more PR planned, the user can optionally set a hash table buffer, to skip list traversal in leaks mode. 

That makes leaks mode O(1) time, fast enough to be enabled 24/7. When a low ram threshold is hit, a user could automatically log or save all leaks to disk. Enabled 24/7!

SoucheSouche:
Hi @chipweinberger, the PR is still waiting for internal review after I changed the use of `_lock_t` instead of `portENTER_CRITICAL`.

SoucheSouche:
Hi @chipweinberger, this feature was just merged! Thanks again for your contribution!

My initial fix for the dump() function issue with `printf` was kept. `_lock_t` couldn't be used because it could lead to deadlocks in certain situations and replacing `printf` with `snprint` couldn't be done because of the huge amount of stack necessary for this solution to work.

So in the end, `esp_rom_printf` is used together with `portENTER_CRITICAL`. To answer your question, `esp_rom_printf` does seem to support USB Serial JTAG.

chipweinberger:
That's great!!

Thanks for doing the "polish" and testing to get this merged. That can be the hardest part sometimes!

chipweinberger:
This PR was merged into master: https://github.com/espressif/esp-idf/commit/26a5117870d751d12a308d5c72f4c9e3cfdde494

PR #9998, opened 2022-10-18
[Linenoise 2] multi-console support & support for replacing stdin/stdout (IDFGH-8546)

chipweinberger:
❌ **Rejected. This PR has been closed in favor of a new PR: https://github.com/espressif/esp-idf/pull/10526** ❌

Related Issue: https://github.com/espressif/esp-idf/issues/9986

**This is definitely an optional PR. Linenoise is already an easy dependency for users to replace if they want to add this functionality. Still, I wanted to share these changes in hopes that Espressif or others may find them useful.**

**Note:** this PR also incorporates the same EWOULDBLOCK linenoise changes from this other PR: https://github.com/espressif/esp-idf/pull/9983

Features:
- support multiple instances of Linenoise, in order to support multiple consoles
- backwards compatible, no code changes required for linenoise1 users
- *optional* **Linenoise2 API** to support linenoise handles, and replacing stdin/stdout 
- in `commands.c`, support calling `esp_console_run()` on multiple independent threads.

Tested:
- on ESP32-S3
- both smart & dumb linenoise modes
- both stdin/stdout, and /dev/usbserjtag

PR #10526, opened 2023-01-12
[ESP Linenoise] new esp version of linenoise with multiconsole support

chipweinberger:
**Related PR:** https://github.com/espressif/esp-idf/pull/9998

Adds new "esp_linenoise_x" functions. 

**Features:**

- support multiple instances of Linenoise, in order to support multiple consoles
- backwards compatible, no code changes required for original linenoise users
- in commands.c, support calling esp_console_run() on multiple independent threads.

**Tested:**
- on ESP32-S3
- basic console example
- 'edit' linenoise mode
- uart


KaeLL:
Completely unrelated question: why do you often create new PRs instead of updating the ones you already opened? Isn't tracking progress one of the features of PRs?

chipweinberger:
Its only if the old PR needs a huge rewrite. 

There's a few reasons:

1. I want to be able to refer back to the old implementation. I need to create a new branch
2. clarity. Multiple PRs is easier to follow than 3 rewrites and dozens of outdated comments.

components/console/commands.c
```diff
@@ -184,27 +169,32 @@ static const cmd_item_t *find_command_by_name(const char *name)
 
 esp_err_t esp_console_run(const char *cmdline, int *cmd_ret)
 {
-    if (s_tmp_line_buf == NULL) {
-        return ESP_ERR_INVALID_STATE;
```
igrr, line 188: Probably still need to return ESP_ERR_INVALID_STATE if esp_console_init wasn't called?

components/console/esp_console.h
```diff
@@ -12,9 +12,7 @@ extern "C" {
 #include <stddef.h>
 #include "sdkconfig.h"
 #include "esp_err.h"
-
-// Forward declaration. Definition in linenoise/linenoise.h.
-typedef struct linenoiseCompletions linenoiseCompletions;
```
igrr, line 17: Could we keep this as a forward declaration instead of pulling in the dependency on esp_linenoise.h header?

components/console/esp_console.h
```diff
@@ -244,9 +242,9 @@ size_t esp_console_split_argv(char *line, char **argv, size_t argv_size);
  *   linenoiseSetCompletionCallback(&esp_console_get_completion);
  *
  * @param buf the string typed by the user
- * @param lc linenoiseCompletions to be filled in
+ * @param lc esp_linenoise_completions_t to be filled in
  */
-void esp_console_get_completion(const char *buf, linenoiseCompletions *lc);
+void esp_console_get_completion(const char *buf, esp_linenoise_completions_t *lc);
```
igrr, line 247: Just to check: with this change, does the `examples/system/console/advanced` example from IDF v5.0 compile okay with this brach?

(I assume the answer is "yes" since you didn't have to modify the example. Is that due to that `#include "esp_linenoise.h" above which you have added?)

components/console/esp_linenoise.c
```diff
@@ -0,0 +1,1894 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
```
igrr, line 4: To reflect the copyrights and the actual license below, suggest:

```suggestion
 * SPDX-FileCopyrightText: 2010-2016 Salvatore Sanfilippo <antirez at gmail dot com>
 * SPDX-FileCopyrightText: 2010-2013 Pieter Noordhuis <pcnoordhuis at gmail dot com>
 * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
 *
 * SPDX-License-Identifier: BSD-2-Clause
```

components/console/esp_linenoise.c
```diff
@@ -0,0 +1,1894 @@
...
+        .history_len = 0, \
+        .history = NULL, \
+        .allow_empty = true, \
+        .stdin_fp = 0, \
+        .stdout_fp = 0, \
+    },  \
+    .state = {0} \
+}
+
+struct esp_linenoise_t stdio = LINENOISE_DEFAULT;
```
igrr, line 201: is it possible to avoid overriding the standard `stdio` name?

(also, can this variable be `static`, and prefixed with `s_`?)

components/console/esp_linenoise.c
```diff
@@ -0,0 +1,1894 @@
...
+    const int in_fd = fileno(opts->stdin_fp);
+
+    int cols = getColumns(handle);
+    if (cols == -1 && errno == EWOULDBLOCK) {
+        return -1;
+    }
+
+    /* Populate the linenoise state that we pass to functions implementing
+     * specific editing functionalities. */
+    linenoiseState_t l;
```
igrr, line 1037:
```suggestion
    linenoiseState_t l = {};
```

components/console/esp_linenoise.c
```diff
@@ -0,0 +1,1894 @@
...
+//
+// Public Api
+//
+//
+
+esp_linenoise_handle_t esp_linenoise_create()
+{   
+    struct esp_linenoise_t def = LINENOISE_DEFAULT;
+    esp_linenoise_handle_t handle = (esp_linenoise_handle_t) malloc(sizeof(struct esp_linenoise_t));
+    memcpy(handle, &def, sizeof(struct esp_linenoise_t));
```
igrr, line 1430:
```suggestion
    esp_linenoise_handle_t handle = (esp_linenoise_handle_t) malloc(sizeof(struct esp_linenoise_t));
    if (handle == NULL) {
        return NULL;
    }
    *handle = (esp_linenoise_t) LINENOISE_DEFAULT;
```

components/console/esp_linenoise.c
```diff
@@ -0,0 +1,1894 @@
...
+    return handle;
+}
+
+char *esp_linenoise(esp_linenoise_handle_t handle, const char *prompt)
```
igrr, line 1434:
```suggestion
char *esp_linenoise_get_line(esp_linenoise_handle_t handle, const char *prompt)
```
or something like this — we don't have to keep the names 1:1 with the original version, and can make them more self-explanatory.

components/console/esp_linenoise.h
```diff
@@ -0,0 +1,237 @@
...
+esp_linenoise_handle_t esp_linenoise_create();
+
+/**
+ * @brief The main API of the linenoise library.
+ * 
+ * @param prompt the command line prompt to show the user when ready to accept input.
+ * 
+ * @returns the console command the user entered, or NULL if failure.
+ */
+char* esp_linenoise(esp_linenoise_handle_t handle, const char *prompt);
```
igrr, line 46: All other functions return esp_err_t, perhaps we could apply the same treatment to this one?

```suggestion
esp_err_t esp_linenoise_get_line(esp_linenoise_handle_t handle, const char *prompt, char** out_line);
```

components/console/esp_linenoise.h
```diff
@@ -0,0 +1,237 @@
...
+typedef void(esp_linenoise_completion_callback_t)(const char *buf, esp_linenoise_completions_t *lc);
+typedef char*(esp_linenoise_hints_callback_t)(const char *buf, int *color, int *bold);
+typedef void(esp_linenoise_free_hints_callback_t)(void *buf);
+
+/**
+ * @brief Create a new linenoise instance.
+ * 
+ * @returns A new linenoise handle.
+ */
+esp_linenoise_handle_t esp_linenoise_create();
```
igrr, line 37:
```suggestion
esp_linenoise_handle_t esp_linenoise_create(void);
```

components/console/esp_linenoise.h
```diff
@@ -0,0 +1,237 @@
...
+ * @brief Free all history.
+ * 
+ * @returns - ESP_OK on success
+ *          - ESP_INVALID_ARG if the handle is invalid
+ */
+esp_err_t  esp_linenoise_history_free(esp_linenoise_handle_t handle);
+
+
+/**
+ * @brief Linenoise does not necessarily allocate with malloc. Any memory returned by
```
igrr, line 216: This maybe was true for the original library, is this still necessary?

igrr (commented):
Thanks for the massive PR @chipweinberger!

I've taken the first look, the structure looks fine to me. To be honest, I have trouble reviewing esp_linenoise.c since I can't tell which code is the original one and which you have modified. I'll try to open both files side by side and do a more detailed review later.

A few other things:

- Should we replace the usage of the old linenoise APIs in components/console/esp_console_repl.c with the new ones? Same for the examples?
- Should the old APIs be marked as `__attribute__((deprecated, "please use esp_linenoise_xxxx instead"))`? If yes, this should also be mentioned under docs/en/migration-guides/release-5.x/5.1.
- Do the new APIs even need to be called "linenoise"? The name is cool, but frankly is not self-explanatory. Perhaps `esp_readline` or `esp_lineedit`? Not sure about this, just an idea. Once we add a new public API, we can't change its name easily.
- Now that the functions don't use a global context, seems like we can unit test them quite easily? You definitely aren't expected to do this, but we would still need this before the PR can be merged internally.
- Probably need to update the docs:
  - add the new esp_linenoise.h header file to docs/doxygen/Doxyfile
  - remove the mention of the old linenoise functions from docs/en/api-reference/system/console.rst
  - add `.. include-build-file:: inc/esp_linenoise.inc` in the same file to add the generated API reference

  Same as with the unit tests, you can leave the documentation changes to us, if you prefer.


Edit: I see I've missed this comment:

> Trying to push this PR along. It is tested working. It would be great for an Espressif employee to do any necessary "finishing touches", any extra docs, desired code cleanup, small api changes, etc. It's difficult to merge large features as a 3rd party developer.

In that case, please ignore the "few other things" and the review nitpicks I've posted, we'll take over the PR. Thank you for the contribution!

chipweinberger:
> To be honest, I have trouble reviewing esp_linenoise.c since I can't tell which code is the original one and which you have modified.

Yes, let me push a new change that splits this into 2 commits, so that there is a diff. 

> Should we replace the usage of the old linenoise APIs in components/console/esp_console_repl.c with the new ones? Same for the examples?

Eventually! That's a lot for single PR!

> Should the old APIs be marked as __attribute__((deprecated, "please use esp_linenoise_xxxx instead"))? If yes, this should also be mentioned under docs/en/migration-guides/release-5.x/5.1.

My vote would be to deprecate it in 5.2. More specifically, I don't have time to do the above ^^ in the 5.1 release window, but perhaps Espressif has the time!

> Do the new APIs even need to be called "linenoise"? The name is cool, but frankly is not self-explanatory. Perhaps esp_readline or esp_lineedit? Not sure about this, just an idea. Once we add a new public API, we can't change its name easily.

Hmm! My vote is keep the `linenoise` name. I agree it's not clear, but there is documentation value in knowing where the code is based from. That said, I like the `esp_lineedit`proposal as well. 

> Now that the functions don't use a global context, seems like we can unit test them quite easily? You definitely aren't expected to do this, but we would still need this before the PR can be merged internally.

Could they use the '0' handle? I have this written so that handle '0' uses a global context. That might be controversial, but it makes the linenoise wrappers very easy to write.

components/console/esp_console.h
chipweinberger: I've done my testing with the "basic" example. But I think you are correct.

components/console/esp_linenoise.c
chipweinberger: Yes good point. I should have noticed this.

components/console/esp_linenoise.h
chipweinberger: Yes I considered that strongly as well. I would be in favor of that.

components/console/esp_linenoise.h
chipweinberger: I had the same thought. Ideally we should eliminate returning memory that we allocate ourself, IMO.

`esp_err_t esp_linenoise_get_line(esp_linenoise_handle_t handle, const char *prompt, char* out_line, size_t len);`

components/console/esp_linenoise.h
chipweinberger: Preferably have the user pass the buffer. But we would need to handle the linenoise wrapper differently.

`esp_err_t esp_linenoise_get_line(esp_linenoise_handle_t handle, const char *prompt, char* out_line, size_t len);`

chipweinberger:
bump

mickeyl:
This looks great, I'd really like this to land in preparation of better console support. Good work, @chipweinberger!

cmorganBE:
Also asking for a bump on this PR. I'd like to implement a second console on my system, with separate commands (as its used for a separate kind of user), and this is a first step towards making that easier.

SoucheSouche:
@chipweinberger, I have recently created a new component [esp_linenoise](https://github.com/espressif/idf-extra-components/tree/c9fe5dd183187adc3f976a6a2bb72d3b045635eb/esp_linenoise) available through the component registry (https://components.espressif.com/components/espressif/esp_linenoise/versions/1.0.0/readme).

It preserves the API of the linenoise version we were keeping in the esp-idf console component but also provides a new API adding support for multi instances and a way to return from `esp_linenoise_get_line()` (replacement function for `linenoise()`) with the new function `esp_linenoise_abort()`.

It will eventually replace the linenoise version inside the esp-idf console component.

Thank you for the changes suggested in this PR. They were taken into consideration while implementing the new library.

I would suggest closing this PR if you don't see any issues with that.

chipweinberger:
wow amazing! it's so satisfying and flattering to see this landed. thank you @SoucheSouche

Issue #10531, opened 2023-01-12
[v4.4.3][SD Card] sdmmc mount sometimes hits infinite loop in sdmmc_host_clock_update_command()  (IDFGH-9131)


chipweinberger:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.3-215-g9f3f82f54b

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32-S3 Dev Kit C

### Power Supply used.

USB

### What is the expected behavior?

1. esp_vfs_fat_sdmmc_mount() should succeed

### What is the actual behavior?

after esp_restart(), `sdmmc_host_set_card_clk`() hits an infinite loop in `sdmmc_host_start_command()`, ~50% of the time,

See logs for details.

```
static void sdmmc_host_clock_update_command(int slot)
{
    // Clock update command (not a real command; just updates CIU registers)
    sdmmc_hw_cmd_t cmd_val = {
        .card_num = slot,
        .update_clk_reg = 1,
        .wait_complete = 1
    };
    bool repeat = true;
    while(repeat) {
        ESP_LOGI(TAG, "sdmmc_host_start_command()");
        sdmmc_host_start_command(slot, cmd_val, 0);
        while (true) {
            // Sending clock update command to the CIU can generate HLE error.
            // According to the manual, this is okay and we must retry the command.
            if (SDMMC.rintsts.hle) {
                SDMMC.rintsts.hle = 1;
                repeat = true;
                break;
            }
            // When the command is accepted by CIU, start_command bit will be
            // cleared in SDMMC.cmd register.
            if (SDMMC.cmd.start_command == 0) {
                repeat = false;
                break;
            }
        }
    }
}

```

### Steps to reproduce.

1. mount sd card with sdmmc
2. esp_restart() into some other app app
3. mount sd card with sdmmc again
4. infinite loop

### Debug Logs.

```plain
I (5079) pd sd card: Mounting SD card. format_if_mount_failed? NO
D (5079) sdmmc_periph: peripheral version 5342270a, hardware config 03c44c83
[...]

PR #10532, opened 2023-01-12
[SDMMC Mount] fix infinite loop when SD card is not responsive (IDFGH-9132)

chipweinberger:
**Related Issue:** https://github.com/espressif/esp-idf/issues/10531

When the sd card is not responsive, we need to have reasonable timeouts to prevent infinite loops.

### DIFF
--- a/components/driver/sdmmc_host.c
+++ b/components/driver/sdmmc_host.c
@@ -25,6 +25,8 @@
 
 #define SDMMC_EVENT_QUEUE_LENGTH 32
 
+#define SDMMC_TIMEOUT_MS 1000
+
 
 static void sdmmc_isr(void* arg);
 static void sdmmc_host_dma_init(void);
@@ -68,16 +70,22 @@ static void configure_pin_iomux(uint8_t gpio_num);
 
 static esp_err_t sdmmc_host_pullup_en_internal(int slot, int width);
 
-void sdmmc_host_reset(void)
+esp_err_t sdmmc_host_reset(void)
 {
     // Set reset bits
     SDMMC.ctrl.controller_reset = 1;
     SDMMC.ctrl.dma_reset = 1;
     SDMMC.ctrl.fifo_reset = 1;
+
     // Wait for the reset bits to be cleared by hardware
+    int t0 = esp_timer_get_time();
     while (SDMMC.ctrl.controller_reset || SDMMC.ctrl.fifo_reset || SDMMC.ctrl.dma_reset) {
-        ;
+        if (esp_timer_get_time() - t0 > SDMMC_TIMEOUT_MS) {
+            return ESP_ERR_TIMEOUT;
+        }
     }
+
+    return ESP_OK;
 }
 
 /* We have two clock divider stages:
@@ -142,7 +150,7 @@ static void sdmmc_host_input_clk_disable(void)
     SDMMC.clock.val = 0;
 }
 
-static void sdmmc_host_clock_update_command(int slot)
+static esp_err_t sdmmc_host_clock_update_command(int slot)
 {
     // Clock update command (not a real command; just updates CIU registers)
     sdmmc_hw_cmd_t cmd_val = {
@@ -152,8 +160,20 @@ static void sdmmc_host_clock_update_command(int slot)
     };
     bool repeat = true;
     while(repeat) {
-        sdmmc_host_start_command(slot, cmd_val, 0);
+
+        esp_err_t err = sdmmc_host_start_command(slot, cmd_val, 0);
+        if (err != ESP_OK) {
+            ESP_LOGE(TAG, "sdmmc_host_start_command() failed");
+            return err;
+        }
+
+        int t0 = esp_timer_get_time();
         while (true) {
+
+            if (esp_timer_get_time() - t0 > SDMMC_TIMEOUT_MS) {
+                return ESP_ERR_TIMEOUT;
+            }
+
             // Sending clock update command to the CIU can generate HLE error.
             // According to the manual, this is okay and we must retry the command.
             if (SDMMC.rintsts.hle) {
@@ -169,6 +189,8 @@ static void sdmmc_host_clock_update_command(int slot)
             }
         }
     }
+
+    return ESP_OK;
 }
 
 void sdmmc_host_get_clk_dividers(const uint32_t freq_khz, int *host_div, int *card_div)
@@ -215,7 +237,11 @@ esp_err_t sdmmc_host_set_card_clk(int slot, uint32_t freq_khz)
 
     // Disable clock first
     SDMMC.clkena.cclk_enable &= ~BIT(slot);
-    sdmmc_host_clock_update_command(slot);
+    esp_err_t err = sdmmc_host_clock_update_command(slot);
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "disable clk failed");
+        return err;
+    }
 
     int host_div = 0;   /* clock divider of the host (SDMMC.clock) */
     int card_div = 0;   /* 1/2 of card clock divider (SDMMC.clkdiv) */
@@ -236,12 +262,20 @@ esp_err_t sdmmc_host_set_card_clk(int slot, uint32_t freq_khz)
             break;
     }
     sdmmc_host_set_clk_div(host_div);
-    sdmmc_host_clock_update_command(slot);
+    err = sdmmc_host_clock_update_command(slot);
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "set clk div failed");
+        return err;
+    }
 
     // Re-enable clocks
     SDMMC.clkena.cclk_enable |= BIT(slot);
     SDMMC.clkena.cclk_low_power |= BIT(slot);
-    sdmmc_host_clock_update_command(slot);
+    err = sdmmc_host_clock_update_command(slot);
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "re-enable clk failed");
+        return err;
+    }
 
     // set data timeout
     const uint32_t data_timeout_ms = 100;
@@ -282,8 +316,11 @@ esp_err_t sdmmc_host_start_command(int slot, sdmmc_hw_cmd_t cmd, uint32_t arg) {
     if (cmd.data_expected && cmd.rw && (SDMMC.wrtprt.cards & BIT(slot)) != 0) {
         return ESP_ERR_INVALID_STATE;
     }
+    int t0 = esp_timer_get_time();
     while (SDMMC.cmd.start_command == 1) {
-        ;
+        if (esp_timer_get_time() - t0 > SDMMC_TIMEOUT_MS) {
+            return ESP_ERR_TIMEOUT;
+        }
     }
     SDMMC.cmdarg = arg;
     cmd.card_num = slot;
@@ -305,7 +342,12 @@ esp_err_t sdmmc_host_init(void)
     sdmmc_host_set_clk_div(2);
 
     // Reset
-    sdmmc_host_reset();
+    esp_err_t err = sdmmc_host_reset();
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "sdmmc_host_reset() failed");
+        return err;
+    }
+
     ESP_LOGD(TAG, "peripheral version %"PRIx32", hardware config %08"PRIx32, SDMMC.verid, SDMMC.hcon);
 
     // Clear interrupt status and set interrupt mask to known state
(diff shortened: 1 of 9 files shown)


components/driver/sdmmc_host.c
```diff
@@ -68,16 +68,22 @@ static void configure_pin_iomux(uint8_t gpio_num);
...
     // Set reset bits
     SDMMC.ctrl.controller_reset = 1;
     SDMMC.ctrl.dma_reset = 1;
     SDMMC.ctrl.fifo_reset = 1;
+
     // Wait for the reset bits to be cleared by hardware
+    uint32_t t0 = esp_log_timestamp();
     while (SDMMC.ctrl.controller_reset || SDMMC.ctrl.fifo_reset || SDMMC.ctrl.dma_reset) {
-        ;
+        if (esp_log_timestamp() - t0 > 1000) {
```
atanisoft, line 81: perhaps use `ESP_RETURN_ON_FALSE` (from esp_check.h) for this sort of check?

components/driver/sdmmc_host.c
```diff
@@ -152,8 +158,20 @@ static void sdmmc_host_clock_update_command(int slot)
     };
     bool repeat = true;
     while(repeat) {
-        sdmmc_host_start_command(slot, cmd_val, 0);
+
+        esp_err_t err = sdmmc_host_start_command(slot, cmd_val, 0);
```
atanisoft, line 162: perhaps use `ESP_RETURN_ON_ERROR` for this and similar usages.

components/driver/sdmmc_host.c
```diff
@@ -152,8 +158,20 @@ static void sdmmc_host_clock_update_command(int slot)
...
+        if (err != ESP_OK) {
+            ESP_LOGE(TAG, "sdmmc_host_start_command() failed");
+            return err;
+        }
+
+        uint32_t t0 = esp_log_timestamp();
         while (true) {
+
+            if (esp_log_timestamp() - t0 > 1000) {
```
atanisoft, line 171: might be good to move the `1000` to a constant and document the time unit.

components/driver/sdmmc_host.c
igrr: Additionally, I wouldn't rely on esp_log_timestamp, it is not guaranteed to be monotonic. I would suggest esp_timer_get_time, instead.

components/driver/sdmmc_host.c
```diff
@@ -236,12 +260,20 @@ esp_err_t sdmmc_host_set_card_clk(int slot, uint32_t freq_khz)
...
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "set clk div failed");
+        return err;
+    }
 
     // Re-enable clocks
     SDMMC.clkena.cclk_enable |= BIT(slot);
     SDMMC.clkena.cclk_low_power |= BIT(slot);
     sdmmc_host_clock_update_command(slot);
+    if (err != ESP_OK) {
```
igrr, line 273: assignment to `err` is missing on the line above.

components/driver/sdmmc_host.c
```diff
@@ -236,12 +260,20 @@ esp_err_t sdmmc_host_set_card_clk(int slot, uint32_t freq_khz)
             break;
     }
     sdmmc_host_set_clk_div(host_div);
-    sdmmc_host_clock_update_command(slot);
+    err = sdmmc_host_clock_update_command(slot);
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "set clk div failed");
```
igrr, line 265: Probably not necessary to log here, since it's not something that can happen in practice. (Clock update commands don't depend on the state of the Card Interface Unit, and as such shouldn't be blocked by the bus state.)

(Same comment for other sdmmc_host_clock_update_command error checks.)

igrr (changes requested):

components/driver/sdmmc_host.c
chipweinberger: I looked at `ESP_RETURN_ON_FALSE`, but I noticed it logs, which we don't need here.

components/driver/sdmmc_host.c
chipweinberger: You must be mistaken?  In my issue, we get stuck in 

`I (5109) sdmmc_periph: sdmmc_host_clock_update_command() - SDMMC.clkena.cclk_enable &= ~BIT(slot);`

chipweinberger:
@igrr , updated the PR with the given feedback.

components/driver/sdmmc_host.c
igrr: Sorry, have overlooked that. In that case the fix might be a bit different than ignoring the fact that the clock update didn't happen. We might be missing to reset some part of the peripheral. I'll check it and come back with more details.

chipweinberger: Yes would be great to fix the underlying issue, in addition to having reasonable timeouts.

components/driver/sdmmc_host.c
atanisoft: @chipweinberger Fair point on logging, I'm on the fence if the timeout should be logged or not. It would be really nice if `ESP_RETURN_ON_FALSE` had a way to pass in a log level or a flag to say "don't log this"

components/driver/sdmmc_host.c
chipweinberger: lmk how things go, Ivan. Thanks.

chipweinberger:
@igrr , I audited all SDMMC code for potential infinite loops & added reasonable timeouts. 

Please re-review.

components/driver/sdmmc_transaction.c
```diff
@@ -153,7 +154,11 @@ esp_err_t sdmmc_host_do_transaction(int slot, sdmmc_command_t* cmdinfo)
     cmdinfo->error = ESP_OK;
     sdmmc_req_state_t state = SDMMC_SENDING_CMD;
     sdmmc_event_t unhandled_events = { 0 };
+    int t0 = esp_timer_get_time();
```
igrr, line 157: The return type of esp_timer_get_time is int64_t, suggest using that for the variable type.

(applies to a bunch of other assignments, as well)

components/sdmmc/sdmmc_cmd.c
```diff
@@ -418,11 +419,16 @@ esp_err_t sdmmc_write_sectors_dma(sdmmc_card_t* card, const void* src,
     }
     uint32_t status = 0;
     size_t count = 0;
+    int t0 = esp_timer_get_time();
     /* SD mode: wait for the card to become idle based on R1 status */
     while (!host_is_spi(card) && !(status & MMC_R1_READY_FOR_DATA)) {
-        // TODO: add some timeout here
+        if (esp_timer_get_time() - t0 > SDMMC_WRITE_SECTORS_DMA_TIMEOUT_MS) {
```
igrr, line 425:
```suggestion
        if (esp_timer_get_time() - t0 > SDMMC_READY_FOR_DATA_TIMEOUT_MS) {
```

(the data has already been written at this point, we are waiting for the card to become idle)

components/sdmmc/sdmmc_cmd.c
```diff
@@ -514,10 +520,15 @@ esp_err_t sdmmc_read_sectors_dma(sdmmc_card_t* card, void* dst,
...
+    int t0 = esp_timer_get_time();
     while (!host_is_spi(card) && !(status & MMC_R1_READY_FOR_DATA)) {
-        // TODO: add some timeout here
+        if (esp_timer_get_time() - t0 > SDMMC_READ_SECTORS_DMA_TIMEOUT_MS) {
+            ESP_LOGE(TAG, "read sectors dma - timeout");
+            return ESP_ERR_TIMEOUT;
+        }
         err = sdmmc_send_cmd_send_status(card, &status);
         if (err != ESP_OK) {
+            ESP_LOGE(TAG, "read: status cmd failed");
```
igrr, line 531: suggest
```suggestion
            ESP_LOGE(TAG, "%s: sdmmc_send_cmd_send_status returned 0x%x", __func__, err);
```
(similar to other error messages in this file)

components/driver/sdmmc_host.c
```diff
@@ -152,8 +158,20 @@ static void sdmmc_host_clock_update_command(int slot)
...
+            ESP_LOGE(TAG, "sdmmc_host_start_command() failed");
+            return err;
+        }
+
+        int t0 = esp_timer_get_time();
         while (true) {
+
+            if (esp_timer_get_time() - t0 > SDMMC_HOST_CLOCK_UPDATE_CMD_TIMEOUT_MS) {
+                return ESP_ERR_TIMEOUT;
+            }
```
igrr, line 173: Can be combined with the loop condition,
```c++
while (esp_timer_get_time() - t0 < SDMMC_HOST_CLOCK_UPDATE_CMD_TIMEOUT_MS) {
```

igrr (changes requested):
The timeout constants are defined with `_MS` suffix (e.g. `SDMMC_INIT_WAIT_DATA_READY_TIMEOUT_MS`) but you are comparing them against the return value of `esp_timer_get_time()`, which is in microseconds.

chipweinberger:
Thanks for the review, Ivan!

components/driver/sdmmc_transaction.c
```diff
@@ -153,7 +154,11 @@ esp_err_t sdmmc_host_do_transaction(int slot, sdmmc_command_t* cmdinfo)
...
     while (state != SDMMC_IDLE) {
+        if (esp_timer_get_time() - t0 > SDMMC_HOST_DO_TRANSACTION_TIMEOUT_MS) {
```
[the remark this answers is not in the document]
chipweinberger: Correct me if I'm wrong, but it is not *just* a state machine. It seems like it does active communication with the SD card, which is why I added a timeout.

For example:

```
static esp_err_t process_events(sdmmc_event_t evt, sdmmc_command_t* cmd,
        sdmmc_req_state_t* pstate, sdmmc_event_t* unhandled_events)

{
    while (next_state != state) {
            state = next_state; // note: this might still be the same state as before!
            switch (state) {

         case SDMMC_SENDING_DATA:
                  .....
                 // it looks like this code will loop until some stop condition is set in the peripheral
                break;
}

```

components/driver/sdmmc_host.c
chipweinberger: True, but I think it's more readable to keep all the checks identical. Unnecessary complexity imo. Plus, would make it easy to forget to set the return value.

components/driver/sdmmc_transaction.c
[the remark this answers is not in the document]
chipweinberger: >  If we stay in the same state, then the function exits and we get back to sdmmc_host_wait_for_event

So, could a broken SD card prevent us from exiting? i.e. by returning new events infinitely?

lmk if I should remove this check, or if it is of some value.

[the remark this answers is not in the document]
igrr: Not really, it's more like if the SDMMC peripheral suddenly stops following its documented state machine. I mean, it's okay to not trust the hardware, but if we sprinkle all the drivers for checks that the chip peripheral is not following its spec, it will be a mess to read and the code size will be impacted. So I do think this one is unnecessary.

(I agree with the other ones which are card behavior dependent, though.)

components/driver/sdmmc_host.c
igrr: Fair enough, I just thought someone might ask me to change this during code review. Let's leave it as is, for now.

components/driver/sdmmc_transaction.c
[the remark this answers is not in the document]
chipweinberger: Okay. removed!

igrr (approved):

igrr:
sha=d7324be0348dd9f4990f4762b193427b4f8e05f2

chipweinberger:
@igrr any updates? this seemed like a good simple change

chipweinberger:
bump. Any blockers?

adokitkat:
Hi @chipweinberger. The MR is currently in review stage needing just a few last accepts. Sorry for the wait. Hopefully it will be merged soon.

AxelLin:
> sha=d7324be0348dd9f4990f4762b193427b4f8e05f2

Just curious what does this comment mean?
People usually think this means the PR is merged internally and will be available publicly soon.
But the fact is it may take yet another months to get the fix in github. (And such long time is not a single case).
Several PRs are in "Sync-merge" state for months.

igrr:
This type of a comment is a trigger for creating an internal merge request. This change is in review now, it was left unattended for a while but was recently picked up again.

adokitkat:
The pull request was merged to master branch. Thank you for your contribution :) I will close this on GH as canceled but your commits are merged and will be shown in commit history after the GH master branch syncs with our internal master.

PR #10533, opened 2023-01-12
Update esp_eth_phy_ksz80xx.c (IDFGH-9134)

jamesarm97:
Updated to work with the KSZ8041RNLI chip we are using that is returning a different ID than the ones in the list. Added KSZ80XX_MODEL_NUMBER_13 = 0x13 to the list of known good models and changed any appropriate places in code.


Alvin1Zhang:
Thanks for your contribution.

Alvin1Zhang:
Thanks for your contribution.

components/esp_eth/src/esp_eth_phy_ksz80xx.c
```diff
@@ -20,6 +20,7 @@
 typedef enum
 {
     KSZ80XX_MODEL_NUMBER_11 = 0x11,     // KSZ8041
+    KSZ80XX_MODEL_NUMBER_13 = 0x13,     // KSZ8041
```
suda-morris, line 23: Out of my interest, I read the datasheet quickly but didn't find out why there's going to be a new model ID for KSZ8041 portfolio. 
image
Maybe that's for KSZ8041RNL**I** only? If so, we'd better update the comment here to "KSZ8041RNLI" in case someone else missed the background and thought it was a typo.

kostaond: I am kind of hesitant to approve this change without any confirmation from Microchip. We need to be sure that KSZ8041RNLI really returns different model number. @jamesarm97, since you have the chip and you made that observation, could you please contact Microchip customer support and report them the chip returns different model number than defined in datasheet?

jamesarm97: Yes, we have this chip on a board and it will not work with the code as is. I did some debugging and it is returning 0x13. The data sheet we have is the same you are showing so we were a little confused because we picked this chip knowing it was in the list of PHYs with a driver for the ESP IDF. All other registers read correctly (acts like the x11 and x21 as far as the OP_MODE_REG and offset go.

igrr: See also https://github.com/torvalds/linux/commit/4bd7b5127bd02c12c1cc837a7a0b6ce295eb2505

jamesarm97: Glad it isn't just me. That seems to match the same ID I am reading. I did send a ticket request to microchip to see what their response is and will post when/if I get something.

jamesarm97: Still waiting for Microchip to get back with me. They said they are very busy and will get to it. Frustrated because I am trying to use the arduino framework on ESP32 with this chip so the ESP-IDF release v4.4 also needs to be patched to work with the arduino framework since that is the highest version it works with. That version PHY code does the Model number check a little differently and isn't a clean solution like the master version. Everything I try seems to be failing  because the arduino esp framework seems to want to pull from the github instead of using my local files. If there are any pointers for that it would be appreciated.

jamesarm97: Here is the correspondence with Microchip so far. They are looking into getting the chip doc updated as well:

“Thank you very much for your patience. I apologize for the delay in my response; my colleagues have informed me that we can now proceed with providing technical support in this case.

Thank you for sharing screenshots taken from the Linux driver. I also checked the driver source code before and noticed the same observation you made. For reference, here is the commit referencing the message mentioned in the case description:

https://github.com/torvalds/linux/commit/4bd7b5127bd02c12c1cc837a7a0b6ce295eb2505

Out of curiosity, I went to the ESP-IDF GitHub page and noticed this pull request which I believe you submitted. Is my assumption correct?

https://github.com/espressif/esp-idf/pull/10533

Other than that, I will reach out to my colleagues in the Business Unit about this observation. I will also ask if the datasheet could be updated to reflect this. In the meantime, I believe proceeding with the change should be okay for development since the Linux driver also has the same change.”

mrengineer7777:
RE: Create `package.json`.  Was this accidental?

kostaond:
> 

Yeah, that seems like an accident. Anyway, thanks for the comment, the PR lost out of the radar somehow and this was nice reminder.

@jamesarm97 could you please make improvements of the comments as suggested by @suda-morris and delete `package.json`. Once fixed, I will proceed with PR merging.

package.json
```diff
@@ -0,0 +1,17 @@
+{
```
kostaond, line 1: Please remove this file.

kostaond (approved):

kostaond:
@jamesarm97 could you please process the review comments, squash commits and sign license so we could proceed with this PR?

kostaond:
@jamesarm97, just a gentle reminder. Could you please process the review comments, squash commits and sign license so we could proceed with this PR? Otherwise, I will have to close this PR and process the fix internally. Thank you for understanding.

kostaond:
I have to close the PR due to reasons stated above. The fix will be processed internally. Thank you @jamesarm97 for pointing the issue out.

PR #10536, opened 2023-01-12
[Virtual eFuse Logs] Add more logging for virtual efuses (IDFGH-9137)

chipweinberger:
I've long had these logs in my own code base and found them useful. Figured I should make a PR. 

- prepend logs related to virtual efuses with "Virtual eFuses:"
- more logs for loading eFuses from flash

**Note**: these logs only are compiled when `CONFIG_EFUSE_VIRTUAL` and/or `CONFIG_EFUSE_VIRTUAL_KEEP_IN_FLASH` is enabled.

**Note2**: Like all other virtual fuse logs, we use LOGW because virtual fuses are not the "normal flow of operation".

### DIFF
--- a/components/efuse/src/esp_efuse_utility.c
+++ b/components/efuse/src/esp_efuse_utility.c
@@ -164,10 +164,26 @@ void esp_efuse_utility_reset(void)
     }
 }
 
-// Burn values written to the efuse write registers
+void esp_efuse_utility_debug_dump_pending()
+{
+    for (int num_block = 0; num_block < EFUSE_BLK_MAX; num_block++) {
+        int subblock = 0;
+        for (uint32_t addr_wr_block = range_write_addr_blocks[num_block].start; addr_wr_block <= range_write_addr_blocks[num_block].end; addr_wr_block += 4) {
+            if (virt_blocks[num_block][subblock++] != REG_READ(addr_wr_block)) {
+                esp_efuse_utility_debug_dump_single_block(num_block);
+                break; // one log per block
+            }
+        }
+    }
+}
+
 esp_err_t esp_efuse_utility_burn_efuses(void)
 {
     ++s_burn_counter;
+#ifdef CONFIG_EFUSE_VIRTUAL_LOG_ALL_WRITES
+    ESP_EARLY_LOGW(TAG, "Burn:");
+    esp_efuse_utility_debug_dump_pending();
+#endif
     esp_err_t err = esp_efuse_utility_burn_chip();
     ++s_burn_counter;
     return err;
@@ -193,13 +209,13 @@ void esp_efuse_utility_update_virt_blocks(void)
 #else
     if (1) {
 #endif
-        ESP_EARLY_LOGW(TAG, "Loading virtual efuse blocks from real efuses");
+        ESP_EARLY_LOGW(TAG, "[Virtual] Loading virtual efuse blocks from real efuses");
         for (int num_block = EFUSE_BLK0; num_block < EFUSE_BLK_MAX; num_block++) {
             int subblock = 0;
             for (uint32_t addr_rd_block = range_read_addr_blocks[num_block].start; addr_rd_block <= range_read_addr_blocks[num_block].end; addr_rd_block += 4) {
                 virt_blocks[num_block][subblock++] = REG_READ(addr_rd_block);
             }
-            ESP_EARLY_LOGD(TAG, "virt_blocks[%d] is filled by EFUSE_BLOCK%d", num_block, num_block);
+            ESP_EARLY_LOGD(TAG, "[Virtual] virt_blocks[%d] is filled by EFUSE_BLOCK%d", num_block, num_block);
         }
 #ifdef CONFIG_EFUSE_VIRTUAL_KEEP_IN_FLASH
         esp_efuse_utility_write_efuses_to_flash();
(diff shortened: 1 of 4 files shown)


components/efuse/src/esp_efuse_utility.c
```diff
@@ -421,58 +421,74 @@ void esp_efuse_init_virtual_mode_in_flash(uint32_t offset, uint32_t size)
 void esp_efuse_utility_erase_efuses_in_flash(void)
 {
     if (esp_efuse_flash_offset == 0) {
-        ESP_EARLY_LOGE(TAG, "no efuse partition in partition_table? (Flash is not updated)");
+        ESP_EARLY_LOGE(TAG, "Virtual eFuse: no efuse partition in partition_table? (Flash is not updated)");
```
mahavirj, line 424: I wonder if it would help if we change the `TAG` itself for the virtual EFuse configuration case?

chipweinberger: ya I agree that sounds better i'll update the PR

chipweinberger: updated.

components/efuse/src/esp_efuse_utility.c
```diff
@@ -15,6 +15,8 @@
 
 static const char *TAG = "efuse";
 
+static const char *TAGV = "efuse (virtual)";
```
mahavirj, line 18: I was actually suggesting something like:

```
#if CONFIG_EFUSE_VIRTUAL
static const char *TAG = "efuse (virtual)";
#else
static const char *TAG = "efuse";
#endif
```

CC @KonstantinKondrashov

chipweinberger: I like that suggestion. Updated the PR.

components/efuse/src/esp_efuse_utility.c
```diff
@@ -452,11 +461,29 @@ bool esp_efuse_utility_load_efuses_from_flash(void)
...
+    esp_efuse_utility_debug_dump_blocks();
+
+    ESP_EARLY_LOGW(TAG, "Write complete");
+}
+
 void esp_efuse_utility_write_efuses_to_flash(void)
 {
+    ESP_EARLY_LOGW(TAG, "Writing fuses to flash: 0x%x (offset)", esp_efuse_flash_offset);
+
+    esp_efuse_utility_debug_dump_blocks();
```
KonstantinKondrashov, line 485: At some point, for example in esp_efuse_init_virtual_mode_in_flash(), we will see two dumps, with the same data.
If we need to print dump after burning or at startup we can directly call esp_efuse_utility_debug_dump_blocks, 
So I suggest to do not putting it here, to not be overloaded by the logs.

components/efuse/src/esp_efuse_utility.c
```diff
@@ -452,11 +461,29 @@ bool esp_efuse_utility_load_efuses_from_flash(void)
...
+    memset(virt_blocks, 0, sizeof(virt_blocks));
+
+    esp_efuse_utility_debug_dump_blocks();
+
+    ESP_EARLY_LOGW(TAG, "Write complete");
+}
+
 void esp_efuse_utility_write_efuses_to_flash(void)
 {
+    ESP_EARLY_LOGW(TAG, "Writing fuses to flash: 0x%x (offset)", esp_efuse_flash_offset);
```
KonstantinKondrashov, line 483: it will be better to print from esp_efuse_init_virtual_mode_in_flash().

components/efuse/src/esp_efuse_utility.c
```diff
@@ -473,6 +500,8 @@ void esp_efuse_utility_write_efuses_to_flash(void)
         ESP_EARLY_LOGE(TAG, "secure_version can not be written to flash. err = 0x%x", err);
         abort();
     }
+
+    ESP_EARLY_LOGW(TAG, "Write complete");
```
KonstantinKondrashov, line 504: I think it will overload the logs. If we have an error then it will be catched by the above conditions. I suggest removing it.

components/efuse/src/esp_efuse_utility.c
```diff
@@ -452,11 +461,29 @@ bool esp_efuse_utility_load_efuses_from_flash(void)
...
 
+void write_efuses_to_ram(void)
+{
+    ESP_EARLY_LOGW(TAG, "Writing fuses to ram");
+
+    memset(virt_blocks, 0, sizeof(virt_blocks));
+
+    esp_efuse_utility_debug_dump_blocks();
+
+    ESP_EARLY_LOGW(TAG, "Write complete");
```
KonstantinKondrashov, line 478: I suggest keeping this function quiet. In this func, we just reset efuses. I think it is ok to keep the original approach.

KonstantinKondrashov (commented):
LGTM! Just a few comments. Thanks for your contribution. (I can update it before merging if you agree with that.)

KonstantinKondrashov:
There is another point of view why we can not use these changes - the log component has `esp_log_level_set(tag, log_level)` API that helps to change the log level for a component by TAG. TAG depends on virtual efuse mode. So if we change the TAG for efuse then this functionality will not work need to update the given tag as well.
And one more thing is - we need to update from `efuse` to `efuse (virtual)` in entire the efuse component not just in utility files.

chipweinberger:
okay.

I'll go back to my original PR that just changed a few of the messages to indicate they were related to virtual efuse.

components/efuse/src/esp_efuse_utility.c
chipweinberger: Writing fuses is very, very, rare, (of course, by design...) so I think there is small risk of overloading.

I've hit conditions where we got stuck (I forget why). I would have appreciated this log.

components/efuse/src/esp_efuse_utility.c
chipweinberger: If you are concerned about too many logs (reasonable), perhaps we can put this in a kConfig? a `EFUSE_VIRTUAL_LOG_ALL_WRITES`?

Virtual eFuse is by definition a _debug_ mode. Users _need_ an easy way to see which fuses would be burned, before burning the real fuses. 

Adding a bunch of `esp_efuse_utility_debug_dump_blocks`() calls to your own code is cumbersome.

chipweinberger:
@KonstantinKondrashov , made some changes based on your feedback. Let me know your thoughts!

KonstantinKondrashov:
Hi @chipweinberger!
- I would rather add it to esp_efuse_utility_burn_efuses() instead of into chip-specific files. The functionality that you added is not chip related. I have another idea, see below.
- You can create a new API function that will print range_write_addr_blocks (it contains efuses only for burning). It will be called under your Kconfig option from esp_efuse_utility_burn_efuses() before the burning. For virt mode and for your needs (as I understand), you just want to print diff of blocks, it does not matter that we need to take blocks before and after the burn. See the code below.
- Looks good to add esp_efuse_utility_debug_dump_single_block().

```c
esp_err_t esp_efuse_utility_burn_efuses(void)
{
    ++s_burn_counter;
#if CONFIG_EFUSE_VIRTUAL_LOG_ALL_WRITES
    esp_efuse_utility_debug_dump_prepared_data();
#endif
    esp_err_t err = esp_efuse_utility_burn_chip();
    ++s_burn_counter;
    return err;
}

#if CONFIG_EFUSE_VIRTUAL_LOG_ALL_WRITES
esp_efuse_utility_debug_dump_prepared_data()
{
    for (int num_block = 0; num_block < EFUSE_BLK_MAX; num_block++) {
        int subblock = 0;
        for (uint32_t addr_wr_block = range_write_addr_blocks[num_block].start; addr_wr_block <= range_write_addr_blocks[num_block].end; addr_wr_block += 4) {
            if (virt_blocks[num_block][subblock++] != REG_READ(addr_wr_block)) {
                esp_efuse_utility_debug_dump_prepared_single_block(num_block );
                break;
            }
        }
    }
}
#endif
```

chipweinberger:
@KonstantinKondrashov 

Yes, that looks more reasonable! Implemented!

KonstantinKondrashov:
Thank you for the update. I left some comments, LGTM!

chipweinberger:
@KonstantinKondrashov I think you forgot to 
"submit" the comments!

components/efuse/src/esp_efuse_utility.c
```diff
@@ -420,57 +441,66 @@ void esp_efuse_init_virtual_mode_in_flash(uint32_t offset, uint32_t size)
...
     if (err != ESP_OK) {
-        ESP_EARLY_LOGE(TAG, "Failed to erase flash. err = 0x%x", err);
+        ESP_EARLY_LOGE(TAG, "[Virtual] Failed to erase flash. err = 0x%x", err);
         abort();
     }
 
     err = bootloader_flash_write(esp_efuse_flash_offset, &virt_blocks, sizeof(virt_blocks), false);
     if (err != ESP_OK) {
-        ESP_EARLY_LOGE(TAG, "secure_version can not be written to flash. err = 0x%x", err);
+        ESP_EARLY_LOGE(TAG, "[Virtual] secure_version can not be written to flash. err = 0x%x", err);
```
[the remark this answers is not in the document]
chipweinberger: what does this mean?

components/efuse/src/esp_efuse_utility.c
```diff
@@ -164,10 +164,27 @@ void esp_efuse_utility_reset(void)
...
+        }
+    }
+}
+#endif
+
 esp_err_t esp_efuse_utility_burn_efuses(void)
 {
     ++s_burn_counter;
+#ifdef CONFIG_EFUSE_VIRTUAL_LOG_ALL_WRITES
+    esp_rom_printf("Burn new fuses:\n");
```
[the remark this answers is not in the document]
chipweinberger: are you sure? I see this which makes me think this is an even earlier function?

```
 .text.esp_efuse_utility_debug_dump_blocks
                0x0000000000000000       0x9a esp-idf/efuse/libefuse.a(esp_efuse_utility.c.obj)
```

chipweinberger:
Appreciate your effort as well!

chipweinberger:
@KonstantinKondrashov should we rename EFUSE_VIRTUAL_LOG_ALL_WRITES to EFUSE_LOG_ALL_WRITES ? It looks like it should work even when virtual fuses are not enabled now?


( or EFUSE_LOG_BURNS?)

also maybe the dump_pending function we added would be better named dump_pending_changes?

PR #10554, opened 2023-01-16
Verify w5500 chip properly (IDFGH-9158)

reinismu:
Datasheet says that 0x4 is only value returned. We should verify that we got correct one.
image

Note: Currently trying to make W5500 work using `esp-idf-svc` in rust. SPI returns garbage data, but no clear error message is returned. So would be nice if we failed fast here.

### DIFF
--- a/components/esp_eth/src/esp_eth_mac_w5500.c
+++ b/components/esp_eth/src/esp_eth_mac_w5500.c
@@ -230,8 +230,8 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
     esp_err_t ret = ESP_OK;
     uint8_t version = 0;
     ESP_GOTO_ON_ERROR(w5500_read(emac, W5500_REG_VERSIONR, &version, sizeof(version)), err, TAG, "read VERSIONR failed");
-    // W5500 doesn't have chip ID, we just print the version number instead
-    ESP_LOGI(TAG, "version=%x", version);
+    // W5500 doesn't have chip ID, we check the version number instead
+    ESP_GOTO_ON_FALSE(version == W5500_CHIP_VERSION, ESP_ERR_INVALID_VERSION, err, TAG, "invalid chip version, expected 0x%x, actual 0x%x", W5500_CHIP_VERSION, version);
 
 err:
     return ret;
--- a/components/esp_eth/src/w5500.h
+++ b/components/esp_eth/src/w5500.h
@@ -16,6 +16,8 @@
 #define W5500_BSB_OFFSET  (3)  // Block Select Bits offset
 #define W5500_RWB_OFFSET  (2)  // Read Write Bits offset
 
+#define W5500_CHIP_VERSION (0x4) // Chip version that VERSIONR returns
+
 #define W5500_BSB_COM_REG        (0x00)    // Common Register
 #define W5500_BSB_SOCK_REG(s)    ((s)*4+1) // Socket Register
 #define W5500_BSB_SOCK_TX_BUF(s) ((s)*4+2) // Socket TX Buffer


components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -233,6 +233,8 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
     // W5500 doesn't have chip ID, we just print the version number instead
     ESP_LOGI(TAG, "version=%x", version);
```
kostaond, line 234: We should change the comment and could remove the LOGI.
```suggestion
    // W5500 doesn't have chip ID, we check the version number instead
```

components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -233,6 +233,8 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
...
 
+    ESP_GOTO_ON_FALSE(version == W5500_CHIP_VERSION, ESP_ERR_INVALID_VERSION, err, TAG, "returned invalid VERSIONR. Should return 0x4");
```
kostaond, line 236:
```suggestion
    ESP_GOTO_ON_FALSE(version == W5500_CHIP_VERSION, ESP_ERR_INVALID_VERSION, err, TAG, "invalid chip version, expected 0x%x, actual 0x%x", W5500_CHIP_VERSION, version);
```

components/esp_eth/src/w5500.h
```diff
@@ -16,6 +16,8 @@
 #define W5500_BSB_OFFSET  (3)  // Block Select Bits offset
 #define W5500_RWB_OFFSET  (2)  // Read Write Bits offset
 
+#define W5500_CHIP_VERSION 0x4 // Chip version that VERSIONR returns
```
kostaond, line 19: Please follow the coding style here.
```suggestion
#define W5500_CHIP_VERSION (0x4) // Chip version that VERSIONR returns
```

kostaond (commented):
Thank you for the contribution. I just left few minor comments. Please have a loook.

kostaond:
Thanks for the update. Could you also please squash the commits into one to follow [contribution rules](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/index.html#how-to-contribute)?

suda-morris (approved):

suda-morris:
sha=bb2f346e55f484edc50f68cb26d164d8a79bcf39

espzav:
Hi @reinismu, please, I have the issue with ethernet MAC W5500 on the latest master (after this PR). I have board with this SPI ethernet chip and it is working right. But after updated IDF to latest master, it fails when initilization - read chip ID returns 0x0 instead of 0x4. Please, could you help me, where can be a problem?

reinismu:
Hi @espzav,
Interesting. Datasheet does say that it will always return 0x4. 
I wonder if there are different W5500 chips. What is your w5500 part number?

espzav:
Hi @reinismu, I bought this: https://www.tme.eu/cz/en/details/w5500/ethernet-interfaces-integrated-circuits/wiznet/ (I can try another board later, which was assmbled in JLCPCB)

reinismu:
They link to the same datasheet where it mentions that 0x4 is the returned version.
You could try writing simple program that just reads the version register and see what happens. Maybe commands before corrupt it.

espzav:
Thank you, I will try it and check the reset state.

espzav:
Hi @reinismu,
I tried reset and still returns 0x0. But when I put 100ms delay before ID check, it works!

Issue #9974, opened 2022-10-14
[Usb Serial JTAG] [Deadlock] [Console Deinit] esp_vfs_usb_serial_jtag_use_nonblocking() causes deadlock


chipweinberger:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.2

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### Development Kit.

ESP32-S3

### Power Supply used.

External 5V

### What is the expected behavior?

No deadlock.

### What is the actual behavior?

Deadlock.

### Steps to reproduce.

Related Issue: https://github.com/espressif/esp-idf/issues/9964

I've debugged the cause of the deadlock, hopefully I can explain it well.

1. USB Serial JTAG as the Console, via Menu Config
2. During boot I Initialize a linenoise console
3. I also then initialize a linenoise thread, `while(!should_stop) { char* line = linenoise(prompt); ....}`
4. At this point, everything works fine. No problems. I can use the console.
5. The problems come when I try and deinit the console. I call code that looks like this, copied from `esp_console_repl.c`:

**Deinit Code:**
```
void pd_console_api_deinit(pd_error_t* error)
{
    pd_err_unset(error);

/*
    // I'VE YET FOUND A WAY TO ACHEIVE THIS!
    // wait for console thread to stop
    ESP_LOGI(TAG, "waiting for console thread to exit...");
    should_stop = true;
    while (!stopped) {
        // write empty line to stdin to unblock the console thread
        //write(0, " \r\n",4); 
        //write(1, " \r\n",4); 
        //fputc('\n', stdin);
        //ungetc('\n',stdin);
        delayMillis(500);
    }
*/

    ESP_LOGI(TAG, "console deinit"); 
    esp_err_t err = esp_console_deinit();
    if (err != ESP_OK){
        pd_err_fail(error, TAG, "esp_console_deinit() %s", esp_err_to_name(err));
        return;
    }

#if CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG

    ESP_LOGI(TAG, "usb nonblocking");

[...]

Issue #9964, opened 2022-10-13
[USB Serial JTAG] calling select() & poll() crashes (IDFGH-8509)


chipweinberger:
**v4.4.2
ESP32-S3**

### What is the expected behavior?

poll() and select() should be implemented for USB Serial JTAG. Is there a reason they are not?

relevant code: https://github.com/espressif/esp-idf/blob/master/components/vfs/vfs_usb_serial_jtag.c

I need poll() or select() to determine if bytes are waiting on stdin.

**Use Case:** If the user enters any input during boot, USB Host is disabled and instead the user enters a Usb Serial Console.

### What is the actual behavior?

The code crashes. Null ptr deref trying to execute the non-existent function.

(This was very annoying to debug due to the virtual pointers! At the least we should log + assert😀)

### Steps to reproduce.

1. Enable Usb Serial JTAG as the Primary Console in Menuconfig
2. call poll() or select() on stdin.

Code:

```
int stdin_has_data()
{
    struct pollfd fds[2];
    fds[0].fd = 0; // 0 is stdin
    fds[0].events = POLLIN;
    int timeout_msecs = 0;
    int ret = poll(fds, 1, timeout_msecs);
    if (ret < 0) {
        ESP_LOGE(TAG, "poll() failed");
        return false;
    }

    return (fds[0].revents & POLLIN) != 0;
}
```

### Debug Logs.

```

    //printf("write %p\n", get_vfs_for_index(primary_vfs_index)->vfs.write);
    //printf("open %p\n", get_vfs_for_index(primary_vfs_index)->vfs.open);
    //printf("fstat %p\n", get_vfs_for_index(primary_vfs_index)->vfs.fstat);
    //printf("close %p\n", get_vfs_for_index(primary_vfs_index)->vfs.close);
    //printf("read %p\n", get_vfs_for_index(primary_vfs_index)->vfs.read);
    //printf("fcntl %p\n", get_vfs_for_index(primary_vfs_index)->vfs.fcntl);
    //printf("fsync %p\n", get_vfs_for_index(primary_vfs_index)->vfs.fsync);
    //printf("access %p\n", get_vfs_for_index(primary_vfs_index)->vfs.access);
    //printf("start_select %p\n", get_vfs_for_index(primary_vfs_index)->vfs.start_select);
    //printf("end_select %p\n", get_vfs_for_index(primary_vfs_index)->vfs.end_select);
[...]

PR #10580, opened 2023-01-20
[Console] fix: console deinit always deadlocks - adds ability to unblock linenoise (try 2) (IDFGH-9188)

chipweinberger:
**Previous PR:** https://github.com/espressif/esp-idf/pull/9983

I am re-opening this PR to get more traction on it. 

This solves: https://github.com/espressif/esp-idf/issues/9974

**Motivation:**
- currently, console deinit causes a deadlock
- without deinit, we can't create consoles "as needed"
- for example, USB Serial JTAG console should be destroyed when entering USB host mode
- for example, "web" consoles should be destroyed when the browser window is closed

### DIFF
--- a/components/console/esp_console_repl.c
+++ b/components/console/esp_console_repl.c
@@ -30,6 +30,8 @@ typedef enum {
     CONSOLE_REPL_STATE_DEINIT,
     CONSOLE_REPL_STATE_INIT,
     CONSOLE_REPL_STATE_START,
+    CONSOLE_REPL_STATE_REPL_STOP_REQUESTED,
+    CONSOLE_REPL_STATE_REPL_TASK_ENDED,
 } repl_state_t;
 
 typedef struct {
@@ -409,13 +411,25 @@ static esp_err_t esp_console_repl_uart_delete(esp_console_repl_t *repl)
     esp_err_t ret = ESP_OK;
     esp_console_repl_com_t *repl_com = __containerof(repl, esp_console_repl_com_t, repl_core);
     esp_console_repl_universal_t *uart_repl = __containerof(repl_com, esp_console_repl_universal_t, repl_com);
+
     // check if already de-initialized
     if (repl_com->state == CONSOLE_REPL_STATE_DEINIT) {
         ESP_LOGE(TAG, "already de-initialized");
         ret = ESP_ERR_INVALID_STATE;
         goto _exit;
     }
+
+    // wait for repl thread to stop, if it is running
+    if (repl_com->state == CONSOLE_REPL_STATE_START) {
+        repl_com->state = CONSOLE_REPL_STATE_REPL_STOP_REQUESTED;
+        while (repl_com->state != CONSOLE_REPL_STATE_REPL_TASK_ENDED) {
+            uart_unblock_reads(uart_repl->uart_channel);
+            vTaskDelay(1);
+        }
+    }
+
     repl_com->state = CONSOLE_REPL_STATE_DEINIT;
+
     esp_console_deinit();
     esp_vfs_dev_uart_use_nonblocking(uart_repl->uart_channel);
     uart_driver_delete(uart_repl->uart_channel);
@@ -429,13 +443,21 @@ static esp_err_t esp_console_repl_usb_cdc_delete(esp_console_repl_t *repl)
     esp_err_t ret = ESP_OK;
     esp_console_repl_com_t *repl_com = __containerof(repl, esp_console_repl_com_t, repl_core);
     esp_console_repl_universal_t *cdc_repl = __containerof(repl_com, esp_console_repl_universal_t, repl_com);
+
     // check if already de-initialized
     if (repl_com->state == CONSOLE_REPL_STATE_DEINIT) {
         ESP_LOGE(TAG, "already de-initialized");
         ret = ESP_ERR_INVALID_STATE;
         goto _exit;
     }
+
+    // TODO: wait for repl thread to stop, if it is running.
+    //   Need to implement a USB CDC driver, and a
+    //   corresponding usb_cdc_unblock_reads() function.
+    //   See other esp_console_repl_X_delete() functions for reference.
+
     repl_com->state = CONSOLE_REPL_STATE_DEINIT;
+
     esp_console_deinit();
     free(cdc_repl);
 _exit:
@@ -454,7 +476,18 @@ static esp_err_t esp_console_repl_usb_serial_jtag_delete(esp_console_repl_t *rep
         ret = ESP_ERR_INVALID_STATE;
         goto _exit;
     }
+
+    // wait for repl thread to stop, if it is running
+    if (repl_com->state == CONSOLE_REPL_STATE_START) {
+        repl_com->state = CONSOLE_REPL_STATE_REPL_STOP_REQUESTED;
+        while (repl_com->state != CONSOLE_REPL_STATE_REPL_TASK_ENDED) {
+            usb_serial_jtag_unblock_reads();
+            vTaskDelay(1);
+        }
+    }
+
     repl_com->state = CONSOLE_REPL_STATE_DEINIT;
+
     esp_console_deinit();
     esp_vfs_usb_serial_jtag_use_nonblocking();
     usb_serial_jtag_driver_uninstall();
@@ -535,6 +568,8 @@ static void esp_console_repl_task(void *args)
         /* linenoise allocates line buffer on the heap, so need to free it */
         linenoiseFree(line);
     }
+
+    repl_com->state = CONSOLE_REPL_STATE_REPL_TASK_ENDED;
     ESP_LOGD(TAG, "The End");
     vTaskDelete(NULL);
 }
--- a/components/driver/uart/uart.c
+++ b/components/driver/uart/uart.c
@@ -1762,6 +1762,16 @@ esp_err_t uart_set_loop_back(uart_port_t uart_num, bool loop_back_en)
     return ESP_OK;
 }
 
+esp_err_t uart_unblock_reads(uart_port_t uart_num)
+{
+    if (p_uart_obj[uart_num]->rx_ring_buf == NULL) {
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    vRingbufferUnblockRx(p_uart_obj[uart_num]->rx_ring_buf);
+    return ESP_OK;
+}
+
 void uart_set_always_rx_timeout(uart_port_t uart_num, bool always_rx_timeout)
 {
     uint16_t rx_tout = uart_hal_get_rx_tout_thr(&(uart_context[uart_num].hal));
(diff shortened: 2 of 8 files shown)


components/esp_ringbuf/include/freertos/ringbuf.h
```diff
@@ -496,6 +496,16 @@ void vRingbufferGetInfo(RingbufHandle_t xRingbuffer,
...
 
+/**
+ * @brief   Unblock any read function that is currently waiting. example: xRingbufferReceiveUpTo()
+ * 
+ * All read functions take a xTicksToWait argument, which can be set up to
+ * to infinity. This function will unblock any threads currently waiting.
+ *
+ * @param[in]   xRingbuffer     The ring buffer who's rx will be unblocked.
+ */
+void vRingbufferUnblockRx(RingbufHandle_t xRingbuffer);
```
chipweinberger, line 507: > If usb_serial_jtag.c starts using something other than a ringbuffer internally (e.g. a StreamBuffer), we would have to add similar API for unblocking there; and given that StreamBuffer is upstream FreeRTOS code we would most likely want to avoid that.

^ From Ivan. 

Can you re-evaluate your position on this? `vRingbufferUnblockRx` seems like the ideal solution.

If we ever switched to `StreamBuffer` we could come up with alternative solutions. For example:
- A) a counting semaphore + StreamBuffer. 'give' the semaphore on new data + then call StreamBuffer with a 0 timeout.
- B) Use a long timeout with StreamBuffer, perhaps 10 seconds, and poll an exit condition
- C) Make a Pull Request on FreeRTOS to add the necessary unblock feature to StreamBuffer

chipweinberger: Alternatively, ^^ solution (A), could be used to mitigate the need for `vRingbufferUnblockRx` in this PR as well. 

I think the solution would look like this:
- **normal exclusive mutex.**  `usb_serial_jtag_read_bytes` will grab this mutex first. _Not entirely necessary._ This makes sure the counting semaphore only has a single thread using it at a time. **i.e.** so the first thread to call `read()` will always be satisfied first.
- **counting semaphore.** `usb_serial_jtag_isr_handler_default` increments this counting semaphore _once_ each time it appends to the ringbuffer, `usb_serial_jtag_read_bytes` will 'take' from the semaphore, in order to wait until new bytes are ready (We could also use `event_groups` instead).
- **ring buffer**, `usb_serial_jtag_read_bytes` calls the ringbuffer with a zero timeout, and will wait on the counting semaphore again until there is enough data.

Of course, I would consider the `vRingbufferUnblockRx` solution better.

SoucheSouche:
Hi @chipweinberger, I started looking into your PR. Can I ask you to rebase and fix the build issues and conflict with upstream?

chipweinberger:
sure, one moment.

chipweinberger:
@SoucheSouche updated. was just a trivial collison

components/esp_ringbuf/ringbuf.c
```diff
@@ -1351,6 +1361,18 @@ void vRingbufferGetInfo(RingbufHandle_t xRingbuffer,
     portEXIT_CRITICAL(&pxRingbuffer->mux);
 }
 
+void vRingbufferUnblockRx(RingbufHandle_t xRingbuffer)
+{
+    Ringbuffer_t *pxRingbuffer = (Ringbuffer_t *)xRingbuffer;
+    configASSERT(pxRingbuffer);
+
+    // is the semaphore taken?
+    if (uxSemaphoreGetCount(rbGET_RX_SEM_HANDLE(pxRingbuffer)) == 0) {
```
SoucheSouche, line 1370: you still need to `#include "freertos/semphr.h"` to use the function.

components/driver/uart/uart.c
```diff
@@ -1762,6 +1762,16 @@ esp_err_t uart_set_loop_back(uart_port_t uart_num, bool loop_back_en)
     return ESP_OK;
 }
 
+esp_err_t uart_unblock_reads(uart_port_t uart_num)
+{
+    if (p_uart_obj[uart_num]->rx_ring_buf == NULL) {
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    vRingbufferUnblockRx(pp_uart_obj[uart_num]->rx_ring_buf);
```
SoucheSouche, line 1771: you left a typo here: `pp_uart_obj` → `p_uart_obj`

components/esp_ringbuf/ringbuf.c
chipweinberger: updated.

components/driver/uart/uart.c
chipweinberger: updated.

components/esp_ringbuf/ringbuf.c
```diff
@@ -1351,6 +1362,18 @@ void vRingbufferGetInfo(RingbufHandle_t xRingbuffer,
...
```
SoucheSouche, line 1371: `rbGET_RX_SEM_HANDLE` doesn't seem to be declared anywhere.

components/console/esp_console_repl.c
```diff
@@ -409,13 +411,25 @@ static esp_err_t esp_console_repl_uart_delete(esp_console_repl_t *repl)
...
         ESP_LOGE(TAG, "already de-initialized");
         ret = ESP_ERR_INVALID_STATE;
         goto _exit;
     }
+
+    // wait for repl thread to stop, if it is running
+    if (repl_com->state == CONSOLE_REPL_STATE_START) {
+        repl_com->state = CONSOLE_REPL_STATE_REPL_STOP_REQUESTED;
+        while (repl_com->state != CONSOLE_REPL_STATE_REPL_TASK_ENDED) {
+            uart_unblock_reads();
```
SoucheSouche, line 426: This function is expecting `uart_num` as a parameter it seems. you should update it to `uart_unblock_reads(uart_repl->uart_channel);`

components/esp_ringbuf/ringbuf.c
chipweinberger: This code must have changed. @SoucheSouche, I'm pretty behind and stressed about my own work at the moment. If you are able to update it that would be a great help.

SoucheSouche: sure no problem!

chipweinberger: I wonder if we should move this from the flags, and into its own variable. That way we are less likely to hit threading problems from multiple flags sharing the same memory location.

chipweinberger: ^ I'm not familiar enough with ringbuf to make this judgement call.

components/console/esp_console_repl.c
chipweinberger: updated. I wonder how I was missing that.

SoucheSouche:
@chipweinberger, there has been some changes in the ring buffer so I am now looking at your PR again since the semaphores for rx and tx were removed.

chipweinberger:
> since the semaphores for rx and tx were removed.

I updated the PR and removed the semaphores. But as I was saying I think we should move this bit: `#define rbBUFFER_UNBLOCK_RX_FLAG    ( ( UBaseType_t ) 32 )` to its own variable, to simplify race condition worries.

SoucheSouche:
The other issue is that [this comment](https://github.com/espressif/esp-idf/pull/9983#discussion_r1003527773) is still valid and your current solution won't be accepted.

chipweinberger:
@SoucheSouche `xTaskAbortDelay` is going to cause a lot of issues IMO, because it does not give a chance for clenup. Memory leaks at minimum, bugs at worse.

chipweinberger:
@SoucheSouche , did you understand this comment? "Thinking about this again, select on a pair of file descriptors (stdin and eventfd) is probably the cleanest solution." I'm not sure what that solution would look like.

If you use `xTaskAbortDelay`, make sure the thread is not blocked in user code.

SoucheSouche:
Hi @chipweinberger,
Sorry that I left this PR not updated for so long. I will get back at it this week. It's been a while so it might take a bit of time to get back into it but I will update you as I go through it.
Hopefully we will finally be able to merge it this time.

SoucheSouche:
Hi @chipweinberger, a little update on the PR.
Since it was fairly old, I first had to rebase it on the latest master and realized that your solution was no longer working.
The concept of adding 2 states to the console repl is great and works just fine but the solution that you were using to force linenoise to return didn't seem to work.

I experimented with `select` as @igrr suggested and ended up finding a solution.
The very drafty logic implemented in `linenoiseDumb()` and `linenoiseRaw()` is the following:
```
[...]
        if (select(in_fd + 1, &read_fds, NULL, &except_fds, NULL) < 0) {
            // select returned with -1, handle the error
        }

        // std_in is either ready to read or an exceptional condition occurred
        if(FD_ISSET(in_fd, &except_fds)) {
            // exceptional condition occurred, return from linenoise back into the while loop
            // of `esp_console_repl_task` where the repl state will be reevaluated
            break;
        }
        else if(FD_ISSET(in_fd, &read_fds)) {
            // do the reading since std_in is ready for reading
        }
        else {
            continue;
        }
```
I kept `uart_unblock_reads` and `usb_serial_jtag_unblock_reads` and I make them fire an "exceptional condition" event that will make the `select` return and `if(FD_ISSET(in_fd, &except_fds))` evaluated to true.

chipweinberger:
seems like a reasonable solution!

impressive :) fun :)

SoucheSouche:
I had to implement the select for the usb serial jtag driver so I will have to first merge this added feature before proceeding with the actual deadlock fix.
I will keep you posted.

chipweinberger:
related: https://github.com/espressif/esp-idf/issues/9964

SoucheSouche:
Hi @chipweinberger,
a little update on the fix.
I am still waiting to merge the select() feature for the usb cdc and usb serial jtag drivers before I can merge the fix for the deadlock.
I will let you know when the select() features have been merged but it might take a little while.

SoucheSouche:
hi @chipweinberger,
Sorry I didn't update you for so long. It was quite a journey to implement and merge the select functionality in the usb serial jtag driver and the usb cdc driver.
But great news, I just merged the last one today so I can finally update the bugfix and submit it for review.

I will try to submit it before the end of the week, so I hope it will be merged before Christmas.

chipweinberger:
Yes I can imagine it was a difficult fix. Congrats!

My product is now shipping: [jamcorder.com](jamcorder.com)

SoucheSouche:
Congratulation to you!
The project looks really good! I wish you the best.

And I couldn't help but notice that you are missing the mouse hover on the FAQ section for the desktop version of the website.

chipweinberger:
very nice!

SoucheSouche:
@chipweinberger it took its time but it is done :)

Issue #10573, opened 2023-01-19
Reading coredump information from flash is verbose (IDFGH-9179)


cbaechler:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.3-321-g38961260d8

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

CMD

### Development Kit.

ESP32S3-WROOM N8R2

### Power Supply used.

USB

### What is the expected behavior?

Verbosity of logs when executing `esp_core_dump_get_summary` should be configurable by setting the level with `esp_log_level_set`. 

Is there any way to silence the logs when reading the coredump partition? 

### What is the actual behavior?

Reading coredump partition produces a big amount of logs with TAGs `esp_core_dump_elf` and `esp_core_dump_port`. The output looks like an actual fault is just happening and can confuse the developer. 

### Steps to reproduce.

1. Enable coredump to flash feature and add corresponding partition
2. Trigger a hardfault condition to fill the coredump partition
3. Read the summary from the partition with `esp_core_dump_get_summary`
4. Observe log output

KonssnoK:
Adding example core dump
```
D (00:00:36.928) modem_comm: IP event 0
I (00:00:36.929) network: Network connected
I (00:00:36.934) mqtt_app: MQTT_EVENT_BEFORE_CONNECT
I (00:00:38.468) mqtt_app: MQTT_EVENT_CONNECTED
I (00:00:38.475) mqtt_app: sent publish announce returned msg_id=7736
D (3741) esp_core_dump_elf: ELF ident 7f E L F
D (3741) esp_core_dump_elf: Ph_num 46 offset 34
D (3746) esp_core_dump_elf: PHDR type 4 off 5f4 vaddr 0 paddr 0 filesz 3440 memsz 3440 flags 6 align 0
D (3755) esp_core_dump_elf: Note NameSZ 8 DescSZ 24c Type 1 name CORE
D (3762) esp_core_dump_elf: Note NameSZ 8 DescSZ 24c Type 1 name CORE
[...]

PR #10583, opened 2023-01-20
espcoredump: add log verbosity option (IDFGH-9191)

cbaechler:
Adds an option to override `LOG_LOCAL_LEVEL` in the espcoredump component. With this, the amount of logs generated when reading the coredump information from the partition (e.g. by `esp_core_dump_get_summary`) can be reduced. 

This PR addresses and solves #10573.

### DIFF
--- a/components/espcoredump/Kconfig
+++ b/components/espcoredump/Kconfig
@@ -122,4 +122,44 @@ menu "Core dump"
         default "disable" if ESP_COREDUMP_DECODE_DISABLE
         default "info" if ESP_COREDUMP_DECODE_INFO
 
+    config ESP_COREDUMP_LOG_LEVEL_OVERRIDE
+        bool "Override LOG_LOCAL_LEVEL"
+        default n
+        depends on !ESP_COREDUMP_ENABLE_TO_NONE
+        help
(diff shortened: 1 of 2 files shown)


components/espcoredump/Kconfig
```diff
@@ -122,4 +122,44 @@ menu "Core dump"
...
+            bool "Error"
+        config ESP_COREDUMP_LOG_LEVEL_WARN
+            bool "Warning"
+        config ESP_COREDUMP_LOG_LEVEL_INFO
+            bool "Info"
+        config ESP_COREDUMP_LOG_LEVEL_DEBUG
+            bool "Debug"
+        config ESP_COREDUMP_LOG_LEVEL_VERBOSE
+            bool "Verbose"
+    endchoice
```
chipweinberger, line 154: I would add `config ESP_COREDUMP_LOG_LEVEL_SAME` as the default option.

No need for the override.

cbaechler:
We found that an option to fully disable logs for the coredump component was merged with https://github.com/espressif/esp-idf/commit/962621e5d76699bde807c1cd9f3eea35c4352cec. Since this is enough for our use-case, I am going to close this PR.

Issue #10594, opened 2023-01-21
[Question] [HTTP Server] 🚀🚀 Handling multiple requests at the same time (IDFGH-9204)


chipweinberger:
### HTTP Server

**Is there a way to handle multiple requests at the same time** for the http server?

is it even possible?

In ESP-IDF today, **only a single request handler can run at a time**. And even with a separate thread, **`httpd_req_t` is freed when the handler completes** so prolonged communication is impossible.

How reasonable is it for me to hack ESP-IDF to support this?

**Previous discussion:** https://esp32.com/viewtopic.php?t=28000

wuyuanyi135:
https://github.com/espressif/esp-idf/blob/49551cc48cb3cdd5563059028749616de313f0ec/examples/protocols/http_server/advanced_tests/main/tests.c#L273

chipweinberger:
Ive seen that example before but I don't understand how it would work in a real situation.

`httpd_queue_work` executes code on the same thread as the HTTP Server, and the example just returns a string synchronously. What is the reason for doing it this way?

Perhaps the example could show how to execute a long 30+ second calculation asynchronously and _then_ return it.

Or more important, how to download a large file over HTTP for a few minutes while still executing other requests.

boarchuz:
+1 for this feature request

wuyuanyi135:
You can save those file descriptor and handle to a structure and pass it to another thread for later use. In this case you cannot use those req-like functions. You have to assemble the http packet yourself. I can show you some c++ code to wrap around these functions when I got home

wuyuanyi135:
```cpp
#define HTTP_UTIL_CHECK(x)                                                                              \
  do {                                                                                                  \
    int ret = x;                                                                                        \
    if (ret == HTTPD_SOCK_ERR_INVALID || ret == HTTPD_SOCK_ERR_TIMEOUT || ret == HTTPD_SOCK_ERR_FAIL) { \
      return ret;                                                                                       \
    }                                                                                                   \
  } while (0)

int httpd_socket_send_common_header(httpd_handle_t hd, int fd, const std::string& content_type, size_t content_length, int status = 200) {
  char buf[80];
  size_t sz = snprintf(
      buf,
      sizeof(buf),
      "HTTP/1.1 %d OK\r\n"
      "Content-Type: %s\r\n"
      "Content-Length: %d\r\n",
      status,
      content_type.c_str(),
      content_length);
[...]

PR #10602, opened 2023-01-24
[HTTPD] support for multiple simultaneous requests (IDFGH-9214)

chipweinberger:
Related Issue: https://github.com/espressif/esp-idf/issues/10594

Add basic support for supporting multiple simultaneous requests in HTTPD server.

**Implementation**
- adds `httpd_req_copy(dest, source)` so that request memory ownership can be given to the user
- adds `httpd_req_claim_ownership(req)` to fully give ownership of the req to another thread, so the socket will not get purged by `lru_purge_enable`, until `httpd_req_relinquish_ownership(req)` is called.

**Open Question 1:** It might be better to remove `httpd_req_copy(dest, source)`, and instead have the server allocate a new request object for each request. Possibly behind a, `httpd_server.new_request_per_handler`, flag. Or, have `httpd_req_claim_ownership(req)` return a copy of the request.

**Open Question 2:** Instead of calling `httpd_req_claim_ownership(req)`, the user's request handler could return `ESP_ERR_NOT_FINISHED`, which would do the same thing. The user would then call `httpd_req_finish(req)` to mark the socket as purgeable again.

In later ESP-IDF versions, we should probably make HTTPD have more "proper" simultaneous request support, at which point we could deprecate these functions perhaps.  Lots of people have use cases for this, and this approach gives lots of control to users.

**Example**
- add an example to ESP-IDF, showing usage.

**Testing**
- tested on a ESP32-S3


chipweinberger:
This PR is more of an experiment, that I wanted to share. 

If we merge something like this, I like the `ESP_ERR_NOT_FINISHED` approach with a `httpd_server.new_request_per_handler` flag.

chipweinberger:
Actually thought of an even better idea. 

Instead of `httpd_server.new_request_per_handler`, a new req obj will be allocated only when ESP_ERR_NOT_FINISHED is returned. Otherwise the same req obj will be reused indefinitely.

chipweinberger:
new PR: https://github.com/espressif/esp-idf/pull/10669

Issue #8662, opened 2022-03-25
pthread allow Stack on Spiram (IDFGH-7045)


f-hoepfinger-hr-agrartechnik:
when you have a lot of Tasks, 
internal Memory might be used up fast. 

now, 
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-spiram-allow-stack-external-memory
allow use of Stack on SPIRAM. 
but not via C++ Pthreads. 
https://github.com/espressif/esp-idf/tree/master/examples/cxx/pthread

would be a nice Extension right ? 

Thanks, 
Franz

0xjakob:
@f-hoepfinger-hr-agrartechnik

> would be a nice Extension right ?

Certainly! But I'm afraid that the [constructor of std::thread](https://en.cppreference.com/w/cpp/thread/thread/thread) does not allow any additional parameters than the actual function to be executed in the thread.

Maybe I missed something, so if you have another idea, please let me know!

franz-ms-muc:
esp_pthread_set_cfg(&cfg) 
The cfg could be extended.

0xjakob:
@f-hoepfinger-hr-agrartechnik @franz-ms-muc I see now (Should have looked at the example more thoroughly). Just to clarify: you would like a possibility to configure `std::thread` via the `esp_pthread` API? If that's correct, then I think it's possible. We would add this to our backlog but we currently have some other high-priority tasks. So it might take some (> 1 month). If you eventually implement something like this yourself, feel free to open a pull request and then we can discuss there.

f-hoepfinger-hr-agrartechnik:
Hi,

we did one implementation here: 
https://github.com/hr-agrartechnik/esp-idf/blob/HR_4.4_ram/components/pthread/pthread.c

but i would not want to open a Pull Request as i feel the Quality is not good enough for a Pull request. 
maybe later when we have done a bunch more test. 

it is "generally" running OK, so we use it in a Prototype.

f-hoepfinger-hr-agrartechnik:
Ported same to 4.4.2
https://github.com/hr-agrartechnik/esp-idf/tree/HR_4.4.2_ram

any outlook that we can get such in ESP-IDF ?

f-hoepfinger-hr-agrartechnik:
will be here shortly.
https://github.com/espressif/esp-idf/pull/10623

PR #10623, opened 2023-01-26
External Memory for Pthread

f-hoepfinger-hr-agrartechnik:
https://github.com/espressif/esp-idf/issues/8662


components/pthread/include/esp_pthread.h
```diff
@@ -21,6 +21,7 @@ extern "C" {
 /** pthread configuration structure that influences pthread creation */
 typedef struct {
     size_t stack_size;  ///< The stack size of the pthread
+    bool externalmemory;
```
igrr, line 24:
```suggestion
    uint32_t stack_alloc_caps;   ///< A bit mask of memory capabilities (MALLOC_CAPS*) to use when allocating the stack. If zero, MALLOC_CAP_DEFAULT is assumed.
```

"External ram" is probably not a very future-proof flag, since some chips might have more types of memory: TCM, internal slower memory, external faster memory, external slower memory, retention-capable memory. Using MALLOC_CAPS macros provides the same amount of flexibility as available in the heap allocation APIs.

components/pthread/pthread.c
```diff
@@ -272,27 +277,40 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
...
-    BaseType_t res = xTaskCreatePinnedToCore(&pthread_task_func,
+    // stack_size is in bytes. This transformation ensures that the units are
+    // transformed to the units used in FreeRTOS.
+    // Note: float division of ceil(m / n) ==
+    //       integer division of (m + n - 1) / n
+    int size_stack = (stack_size + sizeof(StackType_t) - 1) / sizeof(StackType_t);
+    uint32_t memoryLocation = externalmemory ? MALLOC_CAP_SPIRAM : MALLOC_CAP_INTERNAL;
+    StackType_t *stack_for_task = (StackType_t *) heap_caps_calloc(1, size_stack, memoryLocation | MALLOC_CAP_8BIT);
+    if (stack_for_task == NULL) {
+        ESP_LOGE(TAG, "Failed to create task!");
```
igrr, line 288:
```suggestion
        ESP_LOGE(TAG, "Failed to allocate task stack!");
```

components/pthread/pthread.c
```diff
@@ -272,27 +277,40 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
...
+        free(pthread);
+        free(task_arg);
+    	return ENOMEM;
+    }
+    StaticTask_t *taskTC =  (StaticTask_t *) heap_caps_calloc(1, sizeof(StaticTask_t), MALLOC_CAP_INTERNAL|MALLOC_CAP_8BIT);
+    if (taskTC == NULL) {
+        ESP_LOGE(TAG, "Failed to create task!");
+        free(pthread);
+        free(task_arg);
```
igrr, line 297: Need to add `free(stack_for_task);` here

components/pthread/pthread.c
```diff
@@ -272,27 +277,40 @@ int pthread_create(pthread_t *thread, const pthread_attr_t *attr,
...
-                                             &xHandle,
+											 stack_for_task,
+											 taskTC,
                                              core_id);
 
-    if (res != pdPASS) {
+    if (xHandle == NULL) {
         ESP_LOGE(TAG, "Failed to create task!");
         free(pthread);
         free(task_arg);
```
igrr, line 312:
```c++
free(stack_for_task);
free(taskTC);
```
here as well

igrr (changes requested):
Please note that the stack and the TCB of a task allocated using `xTaskCreateStaticPinnedToCore` are not deleted automatically. You also need to modify the pthread deletion code to free the memory, otherwise it would be leaked.

I've also left a few other comments below.

components/pthread/pthread.c
f-hoepfinger-hr-agrartechnik: done

components/pthread/pthread.c
f-hoepfinger-hr-agrartechnik: done

components/pthread/pthread.c
```diff
@@ -121,6 +123,9 @@ static esp_pthread_t *pthread_find(TaskHandle_t task_handle)
 static void pthread_delete(esp_pthread_t *pthread)
 {
     SLIST_REMOVE(&s_threads_list, pthread, esp_pthread_entry, list_node);
+    free(pthread->task_arg);
```
f-hoepfinger-hr-agrartechnik, line 126: this was also not freed before ? 
is this intended, or was it a bug ?

igrr: I think you are right that this should be done together with deleting `pthread` structure. Please remove `free(pthread->task_arg);` from pthread_exit to avoid a double-free.

igrr:
Also please combine the changes into a single commit (using `git rebase -i`) to keep the commit history clean.

components/pthread/pthread.c
f-hoepfinger-hr-agrartechnik: done !

f-hoepfinger-hr-agrartechnik:
> Also please combine the changes into a single commit (using `git rebase -i`) to keep the commit history clean.

like this ?

f-hoepfinger-hr-agrartechnik:
This code fails with a assertion at thread.join();

f-hoepfinger-hr-agrartechnik:
no, still 
`assert failed: prvDeleteTCB tasks.c:4758 (pxTCB->ucStaticallyAllocated == ( ( uint8_t ) 2 ))`

f-hoepfinger-hr-agrartechnik:
> no, still `assert failed: prvDeleteTCB tasks.c:4758 (pxTCB->ucStaticallyAllocated == ( ( uint8_t ) 2 ))`

ah, but this time in the Idle Task. 
need more Help.

f-hoepfinger-hr-agrartechnik:
https://github.com/hr-agrartechnik/esp-idf-examples-cxx-pthread

Crash on join()

It is reproducible.
So need to debug on this minimal example.

components/pthread/include/esp_pthread.h
```diff
@@ -21,6 +21,7 @@ extern "C" {
 /** pthread configuration structure that influences pthread creation */
 typedef struct {
     size_t stack_size;  ///< The stack size of the pthread
+    uint32_t stack_alloc_caps;   ///< A bit mask of memory capabilities (MALLOC_CAPS*) to use when allocating the stack. If zero, MALLOC_CAP_DEFAULT is assumed.
```
0xjakob, line 24: I'm afraid this might cause trouble for users who initialized each member of `esp_pthread_cfg_t` manually like this:
```
    esp_pthread_cfg_t cfg;
    cfg.thread_name = name;
    cfg.pin_to_core = core_id;
    cfg.stack_size = stack;
    cfg.prio = prio;
    cfg.inherit_cfg = false;
``` 
If they would upgrade IDF including this change, `esp_pthread_cfg_t.stack_alloc_caps` is uninitialized. Once the configuration is used, this would produce a crash at runtime at best or a way more subtle memory issue in the worst case.

I don't have a definitive solution for this yet. The `esp_pthread_set_cfg` makes it hard to introduce non-breaking configuration changes.

components/pthread/pthread.c
```diff
@@ -423,11 +455,9 @@ void pthread_exit(void *value_ptr)
     if (!pthread) {
         assert(false && "Failed to find pthread for current task!");
     }
-    if (pthread->task_arg) {
-        free(pthread->task_arg);
-    }
     if (pthread->detached) {
         // auto-free for detached threads
+        vTaskDelete(NULL);
```
0xjakob, line 460: What was the reason for moving `vTaskDelete(NULL)` here? It seems that the semantics are changed significantly: the following two statements won't be executed, the log call and giving the semaphore below won't be executed, either.

0xjakob (commented):
@f-hoepfinger-hr-agrartechnik Thanks for addressing the pthread external memory issue and sorry for the slow turnaround!

I looked at the code and most of it looks good. I have, however two points that I'm not certain about, please see the comments in the code.

We'll talk about the issue with the `esp_pthread_set_cfg()` API (see my comments) and let you know about any results.

components/pthread/include/esp_pthread.h
0xjakob: Sorry, I missed that our documentation is [explicitly advising users to zero-initialize configuration structures](https://docs.espressif.com/projects/esp-idf/en/release-v5.0/esp32/api-reference/api-conventions.html). You can consider this comment resolved, then.

donghengqaz:
> no, still `assert failed: prvDeleteTCB tasks.c:4758 (pxTCB->ucStaticallyAllocated == ( ( uint8_t ) 2 ))`

It looks like the freeRTOS forbidden to delete a full static(static task stack and TCB) task, so [code](https://github.com/espressif/esp-idf/blob/master/components/freertos/FreeRTOS-Kernel/tasks.c#L4954) has `assert` process. @f-hoepfinger-hr-agrartechnik @0xjakob using `xTaskCreateRestrictedStatic` may be solution here.

0xjakob:
@f-hoepfinger-hr-agrartechnik I think saving the stack memory in the struct `esp_pthread_t` does not work since the deletion must be deferred if the thread is deleting itself, this might be the cause of some of the issues you observed. The best way to defer stack deletion seems to be via [`vPortCleanUpTCB()`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-freertos-enable-static-task-clean-up), which needs to be enabled and provided.

I'll take a look at this and provide suggestions this week.

diplfranzhoepfinger:
> I'll take a look at this and provide suggestions this week.

Thanks so much. Very Helpful !

f-hoepfinger-hr-agrartechnik:
> I'll take a look at this and provide suggestions this week.

any news ?

0xjakob:
@f-hoepfinger-hr-agrartechnik We needed to go a long way because we needed to create new private FreeRTOS functions for this use case. We have merged the FreeRTOS part already. Now we're in the process of merging the final pthread MR. It will be quite different compared to your code, but we'll leave a thanks in the commit message for sure!

0xjakob:
@diplfranzhoepfinger It's been merged to master now, PTAL. There's a [unit test which shows how to use the API](https://github.com/espressif/esp-idf/blob/master/components/pthread/test_apps/pthread_psram_tests/main/pthread_psram_tests.c#L87).

diplfranzhoepfinger:
Very cool. Want to test it.

Issue #10626, opened 2023-01-26
Intermitent crash during IDF initialisation (IDFGH-9240)


KaeLL:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

9ee3c8337d3c4f7914f62527e7f7c78d7167be95

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

Custom Board

### Power Supply used.

External 5V

### What is the expected behavior?

To not crash, especially before getting to my code.

### What is the actual behavior?

Intermitenttly crashing [here](https://github.com/espressif/esp-idf/blob/9ee3c8337d3c4f7914f62527e7f7c78d7167be95/components/freertos/tasks.c#L4813).

### Steps to reproduce.

<details>
<summary>defconfig</summary>

```
CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_PERF=y
CONFIG_BOOTLOADER_LOG_LEVEL_NONE=y
CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE=y
CONFIG_BOOTLOADER_SKIP_VALIDATE_ALWAYS=y
CONFIG_ESPTOOLPY_FLASHMODE_QIO=y
CONFIG_ESPTOOLPY_FLASHFREQ_80M=y
CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
# CONFIG_ESPTOOLPY_FLASHSIZE_DETECT is not set
CONFIG_ESPTOOLPY_AFTER_NORESET=y
CONFIG_ESPTOOLPY_MONITOR_BAUD_CONSOLE=y
CONFIG_PARTITION_TABLE_CUSTOM=y
CONFIG_PARTITION_TABLE_CUSTOM_FILENAME="partitions_ota.csv"
CONFIG_PARTITION_TABLE_OFFSET=0xC000
CONFIG_COMPILER_WARN_WRITE_STRINGS=y
CONFIG_BT_ENABLED=y
CONFIG_BTDM_CTRL_BLE_MAX_CONN=1
# CONFIG_BTDM_CTRL_MODEM_SLEEP is not set
[...]

Issue #9441, opened 2022-07-26
mutex assert triggering (IDFGH-7923)


KaeLL:
## Environment

- Module or chip used: ESP32
- IDF version: 36f49f3
- Compiler version: crosstool-NG esp-2022r1-RC1

## Problem Description

I'm updating a [test app](https://github.com/KaeLL/ESP32-W5100) of mine to latest IDF, and the assert checking if a mutex is being released by a different thread keeps triggering. Worth noting that it's very hard to reproduce, but when it happens, it's always after the MQTT log (shown below). 

## Debug Logs

<details>
<summary>Log</summary>

```
I (20:03:20.314) MQTTS_EXAMPLE: [APP] Free memory: 261568 bytes

assert failed: xQueueGenericSend queue.c:837 (pxQueue->pcHead != ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == xTaskGetCurrentTaskHandle())


Backtrace: 0x40081cb2:0x3ffbd6a0 0x40085c09:0x3ffbd6c0 0x4008bb5d:0x3ffbd6e0 0x400864f8:0x3ffbd800 0x40082b69:0x3ffbd840 0x40082c19:0x3ffbd870 0x40082c8a:0x3ffbd890 0x4012551d:0x3ffbd8c0 0x40121414:0x3ffbd8e0 0x401352f4:0x3ffbd900 0x40104422:0x3ffbd970 0x4010499b:0x3ffbd9a0 0x400d8cbc:0x3ffbd9c0 0x400d9beb:0x3ffbe280 0x40088cda:0x3ffbe2a0
0x40081cb2: panic_abort at /esp-idf/components/esp_system/panic.c:409

0x40085c09: esp_system_abort at /esp-idf/components/esp_system/esp_system.c:135

0x4008bb5d: __assert_func at /esp-idf/components/newlib/assert.c:78

0x400864f8: xQueueGenericSend at /esp-idf/components/freertos/FreeRTOS-Kernel/queue.c:837 (discriminator 8)

0x40082b69: lock_release_generic at /esp-idf/components/newlib/locks.c:192

0x40082c19: _lock_release at /esp-idf/components/newlib/locks.c:198

0x40082c8a: __retarget_lock_release at /esp-idf/components/newlib/locks.c:334

[...]

PR #10649, opened 2023-01-30
Use correct recursive mutex for env and regular mutex for tz (IDFGH-9266)

smjothen:
Newlib defines __env_lock_object as a recursive mutex and __tz_lock_object as a regular mutex, but when initialized in esp_newlib_locks_init, __env_lock_object gets initialized as a regular mutex (points to s_common_mutex) and __tz_lock_object gets initialized to a recursive mutex (s_common_recursive_mutex).

This causes a lot of panics due to various mutex ownership checks in the FreeRTOS code, but only seems to appear when there is a lot of usage of the __tz_lock_object and __env_lock_object. I believe this is also causing the crashes referred to in the issues: #10626 and #9441

### DIFF
--- a/components/newlib/locks.c
+++ b/components/newlib/locks.c
@@ -375,9 +375,9 @@ void esp_newlib_locks_init(void)
     extern _lock_t __sinit_lock;
     __sinit_lock = (_lock_t) &s_common_recursive_mutex;
     extern _lock_t __env_lock_object;
-    __env_lock_object = (_lock_t) &s_common_mutex;
+    __env_lock_object = (_lock_t) &s_common_recursive_mutex;
     extern _lock_t __tz_lock_object;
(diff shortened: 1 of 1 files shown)


igrr:
sha=7ca5fa87c0d6e7173a353425b40293f79838859d

igrr:
@smjothen thank you for the fix!

KaeLL:
@igrr it'll be backported to 4.4, right?

ESP-Marius:
@KaeLL Yes, when it is merged we will also backport it to 4.4

AxelLin:
> @KaeLL Yes, when it is merged we will also backport it to 4.4

@ESP-Marius Just remind that v4.4 branch still does not include this fix!

ESP-Marius:
@AxelLin it is already merged in our internal gitlab repo. It will be available the next time v4.4 passes QA and syncs to github.

PR #10669, opened 2023-02-01
[HTTPD] basic support for multiple simultaneous requests (try 2) (IDFGH-9285)

chipweinberger:
**Related Issue:** https://github.com/espressif/esp-idf/issues/10594

**Previous PR:** https://github.com/espressif/esp-idf/pull/10602

Add basic support for supporting multiple simultaneous requests in HTTPD server.

**Note**: This PR does _not_ add _native_ support for simultaneous requests, but instead makes it more ergonomic for users to implement themselves. 

Is espressif interested in improving async http request support? This small code change brings big ergonomic advantages over raw sockets.

**Advantages over raw sockets**
- **easy async reply:** Users can use all the existing http functions (i.e. `http_resp_sendstr`) to send their async response on any thread, instead of raw socket send(), and can access headers, etc
- **LRU safety:** The long running async socket will not get closed when LRU socket purge is enabled 

**Implementation:**

- an async request handler will return ESP_ERR_NOT_FINISHED
  - this gives the users request handler exclusive use of the request object. (The server will allocate a new one.)
  - this will also guarentee socket to not be LRU purged
- async handler must call `httpd_req_async_handler_complete()` when done 
  - this lets the socket be LRU purged
  - this frees our request object.

**Example**

- add an example to ESP-IDF, showing usage.

**Testing**

- still needs to be tested. I tested the last PR, but have not re-tested the new changes, pending feedback first. 

**Feeback**
- please give feedback on the approach.


components/esp_http_server/src/httpd_txrx.c
```diff
@@ -78,6 +78,14 @@ static esp_err_t httpd_send_all(httpd_req_t *r, const char *buf, size_t buf_len)
         buf     += ret;
         buf_len -= ret;
     }
+
+    // is the request complete?
+    if (buf_len >= 2 && buf[buf_len-1] == '\r' && buf[buf_len-1] == '\n') {
+        // Note: we aren't handling the case where '\r' & '\n' are sent 
+        // seperately because this would be extremely rare and not worth the complexity.
+        ra->sd->is_req_unfinished = false;
+    }
```
chipweinberger, line 87: I wont have a need for this code once `httpd_req_delete()` is implemented.

chipweinberger:
@ESP-Marius @ESP-YJM

Would love some feedback on this.

It might not be the number 1 feature Espressif needs to work on right now, but I want to add support for this feature in my product sometime next year.

I think it's a lovely, simple approach to a complicated problem.

Does espressif ever plan to support this?

ESP-Marius:
@chipweinberger Thanks for another contribution and apologies for our rather slow handling of PRs....

I think for this code-area @mahavirj might be able to help find someone that can look at it, as Im not very familiar with it.

mahavirj:
@chipweinberger Sorry for slow turnaround. I will prioritize the review here.

chipweinberger:
@mahavirj It's no problem I know espressif has lots of priorities! Thanks!

hmalpani:
Hello @chipweinberger 
Thanks for adding the support for multiple simultaneous requests in esp_http_server.

I tried building the `async_handlers` example and got several compile time errors. Can you please check those errors and update the PR?

chipweinberger:
@hmalpani apologies about that!

I just updated the PR and it should compile now. If it doesn't "just work", let me know and I can fix it when I get back from vacation in 10 days.

The PR is 90% tested, but I did not test the last change I made to http req ownership. Sorry about it!

hmalpani:
Hello @chipweinberger 
Thanks for updating the PR. I still see a compilation error. Can you fix including the `FreeRTOS.h` header file in main.c of the `async_handlers` example.

When trying the example, `quick` handler was working fine but the application crashed when `long` was called from the browser. Can you please check it once?

chipweinberger:
Yes I'll fix the crash as soon as I return next week.

Thanks for trying it again.

chipweinberger:
@hmalpani 

I've pushed an update. Here is where this PR is now:

- `/long` works
- `/quick` works
- calling `/long` then `/quick` results in a `assert failed: block_locate_free tlsf.c:566 (block_size(block) >= size)`

Let me remind you, I made two PRs:
- Try 1: user api uses `httpd_req_copy`, `httpd_req_claim_ownership`, `httpd_req_relinquish_ownership`: https://github.com/espressif/esp-idf/pull/10602
- Try 2: user api uses `ESP_ERR_NOT_FINISHED` (this PR)

It turns out the `Try 2` approach is more complicated than I suspected, and has resulted in a bigger PR, that still needs some work.

`Try 1` approach is tested and working and a simpler change, however the user API is more complicated IMO (debatably).

Before I fix 'Try 2', can you first decide which API you like better? `Try 1` or `Try 2`? Please take a look at `long_async_hander` in both examples.

chipweinberger:
Closed this PR in favor of a new PR. based on the Try 1 code: https://github.com/espressif/esp-idf/pull/11190

PR #10699, opened 2023-02-05
[HEAP] add support for low memory callback (IDFGH-9319)

chipweinberger:
Adds `esp_err_t heap_caps_register_low_heap_callback(uint32_t caps, uint32_t threshold, esp_heap_low_memory_hook_t callback);`

**Motivation:**
- personally, I want to save a heap trace to an sd card whenever I am running low on heap (i.e. due to a leak)
- for others, they may want to schedule work to free memory
- handling low memory is paramount for reliable systems

**Note:** the callback is invoked before every allocation made while we are/will be under the theshold.
**Note:** current heap size is also returned in the callback. Users can use that to detect the "crossover" point, i.e. when we will first go under threshold, vs an allocation made while already under threshold.

### DIFF
--- a/components/heap/heap_caps.c
+++ b/components/heap/heap_caps.c
@@ -21,6 +21,14 @@ static void *heap_caps_realloc_base( void *ptr, size_t size, uint32_t caps );
 static void *heap_caps_calloc_base( size_t n, size_t size, uint32_t caps );
 static void *heap_caps_malloc_base( size_t size, uint32_t caps );
 
+typedef struct {
+    esp_heap_free_space_hook_t callback;
+    uint32_t caps;
+    size_t threshold;
+} free_space_callback_t;
+
+static free_space_callback_t free_space_callbacks[CONFIG_HEAP_FREE_SPACE_CALLBACK_COUNT];
+
 /*
 This file, combined with a region allocator that supports multiple heaps, solves the problem that the ESP32 has RAM
 that's slightly heterogeneous. Some RAM can be byte-accessed, some allows only 32-bit accesses, some can execute memory,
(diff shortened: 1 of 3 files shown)


chipweinberger:
**Precommit:** ERROR: Package 'isort' requires a different Python: 3.7.15 not in '>=3.8.0'

Alvin1Zhang:
Thanks for your contribution.

components/heap/heap_caps.c
```diff
@@ -77,6 +85,54 @@ esp_err_t heap_caps_register_failed_alloc_callback(esp_alloc_failed_hook_t callb
...
+            return ESP_OK;
+        }
+    }
+
+    return ESP_FAIL;
+}
+
+#if CONFIG_HEAP_LOW_MEMORY_CALLBACK_COUNT
+IRAM_ATTR
+#endif
```
SoucheSouche, line 135: I don't see the reason why adding placing this function in IRAM if callback count > 0.

components/heap/heap_caps.c
```diff
@@ -77,6 +85,54 @@ esp_err_t heap_caps_register_failed_alloc_callback(esp_alloc_failed_hook_t callb
...
+
+    // add
+    for (int i = 0; i < CONFIG_HEAP_LOW_MEMORY_CALLBACK_COUNT; i++) {
+        if (low_memory_callbacks[i].callback == NULL) {
+            low_memory_callbacks[i].caps = caps;
+            low_memory_callbacks[i].threshold = threshold;
+            low_memory_callbacks[i].callback = callback;
+            return ESP_OK;
+        }
+    }
```
SoucheSouche, line 128: Wouldn't it be more efficient to make only 1 loop looking for existing entry and adding one entry if no entry found? Also, you are using the register function to modify an existing entry when `low_memory_callbacks[i].caps == caps` (here we want to update the `threshold` and/or `callback`) but it would be nice to also allow the user to call this function to update the `threshold` and/or `caps` by checking for `low_memory_callbacks[i].callback == callback`.

SoucheSouche:
Also, I would avoid compiling the added functions all together if `CONFIG_HEAP_LOW_MEMORY_CALLBACK_COUNT == 0` so the new feature does not influence the binary size of the heap component if not used.

components/heap/heap_caps.c
chipweinberger: As for 1 loop vs 2, I think we should do whatever is most readable. There are no perf concerns either way imo.

For the first loop, yous suggest this? Yes that sounds good.
`if (low_memory_callbacks[i].caps == caps || low_memory_callbacks[i].callback == callback) {`

components/heap/heap_caps.c
chipweinberger: `heap_caps_match` is used by `heap_caps_get_free_size` which is now called in all of the allocators to check if the threshold is breached.

chipweinberger:
@SoucheSouche , thinking about the future API, I'm wondering if we also want to support a `minimum block size` threshold?

- low heap threshold
- minimum block size threshold

`heap_caps_register_low_heap_callback(uint32_t caps, int free_threshold, int minimum_block_size_threshold , ...);`

I imagine most users would want to set both, but they could set it to -1 to skip it?

chipweinberger:
**Also, thinking about this more, we should probably get rid of the threshold altogether.** "Free Space" is just a pointer look up. It is nearly instant. There is no need for a threshold. Users can implement a threshold themselves.

components/heap/heap_caps.c
SoucheSouche: It makes sense, I didn't see that you also added the `IRAM_ATTR` to `heap_caps_get_free_size()`

SoucheSouche:
Yes, I agree that this approach sounds better and gives the user more freedom.

chipweinberger:
closing this PR.  I'll create a new one today or tomorrow with our new approach.

PR #10708, opened 2023-02-06
Fix pre-commit hook (IDFGH-9328)

Himel55:
This is just a suggestion, but a recent fix to the pre-commit hook changed isort revision to 5.12.0. However this version isn't compatible with python version 3.7 as listed by the release notes https://github.com/PyCQA/isort/releases/tag/5.12.0.
However I believe the oldest supported python version for esp-idf is 3.7.

There was a hot fix to the latest isort which still supports python version 3.7 to resolve the poetry dependency issue.
https://github.com/PyCQA/isort/releases/tag/5.11.5

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,7 @@ repos:
       - id: flake8
         args: ['--config=.flake8', '--tee', '--benchmark']
   - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
+    rev: 5.11.5
     hooks:
       - id: isort
         name: isort (python)


Himel55:
Otherwise I believe this workflow version should be updated: https://github.com/espressif/esp-idf/blob/master/.github/workflows/pre_commit_check.yml#L24

igrr:
Hi @Himel55, thanks for raising the PR. I believe that we do have the same fix internally, it will appear on github soon as commit bfb29022de37d6bc4e286c04b73d91ae977006a4.

igrr:
> However I believe the oldest supported python version for esp-idf is 3.7.

Regarding this, I would like to note that while we support _using_ esp-idf (i.e. building apps based on esp-idf) with python 3.7, developers of esp-idf (i.e. people who need to make commits in esp-idf repo) are encouraged to install newer Python versions, if possible.

mahavirj:
Fixed with https://github.com/espressif/esp-idf/commit/bfb29022de37d6bc4e286c04b73d91ae977006a4

PR #10751, opened 2023-02-11
Add `float` and `double` to nvs_partition_gen.py

jclds139:
This brings the partition generator up to compatibility with the [ESP32-Arduino Preferences](/../../espressif/arduino-esp32/blob/master/libraries/Preferences/) library.
Both types are converted to blobs then written as `binary` entries, same as the `Preferences` library does.

### DIFF
--- a/components/nvs_flash/nvs_partition_generator/README.rst
+++ b/components/nvs_flash/nvs_partition_generator/README.rst
@@ -39,12 +39,13 @@ Each line of a CSV file should contain 4 parameters, separated by a comma. The t
       - 
     * - 3
       - Encoding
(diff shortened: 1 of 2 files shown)


chipweinberger:
what kind of floats are supported?

IEEE754? For persistent storage that should be documented IMO.

jclds139:
It's IEEE754 single (for `float`) or double (for `double`) precision.
The C code used by the Arduino library would already be using those, and the Python functions used force IEEE754 regardless of the platform (see note 4 [here](https://docs.python.org/3/library/struct.html#format-characters)).

igrr:
@jclds139 I'm afraid we can't add `float` and `double` to the partition generator tool and the docs without also implementing `nvs_get_float`, `nvs_set_float`, `nvs_get_double`, `nvs_set_double` functions in IDF. Otherwise the readers of the documentation might think that they can actually create key-value pairs using these types and later access them in the program.

However, if we do implement these NVS functions for the new data types, we wouldn't be aliasing them to the `blob` type — they would be added as new types. This means that Arduino Preferences library will have to change the implementation of these NVS types, which might be a breaking change for existing projects.

cc @me-no-dev

PR #10792, opened 2023-02-17
[Heap] new api for user heap trace callbacks (IDFGH-9424)

chipweinberger:
**Previous PR:** https://github.com/espressif/esp-idf/pull/10699

Adds 2 new hooks for the user to trace heap allocations themselves.
```
typedef void (*esp_heap_trace_alloc_hook_t) (void* ptr, size_t size);
typedef void (*esp_heap_trace_free_hook_t) (void* ptr);
```

Motivation:
- personally, I want to save a heap trace to an sd card whenever I am running low on heap (i.e. due to a leak), without needing a timer constantly monitoring freespace
- for others, they may want to schedule work to free memory when a threshold is hit
- user tooling

### DIFF
--- a/components/heap/heap_caps.c
+++ b/components/heap/heap_caps.c
@@ -31,7 +31,8 @@ possible. This should optimize the amount of RAM accessible to the code without 
 */
 
 static esp_alloc_failed_hook_t alloc_failed_callback;
-
+static esp_heap_trace_alloc_hook_t trace_alloc_callback;
+static esp_heap_trace_free_hook_t trace_free_callback;
 
 #ifdef CONFIG_HEAP_ABORT_WHEN_ALLOCATION_FAILS
 IRAM_ATTR static void hex_to_str(char buf[8], uint32_t n)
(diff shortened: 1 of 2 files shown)


chipweinberger:
@SoucheSouche, perhaps you can do the `inline` changes in a separate commit? 

You are more familiar than I am with the required changes, and logically they are separate "features".  Getting free heap size from any thread is IMO an expected capability.

SoucheSouche:
Hi @chipweinberger,

I had a look at this new PR and I came up with an alternative way to implement it (see the patch to your feature below).
This approach saves a bit of space in RAM when callbacks are implemented (~100 bytes between your implementation and mine when the callbacks are defined) but most importantly, it enforces the definition of those callbacks to be placed in RAM when the user provides a definition.

Also, side effect with this approach, the callbacks will be triggered from boot instead of after registration (i.e., after the main is reached). 

I let you have a look and let me know what you think about it. Otherwise, your implementation looks good if we document that callbacks must be placed in RAM.

[alloc_free_hook.patch](https://github.com/espressif/esp-idf/files/10784010/alloc_free_hook.patch)

chipweinberger:
@SoucheSouche that is undoubtedly a better implementation. You are more knowledgeable about this feature than I am. :) 

I have just looked at the patch briefly on my phone. It is night time where I am.

But that patch looks good to me. If you want to close my PR and run with yours that would be great. 

I appreciate your work in implementing my feature suggestion!

SoucheSouche:
I will sync your PR on Gitlab and apply my patch on top of it. I will keep you posted as usual.

SoucheSouche:
The PR was reviewed. I expect it to be merged this week.
I ended up introducing the two weak functions for allocation and free, and keep the mechanism for failures untouched to not introduce any breaking change. I also made sure that the allocation hook is called only when the allocation is successful to avoid calling 2 callbacks on failures (it also reduces the number of `if hook then call hook` blocks in the `heap_caps.c`).

chipweinberger:
sounds reasonable!

chipweinberger:
what is the status on this PR?

chipweinberger:
@SoucheSouche bump

SoucheSouche:
Hi @chipweinberger, the PR was merged. Sorry for the waiting, I was away from work for a while.

PR #10793, opened 2023-02-17
[Heap Trace] Perf: use hash map to speed up leaks mode 34x (IDFGH-9425)

chipweinberger:
**Previous PR:** https://github.com/espressif/esp-idf/pull/10521

This finishes the perf work on heap trace standalone. Leaks mode is now fast enough to be enabled in production, or during development!

Adds `esp_err_t heap_trace_set_hashmap(heap_trace_hashmap_entry_t *entries_buffer, size_t num_entries);`

### Motivation:
- leaks mode should be fast enough to be enabled *in production*
- users can leave this mode enabled after boot & print leaks automatically (or save a dump to disk) when a low ram threshold is hit
- if enabled _before_ boot, leaks mode is a simple tool to find ways to save heap memory. This PR makes that use case fast.

### Testing:
- tested on an esp32s3. 

# Perf:

Most importantly! using a map makes record add & remove O(1), so we don't see a performance cliff when an old allocation is freed. 34x faster!

**SPIRAM** (NUMTABLE 2500, NUMITER 2, NUMENTRY 6000, NUMREC 3000)
**total 41ms** // no trace
**total 3401ms** // standard trace
**total 109ms**  // with map

For internal ram and a more typical NUMTABLE 500, we are 1.8x faster with the hashmap. 

**internal memory:**
**total 665ms** // no trace
**total 1814ms** // standard trace
**total 1065ms** // with map

For SPIRAM, also NUMTABLE 500, I'm surprised its not as dramatic. But still ~1.4x faster.

**SPIRAM:**
**total 683ms** // no trace
**total 2118ms** // standard trace
**total 1491ms** // with map

```
    const int NUMREC = 600;
    const int NUMENTRY = 600;
    const int NUMTABLE = 500;
    const int NUMITER = 100;

    uint32_t caps = MALLOC_CAP_INTERNAL;
    heap_trace_record_t* r = heap_caps_calloc(NUMREC, sizeof(heap_trace_record_t), caps);
    heap_trace_hashmap_entry_t* e = heap_caps_calloc(NUMENTRY, sizeof(heap_trace_hashmap_entry_t), caps);
    void** table = (void**) calloc(NUMTABLE, sizeof(void*));

    for (int i = 0; i < 3; i++) {

        int64_t t0 = esp_timer_get_time();

        if (i == 1 || i == 2) {
            heap_trace_init_standalone(r, NUMREC);
            if (i == 2) {heap_trace_set_hashmap(e, NUMENTRY);}
            heap_trace_start(HEAP_TRACE_LEAKS);
        }

        for(int j = 0; j < NUMITER; j++) {
            for (int k = 0; k < NUMTABLE; k++) {table[k] = heap_caps_malloc(1000,caps);}
            for (int k = 0; k < NUMTABLE; k++) {free(table[k]);}
        }

        int64_t t1 = esp_timer_get_time();
        printf("total %llims\n", (t1 - t0) / 1000);
        if (i == 1 || i == 2) {
            heap_trace_stop();
            heap_trace_dump();
        }
    }
```

### DIFF
--- a/components/heap/heap_trace_standalone.c
+++ b/components/heap/heap_trace_standalone.c
@@ -56,23 +56,42 @@ typedef struct {
     bool has_overflowed;
 } records_t;
 
+typedef struct {
+
+    /* Buffer used for hashmap entries */
+    heap_trace_hashmap_entry_t *buffer;
+
+    /* length of 'buffer' */
+    size_t count;
+} hashmap_t;
+
 // Forward Defines
 static void heap_trace_dump_base(bool internal_ram, bool psram);
-static void record_deep_copy(heap_trace_record_t *rDest, const heap_trace_record_t* r_src);
+static void record_deep_copy(heap_trace_record_t *r_dest, const heap_trace_record_t *r_src);
 static void list_setup(void);
-static void list_remove(heap_trace_record_t* r_remove);
-static bool list_add(const heap_trace_record_t* r_append);
+static void list_remove(heap_trace_record_t *r_remove);
+static heap_trace_record_t* list_add(const heap_trace_record_t *r_append);
 static heap_trace_record_t* list_pop_unused(void);
-static heap_trace_record_t* list_find_address_reverse(void* p);
+static heap_trace_record_t* list_find_address_reverse(void *p);
+static void map_add(const heap_trace_record_t *r_add);
+static void map_remove(void *p);
+static heap_trace_record_t* map_find(void *p);
 
 /* The actual records. */
 static records_t records;
+
+/* The hashmap */
+static hashmap_t map;
 
 /* Actual number of allocations logged */
 static size_t total_allocations;
 
 /* Actual number of frees logged */
 static size_t total_frees;
+
+/* track hits and misses */
+static size_t total_hashmap_hits;
+static size_t total_hashmap_miss;
 
 /* Used to speed up heap_trace_get */
 static heap_trace_record_t* r_get;
@@ -94,6 +113,19 @@ esp_err_t heap_trace_init_standalone(heap_trace_record_t *record_buffer, size_t 
     return ESP_OK;
 }
 
+
+esp_err_t heap_trace_set_hashmap(heap_trace_hashmap_entry_t *entries_buffer, size_t num_entries)
+{
+    if (tracing) {
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    map.buffer = entries_buffer;
+    map.count = num_entries;
+
+    return ESP_OK;
+}
+
 esp_err_t heap_trace_start(heap_trace_mode_t mode_param)
 {
     if (records.buffer == NULL || records.capacity == 0) {
@@ -105,8 +137,11 @@ esp_err_t heap_trace_start(heap_trace_mode_t mode_param)
     tracing = false;
     mode = mode_param;
 
-    // clear buffer
+    // clear buffers
     memset(records.buffer, 0, sizeof(heap_trace_record_t) * records.capacity);
+    if (map.buffer) {
+        memset(map.buffer, 0, sizeof(heap_trace_hashmap_entry_t) * map.count);
+    }
 
     records.count = 0;
     records.has_overflowed = false;
@@ -114,6 +149,8 @@ esp_err_t heap_trace_start(heap_trace_mode_t mode_param)
 
     total_allocations = 0;
     total_frees = 0;
+    total_hashmap_hits = 0;
+    total_hashmap_miss = 0;
     heap_trace_resume();
 
     portEXIT_CRITICAL(&trace_mux);
@@ -183,16 +220,10 @@ esp_err_t heap_trace_get(size_t index, heap_trace_record_t *r_out)
             }
         }
 
-        // copy to destination
-        if (r_get) {
-            memcpy(r_out, r_get, sizeof(heap_trace_record_t));
-        } else {
-            // this should not happen since we already
-            // checked that index < records.count,
-            // but could be indicative of memory corruption
-            result = ESP_ERR_INVALID_STATE;
-            memset(r_out, 0, sizeof(heap_trace_record_t));
-        }
+        // We already checked that index < records.count,
+        // This could be indicative of memory corruption.
+        assert(r_get != NULL);
+        memcpy(r_out, r_get, sizeof(heap_trace_record_t));
     }
 
     portEXIT_CRITICAL(&trace_mux);
@@ -213,6 +244,8 @@ esp_err_t heap_trace_summary(heap_trace_summary_t *summary)
     summary->capacity = records.capacity;
     summary->high_water_mark = records.high_water_mark;
     summary->has_overflowed = records.has_overflowed;
+    summary->total_hashmap_hits = total_hashmap_hits;
+    summary->total_hashmap_miss = total_hashmap_miss;
     portEXIT_CRITICAL(&trace_mux);
 
     return ESP_OK;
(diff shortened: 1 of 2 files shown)


chipweinberger:
@SoucheSouche  =)

chipweinberger:
This is (I think) unrelated to my PR, because it happens without the hashmap as well. 

But "totalAllocations" is not returning a sensible number when it hits 1919. Could there be a locking issue on master?


This is the exact code I'm using. ESP32-S3. 

[sdkconfig.txt](https://github.com/espressif/esp-idf/files/10772747/sdkconfig.txt)

```
    const int NUMREC = 3000;
    const int NUMENTRY = 6000;
    const int NUMTABLE = 2500;

    uint32_t caps = MALLOC_CAP_SPIRAM;
    heap_trace_record_t* r = heap_caps_calloc(NUMREC, sizeof(heap_trace_record_t), caps);
    heap_trace_hashmap_entry_t* e = heap_caps_calloc(NUMENTRY, sizeof(heap_trace_hashmap_entry_t), caps);
    void** table = (void**) calloc(NUMTABLE, sizeof(void*));

    printf("r %p\n", r);
    printf("e %p\n", e);
    printf("table %p\n", table);

    for (int a = 1917; a < 1921; a+=1) {

        for (int i = 0; i < 3; i++) {

            int64_t t0 = esp_timer_get_time();

            if (i == 1 || i == 2) {
                heap_trace_init_standalone(r, NUMREC);
                if (i == 2) {heap_trace_setup_hashmap(e, NUMENTRY);}
                heap_trace_start(HEAP_TRACE_LEAKS);
            }

            for(int j = 0; j < 1; j++) {
                for (int k = 0; k < a; k++) {table[k] = heap_caps_malloc(1000,caps);}
                for (int k = 0; k < a; k++) {free(table[k]); table[k] = 0;}
            }

            int64_t t1 = esp_timer_get_time();
            printf("total %llims\n", (t1 - t0) / 1000);
            if (i == 1 || i == 2) {
                heap_trace_stop();
                heap_trace_dump();
            }

            vTaskDelay(100);
        }
    }
```

chipweinberger:
I've noticed a weird issue, when NUMTABLE is > 1919, total_allocations in the summary does not go past 1919.

This happens with and without the hashmap, and regardless of SPIRAM vs INTERNAL. It seems worse when using INTERNAL. Could there be a locking issue on master? Maybe the lock we are using does not work?

SoucheSouche:
Hi @chipweinberger, I will get to this PR today :) I will investigate the 1919 total allocs issue as I will review and test the feature.

chipweinberger:
Very curious to know what you find! Hopefully you reproduce it.

chipweinberger:
This is my sdkconfig:
[sdkconfig.txt](https://github.com/espressif/esp-idf/files/10811245/sdkconfig.txt)

components/heap/heap_trace_standalone.c
```diff
@@ -161,38 +198,31 @@ esp_err_t heap_trace_get(size_t index, heap_trace_record_t *r_out)
...
-        } else {
-            // this should not happen since we already
-            // checked that index < records.count,
-            // but could be indicative of memory corruption
-            result = ESP_ERR_INVALID_STATE;
-            memset(r_out, 0, sizeof(heap_trace_record_t));
-        }
+        // We already checked that index < records.count,
+        // This could be indicative of memory corruption.
+        assert(rget != NULL);
```
SoucheSouche, line 225: Is it normal that the copy of `r_get`  to `r_out` was removed here ?

SoucheSouche:
I see that you renamed a lot of variables. The thing is that we are using snake case for variable names is esp-idf so it would be nice if you could update them. Otherwise, I will have to commit on top of your commits to change the name again.

I leave a patch with the fixed variable names below feel free to use it :) I hope I didn't miss any.
[heap_trace.patch](https://github.com/espressif/esp-idf/files/10839640/heap_trace.patch)

SoucheSouche:
`map_add`, `map_remove`, `map_find` and subsequently the static variables and function they utilize can be placed in flash by the compiler but are called from function that are specifically placed in IRAM. Everything should be placed in RAM but this means that - in the state of your PR - regardless of whether the hashmap is used or not, a substantial amount of RAM memory will be used to support the feature.

I think it would be wise to wrap the whole hashmap related logic under a config so that if the hashmap is not used, the impact of the heap trace on the RAM will remain the same.

components/heap/heap_trace_standalone.c
chipweinberger: Oh good spot. That is a bug.

I only intended to remove the error code, because I changed my mind. I think asserting on memory corruption is more appropriate.

SoucheSouche:
I'd keep the `r_` everywhere to be consistent across the file.

chipweinberger:
Okay. Updated the PR.

SoucheSouche:
Concerning the `1919 allocations` behavior, could it be that you simply run out of available heap? After all, you allocate NUMTABLE * sizeof(void*) bytes on the heap at the beginning of your test.

I tried the code on esp32s2 (randomly picked board), I added a `heap_caps_get_free_size(caps)` after the table calloc and got ~217000 bytes free. With a NUMTABLE set to 1000, it means that during the first iteration, I will try to allocate 1000000 bytes.

This is what I get: 
```
------- free size : 217656 ---------------

====== Heap Trace: 0 records (600 capacity) ======
====== Heap Trace Summary ======
Mode: Heap Trace Leaks
0 bytes 'leaked' in trace (0 allocations)
records: 0 (600 capacity, 215 high water mark)
hashmap: 600 capacity (21500 hits, 0 misses)
total allocations: 21500
total frees: 21500
```

Which means that 215000 bytes were successfully allocated. Given that for each allocation, there are a few overhead bytes related to the memory allocator algorithm, I would say that it adds up.

SoucheSouche:
if I reduce the allocation size to 16 bytes, Then I get the expected value of total allocations.
```
------- free size : 217656 ---------------

====== Heap Trace: 0 records (600 capacity) ======
====== Heap Trace Summary ======
Mode: Heap Trace Leaks
0 bytes 'leaked' in trace (0 allocations)
records: 0 (600 capacity, 600 high water mark)
hashmap: 0 capacity (0 hits, 0 misses)
total allocations: 100000
total frees: 100000
(NB: Internal Buffer has overflowed, so trace data is incomplete.)
================================

====== Heap Trace: 0 records (600 capacity) ======
====== Heap Trace Summary ======
Mode: Heap Trace Leaks
0 bytes 'leaked' in trace (0 allocations)
records: 0 (600 capacity, 600 high water mark)
hashmap: 600 capacity (60000 hits, 40000 misses)
total allocations: 100000
total frees: 100000
(NB: Internal Buffer has overflowed, so trace data is incomplete.)
```

chipweinberger:
> could it be that you simply run out of available heap?

of course! 🤦🏻‍♂️ Thanks for investigating it!

chipweinberger:
@SoucheSouche  latest status?

SoucheSouche:
Hi @chipweinberger, sorry for the late reply, I was away from work. However, your PR was reviewed while I was away and a suggestion was made for a more efficient implementation. I will have a look at it today and get back at you to let you know if we will go with the suggested implementation or if we will stick to yours.

chipweinberger:
no problem I hope your time away was enjoyable!

> more efficient

😀 nice!

SoucheSouche:
Hi @chipweinberger!
So I modified your hash map implementation as follow:
- `hash_map` array is now an array of linked list.
- when adding an entry to the hash map, we compute the hash of the entry and add it to the corresponding list in `hash_map`

This is reducing collisions in the `hash_map` array significantly and gives better performances for less RAM used overall.
In the case when `HEAP_TRACE_ALL` is used, it gives even better results (I just put them there for comparisons, but just for your information the tests were run on RISCV architecture):

For 1000000 allocations and free, the code with no tracing executes in 5779ms, The code tracing enabled and no hash map takes  40262ms, the code with your implementation of the has map takes 20632ms and the code with the hash map using the linked list as entries executes in 8129ms.

So the first implementation of the hash map is 2 times faster than without hash map usage.
The implementation of hash map with linked list is  5 times faster.

chipweinberger:
Interesting approach!

Yes in my design you the hash map has to big enough to always have spare room and remain fast.

Your approach seems like a nice simplification for the user.

Good work. I imagine that was tricky to code!

SoucheSouche:
The update was fairly straightforward and sat just well in your design. I mostly had to update the `map_x` functions.

chipweinberger:
Glad to hear it! 

I'm glad Espressif has taken an interest in the Console PRs. Hopefully not too much work - they're fully functional but of course "polish" takes awhile too.

chipweinberger:
merged: https://github.com/espressif/esp-idf/commit/b699033ab31c9bbc4a3304dbd64dc03112497316

Issue #10615, opened 2023-01-25
DPP Auth Failed on dpp-enrollee example (IDFGH-9228)


knight-ryu12:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v4.4.3

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

PowerShell

### Development Kit.

ESP32-WROVER-DevkitC

### Power Supply used.

USB

### What is the expected behavior?

I expected it to correctly authenticate with Accesspoint. 

### What is the actual behavior?

It does not authenticate with accesspoint at all;
Seems that it fails with WPA being `D (66148) wpa: Mgmt Tx Status - 1, Cookie - 0x400e036c`, and returns ESP_ERR_DPP_TX_FAILURE no matter what accesspoint it uses.

### Steps to reproduce.

[Code](https://gist.github.com/knight-ryu12/9d7d811984ce6f3090ec5d6714e2ac77)
It is 1:1 to the example, expect I removed the key part. which should just work without it.

### Debug Logs.

```plain
D (64538) wpa: DPP: Rx Public Action frame: action - 9
I (64538) wpa: DPP: Authentication Request from ac:3e:b1:02:80:e0
D (64538) event: running post WIFI_EVENT:20 with handler 0x400d6adc and context 0x3ffbdcfc on loop 0x3ffbc9f0
0x400d6adc: event_handler(void*, char const*, int, void*) at C:/Users/Alma/sensor-data-update/main/main.cpp:55

E (64548) wpa: curve: secp256r1

E (64558) wpa: prime len is 32

E (64558) wpa: Q_x: - hexdump(len=32):
E (64568) wpa: 2e cc 68 18 67 c2 7a 7f d2 59 20 1f 89 c5 99 28
E (64568) wpa: d2 59 3b 12 b9 07 13 4d 04 02 9a 80 0e 57 69 5c
E (64578) wpa: Q_y: - hexdump(len=32):
E (64578) wpa: 52 00 62 4e d9 33 98 b6 79 86 f6 ba 1c 20 a0 34
[...]

PR #10812, opened 2023-02-21
Wifi dpp fixes

gayafhannah:
Fix `ESP_ERR_DPP_TX_FAILURE` error caused by channel mismatch related issues(to the best of my knowledge) by setting `ESP_WIFI_MBO_SUPPORT` be a dependency of `ESP_WIFI_DPP_SUPPORT`
This has fixed all instances of the `ESP_ERR_DPP_TX_FAILURE` error and I have not been able to replicate the error with this fix yet.
- `components/esp_wifi/Kconfig`
- `examples/wifi/wifi-easy-connect/dpp-enrollee/sdkconfig.defaults`

---
Fix `esp_dpp_handle_config_obj` not copying null terminator of ssid and password.
Without this fix, I've quickly run into issues where unless I deinit and init DPP, subsequent attempts at configuring the network can lead to incorrect configs being sent to the callback.
An example of this is first using dpp to try connect to ExampleAP, but failing to connect due to signal reasons so you try connect it to SecondAP instead, without doing a deinit+init in-between, and as the null terminator of SecondAP isn't copied, on the second run you're still left with SecondAPP
- `components/wpa_supplicant/esp_supplicant/src/esp_dpp.c`

Related Issue: #10615


components/wpa_supplicant/esp_supplicant/src/esp_dpp.c
```diff
@@ -180,12 +180,12 @@ static int esp_dpp_handle_config_obj(struct dpp_authentication *auth,
     wifi_config_t *wifi_cfg = &s_dpp_ctx.wifi_cfg;
 
     if (conf->ssid_len) {
-        os_memcpy(wifi_cfg->sta.ssid, conf->ssid, conf->ssid_len);
+        os_strlcpy((char*)wifi_cfg->sta.ssid, (char*)conf->ssid, sizeof(wifi_cfg->sta.ssid));
```
jasta, line 183: I'd probably separate this stuff into two separate PRs.  You have on one hand trying to fix ESP_ERR_DPP_TX_FAILURE and on the other hand trying to make it work to let you bootstrap and/or start_listen multiple times.  I suspect the os_strlcpy fix is going to sale through more easily where as the MBO support is going to need a lot more careful thought to really understand what's going on there and why that would be relevant.

components/wpa_supplicant/esp_supplicant/src/esp_dpp.c
```diff
@@ -180,12 +180,12 @@ static int esp_dpp_handle_config_obj(struct dpp_authentication *auth,
...
     }
 
     if (dpp_akm_legacy(conf->akm)) {
         if (conf->passphrase[0])
-            os_memcpy(wifi_cfg->sta.password, conf->passphrase,
+            os_strlcpy((char*)wifi_cfg->sta.password, (char*)conf->passphrase,
```
jasta, line 188: This has a subtle bug in that if your first start_listen sequence yields a passphrase and your second one does not, you'll technically yield a wifi_config_t with a passphrase from the previous listen.  I think the fix is overall more consistent if you wipe the wifi_cfg data structure on each call to start_listen.  Same bug below with `pmf_cfg.required = true`.

jasta (changes requested):

components/wpa_supplicant/esp_supplicant/src/esp_dpp.c
jasta: Actually there are other existing bugs in this code that suggest wiping s_dpp_ctx.wifi_cfg on each invocation of esp_dpp_handle_config_obj is better.  The caller even runs this function in a loop which seems clearly broken to me...

Alvin1Zhang:
Thanks for your contribution.

components/esp_wifi/Kconfig
```diff
@@ -495,6 +495,7 @@ menu "Wi-Fi"
         bool "Enable DPP support"
         default n
         select ESP_WIFI_MBEDTLS_CRYPTO
+        select ESP_WIFI_MBO_SUPPORT
```
kapilkedawat, line 498: We are trying to reproduce this issue and fix it. these are two separate features and should have no dependency on each other.

gayafhannah: I know they are very unrelated, however this change is the only one that's let dpp consistently work near perfectly, compared to barely working unless the listen channel was set to that of the ap and phone(dont know why its an issue since the qr has the channel + the esp sends on the correct channel anyways).
Without this fix, `esp_wifi_action_tx_req` would in most cases I tested, fail by sending the 
`WIFI_EVENT_ACTION_TX_STATUS` event with a status=1, which in turn esp_dpp.c would send the `ESP_ERR_DPP_TX_FAILURE` event to the callback.
Without this fix, it would take making the environmental conditions near perfect for me to get a moderate success rate, but with this fix, even going out of my way I cant intentionally cause the `WIFI_EVENT_ACTION_TX_STATUS` related error.

components/wpa_supplicant/esp_supplicant/src/esp_dpp.c
gayafhannah: Ah, I must've missed this thanks to the two networks I tested with having the same passphrase, I'll have to look at wiping wifi_cfg. ty

components/wpa_supplicant/esp_supplicant/src/esp_dpp.c
gayafhannah: Yeah, I still don't personally know why exactly MBO support seems to fix it as well as it does. And it doesn't help that once you get deep enough into wifi related stuff, there isn't much in the way of documentation or anything.

components/esp_wifi/Kconfig
jasta: Fwiw on 4.4.4 I could not reproduce MBO having an effect.  What helped for me most was fixing the fact that retry doesn't work as it is supposed to in the example code (s_dpp_stop_listening is not reset on start_listen), then just spamming retry on my phone.  It is possible this MBO feature does something, but it's gonna take some tough work to figure out why.  What isn't up for debate tho is that DPP is highly unreliable for more than just me, and we definitely need to dig in and understand why.

jasta:
@gayafhannah checking in, are you planning to divide this PR so we can get the non-MBO fixes fast tracked?  If not I might be able to tackle that as I'd really like to base the Rust DPP support on top of a fixed version of esp-idf.  Honestly I'd think the severity of these bugs would warrant a new release in the 4.4.x branch given that this feature was mature in the 4.4.x series and 5.x is so new, though I'm not really sure what Espressif's policy is there.

Also @kapilkedawat would be good to hear from you what your preferences are moving forward here.  Even though we have some basic workarounds that make things work better, we still don't really have a fundamental understanding as to why it is so intermittent or why the tx failure occurs in the first place.  I'm recommending we move forward with the obvious fixes (e.g. s_dpp_stop_listening is not cleared so even the example code never retries as it claims to), but punt to Espressif for more sleuthing on the root cause.  I'm happy to help out there, but I don't have source code for the wifi driver and can't really tell why it would fail.

jasta:
Opened #10865 to tackle just the start_listen change

kapilkedawat:
Its just a coincidence that issue was not observed when MBO was enabled, MBO and DPP are independent features. therefore we cannot merge this PR.

PR #10827, opened 2023-02-22
Add function to just send a UART Break (without preceeding data) (IDFGH-9459)

phatpaul:
Added a function to just send a UART Break (without having to send data before the break).  
This would be useful for LIN bus and DMX512 protocols.  

**BTW, This code doesn't work yet.**  My hope is that someone can help me get this to actually work.
The code in this PR compiles, but when used it throws an assert.  I think it is something to do with the semaphores.

This would hopefully avoid the dreadful hacks that I am currently using to create a break:
```c
void LIN_driver_start_frame(uint8_t pid)
{
    uint8_t master_tx_buf[2];
    master_tx_buf[0] = 0x55; // sync byte
    master_tx_buf[1] = pid;

#if USE_SEND_BREAK_HACK_1
    /* Send break character */
    // this method sucks b/c it must send a dummy char before the break
    // Note: if use this, then +1 to LIN_IN_HEADER_SIZE for dummy byte.
    uart_write_bytes_with_break(lin_uart_num, &(char){0}, 1, LIN_BREAK_LEN);

#elif USE_SEND_BREAK_HACK2
    // this method really sucks because of a blocking delay!
    uart_set_line_inverse(lin_uart_num, UART_INVERSE_TXD);
    ets_delay_us(677);
    uart_set_line_inverse(lin_uart_num, UART_INVERSE_DISABLE);

#elif USE_SEND_BREAK_HACK3
    uart_hal_tx_break(UART2, 0); // crashes b/c of spinlock?
    uart_ll_tx_break(&UART2, 0); // crashes b/c of spinlock?

#elif USE_SEND_BREAK_HACK4 || 1
    // Temporarily change baudrate to slightly lower and send a 0 byte
    // to get the right baudrate, we want 9 bits (1 start + 8 data) to take the time of 13 bits (one break)
    uint32_t baudrate;
    uart_get_baudrate(lin_uart_num, &baudrate);
#define LIN_BREAK_BAUDRATE(BAUD) ((BAUD * 9) / 13)
    uart_set_baudrate(lin_uart_num, LIN_BREAK_BAUDRATE(baudrate));
    uart_write_bytes(lin_uart_num, &(char){0}, 1); // send a zero byte.  This call must be blocking.
    uart_wait_tx_done(lin_uart_num, 2);
    uart_wait_tx_done(lin_uart_num, 2);        // add 2nd uart_wait_tx_done per https://esp32.com/viewtopic.php?p=98456#p98456
    uart_set_baudrate(lin_uart_num, baudrate); // set baudrate back to normal after break is sent
    // we have to fake a break detect for the slave, since it doesn't detect it with this method.
    uart_event_t event = {.type = UART_BREAK};
    xQueueSend(lin_uart_event_queue, (void *)&event, 0);

#endif

    // Now send the sync and PID bytes
    uart_tx_chars(lin_uart_num, (char *)master_tx_buf, sizeof(master_tx_buf));
}
```
EDIT: these workarounds were discussed in this thread: https://esp32.com/viewtopic.php?f=13&t=8181

### DIFF
--- a/components/driver/include/driver/uart.h
+++ b/components/driver/include/driver/uart.h
@@ -540,6 +540,8 @@ int uart_write_bytes(uart_port_t uart_num, const void* src, size_t size);
  */
 int uart_write_bytes_with_break(uart_port_t uart_num, const void* src, size_t size, int brk_len);
 
+int uart_send_break(uart_port_t uart_num, int brk_len);
+
 /**
  * @brief UART read bytes from UART buffer
  *
--- a/components/driver/uart.c
+++ b/components/driver/uart.c
@@ -1239,6 +1239,29 @@ int uart_write_bytes_with_break(uart_port_t uart_num, const void *src, size_t si
     return uart_tx_all(uart_num, src, size, 1, brk_len);
 }
 
+int uart_send_break(uart_port_t uart_num, int brk_len)
+{
+    UART_CHECK((p_uart_obj[uart_num]->tx_buf_size == 0), "require tx_buf_size=0", (-1));
+    // lock for uart_tx
+    xSemaphoreTake(p_uart_obj[uart_num]->tx_mux, (portTickType)portMAX_DELAY);
+    p_uart_obj[uart_num]->coll_det_flg = false;
+
+    // semaphore for tx_fifo available
+    if (pdTRUE == xSemaphoreTake(p_uart_obj[uart_num]->tx_fifo_sem, (portTickType)portMAX_DELAY))
+    {
+        // uint32_t sent = uart_enable_tx_write_fifo(uart_num, (const uint8_t *)src, size);
+    }
+    uart_hal_clr_intsts_mask(&(uart_context[uart_num].hal), UART_INTR_TX_BRK_DONE);
+    UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
+    uart_hal_tx_break(&(uart_context[uart_num].hal), brk_len);
+    uart_hal_ena_intr_mask(&(uart_context[uart_num].hal), UART_INTR_TX_BRK_DONE);
+    UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
+    xSemaphoreTake(p_uart_obj[uart_num]->tx_brk_sem, (portTickType)portMAX_DELAY);
+    xSemaphoreGive(p_uart_obj[uart_num]->tx_fifo_sem);
+    xSemaphoreGive(p_uart_obj[uart_num]->tx_mux);
+    return 0;
+}
+
 static bool uart_check_buf_full(uart_port_t uart_num)
 {
     if (p_uart_obj[uart_num]->rx_buffer_full_flg) {


FloHomi:
Hello I have a solid working function for that now:
uart.h:

```c
int uart_write_bytes_with_break(uart_port_t uart_num, const void* src, size_t size, int brk_len);

/**
 * @brief Send a break signal on the UART TX line without sending any data bytes.
 *
 * This function waits until the TX line is idle (all previously queued/buffered data
 * has been transmitted), then sends a break signal of the given duration.
 * It does not return until the break signal has been sent out.
 *
 * @param uart_num UART port number, the max port number is (UART_NUM_MAX -1).
 * @param brk_len break signal duration(unit: the time it takes to send one bit at current baudrate)
 *
 * @return
 *     - (-1) Parameter error
 *     - (0) Break signal sent successfully
 */
int uart_write_break(uart_port_t uart_num, int brk_len);
```

uart.c:

```c
int uart_write_break(uart_port_t uart_num, int brk_len)
{
    ESP_RETURN_ON_FALSE((uart_num < UART_NUM_MAX), (-1), UART_TAG, "uart_num error");
    ESP_RETURN_ON_FALSE((p_uart_obj[uart_num]), (-1), UART_TAG, "uart driver error");
    ESP_RETURN_ON_FALSE((brk_len > 0 && brk_len < 256), (-1), UART_TAG, "break_num error");

    // The hardware break (txd_brk) is only appended after a data byte is transmitted.
    // To emit a standalone break without sending any data bytes, the idle-high TX line
    // is driven low (space level) for brk_len bit periods by inverting the TXD signal.
    uint32_t baud_rate = 0;
    if (uart_get_baudrate(uart_num, &baud_rate) != ESP_OK || baud_rate == 0) {
        return -1;
    }
    // Duration of the break in microseconds = brk_len bit periods at the current baudrate.
    uint32_t brk_us = (uint32_t)(((uint64_t)brk_len * 1000000ULL + baud_rate - 1) / baud_rate);

    //lock for uart_tx to serialize with other TX operations
    xSemaphoreTake(p_uart_obj[uart_num]->tx_mux, (TickType_t)portMAX_DELAY);
#if PROTECT_APB
    esp_pm_lock_acquire(p_uart_obj[uart_num]->pm_lock);
#endif
    //Wait until any previously queued/buffered data has been fully transmitted
    //so that the break signal is placed on an idle line.
    while (!uart_hal_is_tx_idle(&(uart_context[uart_num].hal))) {
        esp_rom_delay_us(1);
    }
    //Drive the idle-high TX line low to generate the break condition.
    UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
    uart_hal_inverse_signal(&(uart_context[uart_num].hal), UART_SIGNAL_TXD_INV);
    UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
    esp_rom_delay_us(brk_us);
    //Restore the TX line to its normal (non-inverted) idle-high level.
    UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
    uart_hal_inverse_signal(&(uart_context[uart_num].hal), 0);
    UART_EXIT_CRITICAL(&(uart_context[uart_num].spinlock));
#if PROTECT_APB
    esp_pm_lock_release(p_uart_obj[uart_num]->pm_lock);
#endif
    xSemaphoreGive(p_uart_obj[uart_num]->tx_mux);
    return 0;
}
```

phatpaul:
Thanks for sharing your solution! I actually forgot about this and have just been dealing with occasional sync errors. But I am very overdue to upgrade the IDF on my project, and LIN has been one of the hurdles.

I see a couple issues in your solution for my application:
1. In my LIN driver, the master and slave drivers are essentially separate.  I use the UART Rx in my LIN slave driver to sync on the BREAK signal. But I think in this solution, the master would change the polarity of the entire UART to create the BREAK, so the BREAK will not be received by the Rx side. I have tried to work around this with software hacks, but it never works 100% and leads to sync errors.  I did just think of another possible solution - use a separate UART for the Rx which I don't mess with. I.e. UART1 is Tx only and hacked to create BREAKs. UART2 is Rx only and can properly sync to breaks generated by other nodes or generated by the other UART on the same chip. Hopefully I can mux the 2 UART pins together.
2. esp_rom_delay_us() will block the entire cpu core I think. That is definitely going to cause issues somewhere else on my project which is using almost every peripheral and needs to service interrupts in a timely fashion.

This method avoids the esp_rom_delay_us() so other tasks can run during the write:
Perhaps combined with your proper use of the semaphores, locks, and critical sections, I can get it to not crash.
```
 // Temporarily change baudrate to slightly lower and send a 0 byte
    // to get the right baudrate, we want 9 bits (1 start + 8 data) to take the time of 13 bits (one break)
    uint32_t baudrate;
    uart_get_baudrate(lin_uart_num, &baudrate);
#define LIN_BREAK_BAUDRATE(BAUD) ((BAUD * 9) / 13)
    uart_set_baudrate(lin_uart_num, LIN_BREAK_BAUDRATE(baudrate));
    uart_write_bytes(lin_uart_num, &(char){0}, 1); // send a zero byte.  This call must be blocking.
    uart_wait_tx_done(lin_uart_num, 2);
    uart_wait_tx_done(lin_uart_num, 2);        // add 2nd uart_wait_tx_done per https://esp32.com/viewtopic.php?p=98456#p98456
    uart_set_baudrate(lin_uart_num, baudrate); // set baudrate back to normal after break is sent
```

PR #10828, opened 2023-02-22
Make export.fish exit with SUCCESS exit code (IDFGH-9460)

maxslarsson:
The export.fish script currently exits with an exit code of 4. Thus, any checks that make sure the `source` command exited successfully (exit code equal to zero) always fail. This was due to the last line trying to erase the `__main` function using the `set -e` command. In fish, you can only erase _variables_ using the `set` command. By changing the last line to erase the `__main` function with the `functions -e` command, the script now exits with an exit code of 0 instead of 4.

Also, while looking at the script, I noticed that the `unset` function seemed to be unused. My understanding was that the author of the script opted to directly use the `set -e` command in the script instead of the `unset` function, leaving it unused.

### DIFF
--- a/export.fish
+++ b/export.fish
@@ -1,10 +1,5 @@
 # This script should be sourced, not executed.
 
-# `idf_tools.py export --deactivate` create statement, with keyword unset, but fish shell support only `set --erase variable`
-function unset
-    set --erase $argv
-end
-
 function __main
     set script_dir (dirname (realpath (status -f)))
     if not set -q IDF_PATH
--- a/export.fish
+++ b/export.fish
@@ -106,6 +106,3 @@ if test $click_version -lt 8
 else
     eval (env _IDF.PY_COMPLETE=fish_source idf.py)
 end
-
-
-set -e __main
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

mfialaf:
@maxslarsson Thanks for your contribution!
Your change about erasing the `__main` function works well, thanks for that. I squashed those commits together, so it is merged as one change.

On the other hand, I had to remove you commit [Remove unused unset function in export.fish](https://github.com/espressif/esp-idf/pull/10828/commits/a4b3d746f48ff171fca5edc9e4a35fdcb567e47f). This function is used when users are switching from one ESP-IDF version to another in opened terminal. One of subcommand in export script then generate sequence of commands with keyword `unset` for unsetting global variables from the previous version. `unset` is accepted in all other shells but not fish. Because of that, the function is needed in `export.fish`.

maxslarsson:
Ahh I see! Thank you for letting me know and for taking the time to review my changes. Let me know if I can do anything else!

PR #10895, opened 2023-03-03
Fix possible conversion errors by using __builtin_ffsll (IDFGH-9541)

term-est:
Using C++20 with `-Wsign-conversion` causes my build to break as the function `xt_utils_set_watchpoint` takes `size` argument as of type `size_t` and passes it to `__builtin_ffs(unsigned int)`, which takes an `unsigned int` parameter.

Using the function `__builtin_ffsll`  instead should fix this.

Thank you for taking your time to review this pull request and have a great day!

### DIFF
--- a/components/xtensa/include/xt_utils.h
+++ b/components/xtensa/include/xt_utils.h
@@ -152,7 +152,7 @@ FORCE_INLINE_ATTR void xt_utils_set_watchpoint(int wp_num,
 {
     // Initialize DBREAKC bits (see Table 4–143 or isa_rm.pdf)
     uint32_t dbreakc_reg = 0x3F;
-    dbreakc_reg = dbreakc_reg << (__builtin_ffs(size) - 1);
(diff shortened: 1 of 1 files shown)


igrr:
@term-est Could you please describe how to reproduce the build issue? This is so that we could add a regression test along with your PR.

term-est:
> @term-est Could you please describe how to reproduce the build issue? This is so that we could add a regression test along with your PR.

I am sorry for the misunderstanding. `__builtin_ffs` breaks *my* build, and it's not an issue on ESP's part, nor the ESP's build system. 

We are using `-Werror` and `-Wsign-conversion` as part of our build process in our company for our C++ builds. and adding `uart.h` adds `xt_utils.h`, which triggers the build error for us.

I still think possible sign conversion errors might be avoided in the ESP-IDF repository, so I deiced to open up the pull request.
For clarification, this is not an issue on the ESP-IDF build system part, I am sorry for the misunderstanding.

0xjakob:
sha=14724d1cb967f252936bff160a69148489ad43bb

PR #10935, opened 2023-03-07
export: avoid error on unbound variable (IDFGH-9586)

MarcFinetRtone:
Might stop sourcing export.sh when `set -u` is used (typically with direnv):
> /path/to/esp-idf/export.sh:16: IDF_EXPORT_QUIET: unbound variable

### DIFF
--- a/export.sh
+++ b/export.sh
@@ -13,7 +13,7 @@ __realpath() {
 
 
 __verbose() {
-    [ -n "${IDF_EXPORT_QUIET}" ] && return
+    [ -n "${IDF_EXPORT_QUIET-}" ] && return
     echo "$@"
 }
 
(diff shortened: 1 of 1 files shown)


mfialaf:
Hi @MarcFinetRtone ,
Can you please explain in more detail what issue is solving your proposal change?
What is the expected output, and what is the actual without your change?
Thanks

MarcFinetRtone:
> Hi @MarcFinetRtone , Can you please explain in more detail what issue is solving your proposal change? What is the expected output, and what is the actual without your change? Thanks

To describe here in the PR or in the commit message ?

Regarding the change:
- without the change, if `set -u` is used in the shell sourcing export.sh stops on the unbound variable with the following error message:
- with the change sourcing export.sh works

mfialaf:
@MarcFinetRtone Thank you for clearing it up for me, I can see the issue now.

While testing it, I discovered the same problem with the unbound variable with `IDF_PATH` on line 75.
Can you please also fix this variable so we can merge it as one fix?
Thank you.

mfialaf (approved):
@MarcFinetRtone  Thank you for your contribution and cooperation.

dobairoland:
sha=13e59ed7a99f6bf7ee3d851faeee1fd4b3261ea5

Issue #9084, opened 2022-06-02
wear_levelling example don't work with 512K partitions size (IDFGH-7517)


nopnop2002:
## Environment
```
$ idf.py --version
ESP-IDF v5.0-dev-3202-ga2d5041492-dirty
```

## Problem Description

I tried [this](https://github.com/espressif/esp-idf/tree/master/examples/storage/wear_levelling).

I changed partitions_example.csv:

```
# Name,   Type, SubType, Offset,  Size, Flags
# Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
nvs,      data, nvs,     0x9000,  0x6000,
phy_init, data, phy,     0xf000,  0x1000,
factory,  app,  factory, 0x10000, 1M,
storage,  data, fat,            , 512K,
#storage,  data, fat,     ,        1M,
```
example don't work.

## Debug Logs

```
I (27) boot: ESP-IDF v5.0-dev-3202-ga2d5041492-dirty 2nd stage bootloader
I (27) boot: compile time 11:42:42
I (28) boot: chip revision: 1
I (32) boot_comm: chip revision: 1, min. bootloader chip revision: 0
I (39) boot.esp32: SPI Speed      : 40MHz
I (44) boot.esp32: SPI Mode       : DIO
I (48) boot.esp32: SPI Flash Size : 2MB
I (53) boot: Enabling RNG early entropy source...
I (58) boot: Partition Table:
I (62) boot: ## Label            Usage          Type ST Offset   Length
I (69) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (77) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (84) boot:  2 factory          factory app      00 00 00010000 00100000
I (91) boot:  3 storage          Unknown data     01 81 00110000 00080000
I (99) boot: End of partition table
I (103) boot_comm: chip revision: 1, min. application chip revision: 0
I (110) esp_image: segment 0: paddr=00010020 vaddr=3f400020 size=08864h ( 34916) map
I (132) esp_image: segment 1: paddr=0001888c vaddr=3ffb0000 size=02648h (  9800) load
I (136) esp_image: segment 2: paddr=0001aedc vaddr=40080000 size=0513ch ( 20796) load
I (147) esp_image: segment 3: paddr=00020020 vaddr=400d0020 size=1a704h (108292) map
[...]

PR #10945, opened 2023-03-08
Update sdkconfig.defaults (IDFGH-9596)

franz-ms-muc:
#9084 occurs if use the Sample as-is.


igrr:
Alternatively, could we increase the FAT partition size to 528kB?

franz-ms-muc:
Truuuue!

franz-ms-muc:
Alternate Solution here: #10947

PR #10947, opened 2023-03-08
Update partitions_example.csv (IDFGH-9599)

franz-ms-muc:
512K is too small, need to be 528K. see #9084


mahavirj:
sha=510592d86cbaaca40ad1c48841da4cd1f7e3b7dd

PR #10950, opened 2023-03-09
esp_system: Add reset reasons for USB_UART and USB_JTAG (IDFGH-9601)

nebkat:
ESP32 C3, C6, H2, H4 and S3 all have an SOC reset reason for USB UART and USB JTAG resets, but as these are not available in `esp_reset_reason_t`, `ESP_RST_UNKNOWN` is set instead. This is particularly unintuitive during development when it is one of the most common reset reasons.

This PR adds these to `esp_reset_reason_t`. There may be other reset reasons that should also be added, but these seem like a priority to me.


suda-morris:
Hi @nebkat I noticed this PR includes a change in the nimble submodule, should we revert it?

nebkat:
Apologies @suda-morris missed that somehow, have now force pushed with nimble submodule changes removed.

suda-morris:
sha=b48750d0549c6c38a8a75d3757302125d68512c0

Issue #9920, opened 2022-10-06
Support option to resolve hostname as IPv6 with esp_tls_conn_new_xxxx() (IDFGH-8458)


hideakitai:
### Is your feature request related to a problem?

`esp_tls_conn_new_xxxx()` cannot resolve hostname as IPv6 now. Because `esp_tls_hostname_to_fd()` that is called inside of it calls `getaddrinfo()` with fixed option: `hints.ai_family = AF_UNSPEC`. `getaddrinfo()` with `AF_UNSPEC` try to resolve hostname as IPv4 even when IPv4 address is not acquired (e.g. even on IPv6 only network).

https://github.com/espressif/esp-idf/blob/495d35949d50033ebcb89def98f107aa267388c0/components/esp-tls/esp_tls.c#L142-L156

### Describe the solution you'd like.

Could you support an option to `esp_tls_conn_new_xxxx()` to set how to resolve hostname (e.g. AF_UNSPEC, AF_INET, AF_INET6, etc.) ? 

### Additional context.

- IPv6 SLAAC is enabled
- DHCPv6 is disabled

```
CONFIG_LWIP_IPV6=y
CONFIG_LWIP_IPV6_AUTOCONFIG=y
CONFIG_LWIP_IPV6_RDNSS_MAX_DNS_SERVERS=16
```

Harshal5:
Hello, @hideakitai!

Thank you for the issue report. Could you please provide the following info -

- Target chip or module (ESP32, ESP32-S2, ESP32-S3, etc.)
- ESP-IDF version (run git describe --tags) / commit ID
- Example to reproduce the issue (or a minimal code snippet)
- Debug logs
- build sdkconfig

Alvin1Zhang:
Thanks for reporting, will close due to short of feedback, feel free to reopen with more updates. Thanks.

PR #10967, opened 2023-03-11
[esp-tls] Add addr_family option to esp_tls_cfg_t (IDFGH-9620)

mmmspatz:
Presently, esp-tls effectively does not support IPV6, as reported here: https://github.com/espressif/esp-idf/issues/9920
Basically: if DNS is able to return A records, esp-tls will always try to use them even if ipv4 connectivity is not available. However, if an A lookup fails, an AAAA lookup is attempted next, and IPV6 can work.

This behavior comes from passing `AF_UNSPEC` to `getaddrinfo()` [here](https://github.com/espressif/esp-idf/blob/47852846d3e89580ef4da28210b6ffc1cb5eaa9d/components/esp-tls/esp_tls.c#L161), which causes it to prefer IPV4 over IPV6 [here](https://github.com/espressif/esp-lwip/blob/10197b212a95c49c733fb18ffed56cafb0d196d4/src/api/netdb.c#L333). [here](https://github.com/espressif/esp-lwip/blob/10197b212a95c49c733fb18ffed56cafb0d196d4/src/core/dns.c#L1684) is the spot in LwIP where an AAAA lookup is done only after an A lookup fails.

I see two options to improve this:
1. Teach `tcp_connect()` to  make a decision about which address family to connect to. Perhaps if the first call to  `connect()` returns `EHOSTUNREACH`, do a second lookup passing `AF_INET6`, and then make a second connection attempt if an AAAA record is returned.
1. Add a field to `esp_tls_cfg_t` that allows the user to explicitly specify whether to pass `AF_UNSPEC`, `AF_INET`, or `AF_INET6` to `getaddrinfo()`.

In this PR, I have implemented option 2. Here is how I am using it in the application I am currently working on:
``` cpp
#include "esp_tls.h"
#include "lwip/nd6.h"

esp_tls_cfg_t tls_cfg = {
  // insert config here
};

// Test if we have a default ipv6 route, and if so make the connection over ipv6
const ip_addr_t test_dst = IPADDR6_INIT_HOST(0x20010002, 0x0, 0x0, 0x0);
if (nd6_find_route(ip_2_ip6(&test_dst))) {
    tls_cfg.addr_family = ESP_TLS_AF_INET6;
}

esp_tls_t* tls = esp_tls_init();
esp_tls_conn_new_sync(...,  &tls_cfg, tls);
```

One might argue that instead of adding a field to `esp_tls_cfg_t`, we should add an argument to the `esp_tls_conn_new*()` functions. The documentation for `esp_tls_conn_new_sync()` suggests that the `cfg` field can be NULL if "If you wish to open non-TLS connection", though I'm not sure it actually handles a NULL input safely. And that is what the `is_plain_tcp` field is for anyways. My reason for preferring to add a cfg field is that it is backwards compatible: If the `esp_tls_cfg_t` is zero initialized, the field defaults to `ESP_TLS_AF_UNSPEC`, which gives exactly the same behavior as before.

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -141,12 +141,24 @@ esp_tls_t *esp_tls_init(void)
     return tls;
 }
 
-static esp_err_t esp_tls_hostname_to_fd(const char *host, size_t hostlen, int port, struct sockaddr_storage *address, int* fd)
+static esp_err_t esp_tls_hostname_to_fd(const char *host, size_t hostlen, int port, esp_tls_addr_family_t addr_family, struct sockaddr_storage *address, int* fd)
 {
     struct addrinfo *address_info;
     struct addrinfo hints;
     memset(&hints, 0, sizeof(hints));
-    hints.ai_family = AF_UNSPEC;
+
+    switch(addr_family) {
+        case ESP_TLS_AF_INET:
+            hints.ai_family = AF_INET;
+            break;
+        case ESP_TLS_AF_INET6:
+            hints.ai_family = AF_INET6;
+            break;
+        default:
+            hints.ai_family = AF_UNSPEC;
+            break;
+    }
+
     hints.ai_socktype = SOCK_STREAM;
 
     char *use_host = strndup(host, hostlen);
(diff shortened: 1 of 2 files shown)


Alvin1Zhang:
Thanks for your contribution.

mahavirj:
sha=0abd1cb51f0a346adcda932e25182692b1f642e1

AxelLin:
@mahavirj 
Could you backport this to stable branch 5.0+?
https://www.esp32.com/viewtopic.php?f=13&t=33033

PR #11025, opened 2023-03-20
Enable support for C++23 in esp-idf

0xFEEDC0DE64:
std::expected, std::to_underlying, modules and concepts were missing in the esp-idf and this change enables all of them.

### DIFF
--- a/tools/cmake/build.cmake
+++ b/tools/cmake/build.cmake
@@ -129,7 +129,7 @@ function(__build_set_lang_version)
         # Use latest supported versions.
         # Please update docs/en/api-guides/cplusplus.rst when changing this.
         set(c_std gnu17)
-        set(cxx_std gnu++20)
+        set(cxx_std gnu++23)
     else()
         enable_language(C CXX)
         # Building for Linux target, fall back to an older version of the standard
@@ -149,7 +149,7 @@ function(__build_set_lang_version)
                                 "${preferred_c_versions}. Please upgrade the host compiler.")
         endif()
 
-        set(preferred_cxx_versions gnu++20 gnu++2a gnu++17 gnu++14)
+        set(preferred_cxx_versions gnu++23 gnu++20 gnu++2a gnu++17 gnu++14)
         set(ver_found FALSE)
         foreach(cxx_version ${preferred_cxx_versions})
             check_cxx_compiler_flag("-std=${cxx_version}" ver_${cxx_version}_supported)


0xjakob:
@0xFEEDC0DE64 Thanks for the suggestion! We are considering this PR but we first need to run tests on it and make sure that it won't affect anyone else.

0xFEEDC0DE64:
Can we make it a sdkconfig instead so you dont have to do testing? With defaults set to cpp20

igrr:
@0xFEEDC0DE64 To set a custom value of C++ standard for your component (without having to change the default value used for other components) you can use `set_property` CMake command on your component library target. For example, in component CMakeLists.txt file:
```cmake
idf_component_register(
    SRCS main.cpp
)

set_property(TARGET ${COMPONENT_LIB} PROPERTY CXX_STANDARD 23)
```
Then the component library will be compiled with an additional `-std=gnu++23` argument, which will override the `-std=gnu++20` argument added by default.

I've tried to compile the following example main.cpp file and it worked as expected:
```c++
#include <expected>

using std::expected;

extern "C" void app_main(void)
{
}
```

(Not saying that we won't update the default C++ language standard to C++23; just giving you a suggestion how to solve this problem without having to patch IDF or having to wait for us to do this update.)

0xjakob:
sha=a96f5bd5b098feb0e3ac39180bd8b19183f80907

0xFEEDC0DE64:
@igrr we just rebased to the latest esp-idf again, and the compiler is called again with -std=gnu++23 and -std=gnu++20. which results in compiler errors.

We are setting the c++ version with 
```
set(CMAKE_CXX_STANDARD 23)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS ON)
```
in the topmost CMakeLists.txt and I added your set_property() to each component in our firmware project
```
set_property(TARGET ${COMPONENT_LIB} PROPERTY CXX_STANDARD 23)
```

But still I have to patch build.cmake in esp-idf to finally make it work. What has been merged here?

0xFEEDC0DE64:
Maybe COMPONENT_LIB has to be replaced with COMPONENT_TARGET ?

igrr:
I think build.cmake has been updated in https://github.com/espressif/esp-idf/commit/a96f5bd5b098feb0e3ac39180bd8b19183f80907 already. Could you please verify you have this commit in your fork?

0xFEEDC0DE64:
Yes we have that commit, but for some reason the if branch for clang was needed to patch to get it to compile again. I dont understand why this happens.

https://github.com/0xFEEDC0DE64/esp-idf/commit/8e9e12a3b63ea75e1fc41cfc4260978ad2f68c25

0xjakob:
@0xFEEDC0DE64 Sorry for this. We tried to fix an incompatibility with our version of clang, but the code was faulty. We'll take a look ASAP.

0xjakob:
@0xFEEDC0DE64 Meanwhile, could you try the following in your component's CMakeLists.txt?
```
target_compile_options(${COMPONENT_LIB} PRIVATE -std=gnu++23)
```
When I use this, the standard is set twice
* First to gnu++20, by IDF general build commands, 
* The second time to gnu++23, from the component's CMakeLists.txt, which is the one prevailing.

I can then compile C++23 code (which fails compiling without setting the option in the component's CMakeLists.txt).

PR #11028, opened 2023-03-20
Bugfix: Fix compile error for ESP32 FSM ULP GPIO Example (IDFGH-9687)

Kampi:
Fix different compile errors in `ulp_example_main.c` when compiling the code with C++ 20 and above.

### DIFF
--- a/examples/system/ulp_fsm/ulp/main/ulp_example_main.c
+++ b/examples/system/ulp_fsm/ulp/main/ulp_example_main.c
@@ -1,7 +1,5 @@
 /* ULP Example
-
    This example code is in the Public Domain (or CC0 licensed, at your option.)
-
    Unless required by applicable law or agreed to in writing, this
    software is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
    CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


examples/system/ulp_fsm/ulp/main/ulp_example_main.c
```diff
@@ -98,27 +98,27 @@ static void init_ulp_program(void)
...
     const char* count_key = "count";
 
     ESP_ERROR_CHECK( nvs_flash_init() );
     nvs_handle_t handle;
-    ESP_ERROR_CHECK( nvs_open(namespace, NVS_READWRITE, &handle));
+    ESP_ERROR_CHECK( nvs_open(nvs_namespace, NVS_READWRITE, &handle));
     uint32_t pulse_count = 0;
     esp_err_t err = nvs_get_u32(handle, count_key, &pulse_count);
     assert(err == ESP_OK || err == ESP_ERR_NVS_NOT_FOUND);
-    printf("Read pulse count from NVS: %5"PRIu32"\n", pulse_count);
```
0xjakob, line 110: @Kampi What is the reason for this change? The formatting macros have been introduced to fix warnings after switching between compilers for different platforms (e.g. Xtensa-gcc to RISCV-gcc). If it is because of the C++ compiler warning about *invalid suffix on literal*, please add the suggested change instead of changing the macro back to a hard-coded formatter.

0xjakob (commented):
@Kampi Thanks a lot for this fix! It looks good except that I have one question regarding the changes. I think after it is resolved, we can merge it.

examples/system/ulp_fsm/ulp/main/ulp_example_main.c
0xjakob: @Kampi OK, thanks for the fix! Would you mind squashing the commits into one? Then we wouldn't have an intermediate commit that could potentially break builds on machines where the print formatters are different.

Kampi: Hi @0xjakob,
I can change it. What´s the best approach to do it? Shall I open a new pull request?

0xjakob: @Kampi OK, let's leave it as it is then. We will amend the commits to squash them. The only difference is that the PR will be seen as closed instead of merged. The rest stays the same. In particular, your authorship of the commit in IDF will be retained.

0xjakob:
sha=c96429e395ac13064038eee6d9163170725dcf24

examples/system/ulp_fsm/ulp/main/ulp_example_main.c
Kampi: Hi @0xjakob 
It's fine. Thanks!

PR #11041, opened 2023-03-22
Update tools.json (IDFGH-9701)

shreeve:
This fixes a hang on macOS, where the installation just sits there and does nothing. It was spinning because the version check for the esp-rom-elfs at https://github.com/espressif/esp-rom-elfs

### DIFF
--- a/tools/tools.json
+++ b/tools/tools.json
@@ -987,9 +987,10 @@
         "all"
       ],
       "version_cmd": [
-        ""
+        "true",
+        "--version"
       ],
-      "version_regex": "",
+      "version_regex": "(.?)",
       "versions": [
         {
           "any": {


Lapshin:
@shreeve, thanks for your contribution!
 
Could you please elaborate on how to reproduce this issue?
I tried to run `install.sh` script (ab63aaa4a24a05904da2862d627f3987ecbeafd0) for both x86 and arm MacOS and the problem you described not appeared.

Actually, I can't understand how is your fix works, because ROM-ELFs component has attribute ["is_executable": false,](https://github.com/espressif/esp-idf/pull/11041/files#diff-c5616b2ed67010e9d507b92123ad73c8eedb635e58db89d1a5eeee2a59ccfc20R983) and `version_cmd` and `version_regex` fields will be ignored

shreeve:
@Lapshin - Over the past two days, I have been trying to install `esp-idf`, but it had not worked. Finally today, I went to find out why. It ended up being that the software was waiting for the version check of the [esp-rom-elfs](https://github.com/espressif/esp-rom-elfs) and it would just sit there endlessly. I realized it has spawned a subprocess to check the version of the `esp-rom-elfs` and was waiting for the response, which never came... so the code just sat there forever. Once I added the *dummy* "true --version" and the always-matching pattern for the result, everything worked immediately. There must be a reason why the `is_executable` is not being honored and the version check is happening anyway.

I am on a macOS system with an M1 processor and running the latest version of Ventura 13.2.1.

In addition, there was another weird change that was later required. I had to do the following:

```
cd /Users/shreeve/.espressif/python_env
ls -la # idf5.1_py3.11_env shows as the only directory
ln -s idf5.1_py3.11_env idf5.0_py3.9_env
```

Then, the last stage of the install proceeds and everything is installed. These two tweaks fixed everything for me.

I am using v5.0.1, so I'm not sure why the above directory was name `idf5.1_py3.11_env`.

dobairoland:
> Actually, I can't understand how is your fix works, because ROM-ELFs component has attribute ["is_executable": false,](https://github.com/espressif/esp-idf/pull/11041/files#diff-c5616b2ed67010e9d507b92123ad73c8eedb635e58db89d1a5eeee2a59ccfc20R983) and version_cmd and version_regex fields will be ignored

I think this can happen if `idf_tools.py` is used from an older ESP-IDF (which doesn't know the flag yet) with a `tools.json` file from a newer ESP-IDF. 

> I am using v5.0.1, so I'm not sure why the above directory was name idf5.1_py3.11_env.

So actually this is causing the issue. Please try to remove `/Users/shreeve/.espressif/python_env` and don't create any such symbolic links. Make sure that your IDF_PATH points to the right version of ESP-IDF and run the install and export scripts again. Please save and share the logs if you still encounter the issue.

shreeve:
Roland,

The symbolic link that I did was done *after* I did the first fix. Without
the first fix (changing tools.json), I would never have made it to even see
the second problem.

Steve

dobairoland:
>The symbolic link that I did was done *after* I did the first fix. Without the first fix (changing tools.json), I would never have made it to even see the second problem.

I know that the issue is not caused by the symbolic link but explains your issue. You have multiple versions of ESP-IDF (or you have switched between them during or after the installation). That is causing your issue. Please clean up your system as I suggested.

dobairoland:
I'm closing this because this is just a workaround for the issue present in your installation. I believe we have left enough clues for you to solve it. Please let us know if you need more help.

Issue #11049, opened 2023-03-24
Global log enable/disable (IDFGH-9712)


someburner:
Any thoughts on adding a global variable that can be used to enable/disable all logs at runtime?

Use case:

In some applications, it may be desirable to have many logs compiled (e.g. set a high CONFIG_LOG_MAXIMUM_LEVEL), but not impose a performance hit during normal operation, and only do so when instructed. E.g. for remote debugging, or perhaps having conditions in code where you want to start seeing logs.

The log cache seems to be fairly performant, but as an example, in my project if I compile in logs at `LOG_LEVEL_DEBUG`, but disable them at runtime with `esp_log_level_set("*", ESP_LOG_NONE);`, there is enough of a performance hit that it warrants looking into.

```cpp
const int64_t MEASUREMENTS = 1000;

int64_t start = esp_timer_get_time();

for (int retries = 0; retries < MEASUREMENTS; retries++) {
    ESP_LOGI(TAG, "some test data, %d, %d, %d", retries, MEASUREMENTS-retries, 12);
}

int64_t end = esp_timer_get_time();

printf("%lld iterations took %lld microseconds (%lld microseconds per invocation)\n",
   MEASUREMENTS, (end - start), (end - start)/MEASUREMENTS);
```

For me, this results in

```
1000 iterations took 9069 microseconds (9 microseconds per invocation)
```

Not too bad but if you have a lot of logs, it can start to add up.

If instead we add a global log variable and do something like this:

```c
// log.c
bool g_logs_enabled = true;

// esp_log.h
extern bool g_logs_enabled;

#define ESP_LOG_LEVEL_LOCAL(level, tag, format, ...) do {               \
        if ( (g_logs_enabled) && (LOG_LOCAL_LEVEL >= level) ) ESP_LOG_LEVEL(level, tag, format, ##__VA_ARGS__); \
    } while(0)
```

And switch that on/off each time with the test code:

```c
g_logs_enabled = !g_logs_enabled;
```

You get:

```sh
# g_logs_enabled = false
1000 iterations took 90 microseconds (0 microseconds per invocation)
# g_logs_enabled = true (but log_level enabled)
1000 iterations took 9069 microseconds (9 microseconds per invocation)
```

Which is much more acceptable, but adds about 15KB to program size in my case. Any pitfalls with this / is this a bad idea? Should I be measuring differently?

atanisoft:
[...]

PR #11057, opened 2023-03-26
log: add master log option (IDFGH-9721)

someburner:
see https://github.com/espressif/esp-idf/issues/11049

Adds optional `LOG_MASTER_LEVEL` Kconfig option to log component. Results in 100x performance increase on ESP32 at 240Mhz when logs are compiled in, but disabled at runtime. Useful in cases where you only want log output some of the time. Results in a non-negligible increase in code size with lots of logs enabled, but that is also when you have the most to gain performance-wise.

I have not added test cases pending feedback.

Other things worth considering:

- if this should/could be applied to the early log macros
- Putting this in the `noinit` section so setting is preserved across boots.

Test it out for yourself:

- Compile with Max/Default = LOG_LEVEL_DEBUG.

```c
#include <stdio.h>
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "esp_log.h"

static const char * TAG = "sometag";

void app_main(void)
{
    ESP_LOGI(TAG, "app_main init");

    esp_log_level_set("*", ESP_LOG_NONE);

    const int64_t MEASUREMENTS = 1000;

    while (true) {
        vTaskDelay( 1000 / portTICK_PERIOD_MS );

        int64_t start = esp_timer_get_time();

        for (int retries = 0; retries < MEASUREMENTS; retries++) {
            ESP_LOGI(TAG, "some test data, %d, %d, %d", retries, MEASUREMENTS-retries, 12);
        }

        int64_t end = esp_timer_get_time();

        printf("%lld iterations took %lld microseconds (%lld microseconds per invocation)\n",
           MEASUREMENTS, (end - start), (end - start)/MEASUREMENTS);

        if (g_master_log_level == ESP_LOG_NONE) {
            g_master_log_level = ESP_LOG_DEBUG;
        } else {
            g_master_log_level = ESP_LOG_NONE;
        }
    }
}
```

### DIFF
--- a/components/log/Kconfig
+++ b/components/log/Kconfig
@@ -84,6 +84,22 @@ menu "Log output"
         default 4 if LOG_MAXIMUM_LEVEL_DEBUG
         default 5 if LOG_MAXIMUM_LEVEL_VERBOSE
 
+    config LOG_MASTER_LEVEL
+        bool "Enable global master log level"
+        default "n"
+        help
+            Enables an additional global "master" log level check that occurs
+            before a log tag cache lookup. This is useful if you want to
+            compile in a lot of logs that are selectable at runtime, but avoid the
+            performance hit during periods where you don't want log output. Examples
+            include remote log forwarding, or disabling logs during a time-critical
+            or CPU-intensive section and re-enabling them later. Results in
+            larger program size depending on number of logs compiled in.
+
+            If enabled, defaults to LOG_DEFAULT_LEVEL and can be set using
+            esp_log_level_master_set().
+            This check takes precedence over ESP_LOG_LEVEL_LOCAL.
+
     config LOG_COLORS
         bool "Use ANSI terminal colors in log output"
         default "y"
--- a/components/log/include/esp_log.h
+++ b/components/log/include/esp_log.h
@@ -41,6 +41,25 @@ typedef int (*vprintf_like_t)(const char *, va_list);
  */
 extern esp_log_level_t esp_log_default_level;
 
+#ifdef CONFIG_LOG_MASTER_LEVEL
+/**
+ * @brief Master log level.
+ *
+ * Optional master log level to check against for ESP_LOGx macros before calling
+ * esp_log_write. Allows one to set a higher CONFIG_LOG_MAXIMUM_LEVEL but not
+ * impose a performance hit during normal operation (only when instructed). An
+ * application may set esp_log_level_master_set(level) to globally enforce a
+ * maximum log level. ESP_LOGx macros above this level will be skipped immediately,
+ * rather than calling esp_log_write and doing a cache hit.
+ *
+ * The tradeoff is increased application size.
+ */
+extern esp_log_level_t g_master_log_level;
+void esp_log_level_master_set(esp_log_level_t level);
+#else
+#define esp_log_level_master_set(level) do { } while(0)
+#endif //CONFIG_LOG_MASTER_LEVEL
+
 /**
  * @brief Set log level for given tag
  *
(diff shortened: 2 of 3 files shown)


igrr:
@someburner Thanks for the PR!

This feature seems to have an odd interaction with `esp_log_level_set`. It essentially overrides the value that can be set using `esp_log_level_set`, which i think is counter-intuitive, especially if `esp_log_level_set` is called later than `esp_log_level_master_set`.

Could we keep the existing API for setting the global log level? `esp_log_level_set("*", ...);` could set the "global" log level to the desired value. If the user calls `esp_log_level_set("any other tag", ...);` afterwards would cancel the "global" setting and set the specific level for the desired tag.

Something along these lines:
```c++
esp_log_level_set("mytag", ESP_LOG_DEBUG);  // (1) lowers "mytag" level to "DEBUG"
esp_log_level_set("mytag", ESP_LOG_NONE);  // (2) all logs (including "mytag") now are disabled
esp_log_level_set("mytag", ESP_LOG_DEBUG);  // (3) now "mytag" is enabled, everything else is disabled
```

If the new Kconfig option is enabled, after point (2) the new "global" "minimal_log_level" variable would be set to ESP_LOG_NONE, and the checks for logs-disabled would be very quick. After point (3), "minimal_log_level" would again be set to ESP_LOG_DEBUG, so the checks would become slower, but the behavior would still be the same as it is now.

This way, the component authors don't need to make assumptions whether the Kconfig option is set and write code accordingly. The behavior would be the same, just the performance/code size tradeoff would be different.

Please let me know if this makes sense!

someburner:
@igrr that makes sense. I did think the API was a bit funky. So in this case you're suggesting to still keep this feature optional? And is it better to do a new PR or just add commits to this one.

igrr:
I think it should be optional if the global minimal log level is checked inside the log macro — since that leads to a significant increase in binary size.

(If you can tolerate slightly lower performance and check the global log level inside esp_log_write then it doesn't have to be optional.)

You can push more commits to this branch, and eventually squash the commits (git rebase -i) once you are done working on the feature.

someburner:
I feel like this feature is targeted at performance, the 100x->10x hit is significant so I'll leave it optional.

Regarding your suggestion though-  The only issue I see here is that some components call `esp_log_level_set` themselves. It looks like those are mostly guarded with config definitions and are mostly in the drivers folder.

In particular there is `mbedtls_esp_enable_debug_log`, and also in `esp_wifi/src/wifi_init.c`.

So my concern is something else in the SDK calling `esp_log_level_set`, which then unsets  the `esp_log_level_set("*", ESP_LOG_NONE);`, and then the user can't be sure what state it's in.

It does seem like most of those are done in initialization functions. It would need to be clear in documentation exactly which components/options can undo this setting, and when.

KonstantinKondrashov:
Hi @someburner!
I agree that your feature is useful. Thanks for bringing this issue up.

> ... Useful in cases where you only want log output some of the time. 

It makes me think that you just want to enable and disable all logs. So I think a better solution that fits your needs would be to add APIs: esp_log_enable() and esp_log_disable(), without setting the log level.

So I see the usage like this: 
- first, you disable all logs, esp_log_enable()
- then set the correct level using esp_log_level_set("mytag", ESP_LOG_DEBUG)
- then enable the logs according to level settings. esp_log_disable()
From my perspective, the usage of these APIs (esp_log_level_set and esp_log_enable/disable) getting more clear.
And I would prefer to put the logic into the log function to skip code size overhead.

FYI, I am working on refactoring the log component so it will be much easier to add your feature after this work will be done.

igrr:
Regarding esp_log_enable/disable, one remark is that it's difficult to deal with such centralized log control unless you are in control of the entire application code. For example, if one of the components you depend on decides to use esp_log_enable/disable, that might affect logging from other components, and you won't be able to do much about the issue without modifying the component source code.
Ideally, log level settings should be composable and one component should be able to control it's own logs without affecting the rest of the system.

chipweinberger:
In my use case, I want to be able to turn on debug logs at runtime, but have them off by default without a big perf hit.

I think a master switch, which is only called by the user is the best approach.

Use a counting semaphore. +1 on *disable*, -1 on *enable*. And have a semaphore per log level.

PR #11063, opened 2023-03-27
The ESP-NOW frame length (IDFGH-9729)

petrkucerak:
The Issue from ESP-NOW repository https://github.com/espressif/esp-now/issues/59

In ESP-NOW documentation is described the frame format is as:

```
------------------------------------------------------------------------------------------------------------
| MAC Header | Category Code | Organization Identifier | Random Values | Vendor Specific Content |   FCS   |
------------------------------------------------------------------------------------------------------------
  24 bytes         1 byte              3 bytes               4 bytes             7~255 bytes        4 bytes
```

and the Vendor Specific Content:

```
-------------------------------------------------------------------------------
| Element ID | Length | Organization Identifier | Type | Version |    Body    |
-------------------------------------------------------------------------------
    1 byte     1 byte            3 bytes         1 byte   1 byte   0~250 bytes
```

I'd like to ask how is possible that the body length is 250 B, but if the Body is full, Vendor Specific Content is 257 B.

Is a problem in my thought or the specification is wrong?

Link for documentation: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_now.html
The Reddit discusion: https://www.reddit.com/r/esp32/comments/11ruo1w/the_espnow_frame_length/


zhangyanjiaoesp:
@petrkucerak 
ESP-NOW uses vendor-specific action frames.
These contain 1 byte Category, 3 bytes Organization Identifier and Vendor Specific Content.

According to the 802.11 standard the Vendor Specific Content could contain multiple Vendor-Specific Fields.
802.11-2020, 9.6.5 Vendor-specific action details: "...The Vendor Specific Content field contains vendor-specific field(s)."

But right now ESP-NOW is only using one Vendor-Specific IE, which is limited to 255 bytes.
So the `Vendor Specific Content` should be `7~257`. Will change the doc ASAP.

zhangyanjiaoesp:
@petrkucerak The change has been merged into deab8d345552e55b89a13d8f7c1f5e16f6605925

PR #11079, opened 2023-03-30
Fix preprocessor condition in esp_tls_mbedtls (IDFGH-9746)

boarchuz:
CONFIG_LOG_DEFAULT_LEVEL_DEBUG is only defined if that specific level is selected, in which case the condition was unexpectedly evaluating as false.
For all other levels CONFIG_LOG_DEFAULT_LEVEL_DEBUG is undefined, so the condition is equivalent to "CONFIG_LOG_DEFAULT_LEVEL > 0", which was usually unexpectedly true.


laukik-hase:
sha=0c85f7407e84842a290225e1e7aeb0ed0d279ea0

PR #11116, opened 2023-04-03
nvs_host_test: replace sprintf with snprintf (IDFGH-9781)

codysch:
With some libc impls (ie: Xcode macos sdk 13.1), sprintf is marked as deprecated and causes the build with `-Werror` to fail. Switch the `sprintf` usage in test_nvs.cpp to use `snprintf` to allow builds to complete without warning.


Alvin1Zhang:
Thanks for your contribution.

igrr (approved):

PR #11117, opened 2023-04-03
test_nvs_host: fix asan reported bugs (IDFGH-9860)

codysch:
- Use of `sizeof(uint32_t *)` instead of `sizeof(uint32_t)` in arguments
 - Use of `memcmp` without checking that length agreed (add `memeq`)

These were observed when running these tests built with clang 14 on a MacOS 12.6.3 system (Xcode 13.1). I suspect some of them will appear on any system where `uint32_t *` and `uint32_t` are different sizes, and where reading some data _should_ cause the test to fail (due to lack of certain python tooling)


igrr:
sha=630343e4b9a706b25cf06c8c407463c774ad6db7

PR #11118, opened 2023-04-03
nvs: support iteration over namespace by index

codysch:
Users of the nvs API are likely to have `nvs_handle_t` in all cases, but not all of them carry around the partition name and namespace name (as they aren't needed after creating an `nvs_handle_t`).

Allow this common case to use nvs iteration without them tracking additional data by using the `nvs_handle_t` to locate the namespace index and the partition name by introducing an alterate to `nvs_entry_find` called `nvs_entry_find_in_handle`, which operates similarly except that it is given a `nvs_handle_t` instead of partition and namespace strings.

This is somewhat more limited than the `nvs_entry_find` API as one cannot examine all keys in a given partition.


Alvin1Zhang:
Thanks for your contribution.

igrr (approved):

components/nvs_flash/include/nvs.h
```diff
@@ -661,14 +661,52 @@ esp_err_t nvs_entry_find(const char *part_name,
...
+ *  while(res == ESP_OK) {
+ *      nvs_entry_info_t info;
+ *      nvs_entry_info(it, &info); // Can omit error check if parameters are guaranteed to be non-NULL
+ *      printf("key '%s', type '%d' \n", info.key, info.type);
+ *      res = nvs_entry_next(&it);
+ *  }
+ *  nvs_release_iterator(it);
+ * \endcode
+ *
+ * @param[in]   handle          Handle obtained from nvs_open function.
```
igrr, line 680: Over here Doxygen complains that the argument name (handle) doesn't match the argument name in function declaration (c_handle). I'll fix this to use `handle` name in the declaration, consistent with the rest of the functions.

jmesmon: Ah, whoops, ya. I originally wrote this change for the esp-idf 4.x and then ported it to 5.x. Seems I didn't fully fix the doxygen comments after that.

igrr:
Cherry-picked in 352e759d1dc15014e2e8194e6f8113bc697d4c54, thanks for the contribution @codysch!

PR #11136, opened 2023-04-05
Wording change (IDFGH-9801)

idea--list:
Change the description of EXAMPLE_RESET_PROV_MGR_ON_FAILURE to be more clear
I find the original wording misleading as the option not only resets during the provisioning session, but also any time later if the the ESP can not connect to the provisioned WiFi AP.

(As any radio communications have down times per se, not sure if such an automatism would be useful in production by the way)

### DIFF
--- a/examples/provisioning/wifi_prov_mgr/main/Kconfig.projbuild
+++ b/examples/provisioning/wifi_prov_mgr/main/Kconfig.projbuild
@@ -77,7 +77,7 @@ menu "Example Configuration"
     config EXAMPLE_PROV_MGR_MAX_RETRY_CNT
         int
         default 5
-        prompt "Max WiFi connection retries before reseting the provisioning state machine"
+        prompt "Max WiFi connection retries before resetting the provisioning state machine"
         depends on EXAMPLE_RESET_PROV_MGR_ON_FAILURE
         help
             Set the Maximum retry limit to avoid trying to connect to an AP that is not within reach or if credentials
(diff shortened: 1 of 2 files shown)


examples/provisioning/wifi_prov_mgr/main/Kconfig.projbuild
```diff
@@ -69,18 +69,18 @@ menu "Example Configuration"
...
-            Enable reseting provisioned credentials and state machine after session failure.
-            This will restart the provisioning service after retries are exhausted.
+            Enable reseting provisioned credentials and state machine after connection retry limit has been reached.
+            This will restart the provisioning service after WiFi connection retries are exhausted.
 
     config EXAMPLE_PROV_MGR_MAX_RETRY_CNT
         int
         default 5
-        prompt "Max retries before reseting provisioning state machine"
+        prompt "Max WiFi connection retries before reseting the provisioning state machine"
```
mahavirj, line 80:
```suggestion
        prompt "Max WiFi connection retries before resetting the provisioning state machine"
```

mahavirj (commented):
Please fix one existing typo, otherwise looks good to me.

idea--list:
Oh my... made another pull request which is not related to this PR but somehow my 2 PRs interfered...

idea--list:
Tested my PR again and it seems that neither of the changes are valid.
So please do NOT merge this. Sorry for that, something was messed up on my side.

PR #11163, opened 2023-04-08
Make custom bootloader message match actual output (IDFGH-9834)

hasheddan:
Fixes the custom bootloader README.md so that the example output matches what will actually be output by default in the custom bootloader.


ESP-Marius:
Thanks for the fix, LGTM!

ESP-Marius:
sha=f7736a6dd730ee93942cdd4dfcb193009d1916c0

PR #11191, opened 2023-04-13
softap: make AP IP address and netmask configurable w/ kconfig

MarcFinetRtone:
The current method to change the IP address of the SoftAP network is to:
- deactivate the DHCP server
- change the IPinfo
- re-activate the DHCP

All for values defined at compile time.

This commit introduces a way to specify the IPv4 address of the AP and the netmask with length rather than specifying 255 (or more 'strange' values). Defaults values still map to 192.168.4.1/24.

On my project with error handling w/ log, it saves ~1kB code and 200B of data.


components/esp_wifi/Kconfig
```diff
@@ -562,4 +562,56 @@ menu "Wi-Fi"
         help
             Select this to enable unity test for DPP.
 
+
+    menu "WiFi SoftAP specific Network address"
```
xueyunfei998, line 566: hi @MarcFinetRtone 

The proposed change is put under this path.

components/lwip/Kconfig

igrr:
> The current method to change the IP address of the SoftAP network is to:

Another (simpler) option is to replace `esp_netif_create_default_wifi_ap()` with customized code to create the interface, for example:
```c++
    esp_netif_config_t cfg = ESP_NETIF_DEFAULT_WIFI_AP();
    esp_netif_inherent_config_t base_cfg = cfg.base;
    base_cfg.ip_info = (modify the IP address);
    esp_netif_t *netif = esp_netif_new(&cfg);
    esp_netif_attach_wifi_ap(netif);
    esp_wifi_set_default_wifi_ap_handlers();
```

`esp_netif_create_default_wifi_ap` is provided only as the default implementation, however other `esp_netif` APIs are available for you to handle specific use cases.

jack0c:
Close the pull request since esp-idf already provide a way to do this.

PR #11204, opened 2023-04-16
Fix function signature in flash suspend example (IDFGH-9888)

hasheddan:
Fixes misspelling of function signature in the flash suspend example.


igrr (approved):

igrr:
sha=7a192131986745a83f468821a70cc4bc6b41802b

PR #11214, opened 2023-04-17
examples/bluedroid/gattc_demo.c - Use correct struct type (IDFGH-9899)

ckcr4lyf:
If the event is `ESP_GATTC_DIS_SRVC_CMPL_EVT` , then the struct should be `gattc_dis_srvc_cmpl_evt_param` as per https://github.com/espressif/esp-idf/blob/f404fe96b17692e3f1de536a3d73a180cdb53b42/components/bt/host/bluedroid/api/include/api/esp_gattc_api.h#L243-L249

Currently it is using `gattc_cfg_mtu_evt_param` , which is defined as https://github.com/espressif/esp-idf/blob/f404fe96b17692e3f1de536a3d73a180cdb53b42/components/bt/host/bluedroid/api/include/api/esp_gattc_api.h#L98-L105

Through coincidence, it does work since in memory the field `.conn_id` is at the same offset. However I think it is better to use the correct struct type , so it is less confusing to someone reading the example.

### DIFF
--- a/examples/bluetooth/bluedroid/ble/gatt_client/main/gattc_demo.c
+++ b/examples/bluetooth/bluedroid/ble/gatt_client/main/gattc_demo.c
@@ -130,7 +130,7 @@ static void gattc_profile_event_handler(esp_gattc_cb_event_t event, esp_gatt_if_
             break;
         }
         ESP_LOGI(GATTC_TAG, "discover service complete conn_id %d", param->dis_srvc_cmpl.conn_id);
-        esp_ble_gattc_search_service(gattc_if, param->cfg_mtu.conn_id, &remote_filter_service_uuid);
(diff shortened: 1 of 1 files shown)


ckcr4lyf:
I keep signing the CLA, it keeps removing it. Is this a bug?

image

esp-cjh:
@ckcr4lyf We will incorporate your updates in the subsequent versions. Thanks.

weiyuhannnn:
@ckcr4lyf  Thanks for your contribution. The change you suggested has been merged with the following commits:

- master / v5.3 branch: be29217eeb4b2f247c8e6813bf8c23b161a407b8
- v5.2 : daa6a0d5c2f8827cd360724896a818a13ef2bcfd
- v5.1: 8519d3c50981d242231027c6f6046b6d4c05d0be
- v5.0: d388c85fd248ff5343cf984f47a1b3acec607470

PR #11239, opened 2023-04-20
build: make default warnings as errors (IDFGH-9945)

MarcFinetRtone:
The `-Werror=all` activates error for all warnings in `-Wall`, however, it does not activate error for other default warnings, such as:
- `int-conversion` (pointer from integer w/o a cast)
- `incompatible-pointer-types`
- `discarded-qualifiers`

Which are IMO even more important that `-Wall`.

This commit fixes that by activating error for all warnings (i.e. from `-Wall` and default ones) and removing those from `-Wextra`, as the culprit commit seemed to address.

Fixes: 60f29236f6 "Build system: Raise warning level" (2016-11-16)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,6 +184,11 @@ if(CONFIG_COMPILER_DISABLE_GCC12_WARNINGS)
                                 "-Wno-use-after-free")
 endif()
 
+if(CONFIG_COMPILER_DISABLE_DEFAULT_ERRORS)
+    list(APPEND compile_options "-Wno-error"
+                                "-Werror=all")
+endif()
+
 # GCC-specific options
 if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     list(APPEND compile_options "-fstrict-volatile-bitfields"
--- a/Kconfig
+++ b/Kconfig
@@ -504,6 +504,19 @@ mainmenu "Espressif IoT Development Framework Configuration"
 
                 This option can be enabled for RISC-V targets only.
 
+        config COMPILER_DISABLE_DEFAULT_ERRORS
+            bool "Disabe errors for default warnings"
+            default "n"
+            help
+                Enable this option if you want to not consider default warnings as errors,
+                especially when upgrading IDF.
+
+                This is a temporary flag that could help to allow upgrade while having
+                some time to address the warnings raised by those default warnings.
+                Alternatives are: 1) fix code (preferred),  2) remove specific warnings
+                or 3) do not consider specific warnings as error.
+
+
         config COMPILER_DISABLE_GCC12_WARNINGS
             bool "Disable new warnings introduced in GCC 12"
             default "n"
--- a/tools/cmake/build.cmake
+++ b/tools/cmake/build.cmake
@@ -99,12 +99,13 @@ function(__build_set_default_build_specifications)
                                     "-fdata-sections"
                                     # warning-related flags
                                     "-Wall"
-                                    "-Werror=all"
+                                    "-Werror"
                                     "-Wno-error=unused-function"
                                     "-Wno-error=unused-variable"
                                     "-Wno-error=unused-but-set-variable"
                                     "-Wno-error=deprecated-declarations"
                                     "-Wextra"
+                                    "-Wno-error=extra"
                                     "-Wno-unused-parameter"
                                     "-Wno-sign-compare"
                                     # ignore multiple enum conversion warnings since gcc 11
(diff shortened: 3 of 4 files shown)


igrr:
@MarcFinetRtone Thank you for the PR!
I have tried running a test with your changes and found that all of the IDF examples can been built successfully with these changes. (This is expected, since we [build them with -Werror](https://github.com/espressif/esp-idf/blob/df9310ada26123d8d478bcfa203f874d8c21d654/tools/ci/configure_ci_environment.sh#L20-L21) in CI.)
Some other internal test applications which are not built with -Werror yet have predictably failed to build...

I think this is in general the right change, but we need to consider whether it can be included in a minor IDF release (v5.1 or v5.2) or not, since it has the potential to introduce breakage in projects based on IDF.

igrr:
> I understand, breakage could hit users, however, I think it's rather for their good.

I totally understand and agree with this!

However another matter we have to regularly deal with is getting customers to upgrade to a newer version of ESP-IDF. Every single breaking change (even a "good" one!) they meet when trying to upgrade from an EoL (end-of-life) version of IDF to a newer one makes it more likely that they will abandon the attempt, and instead try to force Espressif to implement the features or fixes they are missing in an EoL version they are using. This means more unplanned work for ESP-IDF developers, and less time they can spend addressing, for instance, issues reported here on Github.

One compromise we can have is to introduce this change, but add an easy way to get the old behavior. This is something we did recently when upgrading from GCC 11 to GCC 12, which will be part of the upcoming IDF v5.1 release. Along with the upgrade (which, as usual, brings new or stricter GCC warnings) we have added a Kconfig option to disable the new warnings introduced in GCC 12, as well as an idf.py hint to do this (https://github.com/espressif/esp-idf/commit/0da15a7a19c97d38569ddbe3b919045834408c25).

I think we could add a similar Kconfig option in this case: enable more checks by default, but give less technically savvy users an easy way to get the old behavior, without forcing them to learn about CMake, GCC options and such.

> In your failing tests, do you have seen sensible "violations" (that requires a -Wno rather than a code fix) ?

I have looked at the build logs briefly, here are some of the failures I found:

- "comparison is always true due to limited range of data type [-Werror=type-limits]" — something we should definitely fix. In the particular case, the code still happens to work correctly, but it's a bug waiting to happen.
- "-Werror=incompatible-pointer-types" reported in some cases when a callback function is declared with a different signature than expected. For example, a FreeRTOS task function was declared with an `int arg` argument, instead of a `void* arg`, and `xTaskCreate` call was getting an integer value `(void*) SOME_INT_CONSTANT`. The code worked as expected (SOME_INT_CONSTANT passed to the task function). We would definitely clean it up, though, instead of adding -Wno-incompatible-pointer-types.
- error: passing argument 5 of 'getsockopt' from incompatible pointer type [-Werror=incompatible-pointer-types] because the function expects an `long unsigned int *` argument but we were passing an `unsigned int*`. In this case it's a bit annoying because `unsigned long` and `unsigned` have the same size (4 bytes), so i'd say it's as close to a false positive as it gets. We would still fix it, however, since -Wincompatible-pointer-types is a useful check which can help find more severe issues.

All in all, it's not too bad, for an internal test app we would definitely just go and fix these issues in code instead of disabling the warnings. However in similar cases in the past we had customers complaining that their app worked totally fine and we "broke" it by revealing some warnings which were not actually critical. (This is related to the "upgrading to a newer version of IDF" problem I wrote about above.)

Do you think you could do something along the lines of https://github.com/espressif/esp-idf/commit/0da15a7a19c97d38569ddbe3b919045834408c25 in this PR, and add a migration guide note in `docs/en/migration-guides/release-5.x/5.1/build-system.rst`? (If no, no problem whatsoever — we can take your commit and add these changes on top ourselves.)

zikalino (approved):
I will merge this PR + make sure that it's not enabled by default in 5.x.
We will make it default in 6.0 with option to disable.

KonstantinKondrashov:
Hi @MarcFinetRtone!
No, you do not need to rebase it. I took it into our internal repo and rebased it. When it is merged this PR will be automatically closed. Thanks.

SoucheSouche:
b1768dcd8e8ed67882cdcb94a6eebc91d809b197

MarcFinetRtone:
>  We will make it default in 6.0 with option to disable.

It seems that the 6.0 has started (beta1 is out). It's probably time to change the default ?

MarcFinetRtone:
> > We will make it default in 6.0 with option to disable.
> 
> It seems that the 6.0 has started (beta1 is out). It's probably time to change the default ?

@SoucheSouche @KonstantinKondrashov  ?

KonstantinKondrashov:
Yes, the changes were done and merged here 6efd36f4750a2c5721774cd167c5de3ed6d17caa.

PR #11281, opened 2023-04-27
fix usb resume bug on esp32sx (IDFGH-10000)

paul356:
I have encounter this issue on esp32s3 that my usb hid device can't recover from suspend state after computer went to sleep and waked up. After long time of debugging it turned out that we didn't handle resume interrupt in dcd file for esp32sx. The fix is straight forward. I add USB_WKUPINT related bits in function dcd_init and dcd_int_handler. Having tested on my esp32s3 hid keyboard, I think it also works for other USB devices using esp32sx. Please comment if you need me to do further testing.

### DIFF
--- a/components/tinyusb/additions/src/portable/espressif/esp32sx/dcd_esp32sx.c
+++ b/components/tinyusb/additions/src/portable/espressif/esp32sx/dcd_esp32sx.c
@@ -193,6 +193,7 @@ void dcd_init(uint8_t rhport)
                  USB_RXFLVIMSK_M   |
                  USB_ERLYSUSPMSK_M |
                  USB_USBSUSPMSK_M  |
+                 USB_WKUPINTMSK_M  |
                  USB_USBRSTMSK_M   |
                  USB_ENUMDONEMSK_M |
                  USB_RESETDETMSK_M |
@@ -829,6 +830,7 @@ static void _dcd_int_handler(void* arg)
                   USB_GOUTNAKEFF    |
                   USB_ERLYSUSP_M    |
                   USB_USBSUSP_M     |
+                  USB_WKUPINT_M     |
                   USB_ISOOUTDROP_M  |
                   USB_EOPF_M        |
                   USB_EPMIS_M       |


tore-espressif:
Hello @paul356 , thank you for the contribution!
We'll test on more targets and hosts and then merge, thanks

paul356:
This issue relates to esp32s2 and esp32s3 only. You just need to test these two chips.

paul356:
This issue relates to esp32s2 and esp32s3 only. You just need to test these two chips.

tore-espressif:
Thank you again for you contribution. All tests OK, we will merge in out internal git and this issue will be closed automatically after it get synchronized to GitHub.

For IDF version >= 5.0 this will be fixed in [tinyusb component](https://components.espressif.com/components/espressif/tinyusb)

paul356:
> tinyusb component

paul356:
> Thank you again for you contribution. All tests OK, we will merge in out internal git and this issue will be closed automatically after it get synchronized to GitHub.
> 
> For IDF version >= 5.0 this will be fixed in [tinyusb component](https://components.espressif.com/components/espressif/tinyusb)

Thanks you for the information. I have tried IDF 5.0 several months ago. In that time, I didn't find tinyusb in IDF 5.0. So I went back to IDF 4.4. Is tinyusb still a component as in IDF 4.4 in the "components" folder? If that is the case, I may consider migrate my project to esp-idf 5.0. I also need to fix USB HID device related code in IDF 4.4, which I have previously opened a pull request. I am told that it is fixed in IDF 5.0. It will be better to work without a private IDF version.

paul356:
I find where tinyusb is in IDF 5.0. It is a dynamic component. It is installed through Component Manager.

tore-espressif:
The fix is included in our release of [TinyUSB v0.14.3](https://components.espressif.com/components/espressif/tinyusb/versions/0.14.3) so you can use it with IDFv5.x

related https://github.com/espressif/tinyusb/pull/11

IDF v4.x should wait until the fix is synced to GitHub from our internal servers

PR #11283, opened 2023-04-27
Fix emac ext struct for C++ (IDFGH-10003)

0xFEEDC0DE64:
For some reason this did not compile under C++ because of double definitions

### DIFF
--- a/components/soc/esp32/include/soc/emac_ext_struct.h
+++ b/components/soc/esp32/include/soc/emac_ext_struct.h
@@ -11,7 +11,7 @@ extern "C" {
 
 #include <stdint.h>
 
-typedef volatile struct emac_ext_dev_t {
+typedef volatile struct {
     union {
         struct {
             uint32_t div_num : 4;


components/soc/esp32/include/soc/emac_ext_struct.h
```diff
@@ -11,7 +11,7 @@ extern "C" {
 
 #include <stdint.h>
 
-typedef volatile struct emac_ext_dev_t {
```
suda-morris, line 14: I'm afraid removing it will break the compilation because of this: https://github.com/espressif/esp-idf/blob/master/components/hal/include/hal/emac_hal.h#L162

components/soc/esp32/include/soc/emac_ext_struct.h
```diff
@@ -11,7 +11,7 @@ extern "C" {
...
+typedef volatile struct {
```
suda-morris, line 14: Alternatively we can only add the `volatile` to each member

```suggestion
typedef struct emac_ext_dev_t {
    volatile union {
        struct {
            uint32_t div_num : 4;
            uint32_t h_div_num : 4;
            uint32_t dly_num : 2;
            uint32_t reserved10 : 22;
        };
        uint32_t val;
    } ex_clkout_conf;
    ....
```

suda-morris (changes requested):

components/soc/esp32/include/soc/emac_ext_struct.h
suda-morris: cc @kostaond ^^

kostaond:
@0xFEEDC0DE64, since there are more things to be fixed/polished, I will take care of the update. Anyway, thank you very much to pointing this issue out!

kostaond:
Closing the PR since the issue has been resolved internally and will be merged on GH soon.

PR #11287, opened 2023-04-27
Update esp_hid component to register GATTC callback (IDFGH-10009)

MatthewColvin:
example for hid devices registers the ble gattc callback but it would seem to make scenes to hide this in the HID component
Update example to reflect this change

Hopefully I am not misunderstanding something here thanks


zhp0406:
Thank you for your advice. We will conduct an evaluation.

PR #11297, opened 2023-04-28
don't treat timeout as error in send_cmd_send_op_cond (IDFGH-10020)

maprad:
`send_cmd_send_op_cond` allows for multiple retries before `ESP_ERR_TIMEOUT` is returned and a different number of errors that are ignored before the last error is returned. If `ESP_ERR_TIMEOUT` is treated like any other error, the number of retries is limited by the number of errors that can happen. This makes `nretries` useless.

### DIFF
--- a/components/sdmmc/sdmmc_cmd.c
+++ b/components/sdmmc/sdmmc_cmd.c
@@ -143,7 +143,7 @@ esp_err_t sdmmc_send_cmd_send_op_cond(sdmmc_card_t* card, uint32_t ocr, uint32_t
         }
 
         if (err != ESP_OK) {
-            if (--err_cnt == 0) {
+            if (err != ESP_ERR_TIMEOUT && --err_cnt == 0) {
                 ESP_LOGD(TAG, "%s: sdmmc_send_app_cmd err=0x%x", __func__, err);
                 goto done;
             } else {


igrr:
Hi @maprad, thanks for the PR!

You are right that if ESP_ERR_TIMEOUT occurs, nretries does not help to handle that.

However, according to the SD specification, the card should respond to ACMD41 command. If it is not ready yet, it will still send a response with the busy flag set. `nretries` loop in this code was supposed to handle that case — i.e. card repeatedly returning "busy". `nretries` is not supposed to keep retrying ACMD41 if the card doesn't respond at all.

According to the spec, the card may not respond to ACMD41 only if it is an MMC card, in which case we shouldn't retry sending ACMD41 and should send CMD1 instead.

The Linux kernel seems to agree, and doesn't retry in case there is no response for ACMD41:
https://github.com/torvalds/linux/blob/33afd4b76393627477e878b3b195d606e585d816/drivers/mmc/core/mmc_ops.c#L184-L194

Does this change help fix any issue in your case, or is it an observation based on you reviewing the code?

I don't quite remember the reason why I have added `err_cnt` related logic in c6829fa5b84bfaaeadd6a6da0af0b710a1fdf201, but it could have been related to SPI mode initialization somehow, or could have been a temporary hack I made in an attempt to get something else working, which has later slipped through code review. I'll try to revert this part and see if SD, SD-over-SPI and eMMC still work after that.

maprad:
Hi @igrr,

thank you for the quick clarification. 

I am trying to get SD over SPI  (the example) running and this fix would solve my problem (`Failed to initialize the card (ESP_ERR_TIMEOUT)`).
In my case, the ESP_ERR_TIMEOUT seems to come from `start_command_default` in `sdspi_host.c`, from what I found because of `shift_cmd_response` returning `ESP_ERR_NOT_FOUND`.

igrr:
I see, that could possibly indicate that there is some issue with reading command response the first time. Do you by any chance have a logic analyzer to capture what is happening on the bus? And/or add `ESP_LOG_BUFFER_HEX(TAG, cmd->r1, sizeof(cmd->r1));` at the beginning of `shift_cmd_response` function in components/driver/spi/sdspi/sdspi_host.c. This could help figure out whether the issue is in receiving the response or with finding the response in the buffer.

maprad:
This is the Log output with `ESP_LOG_BUFFER_HEX(TAG, &cmd->r1, sizeof(cmd->r1));` added: 
```
I (467) sdspi_host: 01 
I (487) sdspi_host: 01 
I (487) sdspi_host: 05 
I (487) sdspi_transaction: cmd=5, R1 response: command not supported
I (487) sdspi_host: 01 
I (487) sdspi_host: 01 
I (487) sdspi_host: ff 
I (497) sdspi_host: 01 
I (497) sdspi_host: ff 
I (497) sdspi_host: 01 
I (507) sdspi_host: ff 
E (507) sdmmc_common: sdmmc_init_ocr: send_op_cond (1) returned 0x107
E (517) vfs_fat_sdmmc: sdmmc_card_init failed (0x107).
I (517) gpio: GPIO[5]| InputEn: 1| OutputEn: 0| OpenDrain: 0| Pullup: 0| Pulldown: 0| Intr:0 
E (527) example: Failed to initialize the card (ESP_ERR_TIMEOUT). Make sure SD card lines have pull-up resistors in place.
I (537) main_task: Returned from app_main()

I (507) sdspi_host: ff 
E (507) sdmmc_common: sdmmc_init_ocr: send_op_cond (1) returned 0x107
E (517) vfs_fat_sdmmc: sdmmc_card_init failed (0x107).
I (517) gpio: GPIO[5]| InputEn: 1| OutputEn: 0| OpenDrain: 0| Pullup: 0| Pulldown: 0| Intr:0 
E (527) example: Failed to initialize the card (ESP_ERR_TIMEOUT). Make sure SD card lines have pull-up resistors in place.
I (537) main_task: Returned from app_main()
```

I do have a logic analyzer that I'm using with PulseView. Just tell me how and in which format I should make the capture available for you.
I discovered that everything works fine if I change the `Default log verbosity` to `Verbose`. With other log levels I get the error as seen above.

PR #11326, opened 2023-05-03
correct typo in reference to ESP-IDF repo (IDFGH-10049)

mykmelez:
There's a reference to the ESP-ADF repo that looks like it was intended to be a reference to the ESP-IDF repo. This branch fixes the reference.


ESP-Marius:
Yeah, you are right. Thanks for the fix!

ESP-Marius:
sha=55b38afbbc5ca9f33e897634a62422df70eccfe7

Issue #11331, opened 2023-05-04
W5500 is broken due to enforced version check  (IDFGH-10054)


huming2207:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.1-dev-4528-g420ebd208a

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

Custom board

### Power Supply used.

USB

### What is the expected behavior?

Some W5500 chips may return 0 for version, and just works and connects to the ethernet in ESP-IDF 5.0 (or anything before this commit bb2f346e55f484edc50f68cb26d164d8a79bcf39)

### What is the actual behavior?

Broken with error:  

```
E (3645) w5500.mac: w5500_verify_id(234): invalid chip version, expected 0x4, actual 0x0
E (3655) w5500.mac: emac_w5500_init(703): vefiry chip ID failed
```
...and apparently, no ethernet connection can be established after this error.

### Steps to reproduce.

I've tried commenting out this line, the issue can be fixed: https://github.com/espressif/esp-idf/blob/bb2f346e55f484edc50f68cb26d164d8a79bcf39/components/esp_eth/src/esp_eth_mac_w5500.c#L234

### More Information.

I've sourced our W5500 module from here: https://99tech.com.au/product/mod-w5500-t2/

It seems to be a genuine chip, at least the engraved mark on the chip package is identical to the one on Wiznet's official Raspberry Pi Pico dev W5500 HAT module. It might be just in a different version, and both versions are working fine with ESP-IDF's W5500 driver anyway.

huming2207:
Interestingly, according to W5500's datasheet, the VERSIONR should be always 0x04:

<img width="668" alt="image" src="https://user-images.githubusercontent.com/4463497/236136011-8cf15fa9-0c3e-4cfa-8d28-00157630c8cd.png">

I don't know now, maybe the module bought from 99tech is with a fake chip or engineering sample chip? But it does work fine with the W5500 driver if that line of code is removed.

igrr:
Linking the discussion in https://github.com/espressif/esp-idf/pull/10554#issuecomment-1512821151

[...]

PR #11337, opened 2023-05-05
w5500: mac: poll VERSIONR to ensure the chip is initialised (IDFGH-10061)

huming2207:
See discussion: https://github.com/espressif/esp-idf/issues/11331 and https://github.com/espressif/esp-idf/pull/10554#issuecomment-1512821151

This addresses the fix for the version check. Based on our experiments, we have to wait for at least 10ms (one FreeRTOS tick) to let the W5500 fully come up and talks to the host ESP32. We've also implemented a poll to retry 20 times just in case a simple wait doesn't work in some cases.


suda-morris:
because the `w5500_verify_id` is called directly after the `w5500_reset`, maybe what we need is a better exit path for the `w5500_reset` function? cc @kostaond

huming2207:
> because the `w5500_verify_id` is called directly after the `w5500_reset`, maybe what we need is a better exit path for the `w5500_reset` function? cc @kostaond

Thanks for your reply @suda-morris 

I've read the code again and I wonder if we should enforce a delay ***before*** the MR mode register read instead?

huming2207:
@suda-morris @kostaond Done, I've 1. removed the excessive retry count macro and used the reset timeout value instead, and 2. enforce at least a 10ms wait to poll the `RST` in the mode register just in case that poll read causes any undefined behavior.

kostaond:
> because the `w5500_verify_id` is called directly after the `w5500_reset`, maybe what we need is a better exit path for the `w5500_reset` function? cc @kostaond

Good point, however I am afraid there is no better condition to exit the reset function... Checking the state of autoclear reset bit is kind of standard way. The only thing which crossed my mind was to check the reset bit and the Version ID in the same condition. On the other hand, I don't see much value added in such approach and having it spitted into two calls seems good enough to me.

> I've read the code again and I wonder if we should enforce a delay before the MR mode register read instead?

I would keep it as it was since reading immediately seems to work for majority of chips and it ensures the faster startup for this majority. Polling the Version ID will capture the corner cases when the fastest path does not work.

huming2207:
> I would keep it as it was since reading immediately seems to work for majority of chips and it ensures the faster startup for this majority. Polling the Version ID will capture the corner cases when the fastest path does not work.

Sure, I will revert this change ASAP. Is there anything else I can do for this PR?

huming2207:
@kostaond Done, ready for review. I've also squashed the commits into one to make the commit message cleaner.

huming2207:
Any updates on this? @kostaond @suda-morris

components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -229,12 +229,21 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
...
+        vTaskDelay(pdMS_TO_TICKS(10));
+        ret = w5500_read(emac, W5500_REG_VERSIONR, &version, sizeof(version));
+        if (ret != ESP_OK) {
+            ESP_LOGD(TAG, "Poll VERSIONR failed, retrying...");
+            continue;
+        } else if (version == W5500_CHIP_VERSION) {
+            return ESP_OK;
+        }
+    }
+
```
kostaond, line 245: Please also print actual and expected version.

huming2207:
Any further updates? @kostaond

components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -229,12 +229,22 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
...
+    ESP_LOGE(TAG, "W5500 version mismatched, expected 0x04, got 0x%02x", version);
```
kostaond, line 246: Just a nitpick. Could you please print expected version as an argument `W5500_CHIP_VERSION` as it was done originally? Thank you.

components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -229,12 +229,22 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
...
-    ESP_GOTO_ON_ERROR(w5500_read(emac, W5500_REG_VERSIONR, &version, sizeof(version)), err, TAG, "read VERSIONR failed");
-    // W5500 doesn't have chip ID, we check the version number instead
-    ESP_GOTO_ON_FALSE(version == W5500_CHIP_VERSION, ESP_ERR_INVALID_VERSION, err, TAG, "invalid chip version, expected 0x%x, actual 0x%x", W5500_CHIP_VERSION, version);
 
-err:
-    return ret;
+    ESP_LOGI(TAG, "Waiting W5500 to start & verify version...");
+    uint32_t to = 0;
+    for (to = 0; to < emac->sw_reset_timeout_ms / 10; to++) {
+        vTaskDelay(pdMS_TO_TICKS(10));
```
kostaond, line 236: I missed this one previously because  I though we already discussed it. However, please place the delay at the end of the loop. We should try the fastest way at first.

components/esp_eth/src/esp_eth_mac_w5500.c
```diff
@@ -229,12 +229,22 @@ static esp_err_t w5500_verify_id(emac_w5500_t *emac)
 {
     esp_err_t ret = ESP_OK;
     uint8_t version = 0;
-    ESP_GOTO_ON_ERROR(w5500_read(emac, W5500_REG_VERSIONR, &version, sizeof(version)), err, TAG, "read VERSIONR failed");
-    // W5500 doesn't have chip ID, we check the version number instead
-    ESP_GOTO_ON_FALSE(version == W5500_CHIP_VERSION, ESP_ERR_INVALID_VERSION, err, TAG, "invalid chip version, expected 0x%x, actual 0x%x", W5500_CHIP_VERSION, version);
 
-err:
-    return ret;
+    ESP_LOGI(TAG, "Waiting W5500 to start & verify version...");
```
kostaond, line 233: Remove or make it debug level.

kostaond (commented):
I left two final comments. @suda-morris could you please also have a look?

huming2207:
@kostaond @suda-morris Done, waiting for review

kostaond (changes requested):

huming2207:
> Assuming you're not using a counterfeit chip, you should file an issue on Wiznet forums, since this is something worthy of an update of the datasheet at least.

I think this is necessary but a bit off-topic. We wish we have time to confirm with Wiznet but we've already bought a handful of these W5500 modules and chips. We are also experiencing a short deadline, but we also want to help anyone with the same issue as ours, so we push this fix to the mainline ESP-IDF. As long as this fix works for us (and anyone else, hopefully), it's acceptable anyway.

kostaond (approved):

kostaond:
sha=177f70c056e87b48117646e79b7eca8d920cfd4d

PR #11370, opened 2023-05-10
Improvement in the installation scripts

absoluteabutaj:
1. `tools/detect_python.sh` and `tools/detect_python.sh` will check for `python3` instead of `python`
2. `tools/idf_tools.py` will name the toolchain folder as `.espressif-v4.4.4` instead of `.espressif`


dobairoland:
I'm sorry but we won't accept this PR.

> tools/detect_python.sh and tools/detect_python.sh will check for python3 instead of python

They check for both but first `python` is considered. This is for the reason if it is set to a suitable version then it should be used instead of `python3`. Many users set this when they have various `python3.x` on their system.

> tools/idf_tools.py will name the toolchain folder as .espressif-v4.4.4 instead of .espressif

That directory is used for all ESP-IDF version, not just one. Also please notice that that is just the default one. If you for some reason want to change it then you can do with setting the `IDF_TOOLS_PATH` environment variable.

absoluteabutaj:
@dobairoland - Thanks for your suggestion.

PR #11371, opened 2023-05-10
Add possibility to change pinginterval (IDFGH-10096)

danielsmoofit:
Can change ping interval while websocket is connected.

### DIFF
--- a/components/esp_websocket_client/esp_websocket_client.c
+++ b/components/esp_websocket_client/esp_websocket_client.c
@@ -759,6 +759,22 @@ esp_err_t esp_websocket_client_stop(esp_websocket_client_handle_t client)
     return ESP_OK;
 }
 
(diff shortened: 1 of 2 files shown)


components/esp_websocket_client/esp_websocket_client.c
```diff
@@ -759,6 +759,22 @@ esp_err_t esp_websocket_client_stop(esp_websocket_client_handle_t client)
...
+{
+    if (client == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    if (client->config == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    if (ping_interval_sec == 0) {
+        return ESP_ERR_INVALID_ARG;
+    }
```
KaeLL, line 772: Why the extra verbosity? `if ( !( client && client->config && ping_interval_sec ) )` or some other single if clause variant would suffice.

PR #11457, opened 2023-05-20
Remove semicolons from ESP_IP4ADDR_INIT and ESP_IP6ADDR_INIT macros (IDFGH-10189)

txgk:
There are structs like `esp_netif_dns_info_t` that have member for generic ip for which the presence of semicolons is problematic, because when trying to initialize an `ip` member with one of these macros it results in a syntax error since semicolon is used within curly braces.

Problematic case example:
```
esp_netif_dns_info_t my_dns = {.ip = ESP_IP4ADDR_INIT(0, 0, 0, 0)};
```
Here `ESP_IP4ADDR_INIT` expands to expression with semicolon at the end which is a syntax error.

It's also not possible to do something like:
```
esp_netif_dns_info_t my_dns;
my_dns.ip = ESP_IP4ADDR_INIT(0, 0, 0, 0)
```
because that would be other syntax error (initialization of struct that was already created).

### DIFF
--- a/components/esp_netif/include/esp_netif_ip_addr.h
+++ b/components/esp_netif/include/esp_netif_ip_addr.h
@@ -79,8 +79,8 @@ extern "C" {
 
 #define ESP_IP4TOADDR(a,b,c,d) esp_netif_htonl(ESP_IP4TOUINT32(a, b, c, d))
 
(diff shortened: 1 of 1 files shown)


Alvin1Zhang:
Thanks for your contribution.

AxelLin:
@mahavirj @david-cermak 
I think the slow response here just discourage people to contribute fixes.(This PR was sent on May 21).

david-cermak (approved):
Thank for the fix!
I'm sorry this PR was somehow forgotten, sorry for such a delay

PR #11494, opened 2023-05-24
twai: Don't abort on interrupt allocation failure (IDFGH-10231)

nebkat:
`twai_driver_install()` currently aborts if there is a failure allocation the interrupt handler, unlike for example UART driver which simply returns the error code:

https://github.com/espressif/esp-idf/blob/d00e7b5af897cc5fafe51fae19c57f0313b81edf/components/driver/uart/uart.c#L1588-L1591

With this change execution can continue even if the allocation fails.

### DIFF
--- a/components/driver/twai.c
+++ b/components/driver/twai.c
@@ -10,6 +10,7 @@
 #include "freertos/task.h"
 #include "freertos/queue.h"
 #include "freertos/semphr.h"
+#include "esp_check.h"
 #include "esp_types.h"
 #include "esp_log.h"
 #include "esp_intr_alloc.h"
@@ -98,6 +99,8 @@ static portMUX_TYPE twai_spinlock = portMUX_INITIALIZER_UNLOCKED;
 #define TWAI_EXIT_CRITICAL()        portEXIT_CRITICAL(&twai_spinlock)
 
 static twai_hal_context_t twai_context;
+
+static const char *TWAI_TAG = "twai";
 
 /* -------------------- Interrupt and Alert Handlers ------------------------ */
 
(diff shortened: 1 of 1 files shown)


components/driver/twai.c
```diff
@@ -446,7 +449,8 @@ esp_err_t twai_driver_install(const twai_general_config_t *g_config, const twai_
 
     //Allocate GPIO and Interrupts
     twai_configure_gpio(g_config->tx_io, g_config->rx_io, g_config->clkout_io, g_config->bus_off_io);
-    ESP_ERROR_CHECK(esp_intr_alloc(ETS_TWAI_INTR_SOURCE, g_config->intr_flags, twai_intr_handler_main, NULL, &p_twai_obj->isr_handle));
+    ret = esp_intr_alloc(ETS_TWAI_INTR_SOURCE, g_config->intr_flags, twai_intr_handler_main, NULL, &p_twai_obj->isr_handle);
```
Dazza0, line 452: Prefer if we

- Allocated interrupts immediately after allocating the driver object. This way, if any resource allocation fails, we can return early
- Added the `ESP_INTR_FLAG_INTRDISABLED` flag so that an allocated interrupt remains disabled until we explicitly enable it using `esp_intr_enable()`

So something like...

```suggestion
    //Create a TWAI object (including queues and semaphores)
    p_twai_obj_dummy = twai_alloc_driver_obj(g_config->tx_queue_len, g_config->rx_queue_len);
    TWAI_CHECK(p_twai_obj_dummy != NULL, ESP_ERR_NO_MEM);
    //Allocate interrupt for TWAI driver
    intr_handle_t isr_hdl;
    ret = esp_intr_alloc(ETS_TWAI_INTR_SOURCE, g_config->intr_flags | ESP_INTR_FLAG_INTRDISABLED, twai_intr_handler_main, NULL, &isr_hdl);
    if (ret != ESP_OK) {
        ESP_LOGE(TWAI_TAG, "Interrupt allocation failed");
        goto intr_alloc_err;
    }
```

Dazza0 (commented):
@nebkat Thanks for your contribution, the requested changes looks reasonable to me.

I would probably want to rearrange the order of operations in the `twai_driver_install()` so that all resource allocations are done at the start. Furthermore, would you be able to target this PR for the master branch (it would work better with our backporting workflow)?

nebkat:
Sorry never got around to finishing this, thanks for the change!

Dazza0:
@nebkat No worries. I've made some modifications to your PR commit, but [commit author](9956bfac3e1adbed5cc2b502bb7c04a9102b8eec) is still attributed to you. Thanks for the contribution

PR #11411, opened 2023-05-16
Fix #10804 by running continuous ADC DMA in endless loop instead of restarting after each run to avoid losing samples (IDFGH-10140)

Erlkoenig90:
This fixes #10804 by running the ADC DMA in a continuous loop.

* A function `adc_hal_read_desc_finish` is introduced to set a DMA descriptor's `owner` field to `1` such that it can be re-used by DMA.
* A new DMA callback function `on_descr_err` is added for the `DSCR_ERR` interrput. The interrupt is only enabled if the callback is specified. The continuous ADC driver uses this to print an error to  detect DMA buffer overruns.

PR #11499, opened 2023-05-25
Implement descriptor error callback for GDMA (IDFGH-10236)

Erlkoenig90:
As discussed in #11411 this PR adds a callback to the GDMA driver for the `DESCR_ERR` GDMA interrupt called `on_descr_err`. This can be used to debug drivers that use GDMA (such as adc_continuous) in a circular modes by detecting overruns.

The interrupt is only enabled if the callback is specified.


suda-morris:
sha=5e4ae148fa530dad0311bc46ef0fa3e8c7d144e0

PR #11500, opened 2023-05-25
Fix #10804 by running continuous ADC DMA in endless loop instead of restarting after each run to avoid losing samples (IDFGH-10235)

### LINKED ISSUE #10804: adc_digi_read_bytes() messes the order of channels between read chuncks  (IDFGH-9437)
### Answers checklist.

### IDF version.

v4.4.3

### Operating System used.

Windows

Eclipse IDE

None

### Development Kit.

ESP32-S3

### Power Supply used.

USB

adc_digi_read_bytes() for multiple channels and given number of samples shall return the channel values in the same order  in each data chunk returned by function call.
e.g. for 2 channel ADC1 only, i expect 
chunk1 = **data_chn_a**, **data_chn_b**,  **data_chn_a**, **data_chn_b**,  **data_chn_a**, **data_chn_b**, ...
if chunk1 ends with **data_chn_a**, chunk2 shall start with **data_chn_b**,  data_chn_a, data_chn_b,  data_chn_a, data_chn_b, ...
if chunk1 ends with **data_chn_b**, chunk2 shall start with **data_chn_a**, data_chn_b,  data_chn_a, data_chn_b,  data_chn_a, ...
In this way the assembled buffer will not miss any sample.

Instead the following chunk start randomly with data_chn_a or data_chn_a messing the order in the assembled buffer.
Maybe some samples are lost, which for my case is very critical

### Steps to reproduce.

this is a sample code to reporoduce
```
typedef union
    {
    uint8_t result[NR_SAMPLES * ADC_RESULT_BYTE];
    adc_digi_output_data_t presult[NR_SAMPLES];
    } res_t;

#define    SENSOR_ADC_CHANNEL        (4)
#define CURRENT_ADC_CHANNEL        (3)
static uint16_t adc1_chan_mask = BIT(SENSOR_ADC_CHANNEL) | BIT(CURRENT_ADC_CHANNEL);
static uint16_t adc2_chan_mask = BIT(0);
static adc_channel_t channel[] = {CURRENT_ADC_CHANNEL, SENSOR_ADC_CHANNEL};
static void continuous_adc_init(uint16_t adc1_chan_mask, uint16_t adc2_chan_mask, adc_channel_t *channel, uint8_t channel_num)
    {
    adc_digi_init_config_t adc_dma_config =
        {
        .max_store_buf_size = NR_SAMPLES * ADC_RESULT_BYTE,
        .conv_num_each_intr = NR_SAMPLES,
        .adc1_chan_mask = adc1_chan_mask,
        .adc2_chan_mask = adc2_chan_mask,
        };
    ESP_ERROR_CHECK(adc_digi_initialize(&adc_dma_config));

    adc_digi_configuration_t dig_cfg =
        {
        .conv_limit_en = ADC_CONV_LIMIT_EN,
        .conv_limit_num = 0,
        //.sample_freq_hz = 20 * 1000,
        .sample_freq_hz = 4 * 1000,
        .conv_mode = ADC_CONV_SINGLE_UNIT_1,
        .format = ADC_DIGI_OUTPUT_FORMAT_TYPE2,
        };

    adc_digi_pattern_config_t adc_pattern[4] = {0};
    dig_cfg.pattern_num = channel_num;
    for (int i = 0; i < channel_num; i++)
        {
        uint8_t unit = GET_UNIT(channel[i]);
        uint8_t ch = channel[i] & 0x7;
        adc_pattern[i].atten = ADC_ATTEN_DB_11;
        adc_pattern[i].channel = ch;
        adc_pattern[i].unit = unit;
        adc_pattern[i].bit_width = SOC_ADC_DIGI_MAX_BITWIDTH;

        //ESP_LOGI(TAG, "adc_pattern[%d].atten is :%x", i, adc_pattern[i].atten);
        //ESP_LOGI(TAG, "adc_pattern[%d].channel is :%x", i, adc_pattern[i].channel);
        //ESP_LOGI(TAG, "adc_pattern[%d].unit is :%x", i, adc_pattern[i].unit);
        }
    dig_cfg.adc_pattern = adc_pattern;
    ESP_ERROR_CHECK(adc_digi_controller_configure(&dig_cfg));
    }

void app_main()
     {
...

    continuous_adc_init(adc1_chan_mask, adc2_chan_mask, channel, sizeof(channel) / sizeof(adc_channel_t));
...
       res_t res;
       index = 0;
       ret = ESP_OK;
       uint32_t b2read = NR_SAMPLES * ADC_RESULT_BYTE;
       adc_digi_start();
       while(index < NR_SAMPLES * ADC_RESULT_BYTE && ret == ESP_OK)
           {
           ret = adc_digi_read_bytes(res.result + index, b2read, &ret_num, 1000);
           if (ret == ESP_ERR_TIMEOUT || ret == ESP_ERR_INVALID_STATE)
               break;
           ESP_LOGI(TAG, "read chunk %d %d", index, ret_num);
           index += ret_num;
           b2read -= ret_num;
           }
       adc_digi_stop();
...
}
```
If you look in res.presult at chunk boundary from time to time you'll see the order of values is changed.

### Debug Logs.

_No response_

### More Information.

_No response_

Erlkoenig90:
This fixes #10804 by running the ADC DMA in a continuous loop. A function `adc_hal_read_desc_finish` is introduced to set a DMA descriptor's `owner` field to `1` such that it can be re-used by DMA.

As discussed in #11411, this requires PR #11499 to be merged too as it relies on the new `on_descr_err` callback to detect/debug GDMA buffer overruns.


Erlkoenig90:
@Icarus113 I think the changes introduced in 9bec423 don't guarantee consistency of the DMA buffers and contain a race condition:

`adc_hal_get_reading_result` [now](https://github.com/espressif/esp-idf/blob/9bec4233a48f4988920bc240a234f08507210976/components/hal/adc_hal.c#L320) [sets](https://github.com/espressif/esp-idf/blob/9bec4233a48f4988920bc240a234f08507210976/components/hal/adc_hal.c#L330)  `eof_desc->dw0.owner = 1;` i.e. before the caller can even read the buffer data. This means that the DMA could overwrite this buffer even before `adc_hal_get_reading_result` returns.

E.g. when `s_adc_dma_intr` from `adc_continuous` [calls](https://github.com/espressif/esp-idf/blob/9bec4233a48f4988920bc240a234f08507210976/components/esp_adc/adc_continuous.c#L329) `adc_hal_get_reading_result`, the data in `finished_buffer` might get overwritten by DMA even before it is copied to the ringbuffer by `xRingbufferSendFromISR`.

I think this can't be properly implemented without providing a function `adc_hal_read_desc_finish` as proposed which should be called *after* the user of `adc_hal_get_reading_result` is done with the data.

Since a descriptor error callback is not used, we don't get notified if DMA attempts to overwrite the buffer before setting `owner`.

Icarus113:
Hi @Erlkoenig90 , yes, this is actually the competition between the A. software (dealing with the ready dma desc) and B. hardware (moving new dma data). If the A is less than B then 
- the setting-owner-to-1 operation is dangerous
- if owner isn't DMA, the dma desc err error event will be triggered.

We adopted the DMA rx desc err event PR. For ADC side we didn't adopt the error log logic yet. On chips later than S2, the ADC sampling speed is not very fast, so this shouldn't be an issue. May I know if you observe that under certain condition A is less than B?

Erlkoenig90:
> May I know if you observe that under certain condition A is less than B?

Hi @Icarus113 , I did not observe this in practice (yet). I just thought this might potentially be an issue from looking at the code and that it might be worth fixing before an issue arises. This also depends on CPU load and interrupt load; an application with high CPU load might experience this problem (when the interrupt is processed right before the DMA attempts to re-use the buffer). This would cause data corruption which would be hard to debug.

Icarus113:
The downside for your solution of delaying the `set owner to 1` is:
- it makes the change to a single descriptor scattered, thus further lengthen the A (software dealing speed)

So if it's really in an extreme condition for example 
- A is just a little bit faster than B, this change will make DMA check owner fail, then leading desc err event, and DMA will stops fetching dma descriptors.
- A is much slower than B, DMA may overwrite some descriptors whose owner is set to 1, and it will finally find an owner-equals-0 descriptor then lead to DMA stop with err event

So changing the descriptor update timing isn't a good way for such condition.

A better solution is to avoid such extreme condition. There are few solutions when designing the driver:
- make the conversion frame longer by setting `conv_frame_size` bigger
- as you can see the dma descriptor number is fixed in the current driver, there is a plan to make it configurable by users. (We didn't implement this yet)

So by increasing these we can avoid the extreme condition.

On the other hand, 

>  I just thought this might potentially be an issue from looking at the code and that it might be worth fixing before an issue arises

Having a tip for this condition is acceptable. We can try to plan it then and give some hints accordingly.

@Erlkoenig90

Erlkoenig90:
> * A is just a little bit faster than B, this change will make DMA check owner fail, then leading desc err event, and DMA will stops fetching dma descriptors.
> 
> * A is much slower than B, DMA may overwrite some descriptors whose owner is set to 1, and it will finally find an owner-equals-0 descriptor then lead to DMA stop with err event

I think it is better to get a DMA error rather than having the DMA causing data corruption by overwriting a buffer that is still being copied to the ringbuffer ("silent error").

> There are few solutions when designing the driver:

Yes, but my suggestion does not attempt to solve the issue, it just allows the application developer to *notice* the issue (by getting a DMA error) at all.

PR #11508, opened 2023-05-26
[Heap Trace Standalone] increase Kconfig max stack depth (IDFGH-10245)

chipweinberger:
Increase max stack depth from 10 -> 32

10 is too small a maximum, and is preventing me from knowing who is calling the function that is leaking.

### DIFF
--- a/components/heap/include/heap_trace.inc
+++ b/components/heap/include/heap_trace.inc
@@ -68,9 +68,31 @@ static IRAM_ATTR __attribute__((noinline)) void get_call_stack(void **callers)
     TEST_STACK(7);
     TEST_STACK(8);
     TEST_STACK(9);
+    TEST_STACK(10);
+    TEST_STACK(11);
+    TEST_STACK(12);
+    TEST_STACK(13);
+    TEST_STACK(14);
+    TEST_STACK(15);
+    TEST_STACK(16);
+    TEST_STACK(17);
+    TEST_STACK(18);
+    TEST_STACK(19);
+    TEST_STACK(20);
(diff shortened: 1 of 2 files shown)


components/heap/include/heap_trace.inc
```diff
@@ -39,7 +39,7 @@ inline static uint32_t get_ccount(void)
 #define STACK_OFFSET  2
 
 #define TEST_STACK(N) do {                                              \
-        if (STACK_DEPTH == N) {                                         \
+        if (N >= STACK_DEPTH) {                                         \
             return;                                                     \
```
chipweinberger, line 43: I'm not really sure what this should be. Why was it originally STACK_DEPTH == N?

chipweinberger:
Please add this to v5.1

chipweinberger:
@ESP-Marius , would be great if you could sync this one as well. 

I've been testing/using this change without issue.

ESP-Marius:
sha=bf7020c36fcf1d5dc98afb299605ed7be5ac3af2

ESP-Marius:
LGTM @chipweinberger 

> Please add this to v5.1 =)

I'm afraid 5.1 has already branched of from master and we normally only backport bugfixes. 

If we allowed features to be backported it would greatly increase the number of changes done to stable branches and increase the regression risk (even though I agree for this change the risk is minimal)

chipweinberger:
merged https://github.com/espressif/esp-idf/commit/23cba5fe002ec8205585fa58122dce81ec51031d

PR #11534, opened 2023-05-29
lwip: port: freertos: Use libc malloc for all allocations

ejohnso49:
Mixing lwip and libc allocation functions causes heap failures when LwIPs MEM_STATS feature is enabled. MEM_STATS uses additional bookkeeping space that is hidden from the caller. If previously allocated blocks are free'd without taking this into account the heap will assert. This commit modifies these calls to consistently use libc malloc

### DIFF
--- a/components/lwip/port/freertos/sys_arch.c
+++ b/components/lwip/port/freertos/sys_arch.c
@@ -205,7 +205,7 @@ sys_sem_free(sys_sem_t *sem)
 err_t
 sys_mbox_new(sys_mbox_t *mbox, int size)
 {
-  *mbox = mem_malloc(sizeof(struct sys_mbox_s));
+  *mbox = malloc(sizeof(struct sys_mbox_s));
   if (*mbox == NULL){
     LWIP_DEBUGF(ESP_THREAD_SAFE_DEBUG, ("fail to new *mbox\n"));
     return ERR_MEM;
@@ -517,7 +517,7 @@ sys_thread_sem_free(void* data) // destructor for TLS semaphore
 sys_sem_t*
 sys_thread_sem_init(void)
 {
-  sys_sem_t *sem = (sys_sem_t*)mem_malloc(sizeof(sys_sem_t*));
+  sys_sem_t *sem = (sys_sem_t*)malloc(sizeof(sys_sem_t*));
 
   if (!sem){
     ESP_LOGE(TAG, "thread_sem_init: out of memory");


chipweinberger:
What is the reason for `mem_malloc` in the first place? Is this going to hurt perf?

xueyunfei998:
hi @ejohnso49 
Thanks for submitting the fix. I will review this code later.

ejohnso49:
> What is the reason for `mem_malloc` in the first place? Is this going to hurt perf?

To the best of my knowledge, there should not be any effect on performance. LwIP defines `mem_malloc` to give users more flexibility regarding what implements memory allocation for LwIP. There are a few other options like using LwIP's memory heap implementation, a pool-based allocation, a pool-based allocation using `malloc`, etc.

The configuration is set [here](https://github.com/espressif/esp-idf/blob/903af13e847cd301e476d8b16b4ee1c21b30b5c6/components/lwip/port/include/lwipopts.h#LL92C42-L92C42) to use `malloc`. Then in `mem.c`, `mem_malloc` and `mem_free`, wrap the libc functions with configurable stat counters. These default to off so as not to affect performance.

I do not think the allocations done in `sys_arch.c` are required to be counted by the LwIP stats, so my thinking was to convert back to plain libc calls where required to resolve the issue.

xueyunfei998:
hi  @ejohnso49 

 when LwIPs MEM_STATS feature is enabled

#define MEM_STATS                       ((MEM_LIBC_MALLOC == 0) && (MEM_USE_POOLS == 0))

These two macros are all equal to 0. Will there be a problem of mixing lib malloc and lwip malloc?

ejohnso49:
> hi @ejohnso49
> 
> when LwIPs MEM_STATS feature is enabled
> 
> #define MEM_STATS ((MEM_LIBC_MALLOC == 0) && (MEM_USE_POOLS == 0))
> 
> These two macros are all equal to 0. Will there be a problem of mixing lib malloc and lwip malloc?

MEM_STATS can also be enabled separately by setting it to 1 manually, that line is just the default value.

For any of the combinations of `MEM_LIBC_MALLOC` and `MEM_USE_POOLS` I do not think that introduces a problem because all objects created in `arch_sys.c` will use the correct matching allocation functions (libc malloc + free) with this change. The current implementation excludes these objects from being tracked by LwIPs mem stats because it uses the FreeRTOS dynamic allocation to create them.

david-cermak:
Thanks for this PR!
Yes, we indeed mix lwip allocations with standard malloc/free calls. But in this instance, I'd probably use lwip based allocs, as this port layer logically belongs to lwip and the stat module can also keep track of the port layer allocation (which makes sense IMO).
What you think of changing

https://github.com/espressif/esp-idf/blob/8760e6d2a7e19913bc40675dd71f374bcd51b0ae/components/lwip/port/freertos/sys_arch.c#L379

to `mem_free()` ?

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/d744d16fb064bf93affab679e672e407cbcba748

ejohnso49:
> Thanks for this PR! Yes, we indeed mix lwip allocations with standard malloc/free calls. But in this instance, I'd probably use lwip based allocs, as this port layer logically belongs to lwip and the stat module can also keep track of the port layer allocation (which makes sense IMO). What you think of changing
> 
> https://github.com/espressif/esp-idf/blob/8760e6d2a7e19913bc40675dd71f374bcd51b0ae/components/lwip/port/freertos/sys_arch.c#L379
> 
> to `mem_free()` ?

Thanks for fixing this in your PR, sorry I totally missed your comment a while back!

PR #11562, opened 2023-06-01
[Console] add setting to use PSRAM (IDFGH-10303)

chipweinberger:
Add an option to use SPIRAM for esp console allocations, otherwise fallback to default allocation scheme.

I have 215 commands, so **this was taking up nearly 20KB of internal ram**.

### DIFF
--- a/components/console/commands.c
+++ b/components/console/commands.c
@@ -8,6 +8,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <sys/param.h>
+#include "esp_heap_caps.h"
 #include "esp_log.h"
 #include "esp_console.h"
 #include "esp_system.h"
@@ -45,6 +46,9 @@ static esp_console_config_t s_config;
 /** temporary buffer used for command line parsing */
 static char *s_tmp_line_buf;
 
+/** for internal allocations */
+static uint32_t s_heap_alloc_caps = MALLOC_CAP_DEFAULT;
+
 static const cmd_item_t *find_command_by_name(const char *name);
 
 esp_err_t esp_console_init(const esp_console_config_t *config)
@@ -59,7 +63,10 @@ esp_err_t esp_console_init(const esp_console_config_t *config)
     if (s_config.hint_color == 0) {
         s_config.hint_color = ANSI_COLOR_DEFAULT;
     }
-    s_tmp_line_buf = calloc(config->max_cmdline_length, 1);
+    if (config->heap_alloc_caps != 0) {
+        s_heap_alloc_caps = config->heap_alloc_caps;
+    }
+    s_tmp_line_buf = heap_caps_calloc(config->max_cmdline_length, 1, s_heap_alloc_caps);
     if (s_tmp_line_buf == NULL) {
         return ESP_ERR_NO_MEM;
     }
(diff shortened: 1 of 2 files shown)


suda-morris:
maybe we can also think of adding a new config entry in the `esp_console_config_t` instead of relying on the Kconfig :thinking: As you know, the Kconfig is not **configurable** in Arduino.

chipweinberger:
Okay, let me know espressif's decision.

chipweinberger:
@suda-morris , the PR has been updated

chipweinberger:
@suda-morris bump

chipweinberger:
@suda-morris bump =) just hoping to catch you when you have a free moment.

chipweinberger:
@igrr could we get this assigned to someone? It's such a small feature.

ESP-Marius:
@chipweinberger sorry for that no-one follow up on this one. I'll assign it to myself. 

Looks like a small and use-full feature. I think we can get this one merged quickly.

chipweinberger:
I appreciate you! =)

components/console/esp_console.h
```diff
@@ -22,6 +22,7 @@ typedef struct linenoiseCompletions linenoiseCompletions;
 typedef struct {
     size_t max_cmdline_length;  //!< length of command line buffer, in bytes
     size_t max_cmdline_args;    //!< maximum number of command line arguments to parse
+    bool use_psram;             //!< use psram for internal allocations
```
ESP-Marius, line 25: What do you think about making `uint32_t alloc_caps //!< heap allocator caps (e.g. MALLOC_CAP_SPIRAM) used  when allocating console objects like cmds, defaults to MALLOC_CAP_DEFAULT` instead? 

This way the API is more flexible and would work even if we were to add some non psram external memory.

This way we can also simply the code (`static s_heap_alloc_caps = MALLOC_CAP_DEFAULT`,  and all the allocs can simply use `s_heap_alloc_caps` when allocating)

What do you think?

chipweinberger: Yes, I was actually just thinking the same thing. Let me update the PR.

chipweinberger: updated.

chipweinberger: updated again

components/console/commands.c
```diff
@@ -59,7 +63,8 @@ esp_err_t esp_console_init(const esp_console_config_t *config)
     if (s_config.hint_color == 0) {
         s_config.hint_color = ANSI_COLOR_DEFAULT;
     }
-    s_tmp_line_buf = calloc(config->max_cmdline_length, 1);
+    s_heap_alloc_caps = config->heap_alloc_caps;
```
ESP-Marius, line 66: I think we should only set this if `config->heap_alloc_caps != 0` , some users might not be using `ESP_CONSOLE_CONFIG_DEFAULT`. So if users havent set  `heap_alloc_caps` we just keep using the default value `MALLOC_CAP_DEFAULT`

chipweinberger: good point

chipweinberger: updated

ESP-Marius:
Thanks a lot for the super quick updates @chipweinberger, no more comments from me. After the last one is resolved i'll pull to our internal repo!

chipweinberger:
should be all resolved now.

ESP-Marius:
sha=9178748ff3c0f4595a0ece6a0584f8cf38728504

chipweinberger:
merged: https://github.com/espressif/esp-idf/commit/54f327a380ac9aaf66f4db53b082c50370300dfc

PR #11637, opened 2023-06-09
Fix for IDFGH-10379: Delay in esp_timer task callback prevents dispatch of ESP_TIMER_ISR callbacks

acf-bwl:
A delay in an esp_timer task-dispatched callback can currently prevent ISR-dispatched callbacks from running until the former returns. This pull request introduces a fix for the issue. See IDFGH-10379


Alvin1Zhang:
Thanks for your contribution.

KonstantinKondrashov:
Hi @acf-bwl! Thanks for the fix. That helps for the case that you described. I will take your changes. I will add some tests for this case and change your approach a little bit to reduce the amount of changes for the rest chips that use systimer.

PR #11638, opened 2023-06-09
Fix for IDFGH-10379: Delay in esp_timer task callback prevents dispatch of ESP_TIMER_ISR callbacks (v4.4 backport) (IDFGH-10381)

acf-bwl:
A delay in an esp_timer task-dispatched callback can currently prevent ISR-dispatched callbacks from running until the former returns. This pull request introduces a fix for the issue. See IDFGH-10379
-
We run the v4.4 release internally - I am not sure if patches for bugs against this release branch are helpful or not. Pull request against master is #11637 .


KonstantinKondrashov:
The issue is not critical so according to our backport rules we do not provide backport for v4.4.
(5.0 includes this changes, see https://github.com/espressif/esp-idf/pull/11637)

PR #11652, opened 2023-06-12
Fix ULP FSM register macros with addr[9:0] > 0xFF (IDFGH-10397)

boarchuz:
The following was prematurely constraining reg to 8 bits, effectively masking off addr[7:6]:
`.addr = (reg & 0xff) / sizeof(uint32_t),`
This was an issue when reading from or writing to registers with bits[9:8] != 0.

For example, on S3 operations with the following register:
`#define RTC_CNTL_TOUCH_DAC1_REG          (DR_REG_RTCCNTL_BASE + 0x150)`
MIstakenly interact with the following register instead (as 0x150 & 0xFF == 0x50):
`#define RTC_CNTL_STORE0_REG          (DR_REG_RTCCNTL_BASE + 0x50)`

Fixed by applying the bitmask after the division, instead of before.


sudeep-mohanty:
Hello @boarchuz,
Thank you for noticing the bug and fixing it. Changes LGTM! I shall be proceeding to merge the PR to our internal repository for further sanity tests. Thanks!

sudeep-mohanty:
sha=07332abbaad63a74b22a6641488d9f6db000a2e2

sudeep-mohanty (approved):

PR #11662, opened 2023-06-13
fix proto cmd crash problem in iperf example (IDFGH-10408)

idrey:
Problem:
If user use "proto" cmd at the beginning of iperf example, the prompt will crash and restart

Cause:
Chip is in wifi_null_mode at first, the code of proto cmd didn't check that

Fix:
Add wifi mode check so prompt will not crash


xuxiao111:
Hi @idrey,
Thanks for adding the fix. And this code change has been merged.

PR #11724, opened 2023-06-21
esp_http_client: Fix unsigned integer overflow (IDFGH-10477)

hwqchi:
Fix unsigned integer overflow when `content_length` is -1. 
The proposed change is to modify the format specifier for content_length from PRIu64 to PRId64, which will print the signed value as a 64-bit integer instead.


hmalpani:
@hwqchi Thanks for adding the fix.

hmalpani:
sha=482f6340b96a0a5ab5659ad7da95df32daf8f5e7

PR #11785, opened 2023-07-01
feat(nvs_flash): nvs_gen_partition_image CMake function (IDFGH-10542)

nebkat:
Adds a CMake function `nvs_gen_partition_image` similar to `spiffs_create_partition_image`.

Will add examples and documentation shortly.


higaski:
Can we move forward with any of the two PRs? Personally, I like my function name better because it's almost identical to the SPIFFS one. But in the end I really don't care, just pick one or provide feedback for further improvements please.

Issue #11783, opened 2023-06-30
Multiple build configurations improvements (IDFGH-10539)


nebkat:
### Is your feature request related to a problem?

I'd like to revisit the topic of multiple build configurations and continue the discussion from #5658.

Multiple build configurations are currently well supported using:

`idf.py -B build-debug -DSDKCONFIG=build-production/sdkconfig`
`idf.py -B build-production -DSDKCONFIG=build-production/sdkconfig -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.prod"`

Or even better with the following line:

```cmake
set(SDKCONFIG "${CMAKE_BINARY_DIR}/sdkconfig")
```

However, the syntax remains quite awkward. Commands that would often be run quickly in succession e.g. `idf.py build`, `idf.py app-flash`, `idf.py monitor` all need the arguments to be provided, making it quite difficult to use without setting up custom scripts or environment variables.

This in turn is resulting in very low adoption of an otherwise excellent feature: just [11 instances](https://github.com/search?q=set%28SDKCONFIG+%22%24%7BCMAKE_BINARY_DIR%7D%2Fsdkconfig%22%29&type=code) of the above CMake snippet are found on GitHub. There are no doubt hundreds of developers making multiple clones of their repos in order to change a few menuconfig options (and then gitignoring sdkconfig because it is different between two targets).

### Describe the solution you'd like.

Perhaps the simplest way this could be improved would be the adoption of a syntax similar to `esptool`, where a file path prefixed with `@` would be appended to the list of arguments.

One could then define files such as:
```
// profiles/prod
-B build-production -DSDKCONFIG=build-production/sdkconfig -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.prod"

// profiles/debug
-B build-debug -DSDKCONFIG=build-debug/sdkconfig -DSDKCONFIG_DEFAULTS="sdkconfig.defaults;sdkconfig.debug"
```

And then call:

```
idf.py @profile/prod build
idf.py @profile/debug build
```

This would be an improvement even if the files simply contained e.g. `-DAPP_PRODUCTION=1` and then CMake script did the rest as the interface with `idf.py` would be much improved.

CMake [already supports](https://cmake.org/cmake/help/v3.25/manual/cmake.1.html#cmdoption-cmake-C) a similar feature, though it is only for cache entries. `-C` is used in `idf.py` as the project path, so perhaps `--initial-cache` could be used instead.

[...]

PR #11821, opened 2023-07-06
feat(idf.py): Allow adding arguments from file via @filename.txt (#11783) (IDFGH-10584)

nebkat:
See discussion in #11783.

Replicates functionality from `esptool.py` allowing arguments to be loaded from a file, simplifying multiple build profiles.

See espressif/esptool@d9b82fa7d2e727b0f8d989e29b06a49b2c22a871

### DIFF
--- a/tools/idf.py
+++ b/tools/idf.py
@@ -18,6 +18,7 @@ import json
 import locale
 import os
 import os.path
+import shlex
 import subprocess
 import sys
 from collections import Counter, OrderedDict, _OrderedDictKeysView
@@ -695,7 +696,7 @@ def init_cli(verbose_output: List=None) -> Any:
     return CLI(help=cli_help, verbose_output=verbose_output, all_actions=all_actions)
 
 
-def main() -> None:
+def main(argv=None) -> None:
     # Check the environment only when idf.py is invoked regularly from command line.
     checks_output = None if SHELL_COMPLETE_RUN else check_environment()
 
@@ -713,7 +714,54 @@ def main() -> None:
         else:
             raise
     else:
-        cli(sys.argv[1:], prog_name=PROG, complete_var=SHELL_COMPLETE_VAR)
+        argv = expand_file_arguments(argv or sys.argv[1:])
+
+        cli(argv, prog_name=PROG, complete_var=SHELL_COMPLETE_VAR)
+
+
+def expand_file_arguments(argv):
+    """
+    Any argument starting with "@" gets replaced with all values read from a text file.
+    Text file arguments can be split by newline or by space.
+    Values are added "as-is", as if they were specified in this order
+    on the command line.
+    """
+    visited = set()
+    expanded = False
+
+    def expand_args(args, parent_path, file_stack):
+        expanded_args = []
+        for arg in args:
+            if not arg.startswith("@"):
+                expanded_args.append(arg)
+            else:
+                nonlocal expanded, visited
+                expanded = True
+
+                file_name = arg[1:]
+                rel_path = os.path.normpath(os.path.join(parent_path, file_name))
+
+                if rel_path in visited:
+                    file_stack_str = ' -> '.join(['@' + f for f in file_stack + [file_name]])
+                    raise FatalError(f'Circular dependency in file argument expansion: {file_stack_str}')
+                visited.add(rel_path)
+
+                try:
+                    with open(rel_path, "r") as f:
+                        for line in f:
+                            expanded_args.extend(expand_args(shlex.split(line), os.path.dirname(rel_path), file_stack + [file_name]))
+                except IOError:
+                    file_stack_str = ' -> '.join(['@' + f for f in file_stack + [file_name]])
+                    raise FatalError(f"File '{rel_path}' (expansion of {file_stack_str}) could not be opened. "
+                                     "Please ensure the file exists and you have the necessary permissions to read it.")
+        return expanded_args
+
+    argv = expand_args(argv, os.getcwd(), [])
+
+    if expanded:
+        print(f'Running: idf.py {" ".join(argv)}')
+
+    return argv
 
 
 def _valid_unicode_config() -> Union[codecs.CodecInfo, bool]:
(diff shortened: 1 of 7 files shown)


tools/idf.py
```diff
@@ -713,7 +713,32 @@ def main() -> None:
...
+
+def expand_file_arguments(argv):
+    """
+    Any argument starting with "@" gets replaced with all values read from a text file.
+    Text file arguments can be split by newline or by space.
+    Values are added "as-is", as if they were specified in this order
+    on the command line.
+    """
+    new_args = []
+    expanded = False
```
dobairoland, line 729: I think this flag is unnecessary. `if expanded` could be replaced by `if len(new_args) > 0`.

tools/idf.py
```diff
@@ -713,7 +713,32 @@ def main() -> None:
...
+    for arg in argv:
+        if arg.startswith("@"):
+            expanded = True
+            with open(arg[1:], "r") as f:
+                for line in f.readlines():
+                    new_args += shlex.split(line)
+        else:
+            new_args.append(arg)
+    if expanded:
+        print("idf.py %s" % (" ".join(new_args[1:])))
```
dobairoland, line 739: Could you use please `.format()` instead of the old `%`?

dobairoland (changes requested):
Thank you @nebkat for the contribution. Could you please add a test for this in https://github.com/espressif/esp-idf/blob/master/tools/test_idf_py/test_idf_py.py?

tools/idf.py
kumekay: In python, it's sufficient to check `if new_args`, no need to use len

tools/idf.py
```diff
@@ -713,7 +713,32 @@ def main() -> None:
...
+    Text file arguments can be split by newline or by space.
+    Values are added "as-is", as if they were specified in this order
+    on the command line.
+    """
+    new_args = []
+    expanded = False
+    for arg in argv:
+        if arg.startswith("@"):
+            expanded = True
+            with open(arg[1:], "r") as f:
```
kumekay, line 733: I would suggest printing a meaningful error for the user in the case when the file doesn't exist instead of python stack trace.

kumekay (commented):
@nebkat Thank you for bringing this useful feature to idf.py

tools/idf.py
```diff
@@ -713,7 +713,32 @@ def main() -> None:
...
+                for line in f.readlines():
+                    new_args += shlex.split(line)
```
mfialaf, line 735: There is a missing `import shlex` at the beginning of the file.

tools/idf.py
mfialaf: I would prepend the output with `Running: idf.py ...`

It's probably a little mistake, but if not, please explain why you are skipping the first argument in new_args when printing.

tools/idf.py
```diff
@@ -713,7 +713,32 @@ def main() -> None:
...
+        else:
+            new_args.append(arg)
+    if expanded:
+        print("idf.py %s" % (" ".join(new_args[1:])))
+        return new_args
+    return argv
```
mfialaf, line 741: `new_args` and `argv` should be the same even when there is no `@file.txt`
```suggestion
    return new_args
```

mfialaf (commented):
@nebkat Thank you for your improvement. I left a couple of notes to improve

nebkat:
Will make these changes now + will introduce possibility of recursive @ parameters (with check for circular dependency of course).

Re the various changes suggested - this was copied directly from `esptool.py`, so some of these should possibly be applied there too:

https://github.com/espressif/esptool/blob/091c4a4149943828a7d8ace01ea3e6a448a8da2f/esptool/__init__.py#L926-L938

tools/idf.py
nebkat: `new_args` also contains the non-file arguments in `argv`, so the flag is needed to detect if there were any expansions - but could probably rename `new_args` as it implies it only contains the newly expanded args.

nebkat:
Now supports recursive expansion (e.g. @profile/device_v2_release contains @release @device_v2 @partition_16mb).

Tests written but I couldn't get them to run here, let me know if any changes needed.

tools/idf.py
```diff
@@ -713,7 +714,54 @@ def main() -> None:
...
+                    file_stack_str = " -> ".join(["@" + f for f in file_stack + [file_name]])
+                    raise FatalError(f"Circular dependency in file argument expansion: {file_stack_str}")
+                visited.add(rel_path)
+
+                try:
+                    with open(rel_path, "r") as f:
+                        for line in f:
+                            expanded_args.extend(expand_args(shlex.split(line), os.path.dirname(rel_path), file_stack + [file_name]))
+                except IOError:
+                    file_stack_str = " -> ".join(["@" + f for f in file_stack + [file_name]])
```
mfialaf, line 754: Please rebase to the current master branch. The fix for the failing pipeline is present there.

The pre-commit will maybe raise some errors that should be fixed. One of them will be to change the double quotations `"` for single quotations `'` in ifd.py where possible.
```suggestion
                    file_stack_str = ' -> '.join(['@' + f for f in file_stack + [file_name]])
```

tools/idf.py
```diff
@@ -713,7 +714,54 @@ def main() -> None:
...
+                    raise FatalError(f"File '{rel_path}' (expansion of {file_stack_str}) could not be opened. "
```
mfialaf, line 755: Here the `"` is okay, since `'` are used in the string.

mfialaf (commented):
@nebkat Thank you for addressing the comments.  I left few more with the solution how to fix the failing test and pipeline. Please address also those, then we can merge your PR.

Btw very nice work! We appreciate it.

dobairoland:
sha=4c1f574c3d495d292568a91dd4990d796f1cf3ee

mfialaf:
@nebkat Thanks for the contribution again. We will take care of it from now. No need to fix the pipeline.

Issue #11877, opened 2023-07-14
MCPWM: Configurable Pull up/down in `mcpwm_generator_config_t` (IDFGH-10646)


abrondijk:
### Is your feature request related to a problem?

I'm attempting to control a motordriver on a pcb with external pull-downs on the PWM lines.
There's currently no way to disable the internal pull-ups on the generators in the mcpwm component: https://github.com/espressif/esp-idf/blob/36908cc5fde2a67714483ad4eb65b9a7cf1b984d/components/driver/mcpwm/mcpwm_gen.c#L92

### Describe the solution you'd like.

A solution similar to that of the `mcpwm_gpio_fault_config_t`, `mcpwm_capture_channel_config_t` and `mcpwm_gpio_sync_src_config_t`.

These have a `pull_up` and `pull_down` field.

PR #11878, opened 2023-07-14
feat(mcpwm): Allow for pull up/down to be configurable on generators (IDFGH-10647)

abrondijk:
Implements feature requested in https://github.com/espressif/esp-idf/issues/11877.

Copies config fields from other gpio config types out of the MCPWM component.


jrahlf:
Would be great if you could also specify whether you want a push pull or an open drain output...

suda-morris:
sha=aa2c0a10c3ef43d3a8da527fab62d137a6a7486b

PR #11923, opened 2023-07-20
fix pointer cast + condition (IDFGH-10697)

alexxmos:
Fixed issues that prevented the correct use of ATECC608 in custom mode:
- Updated condition to use the right data for comparison
- Added missing brackets when casting variable

### DIFF
--- a/components/esp-tls/esp_tls_mbedtls.c
+++ b/components/esp-tls/esp_tls_mbedtls.c
@@ -973,7 +973,7 @@ static esp_err_t esp_set_atecc608a_pki_context(esp_tls_t *tls, const void *pki)
     }
     mbedtls_x509_crt_init(&tls->clientcert);
 
-    if(((esp_tls_pki_t *)pki)->publiccert_pem_buf != NULL) {
+    if (((esp_tls_pki_t *)pki)->publiccert_pem_buf != NULL) {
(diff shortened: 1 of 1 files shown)


components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -973,8 +973,8 @@ static esp_err_t esp_set_atecc608a_pki_context(esp_tls_t *tls, const void *pki)
     }
     mbedtls_x509_crt_init(&tls->clientcert);
 
-    if(cfg->clientcert_buf != NULL) {
-        ret = mbedtls_x509_crt_parse(&tls->clientcert, (const unsigned char*)((esp_tls_pki_t *)pki->publiccert_pem_buf), (esp_tls_pki_t *)pki->publiccert_pem_bytes);
+    if(((esp_tls_pki_t *)pki)->publiccert_pem_buf != NULL) {
```
AdityaHPatwardhan, line 976:
```suggestion
    if (((esp_tls_pki_t *)pki)->publiccert_pem_buf != NULL) {
```

AdityaHPatwardhan (commented):
Hi @alexxmos
Just one nitpick, since we are anyways updating that line.
The MR looks good to me,
Thanks for finding out the issue and raising the fix.

AdityaHPatwardhan:
@alexxmos Can you please update the changes in the same commit.
If I edit the commit then the commit SHA would change and it may lead to your contribution not being properly displayed.

alexxmos:
@AdityaHPatwardhan thanks, just applied the suggestion, let me know if that's ok.

mahavirj:
@alexxmos Please squash the commits. Also it would be good if you could add more details in the commit message. Thanks.

PR #11926, opened 2023-07-21
Support NVS image generation from CMake (IDFGH-10702)

higaski:
In the spirit of [spiffs_create_partition_image](https://github.com/espressif/esp-idf/blob/master/components/spiffs/project_include.cmake) I thought it might be a good idea to create a similar function for NVS storage. In my particular case it releases me from the burden to explicitly check in code if certain settings are either

1. initialized at all or
2. lie within certain bounds 

The API is strongly based on the SPIFFS variant. Users simply pass the partition, the CSV file containing the entries and an optional FLASH_IN_PROJECT parameter to add a dependency to `idf.py flash`.

```cmake
nvs_create_partition_image(nvs ../sample_val.csv FLASH_IN_PROJECT)
```

I've used the [sample_val.csv](https://github.com/espressif/esp-idf/blob/master/components/nvs_flash/nvs_partition_generator/sample_val.csv) file to test this in an example [here](https://github.com/higaski/nvs_create_partition_image).

I would also write the documentation for it, but ask for guidelines where the relevant part is desired.

/edit
The english documentation that is, obviously. :smile:

### DIFF
--- a/components/nvs_flash/project_include.cmake
+++ b/components/nvs_flash/project_include.cmake
@@ -2,10 +2,11 @@
 #
 # Create a NVS image of the specified CSV on the host during build and
 # optionally have the created image flashed using `idf.py flash`
-function(nvs_create_partition_image partition csv_file)
+function(nvs_create_partition_image partition csv)
     set(options FLASH_IN_PROJECT)
     set(one VERSION)
-    cmake_parse_arguments(arg "${options}" "${one}" "" "${ARGN}")
+    set(multi DEPENDS)
+    cmake_parse_arguments(arg "${options}" "${one}" "${multi}" "${ARGN}")
 
     # Default to version 2
     if(NOT DEFINED arg_VERSION)
@@ -18,7 +19,7 @@ function(nvs_create_partition_image partition csv_file)
         ${idf_path}/components/nvs_flash/nvs_partition_generator/nvs_partition_gen.py
        )
 
-    get_filename_component(csv_file_full_path ${csv_file} ABSOLUTE)
+    get_filename_component(csv_full_path ${csv} ABSOLUTE)
 
     partition_table_get_partition_info(size "--partition-name ${partition}" "size")
     partition_table_get_partition_info(offset "--partition-name ${partition}" "offset")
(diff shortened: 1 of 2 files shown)


igrr:
Thanks! Could you take a look at another PR https://github.com/espressif/esp-idf/pull/11785?

higaski:
> Thanks! Could you take a look at another PR #11785?

Well, looks almost identical although you cant pass a version there.

I now also wonder if all the encryption options should be considered...?

adokitkat:
Hello. Thank you for the contribution! Your PR will be used because it is based on our already existing APIs, but also add @nebkat from #11785 as a co-author since his PR was created earlier and it is very similar.

adokitkat:
sha=1827a968c29b3dbcc30ede3b0a3082b48081f2af

higaski:
> Hello. Thank you for the contribution! Your PR will be used because it is based on our already existing APIs, but I will also add @nebkat from #11785 as a co-author since his PR was created earlier and it is very similar.

Glad the PR got accepted. How about documentation though?

I've thought about adding some below the [NVS Partition Generator Utility](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s3/api-reference/storage/nvs_flash.html#nvs-partition-generator-utility) entry.

https://github.com/espressif/esp-idf/blob/28167ea5a3cdd49227216c4131a9df7f4eda260e/docs/en/api-reference/storage/nvs_flash.rst?plain=1#L90-L94

nebkat:
Apologies for not continuing my PR, have been on vacation.

Either PR is fine of course, I would just mention - I believe the `add_custom_command` instead of (or in in addition to) `add_custom_target` allows the image to be built only when needed due to the `OUTPUT` clause.

In SPIFFS generation we have to create the image every time as there is no way to have a dependency on an entire directory, but that is not the case here.

It is very quick either way so not a major problem, but nice to avoid additional build steps whenever possible.

higaski:
> Either PR is fine of course, I would just mention - I believe the `add_custom_command` instead of (or in in addition to) `add_custom_target` allows the image to be built only when needed due to the `OUTPUT` clause.

You're right, I've copied your version which uses `add_custom_command`. Not only is it one build step less, it also outputs less noise on the command line.

I've also added a some docs.

nebkat:
Thanks @higaski. 

For docs: if `FLASH_IN_PROJECT` is not specified you can flash with `${partition}-flash`, no need for a custom build target.

higaski:
> Thanks @higaski.
> 
> For docs: if `FLASH_IN_PROJECT` is not specified you can flash with `${partition}-flash`, no need for a custom build target.

Thanks, updated.

adokitkat:
sha=0edf332062a2254e1aaa4f39158fca4c5bb8a110

Issue #11591, opened 2023-06-05
console: "help <command>" in addition to plain "help"


crackwitz:
### Is your feature request related to a problem?

I have a lot of commands. Calling "help" confronts the user with a listing several pages long.

### Describe the solution you'd like.

I'd like `help_command(int argc, char **argv)` to accept an optional first argument.

Given that first argument, the `help` command will print just that command's `help` string.

### Describe alternatives you've considered.

I've considered implementing `help` myself but that'll probably have no access to the list of registered commands and their argtables. Tthe `s_cmd_list` in `help_command` in `components/console/commands.c` isn't exposed.

### Additional context.

I'd be willing to propose the code for `help_command`. I don't anticipate it being more than a fistful of lines. It'd merely have to filter the iteration, skipping each command that doesn't match the query.

Would be nice if prefix/infix matches were allowed. `strstr()` seems useful.

igrr:
@crackwitz sounds like a good idea! You are welcome to open a pull request.

jnorgan:
I also had this same need and had done something similar. Not doing any wildcarding, but for my purposes, full command name was sufficient.

crackwitz:
I hope to find some time to get this done tomorrow.

jnorgan:
@crackwitz Here's the PR I submitted: https://github.com/espressif/esp-idf/pull/11945

jnorgan:
Commit https://github.com/espressif/esp-idf/commit/0b595f3c77d52a334e80d918a08040982bb60cfc seems to resolve this issue.

ESP-Marius:
As already mentioned, the merged PR seems to address this.

Closing the issue for now, feel free to re-open if you feel there is still something left unresolved.

PR #11945, opened 2023-07-24
Added Help command parameter for displaying only a specific command (IDFGH-10725)

jnorgan:
This adds a capability to filter the `Help` command output in ESP Console as described by https://github.com/espressif/esp-idf/issues/11591.


components/console/commands.c
```diff
@@ -240,6 +258,9 @@ static int help_command(int argc, char **argv)
 
 esp_err_t esp_console_register_help_command(void)
 {
+    help_args.help_cmd = arg_str0(NULL, NULL, "<string>", "Name of command");
+    help_args.end = arg_end(1);
+
     esp_console_cmd_t command = {
         .command = "help",
         .help = "Print the list of registered commands",
```
igrr, line 266: Probably we need to add
```
.argtable = &help_args
```
so that the output of `help` describes the `help` command correctly :)

components/console/commands.c
```diff
@@ -209,15 +209,33 @@ esp_err_t esp_console_run(const char *cmdline, int *cmd_ret)
...
 
     /* Print summary of each command */
     SLIST_FOREACH(it, &s_cmd_list, next) {
         if (it->help == NULL) {
             continue;
         }
+        if (len > 0 &&
+                (strlen(it->command) != len ||
+                 strcmp(help_args.help_cmd->sval[0], it->command) != 0)) {
+            continue;
```
igrr, line 237: Not sure there's a lot of time saved by checking strlen first and then calling strcmp... For different commands, strcmp will bail out very quickly, while strlen will always scan the command name until the end. Perhaps we can simplify things slightly by removing `size_t len` and simply using `strcmp` here?

jnorgan: Removing the `size_t len`, but keeping the length check in the conditional because we still want the `help` command with no parameters to dump the entire command list

0xjakob (changes requested):
@jnorgan Code looks good, thanks for the contribution! Please squash and (ideally) rebase the commits so we can apply the commit directly. If we need to change your history, your authorship will be retained but this PR will later only be marked as "closed" instead of "merged".

0xjakob:
sha=dec47be2532355fdeda4a9afcf12e1b5ae10c576

0xjakob:
sha=0ef3c8f3600ad1e4888feb22d619882f512ab9fd

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/0b595f3c77d52a334e80d918a08040982bb60cfc.

PR #11987, opened 2023-08-01
idf-docker-image.rst: Run with current user ID instead of root. (IDFGH-10773)

jimmo:
I'm from the MicroPython project, and I noticed an issue with someone trying to follow the instructions at https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-docker-image.html to build MicroPython. Specifically, our build uses `git` to embed version information. They also noticed that the generated build directory was owned by root.

I'm not a docker expert at all, so there might be better ways to solve this, but I've been using exactly this mechanism of using containers for building firmware for many years without trouble. (Although I typically use `-v $HOME:$HOME -w $PWD -u $UID` instead, but I think in this case using `/project` makes more sense).

---

Rather than running the command inside the container as root, which will mean that any build artifacts created will be owned by root on the host, run the command as the current user.

This requires setting a temporary home directory as idf.py will try to access e.g. ~/.cache, so just use /tmp inside the container which is ephemeral anyway.

This also allows the command to use `git`. without setting the user ID, `docker run ... git status` will fail with
    fatal: detected dubious ownership in repository at '/project'

Also added the missing explanation for `-w /project`.

### DIFF
--- a/docs/en/api-guides/tools/idf-docker-image.rst
+++ b/docs/en/api-guides/tools/idf-docker-image.rst
@@ -53,13 +53,16 @@ In the project directory, run:
 
 .. code-block:: bash
 
-    docker run --rm -v $PWD:/project -w /project espressif/idf idf.py build
+    docker run --rm -v $PWD:/project -w /project -u $UID -e HOME=/tmp espressif/idf idf.py build
 
 The above command explained:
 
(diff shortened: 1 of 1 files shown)


dobairoland:
Thank you for the contribution!

dobairoland:
sha=c7d511277163726434e08725b250cddb38c5c660

PR #11991, opened 2023-08-01
initialize dataset with otDatasetCreateNewNetwork (IDFGH-10778)

FHFS:
See https://github.com/espressif/esp-thread-br/issues/25

### DIFF
--- a/components/openthread/src/esp_openthread.cpp
+++ b/components/openthread/src/esp_openthread.cpp
@@ -92,7 +92,8 @@ esp_err_t esp_openthread_auto_start(otOperationalDatasetTlvs *datasetTlvs)
         otOperationalDataset dataset;
         size_t len = 0;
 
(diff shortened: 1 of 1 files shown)


zwx1995esp:
Thanks for your commit! We also have a MR to fix that issue on our internal branch.

But actually, we find we also need to set mesh local prefix for the dataset, if do not, the leader will generate a random one for thread network, and the device which is going to join this network will use another generate mesh local prefix, so when transmit UDP packets, the checksum will be calculated using different IP address. So there should be the issue `dropped udp packets with bad checksums`.

FHFS:
I did not have any dropped udp packets when I tested it. But if it is already in your MR I will close this PR. thanks.

zwx1995esp:
Ok, I will sync th info to you when the MR is merged.

PR #11992, opened 2023-08-01
Fix `rx_buffer` type in TCP example (IDFGH-10780)

Daniel-Chin:
In the example projects for protocol/socket/TCP..., the `rx_buffer` was of type `char*`. I changed it to `uint8_t*`.  

Semantically, TCP transmits a byte stream, not a char stream. Using `char*` in the example can potentially confuse beginners.  

Let me know if there are further considerations.

### DIFF
--- a/examples/protocols/sockets/tcp_client/main/tcp_client_v4.c
+++ b/examples/protocols/sockets/tcp_client/main/tcp_client_v4.c
@@ -30,7 +30,7 @@ static const char *payload = "Message from ESP32 ";
 
 void tcp_client(void)
 {
-    char rx_buffer[128];
+    uint8_t rx_buffer[128];
     char host_ip[] = HOST_IP_ADDR;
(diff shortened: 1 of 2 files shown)


espressif-abhikroy:
I concur with your assessment regarding a standard TCP connection. However, in the context of this specific example, the payload has been hardcoded as an ASCII string. Therefore, the present implementation is accurate and does not require any alterations.

espressif-abhikroy:
I concur with your assessment regarding a standard TCP connection. However, in the context of this specific example, the payload has been hardcoded as an ASCII string. Therefore, the present implementation is accurate and does not require any alterations.

Daniel-Chin:
> I concur with your assessment regarding a standard TCP connection. However, in the context of this specific example, the payload has been hardcoded as an ASCII string. Therefore, the present implementation is accurate and does not require any alterations.

Understood.

PR #12039, opened 2023-08-08
Update iperf example documentation with console interface details. (IDFGH-10832)

StevenMacias:
I wanted to try `iperf` on a custom project but the serial seemed stuck after booting. I am using USB to power the board, flash and see the monitor traces. I deployed the example on a `ESP32-S3-DevKitC-1` with the same results since I was connected to the micro-usb labeled as USB. After taking a brief look to the code I noticed that the console configuration is dependent of `menuconfig` values. 

I updated the `iperf` documentation with some details about the `menuconfig` configuration to warn users about which micro-usb interface is being used by default and how to change this configuration in case they want to use the other connector like me.


MaxwellAlan:
@StevenMacias LGTM, will sync this PR to internal.

PR #12052, opened 2023-08-10
Two small patches for build system and fatfs (IDFGH-10849)

jendo42:
- idf.py gdb action incorrectly generated EOL in gdbinit
- fatfs routines incorrectly report state for read-only filesystem

### DIFF
--- a/tools/idf_py_actions/debug_ext.py
+++ b/tools/idf_py_actions/debug_ext.py
@@ -299,7 +299,7 @@ def action_extensions(base_actions: Dict, project_path: str) -> Dict:
             r.append('set confirm on')
             r.append('end')
             r.append('')
-            return os.linesep.join(r)
+            return '\n'.join(r)
         raise FatalError(f'{ESP_ROM_INFO_FILE} file not found. Please check IDF integrity.')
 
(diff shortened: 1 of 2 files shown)


tools/idf_py_actions/debug_ext.py
```diff
@@ -270,7 +270,7 @@ def generate_gdbinit_rom_add_symbols(target: str) -> str:
             r.append('set confirm on')
             r.append('end')
             r.append('')
-            return os.linesep.join(r)
+            return '\n'.join(r)
```
igrr, line 273: Did `os.linesep` cause any issue when you used it? According to the docs it should be the correct line ending depending on the host OS.

igrr (commented):
Thank you for the PR! Fatfs change looks good to me, just one question about the gdbinit part.

igrr:
@jendo42 One more request, could you please target `master` branch and not `release/v5.1`? Features and bug can be accepted on `master` branch only. We can backport the change to a release branch after it is merged into `master`.

igrr:
sha=bcda40fcf22746f91f301bfcbba768d4a1747922

Issue #12053, opened 2023-08-10
adc_continuous_read returning ESP_ERR_TIMEOUT (IDFGH-10857)


puyoulu:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.0.3

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

esp32-s3-wroom-1

### Power Supply used.

External 5V

### What is the expected behavior?

adc_continuous_read is quite easy (80% possibility) to return ESP_ERR_TIMEOUT

### What is the actual behavior?

adc_continuous_read is quite easy (80% possibility) to always return ESP_ERR_TIMEOUT (conv_frame_size set to 512), after power on reset, and s_adc_dma_intr is then never being called.
This occurs when running v5.0.3, but never occurs on v5.0.2.

I checked the history, seems this bug was introduced by commit 5da5e1801b and d8ee45ce08.
Wish you guys can confirm this.

I tried below patch, and seems can fix:
```
diff --git a/components/esp_adc/adc_continuous.c b/components/esp_adc/adc_continuous.c
index 643dfef..ed1f760 100644
--- a/components/esp_adc/adc_continuous.c
+++ b/components/esp_adc/adc_continuous.c
@@ -355,6 +355,11 @@ static IRAM_ATTR bool s_adc_dma_intr(adc_continuous_ctx_t *adc_digi_ctx)
         }
     }
 
+    if (status == ADC_HAL_DMA_DESC_NULL) {
+        //start next turns of dma operation
+        adc_hal_digi_start(&adc_digi_ctx->hal, adc_digi_ctx->rx_dma_buf);
+    }
+
     return need_yield;
 }
 
diff --git a/components/hal/adc_hal.c b/components/hal/adc_hal.c
index 3633816..2051ce6 100644
--- a/components/hal/adc_hal.c
+++ b/components/hal/adc_hal.c
@@ -238,7 +238,6 @@ static void adc_hal_digi_dma_link_descriptors(dma_descriptor_t *desc, uint8_t *d
     HAL_ASSERT(((uint32_t)data_buf % 4) == 0);
[...]

PR #12065, opened 2023-08-12
adc: fix adc dma interrupt stops under heavy load (IDFGH-10863)

puyoulu:
The adc rx dma eof interrupt will stop generating after all rx eof descriptor is owered by DMA engine. This may happen under heavy load (eg. wifi initialization), and eof_desc_addr should not be used to check the current eof descriptor anymore (for there maybe long delays during adc_dma_intr_handler execution).
Fix this by disabling owner check of descriptor in the DMA engine side, and CPU determines valid eof descriptor by polling the owner bit.

closing #12053

### DIFF
--- a/components/esp_adc/adc_continuous.c
+++ b/components/esp_adc/adc_continuous.c
@@ -201,7 +201,7 @@ esp_err_t adc_continuous_new_handle(const adc_continuous_handle_cfg_t *hdl_confi
 
     gdma_strategy_config_t strategy_config = {
         .auto_update_desc = true,
-        .owner_check = true
+        .owner_check = false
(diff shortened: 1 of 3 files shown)


Alvin1Zhang:
Thanks for your contribution.

railbox:
Any reason why this was closed? Looks like it resolve same bug in my case too.

Icarus113:
> Any reason why this was closed? Looks like it resolve same bug in my case too.

A [PR](https://github.com/espressif/esp-idf/pull/11411) with similar changes have been accepted before. You may upgrade to use a newer version.

PR #12097, opened 2023-08-18
docs(jtag-debugging): add links to gdbstub for convenience (IDFGH-10898)

chipweinberger:
gdbstubs are kind of a hidden feature.

If you google "esp32 debugging", you'll find nothing about GDB Stubs. 

They need to be linked more. They should probably even have their own article. 

---

**suggestion**: There should be a separate **ESP32 debugging** article, that links to
-  JTAG debugging
- GDB Stubs debugging

### DIFF
--- a/docs/en/api-guides/jtag-debugging/index.rst
+++ b/docs/en/api-guides/jtag-debugging/index.rst
@@ -2,7 +2,15 @@ JTAG Debugging
 ==============
 :link_to_translation:`zh_CN:[中文]`
 
-This document provides a guide to installing OpenOCD for {IDF_TARGET_NAME} and debugging using GDB. The document is structured as follows:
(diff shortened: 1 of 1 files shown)


igrr (commented):
Thanks for adding this @chipweinberger! I agree in general, but I think placing this ahead of all the content makes the note look more important than anything else in the document.

Would you be okay with it being placed somewhere lower in the document?

chipweinberger:
Maybe like this?

## JTAG Debugging

This document provides a guide to installing OpenOCD for ESP32 and debugging using GDB. 

> Note: You can also debug your ESP32 with GDB when using ``idf.py monitor`` without needing to setup JTAG or OpenOCD. See: :doc:`IDF Monitor <../../api-guides/tools/idf-monitor.html>` and :ref:CONFIG_ESP_SYSTEM_GDBSTUB_RUNTIME

The document is structured as follows:

PR #12105, opened 2023-08-20
espcoredump: Support for not overwriting existing core dump in flash (IDFGH-10907)

### LINKED ISSUE #12027: Option to limit the number of core dump writes to flash
### Is your feature request related to a problem?

I'd like to be able to enable the `coredump` partition but during development there will be times where the application crashes immediately on startup. This results in a fast loop of repeatedly overwriting a core dump to the same area of flash.

I'm concerned about wearing out the flash when an crash involving a reboot loop occurs.

### Describe the solution you'd like.

I'd like to be able to limit the number of core dump writes to flash, so that after a core dump has been written no more writes will occur until the existing core dump is erased.

### Describe alternatives you've considered.

_No response_

### Additional context.

Currently only one core dump to flash is supported so the only valid limit (if enabled) will be `1` until storage of multiple core dumps becomes a feature.

The only change required to implement this would be to check if there is already a core dump present in `esp_core_dump_to_flash()` and `return` instead of writing the core dump.

nomis:
If there's an unattended boot loop or a crash that causes another crash on the next boot, it needs to be possible to avoid overwriting a saved core dump with another core dump.

Add an option to do this and skip writing core dumps if the partition isn't empty.

Fixes #12027.

### DIFF
--- a/components/espcoredump/src/core_dump_flash.c
+++ b/components/espcoredump/src/core_dump_flash.c
@@ -27,6 +27,10 @@ typedef struct _core_dump_partition_t
     uint32_t size;
     /* Flag set to true if the partition is encrypted. */
     bool encrypted;
+#if CONFIG_ESP_COREDUMP_FLASH_NO_OVERWRITE
+    /* Flag set to true if the partition is empty. */
+    bool empty;
+#endif
 } core_dump_partition_t;
 
 typedef uint32_t core_dump_crc_t;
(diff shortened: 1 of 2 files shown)


components/espcoredump/src/core_dump_flash.c
```diff
@@ -82,6 +86,17 @@ void esp_core_dump_flash_init(void)
     s_core_flash_config.partition.start      = core_part->address;
     s_core_flash_config.partition.size       = core_part->size;
     s_core_flash_config.partition.encrypted  = core_part->encrypted;
+
+#if CONFIG_ESP_COREDUMP_FLASH_NO_OVERWRITE
+    uint32_t core_size;
+    esp_err_t err = esp_partition_read(core_part, 0, &core_size, sizeof(core_size));
+    if (err == ESP_OK && core_size == BLANK_COREDUMP_SIZE) {
```
o-marshmallow, line 93: If the read partition caused an error it may be better to print an error:
```
if (err == ESP_OK) {
    s_core_flash_config.partition.empty = (core_size == BLANK_COREDUMP_SIZE);
} else {
    ESP_COREDUMP_LOGE("Could not read coredump partition size (%d)", err);
}
```

nomis: I've done this and reused the same message as elsewhere in the file.

chipweinberger:
> "Not overwriting existing core dump"

Think this would be more clear:

"Existing coredump found. Will not overwrite."

nomis:
I've rewritten the warning message to be clear about what happened

o-marshmallow:
sha=ee3e01cf5c556e80615fd58b1baacf570865f679

o-marshmallow:
The MR has been merged internally, it will be synced to Github after passing the internal CI

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/1feb3c9f9b4a9f193b5df7f8abdb19a772817018.

PR #12119, opened 2023-08-22
jtag serial drive write function loses characters when rx ring buffer… (IDFGH-10925)

jjsch-dev:
upon examining the code, I noticed that the [usbjtag_tx_char_via_driver](https://github.com/espressif/esp-idf/blob/32472536715d674c160a2565795895e0273f8cde/components/vfs/vfs_usb_serial_jtag.c#L406) function contains a comparison that checks whether an immediate send operation succeeds. However, the return value from [usb_serial_jtag_write_bytes](https://github.com/espressif/esp-idf/blob/32472536715d674c160a2565795895e0273f8cde/components/driver/usb_serial_jtag/usb_serial_jtag.c#L158) is always 1, which prevents the intended logic to try blocking for 50 mS, but also a return that always skyp the bloking try.

### DIFF
--- a/components/driver/usb_serial_jtag/usb_serial_jtag.c
+++ b/components/driver/usb_serial_jtag/usb_serial_jtag.c
@@ -168,8 +168,8 @@ int usb_serial_jtag_write_bytes(const void* src, size_t size, TickType_t ticks_t
     // Now trigger the ISR to read data from the ring buffer.
     usb_serial_jtag_ll_ena_intr_mask(USB_SERIAL_JTAG_INTR_SERIAL_IN_EMPTY);
        
-    // Return -1 on failure, 0 on success.
-    return (result == pdFALSE) ? -1 : 0;
+    // Return 0 on failure, size on success.
+    return (result == pdFALSE) ? 0 : size;
 }
 
 esp_err_t usb_serial_jtag_driver_uninstall(void)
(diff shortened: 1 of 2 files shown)


ginkgm:
usb_serial_jtag_write_bytes never returning 0 is a real issue. We will fix this.

The API will return 0 when write fails. Change this to -1 is a breaking change that we can't do.

From the VFS side, losing data when ringbuffer full is a expected behavior. We can't detect the connection to host, so we use the fifo overflow as indication.

jjsch-dev:
Hi @ginkgm, thanks for the explanation, if you want I'll close the PR.

jjsch-dev:
Hi @ginkgm, I modified the return of usb_serial_jtag_write_bytes to return 0 on failure and the size on success, and reverted the change to usbjtag_tx_char_via_driver.

chipweinberger:
@jjsch-dev thanks for creating the PR!

jjsch-dev:
@chipweinberger, thanks for the comment.

mythbuster5:
@jjsch-dev  Yes, this time changes is absolutely correct, I already have an internal PR to fix this, and that will be merged soon. Thanks for your contribution.

jjsch-dev:
@mythbuster5, thanks for the confirmation, I will wait the merge.

PR #12131, opened 2023-08-23
Add support for X-WWW-Authenticate header (IDFGH-10936)

0xFEEDC0DE64:
I also need the support for the http header X-WWW-Authenticate in my application

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -250,7 +250,8 @@ static int http_on_header_value(http_parser *parser, const char *at, size_t leng
     } else if (strcasecmp(client->current_header_key, "Transfer-Encoding") == 0
                && memcmp(at, "chunked", length) == 0) {
         client->response->is_chunked = true;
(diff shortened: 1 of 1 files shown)


hmalpani:
Hello @0xFEEDC0DE64 

As a part of esp-idf, by default, only the standard HTTP headers are supported. Adding the support `X-WWW-Authenticate` cannot be done in the way it is done in this PR. But I understand that users should have some method to support custom headers as well.
I am sharing a patch with you which have support for custom authentication headers. You can use the new API introduced in the patch to set the auth data. You have to call the new API in [event handler](https://github.com/espressif/esp-idf/blob/master/examples/protocols/esp_http_client/main/esp_http_client_example.c#L65-L67):

```
case HTTP_EVENT_ON_HEADER:
            if (strcasecmp(client->current_header_key, "X-WWW-Authenticate") == 0) {
                        esp_http_client_set_auth(event->client, event->data, event->data_len);
            }
            break;
```

Can you try this patch and let me know if this works for you?

Thanks!

hmalpani:
Hello @0xFEEDC0DE64 
Have you tried the patch shared above? Please let know if it works for you.

PR #12177, opened 2023-09-01
mbedtls: define MBEDTLS_SSL_CID_TLS1_3_PAD_GRANULARITY for CID padding

hasheddan:
Updates config to define the new MBEDTLS_SSL_CID_TLS1_3_PAD_GRANULARITY
option, which replaced the previously used
MBEDTLS_SSL_CID_PADDING_GRANULARITY. The old option is continuing to be
used as the new one exceeds the maximum length for an option name in
esp-idf.

See https://github.com/Mbed-TLS/mbedtls/pull/4490 for more information.

Signed-off-by: Daniel Mangum <georgedanielmangum@gmail.com>

> Note: enabling `MBEDTLS_SSL_CID_PADDING_GRANULARITY` without this fix currently breaks builds.

### DIFF
--- a/components/mbedtls/Kconfig
+++ b/components/mbedtls/Kconfig
@@ -280,13 +280,13 @@ menu "mbedTLS"
                     Maximum length of CIDs used for outgoing DTLS messages
 
             config MBEDTLS_SSL_CID_PADDING_GRANULARITY
-                int "Record plaintext padding (for DTLS 1.2)"
+                int "Record plaintext padding"
                 default 16
                 range 0 32
                 depends on MBEDTLS_SSL_DTLS_CONNECTION_ID
                 help
-                    Controls the use of record plaintext padding when
-                    using the Connection ID extension in DTLS 1.2.
+                    Controls the use of record plaintext padding in TLS 1.3 and
+                    when using the Connection ID extension in DTLS 1.2.
 
                     The padding will always be chosen so that the length of the
                     padded plaintext is a multiple of the value of this option.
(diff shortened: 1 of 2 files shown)


hasheddan (commented):
Hmm seeing the following:
```
MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY is 42 characters long and it should be 40 at most
```
Anyway to override so that we can match the underlying mbedtls config?

hasheddan:
> You would have to shorten the Kconfig option name to 40 characters 

@laukik-hase is there any possibility of overriding the pre-commit hook? I think it would be unfortunate to not mirror the `mbedTLS` option here.

hasheddan (commented):
@laukik-hase any updates here? Is shortening the config option an absolute requirement? Would love to get this merged as Connection ID support is currently broken without it.

igrr:
I'm afraid no, this limit is currently enforced globally and I would prefer not to change it just for this case. Please see if you can shorten the name.

Additionally, since you are renaming a Kconfig option, to ensure compatibility please create components/mbedtls/sdkconfig.rename and add the old and the new option name there ([docs](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html?#backward-compatibility-of-kconfig-options), [example](https://github.com/espressif/esp-idf/blob/master/components/esp_gdbstub/sdkconfig.rename)).

hasheddan:
@igrr in that case, would it make sense to just keep the previous option and have it select the new one?

hasheddan:
@igrr I have pushed an update here reflecting the use of the existing option to define the new mbedTLS option

laukik-hase:
sha=7c63769edefda724ef5c7f7d5e20b33f9e3036ee

hasheddan:
@igrr @laukik-hase anything else needed here to merge?

laukik-hase:
@hasheddan, the PR has been merged internally - it will close automatically when it is synced to GitHub.

laukik-hase:
Merged with https://github.com/espressif/esp-idf/commit/a57c8dc93841575c9089525dc88c377494d08762,closing this PR!

PR #12179, opened 2023-09-01
Update uart_ll.h - Fixing division by zero (IDFGH-10990)

educampos28:
I have faced the division by zero when the baud rate is higher than 1Mbits/s.    It happens because the integer division returns zero for sclk_div while slightly below 1.0.

I also added 0.5 to improve the roundup process of converting float to integer.

With that fix, I can reach 3.125Mbits/s.

### DIFF
--- a/components/hal/esp32s3/include/hal/uart_ll.h
+++ b/components/hal/esp32s3/include/hal/uart_ll.h
@@ -131,9 +131,9 @@ FORCE_INLINE_ATTR void uart_ll_get_sclk(uart_dev_t *hw, soc_module_clk_t *source
  */
 FORCE_INLINE_ATTR void uart_ll_set_baudrate(uart_dev_t *hw, uint32_t baud, uint32_t sclk_freq)
 {
-#define DIV_UP(a, b)    ((float)((a) + (b) - 1) / (b))
+#define DIV_UP(a, b)    ((uint64_t)((a) + (b) - 1) / (b))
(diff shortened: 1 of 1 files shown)


songruo:
Hi @educampos28, 

`DIV_UP(a, b)` should already guarantee that the result is equal or larger than 1, as long as `a >= 1` (it is always true for `sclk_freq`).

Also I have tried on ESP32S3 with a baud rate of 3.125Mbits/s, the code runs fine without crash.

Can you provide more details, such as compile option, sdkconfig, UART source clock selection, baudrate configurations, that would lead to the "division by zero" error?

educampos28:
I removed my changes and ran the code again.  I think that image will help a lot.  


image
image

songruo:
@educampos28 Thanks for the debug! Do you want to fix your PR?

I guess casting `max_div * baud` to `uint64_t` would fix the problem.

And would you also apply the fix to all other targets?

Then we can go through the PR process!

educampos28:
it will fix.  I prefer to use float in this case to guarantee a better result while rounding the number.  Also, that function is called once when the user start their system, which the float math won't be a firmware overload for the firmware.

Sorry, I don't have time to push this fix to other targets.

songruo:
sha=3f40c80c06b2d9057fd651db98f6c5b18f64a8b6

suda-morris:
merged in https://github.com/espressif/esp-idf/commit/b0bde58521b1869da6d3aa7d6fcc0f8abfb71060

Issue #12183, opened 2023-09-03
ESP32-C3 esp_deep_sleep_start changes enable of internal pullup and pulldown internal resistor (IDFGH-10994)


spilz87:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

V4.4.5

### Operating System used.

Windows

### How did you build your project?

Command line with Make

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32-C3

### Power Supply used.

USB

### What is the expected behavior?

I'm building a remote control with an ESP32-C3.
As it works on battery, low consumption is verry important.

I want to wake up from deep sleep when a switch is relased on GPIO 0. I use 100k pull up resistors for the switch to consume less as possible.

when I enter in sleep mode, I want to disable internal pull-up/pull-down resistors as I use the external ones and I expect the ESP32-C3 wakes up from LOW to HIGH on GPIO 0.

### What is the actual behavior?

when ESP32-C3 is deep sleep mode and I relase the switch, I mesure only 0.4v on GPIO 0 instead of 3.3V (cf external 100k pull-up resistor) so ESP32-C3 doesn't wake up.

### Steps to reproduce.

Step 1 : connect a 100k pull-up resistor on GPIO 0 
Step 2 : connect a switch between GPIO 0 and GND
Step 3 : close the switch
Step 4 : enter in deep sleep mode with GPIO wake up on HIGH for GPIO 0
Step 5 : open the switch
Step 6 : ESP32-C3 doesn't wake up
Step 7 : measure the voltage on GPIO 0 : it will be 0.4V instead of 3.3V

![switch](https://github.com/espressif/esp-idf/assets/20712939/5650bf7e-6813-4e5a-8c3c-bffb24c8de22)

### More Information.

On 4.4.4 the issue was only in the first enter in deepsleep mode
On 4.4.5 the issue is alway present.

I discover that **esp_deep_sleep_start** enables internal pull-down resistor (~10k) which is not compatible with the external 100k pull-up resistor.

We should be able to choose if we want to use or not the internal resistor.

this issue is linked to #11024 which constat the issue but doesn't really solve it.
[...]

PR #12186, opened 2023-09-04
Spilz87 patch 1 (IDFGH-10997)

spilz87:


spilz87:
> Seems this PR was created by mistake, thanks for contribution, feel free to reopen.

Actually it’s a real proposal but I don’t know how to do it, I never do it before. 

The explanation of the issue and this modification are here  #12183

If you can help me to propose it properly, it could be very nice 

thanks in advance

PR #12247, opened 2023-09-14
Keep export.sh Bourne Shell compatible (IDFGH-11073)

panduit-mik:
The '[[' operator is not supported by /bin/sh
Use '[' instead to keep it compatible.

### DIFF
--- a/export.sh
+++ b/export.sh
@@ -19,7 +19,7 @@ __verbose() {
 
 __script_dir(){
     # shellcheck disable=SC2169,SC2169,SC2039,SC3010,SC3028  # unreachable with 'dash'
-    if [[ "$OSTYPE" == "darwin"* ]]; then
+    if [ `uname -s` = "Darwin" ]; then
         # convert possibly relative path to absolute
         script_dir="$(__realpath "${self_path}")"
         # resolve any ../ references to make the path shorter


JeremyTubongbanua (approved):

mfialaf:
Hi @panduit-mik ,

Thank you for your contribution.
I've tried your suggestion, and it won't work. The double square brackets with `==` and `*` are used for string comparison. Your suggestion would work without the `*` in the end, but this quantifier serves the same as in regular expressions, so $OSTYPE names like `darwin18.7.0` would also pass the condition.

In addition, $OSTYPE is not recognized by the original Bourne shell, so it would not fix your problem.

Please use newer shells like bash, dash, zsh etc..

Let me know, if you find better working solution.

panduit-mik:
Some build environments (like yocto), use bourne shell for widest compatibility which is why it would be better to continue to keep this script bourne shell compatible.
To get the equivalent pattern match in bourne shell:

if [ "${OSTYPE##darwin}" != "${OSTYPE}" ]; then

Note also that OSTYPE is also not recognized by dash on some platforms
On MacOS, /bin/sh defines $OSTYPE so at least for this case of checking for MacOS, this will work

uname is really the best most portable way to differentiate Unix OS flavors.

if [ `uname -s` = "Darwin" ];

mfialaf:
@panduit-mik Could you update this pull request in accordance with your latest comment? This will enable me to carry out testing and proceed with the merge.

If not, we will close this pull request and revisit your changes at a later time.

mfialaf:
@panduit-mik Thank you! 

And last, please squash all commits into one since it is one change. 
Thanks for the contribution.

mfialaf:
sha=c4da3d4d32633a4a27e18cf3e09c8e13e4a544b1

PR #12256, opened 2023-09-15
Less tcp packet fragmentation

0xFEEDC0DE64:
I noticed that our products waste about 40% of their tcp traffic, because we send every second a very short status update through a websocket client to our cloud and esp_websocket_client keeps sending two tcp packets, one for the websocket header, and one for the websocket payload.

I tried merging those two buffers into a single one by providing additional websocket methods that expect a unused 16 byte prefix buffer to put the websocket header into, meaning the user code has to preallocate these bytes already.

websocket_client from esp-protocols also has a patch available, which uses those new optimized websocket sending mechanisms and my wireshark recording shows great success! all websocket headers are now merged together with the websocket payload.

I expect lots of review comments and change requests, I just want to start the discussion how to integrate this properly in the esp-idf

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -123,7 +123,7 @@ static ssize_t tcp_read(esp_tls_t *tls, char *data, size_t datalen)
 
 static ssize_t tcp_write(esp_tls_t *tls, const char *data, size_t datalen)
 {
-    return send(tls->sockfd, data, datalen, 0);
+    return send(tls->sockfd, data, datalen, MSG_MORE);
 }
 
 ssize_t esp_tls_conn_read(esp_tls_t *tls, void  *data, size_t datalen)
--- a/components/esp_http_server/src/httpd_txrx.c
+++ b/components/esp_http_server/src/httpd_txrx.c
@@ -56,7 +56,7 @@ int httpd_send(httpd_req_t *r, const char *buf, size_t buf_len)
     }
 
     struct httpd_req_aux *ra = r->aux;
-    int ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, 0);
+    int ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, MSG_MORE);
     if (ret < 0) {
         ESP_LOGD(TAG, LOG_FMT("error in send_fn"));
         return ret;
@@ -70,7 +70,7 @@ static esp_err_t httpd_send_all(httpd_req_t *r, const char *buf, size_t buf_len)
     int ret;
 
     while (buf_len > 0) {
-        ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, 0);
+        ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, MSG_MORE);
         if (ret < 0) {
             ESP_LOGD(TAG, LOG_FMT("error in send_fn"));
             return ESP_FAIL;
--- a/components/tcp_transport/include/esp_transport_ws.h
+++ b/components/tcp_transport/include/esp_transport_ws.h
@@ -14,6 +14,11 @@
 extern "C" {
 #endif
 
+// this define will also be used for optimized websocket sends with the
+// optimized version to prepend the required space for the websocket
+// header
+#define MAX_WEBSOCKET_HEADER_SIZE   16
+
 typedef enum ws_transport_opcodes {
     WS_TRANSPORT_OPCODES_CONT =  0x00,
     WS_TRANSPORT_OPCODES_TEXT =  0x01,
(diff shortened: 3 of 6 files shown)


components/esp-tls/esp_tls.c
```diff
@@ -123,7 +123,7 @@ static ssize_t tcp_read(esp_tls_t *tls, char *data, size_t datalen)
 
 static ssize_t tcp_write(esp_tls_t *tls, const char *data, size_t datalen)
 {
-    return send(tls->sockfd, data, datalen, 0);
+    return send(tls->sockfd, data, datalen, MSG_MORE);
```
david-cermak, line 126: I don't think we could hard-code this flag here. It makes sense for this specific use-case, but it's not universally the case for other usages. In fact, users usually prefer that the data are sent immediately, even for TCP, so we typically suggest that people set `TCP_NODELAY` to disable naggling.

I'd suggest adding this to `esp_tls` config options, something like:
```patch
@@ -200,6 +200,7 @@ typedef struct esp_tls_cfg {
     esp_tls_addr_family_t addr_family;      /*!< The address family to use when connecting to a host. */
     const int *ciphersuites_list;           /*!< Pointer to a zero-terminated array of IANA identifiers of TLS ciphersuites.
                                                 Please check the list validity by esp_tls_get_ciphersuites_list() API */
+    int socket_flags;
 } esp_tls_cfg_t;
```

components/esp_http_server/src/httpd_txrx.c
```diff
@@ -56,7 +56,7 @@ int httpd_send(httpd_req_t *r, const char *buf, size_t buf_len)
     }
 
     struct httpd_req_aux *ra = r->aux;
-    int ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, 0);
+    int ret = ra->sd->send_fn(ra->sd->handle, ra->sd->fd, buf, buf_len, MSG_MORE);
```
david-cermak, line 59: same here, suggest making this configurable

components/tcp_transport/include/esp_transport_ws.h
```diff
@@ -139,6 +144,37 @@ esp_err_t esp_transport_ws_set_config(esp_transport_handle_t t, const esp_transp
  */
 int esp_transport_ws_send_raw(esp_transport_handle_t t, ws_transport_opcodes_t opcode, const char *b, int len, int timeout_ms);
 
+/**
+ * @brief               Sends websocket raw message with custom opcode and payload in a optimized way
+ *
+ * This method is similar to esp_transport_ws_send_raw(), but
+ * it assumes, that the first MAX_WEBSOCKET_HEADER_SIZE bytes
```
david-cermak, line 151: Maybe add this note as a `@pre` (prerequisite for using this API) that it expects the buffer to hold some space for headers.

components/tcp_transport/include/esp_transport_ws.h
```diff
@@ -139,6 +144,37 @@ esp_err_t esp_transport_ws_set_config(esp_transport_handle_t t, const esp_transp
...
+ * Note that generic esp_transport_write for ws handle sends
+ * binary massages by default if size is > 0 and
+ * ping message if message size is set to 0.
+ * This API is provided to support explicit messages with arbitrary opcode,
+ * should it be PING, PONG or TEXT message with arbitrary data.
+ *
+ * @param[in]  t           Websocket transport handle
+ * @param[in]  opcode      ws operation code
+ * @param[in]  buffer      The buffer
+ * @param[in]  len         The length
```
david-cermak, line 169: It should be made clear that the `len` parameter refers to the entire buffer with headers, so the actual (payload) length is `len-MAX_WEBSOCKET_HEADER_SIZE`

components/tcp_transport/include/esp_transport_ws.h
```diff
@@ -139,6 +144,37 @@ esp_err_t esp_transport_ws_set_config(esp_transport_handle_t t, const esp_transp
...
+ * @param[in]  timeout_ms  The timeout milliseconds (-1 indicates block forever)
+ *
+ * @return
+ *  - Number of bytes was written
+ *  - (-1) if there are any errors, should check errno
+ */
+int esp_transport_ws_send_raw_optimized(esp_transport_handle_t t, ws_transport_opcodes_t opcode, const char *b, int len, int timeout_ms);
```
david-cermak, line 176: It's just a matter of taste, but I much don't like API names suffixed with `..._optimized()`. Could we think of a better name that would reflect the actual purpose? Maybe `..send_raw_once()`, `...send_raw_ext()`.

components/tcp_transport/transport_ws.c
```diff
@@ -366,6 +365,92 @@ static int _ws_write(esp_transport_handle_t t, int opcode, int mask_flag, const
...
+        ws_header[header_len++] = (uint8_t)((len2 >> 16) & 0xFF);
+        ws_header[header_len++] = (uint8_t)((len2 >> 8) & 0xFF);
+        ws_header[header_len++] = (uint8_t)((len2 >> 0) & 0xFF);
+    }
+
+    if (mask_flag) {
+        ws_header[header_len++] = 0;
+        ws_header[header_len++] = 0;
+        ws_header[header_len++] = 0;
+        ws_header[header_len++] = 0;
```
david-cermak, line 418: This part could be made a `static` function that would be called from both variants of `send_raw()` functions

david-cermak (commented):
@0xFEEDC0DE64 Nice work! The changes make a lot of sense to me and this PR is a good start, but still needs some polishing.

david-cermak:
>  I just want to start the discussion how to integrate this properly in the esp-idf

Please note that IDF uses this concept of "code owners" defined in 

https://github.com/espressif/esp-idf/blob/8fc8f3f47997aadba21facabc66004c1d22de181/.gitlab/CODEOWNERS#L143

and your PR touches two different areas with two owners. I'd suggest that you split your changeset into a smaller per-component PRs, with very few changes. This way it could be very easily reviewed by responsible code-owners from Espressif side.

You can use this PR for changes in `tcp_transport`, I've posted some comments to the code, but this LGTM overall and could be merged almost immediately after you address the comments.

david-cermak:
@0xFEEDC0DE64 Any news from your side? The ideas you shared in this PR look very good to me and I think would be useful to others, we just need to formally proceed with the changes.

Could you please check if you can split the change in two PRs?

0xFEEDC0DE64:
sorry didnt have enough time, and we didnt do a rebase for a long time, maybe i find time to do this finally in the next few days, we are just doing a rebase of everything right now

david-cermak:
@0xFEEDC0DE64 Will you have time to check the status of this PR and update your work per my comments?
If not, just let me know, please. I can take care of these little fixes on top of your commit and proceed with accepting it, as I still think the proposed changes make perfect sense and should be merged.

0xFEEDC0DE64:
Sorry, I cant do it at the moment, our esp-idf fork already has sooo many fixes and improvements ontop that I dont even know which commits I have to cherry pick to here to upstream them to you. And I dont have enough time at the moment to do this. Lets close this pull request

PR #12288, opened 2023-09-22
Test shellcheck in CI

mfialaf:


dobairoland:
@mfialaf I think this can be closed now, right?

PR #12369, opened 2023-10-08
allows provisioner node to use esp_ble_mesh_server_model_send_msg

vecais-dumais-laacis:
I was trying to alter ble_mesh_vendor_model example to be able to send ESP_BLE_MESH_VND_MODEL_OP_SEND from both devices, but it turned out provisioner node can't act as server. 

This patch allows ble mesh provisioner node to also act as ble mesh server otherwise esp_ble_mesh_server_model_send_msg fails with error.

### DIFF
--- a/components/bt/esp_ble_mesh/mesh_common/mesh_common.c
+++ b/components/bt/esp_ble_mesh/mesh_common/mesh_common.c
@@ -77,6 +77,10 @@ uint8_t bt_mesh_get_device_role(struct bt_mesh_model *model, bool srv_send)
     bt_mesh_client_user_data_t *client = NULL;
 
     if (srv_send) {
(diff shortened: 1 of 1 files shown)


v-d-l:
[examples.zip](https://github.com/espressif/esp-idf/files/12848234/examples.zip)
attached example
"provisioner" is based on vendor_client ( + added code from vendor_server )
"node" is based on vendor_server ( + added code from vendor_client )

vecais-dumais-laacis:
attached example code [examples.zip](https://github.com/espressif/esp-idf/files/12848250/examples.zip)
"provisioner" is based on vendor_client ( + added code from vendor_server )
"node" is based on vendor_server ( + added code from vendor_client )

PR #12370, opened 2023-10-08
ESP32 Add lcd sleep command (IDFGH-11203)

AndreaGreco:
Add code to set display in sleep mode.

### Abstract
Some driver (ST7789) support sleep mode that reduce a lots power consumption.
In my case with [2inc-display](https://www.waveshare.com/wiki/2inch_LCD_Module) Power consumption from 5mA to 0.3mA


suda-morris:
sha=71a91f104636add797b33609633b1eb885408323

suda-morris:
Thanks for the contribution. Merged.

PR #12375, opened 2023-10-09
fix: memalign: respect `malloc_alwaysinternal_limit` (IDFGH-11209)

codysch:
This changes `memalign` (and `posix_memalign`) so that it uses an allocation method with the same selection criteria (checking `malloc_alwaysinternal_limit` and picking one of:

 - always MALLOC_CAP_INTERNAL
 - MALLOC_CAP_INTERNAL first with fallback
 - MALLOC_CAP_SPIRAM first with fallback

`malloc_alwaysinternal_limit` is in turn set by the options `CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL` and
`CONFIG_SPRIAM_USE_CAPS_ALLOC`.

This notably affects folks using esp-rs to build rust code for the esp-idf, as all allocations from rust use `memalign`.

### DIFF
--- a/components/heap/heap_caps.c
+++ b/components/heap/heap_caps.c
@@ -655,30 +655,9 @@ size_t heap_caps_get_allocated_size( void *ptr )
     return size;
 }
 
-HEAP_IRAM_ATTR void *heap_caps_aligned_alloc(size_t alignment, size_t size, uint32_t caps)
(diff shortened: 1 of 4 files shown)


Alvin1Zhang:
Thanks for contribution again, changes merged with https://github.com/espressif/esp-idf/commit/9e1e2454770bd3014addf2fc5db141856feadc00.

PR #12424, opened 2023-10-18
fix(driver/gpio): Fix SIG_GPIO_OUT_IDX comment for signal connect

lucasssvaz:
This PR aims to fix the comment regarding `SIG_GPIO_OUT_IDX` in `esp_rom_gpio_connect_out_signal` by commenting on the `signal_idx` parameter and not on `gpio_num`. This avoids the misunderstanding that the GPIO number needs to be set to `SIG_GPIO_OUT_IDX` rather than the signal number.


suda-morris (approved):

suda-morris:
sha=64f8222ec6badc13bce0759bac1afd80efd91acb

PR #12436, opened 2023-10-20
feature(console): add command user context support (IDFGH-11280)

alonbl:
**Outline**

Current implementation implicitly forces the developer to use global variables
to enter its context during the command invocation, this change enables each
module to register a context for command to find without the need to manage
global variables.

No API breakage.

**Fields added**

```c
   esp_console_cmd_t::func_context    - (*)(int argc, char **argv, void *context)
```

**Functions added**

```c
   esp_err_t esp_console_cmd_set_context(const char *cmd, void *context)
```

**Usage**

```c
   esp_console_cmd_register(&cmd));
   esp_console_cmd_set_context(cmd.command, (void *)"context"));
```

### DIFF
--- a/components/console/esp_console.h
+++ b/components/console/esp_console.h
@@ -157,6 +157,15 @@ esp_err_t esp_console_deinit(void);
 typedef int (*esp_console_cmd_func_t)(int argc, char **argv);
 
 /**
+ * @brief Console command main function
+ * @param context a user context given at invocation.
+ * @param argc number of arguments
+ * @param argv array with argc entries, each pointing to a zero-terminated string argument
+ * @return console command return code, 0 indicates "success"
+ */
+typedef int (*esp_console_cmd_func_context_t)(void *context, int argc, char **argv);
+
+/**
  * @brief Console command description
  */
 typedef struct {
@@ -188,6 +197,10 @@ typedef struct {
      * Only used for the duration of esp_console_cmd_register call.
      */
     void *argtable;
+    /**
+     * Pointer to a context aware function which implements the command.
+     */
+    esp_console_cmd_func_context_t func_context;
 } esp_console_cmd_t;
 
 /**
@@ -199,6 +212,16 @@ typedef struct {
  *      - ESP_ERR_INVALID_ARG if command description includes invalid arguments
  */
 esp_err_t esp_console_cmd_register(const esp_console_cmd_t *cmd);
+
+/**
+ * @brief Register context for a command
+ * @param cmd pointer to the command name
+ * @return
+ *      - ESP_OK on success
+ *      - ESP_ERR_NOT_FOUND if command was not found
+ *      - ESP_ERR_INVALID_ARG if invalid arguments
+ */
+esp_err_t esp_console_cmd_set_context(const char *cmd, void *context);
 
 /**
  * @brief Run command line
--- a/components/console/test_apps/console/main/test_console.c
+++ b/components/console/test_apps/console/main/test_console.c
@@ -13,6 +13,18 @@
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
 
+typedef struct {
+        const char *in;
+        const char *out;
+} cmd_context_t;
+
+static int do_hello_cmd_with_context(void *context, int argc, char **argv)
+{
+    cmd_context_t *cmd_context = (cmd_context_t *)context;
+    cmd_context->out = cmd_context->in;
+    return 0;
+}
+
 static int do_hello_cmd(int argc, char **argv)
 {
     printf("Hello World\n");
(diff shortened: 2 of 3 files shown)


alonbl:
Dear Sirs,

I do not understand why I need to permit CLA github permissions on my behalf, why do you need this to sign CLA? For sure you do not need to access my teams, nor webhoos well, apart of my login and email you do not need anything, this is how others play the game.

```
CLA assistant by CLA assistant
would like additional permissions to
```

|What|Permission
|---|---
|Gists 								|Read and write access
|Organizations and teams				|Read-only access
|Commit statuses						|Read and write access
|Repository webhooks and services	        |Read and write access

Alon

components/console/esp_console.h
```diff
@@ -156,6 +157,15 @@ esp_err_t esp_console_deinit(void);
...
 typedef int (*esp_console_cmd_func_t)(int argc, char **argv);
 
+/**
+ * @brief Console command main function
+ * @param argc number of arguments
+ * @param argv array with argc entries, each pointing to a zero-terminated string argument
+ * @param context a user context given at invocation.
+ * @return console command return code, 0 indicates "success"
+ */
+typedef int (*esp_console_cmd_func_context_t)(int argc, char **argv, void *context);
```
alonbl, line 167: BTW: I prefer the context to be placed as 1st parameter, please tell me what you prefer.

JimmyPedersen: yes please, context always first

0xjakob:
@alonbl We're taking a look at this now and will come back here soon.

0xjakob:
@alonbl We discussed the PR and in general, it looks good! We have one concern, however: The current version would add a global context value which is passed to any command callback registered with `esp_console_cmd_t::func_context `. What is the difference to having a global context variable anywhere in the user code? What exact problem are you trying to solve?

alonbl:
> @alonbl We discussed the PR and in general, it looks good! We have one concern, however: The current version would add a global context value which is passed to any command callback registered with `esp_console_cmd_t::func_context `. What is the difference to having a global context variable anywhere in the user code? What exact problem are you trying to solve?

Hi,

Global variables are evil, the libraries should be designed to minimize their use, this helps to keep track of resources. Look at the http_server implementation as a reference which is done nicely. Having application context managed in several places, aka passed as parameters to functions (at is should) and have it also copied in global variable is bad practice as the two (or more) require synchronization.

The next stage of this patch would be the ability to register a context per command using `esp_console_cmd_register()`, and the next one is to have a context per console instance.

The end use case is the following: In application each module registers its own console commands as decoupled from one from another, a command of each module should be invoked in the correct context.

As you well know, the console module is very important to embedded system and is missing a lot of love, for example:

1. having a console instance (vs global) as best practice to manage multiple ports (#10526)
2. to be able to provide different console behavior to different devices (aka console context, just like http_server)
3. unregister command to allow resource cleanups of modules.
4. ability to stop the console task nicely from outside of the console (#9974)
5. ability to invoke a command within specific context (decouple).

What do you think? I can workout a modified patch that will allow a context per command (step#2), will it be better?

Thanks,

0xjakob:
Yes, one context per command sounds more flexible to us.

If you have any question, feel free to come back to us, we'll provide help where possible. Thanks a lot already!

igrr:
Don't the two contexts ("per command" and "per esp_console_run invocation") solve different kinds of problems?
- Context per command: potentially allows having a single command implementation handle multiple commands. For example, you can have a generic implementation (`cmd_write`) of a set of commands (`write_str`, `write_int`, `write_blob`), and register multiple commands all using the same handler function. The handler will distinguish which command to run based on the context provided during registration.
- Context per esp_console_run call: allows the command to access variables specific to the console instance. For example, when running two instances of console in different tasks, the command would have access to the state related to the current instance.

So one solution doesn't seem to be "better" or "worse" than the other, they seem to solve different problems.

> The end use case is the following: In application each module registers its own console commands as decoupled from one from another, a command of each module should be invoked in the correct context.

Could you please clarify what specific issue (other than global variables being evil) did you run into trying to implement this? For example, consider multiple "modules" providing console commands:
- [nvs](https://github.com/espressif/esp-idf/tree/master/examples/system/console/advanced/components/cmd_nvs)
- [wifi](https://github.com/espressif/esp-idf/tree/master/examples/system/console/advanced/components/cmd_wifi)
- [system](https://github.com/espressif/esp-idf/tree/master/examples/system/console/advanced/components/cmd_system)

Each module is decoupled from the other module, and each has its own set of commands. There is shared state (static variables) within each module, but the modules don't have shared state between themselves. I guess your use case is somewhat different, could you please explain in more detail?

(I am not debating whether having user-supplied context in the API is good or not; I would just like to understand the issue you are running to better.)

alonbl:
Hello @igrr  and @0xjakob,

Thank you for your feedback, I pushed an alternate version in which the context is per command.
Can you please review this one? it should answer most of your concerns and be usable for OOP like command model.

Implementation note: I did not want to add the context to `esp_console_cmd_t` because it usually lives in static context.

Regards,
Alon

components/console/test_apps/console/main/test_console.c
```diff
@@ -102,3 +107,52 @@ TEST_CASE("esp console init/deinit test, minimal config", "[console]")
...
+        TEST_ESP_OK(esp_console_cmd_register(&cmds[i]));
+    }
+    TEST_ESP_OK(esp_console_cmd_set_context(cmds[2].command, cmds[2].command));
+    TEST_ESP_OK(esp_console_cmd_set_context(cmds[3].command, cmds[3].command));
+    TEST_ASSERT_EQUAL(esp_console_cmd_set_context(NULL, NULL), ESP_ERR_INVALID_ARG);
+    TEST_ASSERT_EQUAL(esp_console_cmd_set_context("invalid", NULL), ESP_ERR_NOT_FOUND);
+    TEST_ESP_OK(esp_console_cmd_register(&s_quit_cmd));
+    TEST_ESP_OK(esp_console_register_help_command());
+
+    TEST_ESP_OK(esp_console_start_repl(s_repl));
```
igrr, line 156: I think for unit test purposes, you don't need to start the whole REPL. You can call `esp_console_run` and verify that the expected callback got called with the expected context. (Either by setting some variable from the callback, or by temporarily substituting `stdout` using `open_memstream` and then asserting that the expected string was printed by the callback.)

alonbl: Done, I think.

0xjakob (approved):
@alonbl The current approach looks good. Thanks for adding the unit test! We'll handle this and after going through internal CI and reviewing, it will appear on master if there are no concerns. We'll keep you updated.

0xjakob:
sha=867f9d0dbbe130f4ea46bdf87a0b28d1cbcd3c91

0xjakob:
@alonbl We've merged the code to our internal code base, from where it will be synced to the master branch soon. The code looks slightly different now as we adjusted the naming a bit, but your authorship has been retained, of course.

0xjakob:
@alonbl The change has been merged to master. However, we'll change the API once more so that the context can be set with `esp_console_cmd_set_context()` instead of having a separate setting function. The new version should open less room for errors.

alonbl:
> @alonbl The change has been merged to master. However, we'll change the API once more so that the context can be set with `esp_console_cmd_set_context()` instead of having a separate setting function. The new version should open less room for errors.

Hi @0xjakob, 
Thank you for your great work!
I do not understand the `esp_console_cmd_set_context()` is now used to set the context what is the change?
Regards,

0xjakob:
@alonbl Sorry, it's a typo. I edited my post. It means that `esp_console_cmd_register()` now sets the context, based on the arugments in `const esp_console_cmd_t *cmd`, given to that function.

0xjakob:
All the relevant code is merged. Hence, we're closing this PR (to mark it as finished).

PR #12449, opened 2023-10-23
Use correct clang flag for size optimization (IDFGH-11295)

storoj:
See https://clang.llvm.org/docs/CommandGuide/clang.html:
``` 
Code Generation Options
   -Oz Like -Os (and thus -O2), but reduces code size further.
```
Without -Oz enabled clang produced binaries that were too large.


ESP-Marius:
After looking into this, changing to `-Oz` does indeed seem reasonable. We;ll discuss it further internally. 

Thanks for your contribution!

ESP-Marius:
sha=4acefd9df00aff1d926937840d327c890df1803a

Issue #9277, opened 2022-07-03
LVGL Font corruption on SSD1306 display


VikOlliver:
## Environment

- Development Kit:      [none]
- Module or chip used:  [ESP32-WROOM-32]
- IDF version (run ``git describe --tags`` to find it): v5.0-dev-3654-gc2ccc383da
- Build System:         [Make|CMake|idf.py]
- Compiler version (run ``xtensa-esp32-elf-gcc --version`` to find it): 1.22.0-80-g6c4433a
- Operating System:     [Linux]
- Using an IDE?: [No]
- Power Supply:         [USB]

## Problem Description
Default UNSCII fonts are displayed half height when rendered using the stock example code on a 128x32 SSD1306 monochrome OLED display.

Problem is exhibited by other monospace fonts, but the stock examples are probably better for illustrating the bug.

### Expected Behavior
UNSCII 8 font displayed 8 pixels high, UNSCII 16 font displayed 16 pixels high

### Actual Behavior
UNSCII 8 font displayed 4 pixels high and completely illegible.
UNSCII 16 font displayed 8 pixels high, 16 pixels wide.

### Steps to reproduce

1. Set appropriate LVGL UNSCII font in idf.py menuconfig
2. build/flash esp-idf/examples/peripherals/lcd/i2c_oled

// If possible, attach a picture of your setup/wiring here.
Unable to attach. Github reports "something has gone seriously wrong" when attempting to attach a 294KiB JPEG

### Code to reproduce this issue
Stock example.

## Debug Logs
No errors logged.
I (388) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (398) example: Initialize I2C bus
I (408) example: Install panel IO
I (408) example: Install SSD1306 panel driver
I (508) example: Initialize LVGL library
I (508) example: Register display driver to LVGL
I (508) example: Install LVGL tick timer
I (508) example: Display LVGL Scroll Text

VikOlliver:
![font_bug](https://user-images.githubusercontent.com/4644768/177062404-12ddd335-da43-414b-8bea-9b622ba32de6.png)

OK, managed to upload images of the corrupt and squashed font on the OLED display by using Firefox in debug mode :)

VikOlliver:
[...]

PR #12450, opened 2023-10-23
Support SSD1306 128x32 (IDFGH-11296)

storoj:
SSD1306 may have different sizes and resolutions: 128x32, 128x64, 96x16
and others.

`SSD1306_CMD_SET_MULTIPLEX` (0xA8) initialization command effectively
sets the height of the display's resolution. The value for this command
is the height in pixels minus one.

```
Set MUX ratio to N+1 MUX
N=A[5:0] : from 16MUX to 64MUX, RESET= 111111b (i.e. 63d, 64MUX)
A[5:0] from 0 to 14 are invalid entry.
```

`SSD1306_CMD_SET_COMPINS` (0xDA) configures the communication pins.
According to the datasheet it should be `0x2` with other optional bits
set.

```
A[4]=0b, Sequential COM pin configuration
A[4]=1b(RESET), Alternative COM pin configuration

A[5]=0b(RESET), Disable COM Left/Right remap
A[5]=1b, Enable COM Left/Right remap
```

* https://github.com/adafruit/Adafruit_SSD1306;
* https://www.digikey.com/htmldatasheets/production/2047793/0/0/1/ssd1306.html

### DIFF
--- a/components/esp_lcd/include/esp_lcd_panel_vendor.h
+++ b/components/esp_lcd/include/esp_lcd_panel_vendor.h
@@ -58,6 +58,33 @@ esp_err_t esp_lcd_new_panel_st7789(const esp_lcd_panel_io_handle_t io, const esp
 esp_err_t esp_lcd_new_panel_nt35510(const esp_lcd_panel_io_handle_t io, const esp_lcd_panel_dev_config_t *panel_dev_config, esp_lcd_panel_handle_t *ret_panel);
 
 /**
+ * @brief SSD1306 configuration structure
+ *
+ * To be used as esp_lcd_panel_dev_config_t.vendor_config.
+ * See esp_lcd_new_panel_ssd1306().
+ */
+typedef struct {
+   /**
+    * @brief Multiplex ratio: (0..63)
+    *
+    * Display's height minus one.
+    */
+   uint8_t mux_ratio;
+
+   /**
+    * @brief Enables alternative COM pin configuration.
+    *
+    * When unset then Sequential COM pin configuration is used.
+    */
+   bool com_pin_alt;
+
+   /**
+    * @brief COM Left/Right remap.
+    */
+   bool com_lr_remap;
+} esp_lcd_panel_ssd1306_config_t;
+
+/**
  * @brief Create LCD panel for model SSD1306
  *
  * @param[in] io LCD panel IO handle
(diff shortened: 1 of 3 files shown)


storoj:
This change fixes the issue https://github.com/espressif/esp-idf/issues/9277.

storoj:
Sorry, there's one more initialization command that's missing. I will update my PR shortly.

components/esp_lcd/include/esp_lcd_panel_vendor.h
```diff
@@ -67,6 +94,24 @@ esp_err_t esp_lcd_new_panel_nt35510(const esp_lcd_panel_io_handle_t io, const es
...
  *          - ESP_OK                on success
+ *
+ * @note The default panel size is 128x64.
+ * @note Use esp_lcd_panel_ssd1306_config_t to set the correct size.
+ * Example usage:
+ * @code {c}
+ * 
+ * esp_lcd_panel_ssd1306_config_t ssd1306_config = {
+ *     .width = 128,
+ *     .height = 32
```
suda-morris, line 105: seems like this example usage is out of date.

suda-morris: But compared to the 

```c
typedef struct {
   /**
    * @brief Multiplex ratio: (0..63)
    *
    * Display's height minus one.
    */
   uint8_t mux_ratio;

   /**
    * @brief Enables alternative COM pin configuration.
    *
    * When unset then Sequential COM pin configuration is used.
    */
   bool com_pin_alt;

   /**
    * @brief COM Left/Right remap.
    */
   bool com_lr_remap;
} esp_lcd_panel_ssd1306_config_t;
```

This example version looks more intuitive.

storoj: @suda-morris thanks for your comments, good catch on mismatched documentation comment. I had several iterations on the driver and forgot to update the docs.

First I wanted to provide simpler interface with just `width` and `height` to make it easier to use. However later I reconsidered this approach because:
* width seems to never be needed;
* I'm not sure that height is enough to reliably configure a display;
* ... which may be even more confusing to developers when they first try the driver with the default settings (128x64), it fails, then they find this config, set the correct values and it fails again because for example it's a display from another manufacturer. In that case it will be quite hard to override lcd commands sent to display.

I decided to switch to more "generic" (but less intuitive) API to provide developers more flexibility while still being high-level enough to not send raw lcd commands. Maybe I should have mentioned typical config parameters for the most popular display sizes (128x64, 128x32, 96x16 and maybe others) in the documentation header.

Or it could be a good idea to provide public constants like 
```
extern const esp_lcd_panel_ssd1306_config_t esp_lcd_panel_ssd1306_128x32 = {
  .mux_ratio = 0x2,
  .com_pin_alt = false,
  .com_lr_remap = false
}
```
... but I'm not quite sure about naming and overall coding style, that's my very first touch of the embedded area.

@suda-morris what do you think I should prefer:
* mention typical configs for different screens in the documentation comment;
* provide constants with configs for known display types;

I lean towards the constants path. If you agree could you please guide me on the coding style and naming I should use? I struggle with that so far.

suda-morris: Hi @storoj I would prefer the first approach. As you also said, different manufacturers may require different settings. We can't provide the setting for all known display types as well, because that's the job of a BSP package. Ideally, the user will define his own BSP package like [these](https://components.espressif.com/components?q=Board+Support+Package), and in there, he can predefine the value of esp_lcd_panel_ssd1306_config_t according to his manufacture.

espzav:
@storoj Please install and use pre-comit hook for pass tests: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html

suda-morris:
sha=8f75c4c5249c3a85d1e6e89cd7bd744183155d35

suda-morris:
merged in https://github.com/espressif/esp-idf/commit/c4dc16c612a920f206ba1f3a1c6ec24441aaa199

nicklasb:
For those that don't want to use another library just to get their 32px stuff going until this is this code is actually released and included everywhere... I just did this workaround that worked fine to me (stole the code from the esp-bsp SSD1306 library if it looks familiar): https://github.com/RobustoFramework/Robusto/blob/main/components/robusto/ui/src/robusto_32px.c

I call it here, just after the initialization to overwrite some of the settings the normal flow sends:
```
    ESP_ERROR_CHECK(esp_lcd_panel_init(panel_handle));
    ssd1306_init(CONFIG_ROBUSTO_UI_I2C_PORT, CONFIG_ROBUSTO_UI_I2C_HW_ADDR);
    ESP_ERROR_CHECK(esp_lcd_panel_disp_on_off(panel_handle, true));
```

It is not very invasive and easy to remove when done.
It doesn't restrict the screen size or anything like that, so you'd have to keep within the top 32 px, obviously.

PR #12462, opened 2023-10-25
reusing the OTA update facility to update the SPIFFS partition (IDFGH-11309)

kohait00:
In an attempt to reuse the OTA functionality which is currently hard wired to only support ESP_PARTITION_TYPE_APP partitions, 

this fix loosens the grip on that and enables the direct specification of the destination update_partition  via config.

the respective plausibility checks are only performed on APP type partitions now.

benefit: encouraging users to develop OTA updates of other parts of the system without bloating the code each time


mahavirj:
Few questions:

- How would the selection between active/passive custom partition happen? Or that logic is completely application (higher layer) specific?
- Is it possible to share your application code which utilizes this PR to handle the custom partition OTA updates?

kohait00:
I am currently using this to update my spiffs image..
the code is not ripe yet, but I can provide the basic building blocks to see what how it is supposed to work

updating the spiffs would be the most probable usecase i suppose, but i cant find a a way to check the first chunk received for some kind of header info, consistancy or the like.

OTA app update bins contain a version header, SPIFFS images dont, but they have some kind of MAGIC numbers scattered across the fsif I could see at least one of them... 
Currently looking for this. If you have any suggestions... I'd appreciate it.

kohait00:
> Few questions:
> 
>     * How would the selection between active/passive custom partition happen? Or that logic is completely application (higher layer) specific?
> 
>     * Is it possible to share your application code which utilizes this PR to handle the custom partition OTA updates?

The steps are baaicaly these:

`const esp_partition_t* esp_ota_get_data_partition(esp_partition_subtype_t subtype, const char *label)
{
    const esp_partition_t * p = esp_partition_find_first(ESP_PARTITION_TYPE_DATA, subtype, label);
    return esp_partition_verify(p);
}
`

The OTA Config (advanced example) would look like

`esp_https_ota_config_t ota_config = {
...
		.update_partition = esp_ota_get_data_partition(ESP_PARTITION_SUBTYPE_DATA_SPIFFS, WWW_PARTITION),
    };`


now the OTA Task just does as usual, with some user code, to check if the "to be received OTA" is valid (this is my current working site)

`
...
    esp_app_desc_t app_desc;
    err = esp_https_ota_get_img_desc(https_ota_handle, &app_desc);
    if (err != ESP_OK) {
        ESP_LOGE(TAG, "esp_https_ota_read_img_desc failed");
        goto ota_end;
    }
`

`
    err = validate_fs_image_header(&app_desc);
    if (err != ESP_OK) {
        ESP_LOGE(TAG, "image header fs verification failed with 0x%x", err);
        goto ota_end;
    }
...
`

kohait00:
just comitted 2 more changes so esp_https_ota_get_img_desc and esp_https_ota_perform dont complain about non ESP_PARTITION_TYPE_APP partition type..

I could not figure out how to check the first bytes of spiffs image for plausibility...
If anyone has an idea, please..
the spiffs code is not really helpfull in this regard, the check functions available there are to check the overall file system

hyansuper:
This is great feature for data/file_system partition OTA !
I think a more structured way to do it is move `components/app_update/esp_ota_ops.c` to another file maybe called "partition_update.c" and remove partiton_type and magic_word checking, then include this file into `esp_ota_ops.c` and add more strict checking

kohait00:
I was able to use this feature to even update bootloader and, just for fun, the partition table (yes just for the fun of it). but this needs some registering of subtypes.

I'd welcome a more generic approach of course. Judging by the now more concrete possibilities of littlefs on the horizon, we will have even more partitions to update.

I currently have a generic approach of updating the partitions with a rest api, without the need of a cloud which is small enough and using hardcoded HTML / scripts.. Nice for maintaining control of the device without internet connection.

hyansuper:
@kohait00 for robustness, you may also want to store a var in nvs to record the state of data partition, like [app partition state](https://docs.espressif.com/projects/esp-idf/zh_CN/stable/esp32/api-reference/system/ota.html#id4)

hyansuper:
Just to show another possibility, I can store data in the non-active OTA partition.
When update is performed, I first update app code, reboot to the other partition, then update data partition

hmalpani:
Hello @kohait00 
Thanks for the PR.
However I think it will be good to have a new component to support OTA for non-app partitions. We have started working on it. I think we should use `app_update` component only to update the app partitions as the name suggests. I will soon create a PR on https://github.com/espressif/idf-extra-components which will add support for OTA for non app partitions.

hmalpani:
Hello @kohait00 
I have created a PR for this feature. You can try it from here: https://github.com/espressif/idf-extra-components/pull/328

hmalpani:
I am closing this MR. This feature is being developed at https://github.com/espressif/idf-extra-components/pull/328

PR #12468, opened 2023-10-26
Factory Reset: Bootloader with a little Wait Time to enable Usage of GPIO 0 (BOOT) button  (IDFGH-11320)

kohait00:
Using GPIO 0 (BOOT) button to perform Factory Reset is pretty handy, as this GPIO comes already as a button on most ESP32 based Dev Kits of any sort. No need to solder / connect an extra button for that purpose..

The only problem: 

GPIO 0 (BOOT) has a double use already, if held down during reset, the ROM Loader uses it to determine UART BOOT condition, so the esp bootloader wont even start. On the other hand, the bootloader boots so quickly that it passes the         

..
if (bootloader_common_check_long_hold_gpio_level(CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET, CONFIG_BOOTLOADER_HOLD_TIME_GPIO, reset_level) == GPIO_LONG_HOLD) 
..
and wont ever see a GPIO 0 (BOOT) pressed and held "right after reset"

Solution:

introducing a Kconfig configurable wait time in the bootloader, right before this very if condition.  It enables to comfortly reset the ESP and apply the factory reset condition just after it (pressing and holding as usual)..

typical USE CASE:

+               * set GPIO 0 (BOOT) as Factory Reset, which is already a pre-installed button on almost every ESP32 board
+               * for the GPIO 0 (BOOT): set 'Reset on GPIO Low'
+               * enable the Waiting Time after EN
+               * press RESET (EN) --> ROM Loader determines normal flash boot
+               * shortly after, press and hold GPIO 0 (BOOT) button until the Factory reset condition is recognized
+               * boot continues, Factory reset is applied, if no factory image present: Reset is performed and a UART BOOT is recognized (BOOT still asserted) --> release BOOT, perform reset

Regarding the change:

* it is a configurable busy wait (what else is there to do in mean time?) prior to checking the factory reset condition
* the code adds around 0x40 byte, so there is like 0x10 byte left :) I put 2 LOGI outputs together to make some space. the LOGs really add to the bin size.

### DIFF
--- a/components/bootloader/Kconfig.projbuild
+++ b/components/bootloader/Kconfig.projbuild
@@ -172,7 +172,7 @@ menu "Bootloader config"
     config BOOTLOADER_WAIT_AFTER_EN_TIME_MS
         int "Time to wait before the Factory Reset GPIO probe (ms)"
         depends on BOOTLOADER_WAIT_AFTER_EN
-        default 3000
+        default 1000
         range 0 10000
         help
             The specified waiting time after booting the bootloader and prior to probing the Factory Reset GPIO
--- a/components/bootloader/Kconfig.projbuild
+++ b/components/bootloader/Kconfig.projbuild
@@ -151,6 +151,32 @@ menu "Bootloader config"
             bool "Reset on GPIO high"
     endchoice
 
+    config BOOTLOADER_WAIT_AFTER_EN
+        bool "Enable Waiting time before Factory Reset GPIO probe (especially usefull for GPIO 0 as factory reset)"
+        default y
+        depends on BOOTLOADER_FACTORY_RESET
+        help
+            Enable Wait Time in bootloader prior to probing for Factory Reset condition.
+            This is especially usefull if GPIO 0 (BOOT button on most Dev Boards) is set as Factory Reset GPIO.
+            Without the wait time, the GPIO 0 / BOOT, if held down, during boot, is already probed by the ROM loader
+            which would be selecting UART boot mode.
+
+            The intended use case is as follows:
+               * set GPIO 0 (BOOT) as Factory Reset, which is already a pre-installed button on almost every ESP32 board
+               * for the GPIO 0 (BOOT): set 'Reset on GPIO Low'
+               * enable the Waiting Time after EN
+               * press RESET (EN) --> ROM Loader determines normal flash boot
+               * shortly after, press and hold GPIO 0 (BOOT) button until the Factory reset condition is recognized
+               * boot continues, Factory reset is applied
+
+    config BOOTLOADER_WAIT_AFTER_EN_TIME_MS
+        int "Time to wait before the Factory Reset GPIO probe (ms)"
+        depends on BOOTLOADER_WAIT_AFTER_EN
+        default 3000
+        range 0 10000
+        help
+            The specified waiting time after booting the bootloader and prior to probing the Factory Reset GPIO
+
     config BOOTLOADER_OTA_DATA_ERASE
         bool "Clear OTA data on factory reset (select factory partition)"
         depends on BOOTLOADER_FACTORY_RESET
--- a/components/bootloader/subproject/main/bootloader_start.c
+++ b/components/bootloader/subproject/main/bootloader_start.c
@@ -96,14 +96,13 @@ static int selected_boot_partition(const bootloader_state_t *bs)
 #endif
 
 #ifdef CONFIG_BOOTLOADER_WAIT_AFTER_EN
-        ESP_LOGW(TAG, "Use GPIO %d for more than %d s to perform factory reset!",
+        ESP_LOGW(TAG, "Use GPIO %d for more than %d s to perform Factory Feset!",
             (int)CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET,
             (int)(CONFIG_BOOTLOADER_HOLD_TIME_GPIO + BL_HOLDOFF_TIME_MS/1000)
         );
 
         //give the user a chance to react after a power up to hold BOOT button
-        uint32_t tm_start = esp_log_early_timestamp();
-        do { } while (BL_HOLDOFF_TIME_MS > ((esp_log_early_timestamp() - tm_start) ));
+        esp_rom_delay_us(BL_HOLDOFF_TIME_MS * 1000u);
 #endif
 
         if (bootloader_common_check_long_hold_gpio_level(CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET, CONFIG_BOOTLOADER_HOLD_TIME_GPIO, reset_level) == GPIO_LONG_HOLD) {
--- a/components/bootloader/subproject/main/bootloader_start.c
+++ b/components/bootloader/subproject/main/bootloader_start.c
@@ -77,21 +77,6 @@ static int select_partition_number(bootloader_state_t *bs)
     return selected_boot_partition(bs);
 }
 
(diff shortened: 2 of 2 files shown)


components/bootloader/Kconfig.projbuild
```diff
@@ -151,6 +151,32 @@ menu "Bootloader config"
             bool "Reset on GPIO high"
     endchoice
 
+    config BOOTLOADER_WAIT_AFTER_EN
+        bool "Enable Waiting time before Factory Reset GPIO probe (especially usefull for GPIO 0 as factory reset)"
```
o-marshmallow, line 155:
```suggestion
        bool "Enable Waiting time before Factory Reset GPIO probe (especially useful for GPIO 0 as factory reset)"
```
Small typo

components/bootloader/Kconfig.projbuild
```diff
@@ -151,6 +151,32 @@ menu "Bootloader config"
...
+        default y
+        depends on BOOTLOADER_FACTORY_RESET
+        help
+            Enable Wait Time in bootloader prior to probing for Factory Reset condition.
+            This is especially usefull if GPIO 0 (BOOT button on most Dev Boards) is set as Factory Reset GPIO.
```
o-marshmallow, line 160:
```suggestion
            This is especially useful if GPIO 0 (BOOT button on most Dev Boards) is set as Factory Reset GPIO.
```

components/bootloader/Kconfig.projbuild
```diff
@@ -151,6 +151,32 @@ menu "Bootloader config"
...
+               * for the GPIO 0 (BOOT): set 'Reset on GPIO Low'
+               * enable the Waiting Time after EN
+               * press RESET (EN) --> ROM Loader determines normal flash boot
+               * shortly after, press and hold GPIO 0 (BOOT) button until the Factory reset condition is recognized
+               * boot continues, Factory reset is applied
+
+    config BOOTLOADER_WAIT_AFTER_EN_TIME_MS
+        int "Time to wait before the Factory Reset GPIO probe (ms)"
+        depends on BOOTLOADER_WAIT_AFTER_EN
+        default 3000
```
o-marshmallow, line 175: 3 seconds sounds too much in my opinion as this would slow down the boot time by a lot

components/bootloader/subproject/main/bootloader_start.c
```diff
@@ -87,14 +94,25 @@ static int selected_boot_partition(const bootloader_state_t *bs)
 #if CONFIG_BOOTLOADER_FACTORY_RESET_PIN_HIGH
         reset_level = true;
 #endif
+
+#ifdef CONFIG_BOOTLOADER_WAIT_AFTER_EN
+        ESP_LOGW(TAG, "Use GPIO %d for more than %d s to perform factory reset!",
```
o-marshmallow, line 99: Does this message really need to be a warning? I think it would make more sense to have it as an `INFO`.

components/bootloader/subproject/main/bootloader_start.c
```diff
@@ -87,14 +94,25 @@ static int selected_boot_partition(const bootloader_state_t *bs)
...
+            (int)CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET,
+            (int)(CONFIG_BOOTLOADER_HOLD_TIME_GPIO + BL_HOLDOFF_TIME_MS/1000)
+        );
+
+        //give the user a chance to react after a power up to hold BOOT button
+        uint32_t tm_start = esp_log_early_timestamp();
+        do { } while (BL_HOLDOFF_TIME_MS > ((esp_log_early_timestamp() - tm_start) ));
```
o-marshmallow, line 106: You should be able to use `esp_rom_delay_us(CONFIG_BOOTLOADER_WAIT_AFTER_EN_TIME_MS * 1000);` instead

KonstantinKondrashov:
Hi @kohait00!
FYI, there is an additional way to achieve the same goal, defining a custom `bootloader_after_init()` function:
```c
void bootloader_after_init(void) { 
    if (esp_rom_get_reset_reason(0) != RESET_REASON_CORE_DEEP_SLEEP) {
        esp_rom_delay_us(2000);
} 
```
See an example here - [examples/custom_bootloader/bootloader_hooks](https://github.com/espressif/esp-idf/blob/master/examples/custom_bootloader/bootloader_hooks/bootloader_components/my_boot_hooks/hooks.c#L17-L19)

components/bootloader/subproject/main/bootloader_start.c
kohait00: judging by the fact the there is a "otherwise unnecessary" waiting time introduced, that substantially slows down the boot process (see review comment above about the 3 sec), I'd say that it should always be present, to remind to think about probably removing / resolving it..
But I get the reasoning behind it, Warnings make worry..
I dont have a problem lowering it to Info.. Just thought it shouldnt be "invisible", which it would be once you set the overall Log level to Warning..

components/bootloader/subproject/main/bootloader_start.c
kohait00: niiice, will try it out, maybe it will free some more bytes here..

kohait00:
> Hi @kohait00! FYI, there is an additional way to achieve the same goal, defining a custom `bootloader_after_init()` function:
> 
> ```c
> void bootloader_after_init(void) { 
>     if (esp_rom_get_reset_reason(0) != RESET_REASON_CORE_DEEP_SLEEP) {
>         esp_rom_delay_us(2000);
> } 
> ```
> 
> See an example here - [examples/custom_bootloader/bootloader_hooks](https://github.com/espressif/esp-idf/blob/master/examples/custom_bootloader/bootloader_hooks/bootloader_components/my_boot_hooks/hooks.c#L17-L19)

I admit it is intriguing to have it as a hook. I users might find it generally usefull. hence the merge request.
in case it is denied, I would go the hook way, obviosly. thanks for the pointer.

components/bootloader/Kconfig.projbuild
kohait00: 1 s would be enough, you are right, it should be enough to release reset and pres
ss hold the boot button.. feel free to suggest a different value

kohait00:
> Hi @kohait00! FYI, there is an additional way to achieve the same goal, defining a custom `bootloader_after_init()` function:
> 
> ```c
> void bootloader_after_init(void) { 
>     if (esp_rom_get_reset_reason(0) != RESET_REASON_CORE_DEEP_SLEEP) {
>         esp_rom_delay_us(2000);
> } 
> ```
> 
> See an example here - [examples/custom_bootloader/bootloader_hooks](https://github.com/espressif/esp-idf/blob/master/examples/custom_bootloader/bootloader_hooks/bootloader_components/my_boot_hooks/hooks.c#L17-L19)

tested, worked.. comitted as new version

RFC

components/bootloader/Kconfig.projbuild
kohait00: changed to 1 s, see current files state in merge request

components/bootloader/subproject/main/bootloader_start.c
kohait00: it worked, but no size improvements

KonstantinKondrashov:
Hi @kohait00!
Thank you for implementing this feature.
Finally, we decided not to merge it, once it is possible to reach it by using [examples/custom_bootloader/bootloader_hooks](https://github.com/espressif/esp-idf/blob/master/examples/custom_bootloader/bootloader_hooks/bootloader_components/my_boot_hooks/hooks.c#L17-L19).

Issue #12480, opened 2023-10-30
Many places in the ESP_SYSTEM are using CONFIG_FREERTOS_UNICORE instead of CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE


FL0WL0W:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Components of the ESP system that are unrelated to FREERTOS should use the appropriate build definitions. Since these are part of the ESP system they should be using CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE instead of CONFIG_FREERTOS_UNICORE. 

Since CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE is directly set by CONFIG_FREERTOS_UNICORE, there will functionally be no difference to how it is now. So there is no risk in making these updates

I am working on making Asymetric Multicore Processing (AMP) work using FREERTOS on one core and bare metal application on the other. This is the first step towards making that possible. This change alone will not make this possible. I will create another issue and PR for making that possible as there are more risky changes the need to be made for that.

FL0WL0W:
fix pulled in

Issue #10410, opened 2022-12-20
AMP(Asymmetric Multi-Processing) Support (IDFGH-8999)


5ami:
### Is your feature request related to a problem?

 Yes. It could be a problem depending on the application, as there is no official support for BareMetal programming in esp-idf.

### Describe the solution you'd like.

For now, We would like to Have AMP (i.e., running FreeRTOS on one core and a bare metal/other OS on the other).

### Describe alternatives you've considered.

I considered this online article, https://hackaday.io/project/174521-bare-metal-second-core-on-esp32, but the implementation is highly experimental and not reliable especially in esp32s3.

### Additional context.

![image](https://user-images.githubusercontent.com/43592883/208635873-b164ec5a-83e2-471c-9ede-938e5fb2b171.png)

igrr:
Hi @5ami,
The AMP solution for Espressif chips will be offered based on Zephyr project. You can check [this example](https://docs.zephyrproject.org/latest/samples/drivers/ipm/ipm_esp32/README.html) watch the [DevCon video](https://www.youtube.com/watch?v=oble9ObAqxM) for more background.
For bare metal development for Espressif chips you can consider https://github.com/esp-rs/esp-hal.

At this time, we are not planning to add AMP support to ESP-IDF.

5ami:
Hi Igrr, thank you for the reply. 
**Is there a way around  prevent freeRTOS during startup from starting on the app_cpu ?  
---------------------------------------------------------------------------------------------------------------------------------------
**for  https://github.com/esp-rs/esp-hal, I tried to consider Rust, but As far as I know,  for safety-critical markets, such as automotive, there is currently no safety-certified Rust toolchain.

igrr:
> Is there a way around prevent freeRTOS during startup from starting on the app_cpu ?

There isn't a straightforward way of doing this while still using the 2nd CPU to run some code. ESP-IDF is written under assumption that either the 2nd CPU is also running FreeRTOS, or the 2nd CPU is not used at all. This is especially a problem for ESP32 where various silicon bug workarounds require cooperation between the code running on both CPUs. Even for ESP32-S3, several aspects of the code (watchdogs, panic handling, cache & flash operations) assume that when CONFIG_FREERTOS_UNICORE is set (i.e. FreeRTOS runs on 1st CPU only) the 2nd CPU is not running any code.

[...]

PR #12481, opened 2023-10-30
Many places in the ESP_SYSTEM are using CONFIG_FREERTOS_UNICORE instead of CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE (IDFGH-11333)

FL0WL0W:
Components of the ESP system that are unrelated to FREERTOS should use the appropriate build definitions. Since these are part of the ESP system they should be using CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE instead of CONFIG_FREERTOS_UNICORE.

Since CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE is directly set by CONFIG_FREERTOS_UNICORE, there will functionally be no difference to how it is now. So there is no risk in making these updates

I am working on making Asymetric Multicore Processing (AMP) work using FREERTOS on one core and bare metal application on the other. This is the first step towards making that possible. This change alone will not make this possible. I will create another issue and PR for making that possible as there are more risky changes the need to be made for that.

See Issue https://github.com/espressif/esp-idf/issues/12480


components/esp_mm/esp_mmu_map.c
```diff
@@ -424,7 +424,7 @@ static void IRAM_ATTR NOINLINE_ATTR s_do_mapping(mmu_target_t target, uint32_t v
 
     cache_bus_mask_t bus_mask = cache_ll_l1_get_bus(0, vaddr_start, size);
     cache_ll_l1_enable_bus(0, bus_mask);
-#if !CONFIG_FREERTOS_UNICORE
+#if !CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE
```
igrr, line 427: @FL0WL0W In your AMP concept, what meaning do you associate with CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE?
1. Does it mean that only one core is used — FreeRTOS running on CPU 0, and nothing at all runs on CPU1
2. Or does it mean that only one core is used by an IDF app, but there may possibly be a bare-metal app running on CPU1?

Or, putting the same question the other way, what will be the value of CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE in an AMP scenario?

FL0WL0W: Hi @igrr

I will be creating another pull request for AMP. This one is more of a logical reason. These places are not FreeRTOS bits of code and are part of the ESP_System.

To answer your question. In AMP, FREERTOS_UNICORE will be defined and ESP_SYSTEM_SINGLE_CORE_MODE will not be defined. FreeRTOS will be running on one core, while the other core runs a bare metal application.

FREERTOS_UNICORE means FreeRTOS is only running on a single core, but should not necessarily mean nothing is running on the other core
ESP_SYSTEM_SINGLE_CORE_MODE means only 1 core is running code and the other core is disabled and held in reset.

Thank you

igrr: In that case, I don't think all the changes in this PR are correct. For example, when we disable the cache during flash operations, we should only synchronize with the CPU1 if there is an ipc1 task running there. However, in the AMP case there is no ipc1 task running on CPU 1, so the task writing to flash on CPU 0 will get blocked forever.

FL0WL0W: Yes you are correct, I switched that to use ipc_isr instead when in AMP mode. But that is a larger change that will need more effort to review. I want to get this PR finished first since it is effectively not changing anything since currently ESP_SYSTEM_SINGLE_CORE_MODE = FREERTOS_UNICORE

o-marshmallow:
Hello @FL0WL0W ,

Thank you for your contribution! Your MR does correct the assumptions that `ESP_SYSTEM_SINGLE_CORE_MODE == FREERTOS_UNICORE` while writing the drivers, we would like to see it merged once you consider it ready.

Regarding the future AMP feature, it has been discussed in this Github issue in the past: https://github.com/espressif/esp-idf/issues/10410

While merge this AMP feature as part of IDF experimental features is theoretically feasible, we would rather prefer it to be an external component since we don't plan on supporting or maintaining it.

Of course, AMP feature would still be supported by Espressif Zephyr team as part of the Zephyr Project.

FL0WL0W:
Hi @o-marshmallow 

I consider this PR as ready for merging.

I understand that Espressif is not interested in supporting the AMP feature. There are a lot of moving parts that make it difficult to support it. As @igrr mentioned in #10410 there are silicon bug workarounds and memory considerations that make it impossible to make this a separate component from the IDF Kernel.

I would be happy to have AMP merged as an experimental feature, with the expectation that many features outside of basic GPIO and heap memory allocation may not work as expected. We can talk more about this in the upcoming PR.

Thank you

FL0WL0W:
Does "Selected For Development" mean you want me to continue developing the AMP feature? or that you are testing this change in your internal development GITLab?


Looking at the best way to implement the AMP feature, What is the logical difference between portNUM_PROCESSORS and configNUM_CORES? they are always the same value and seem to be used interchangeably. Can i assume that portNUM_PROCESSORS should be ESP_SYSTEM related and configNUM_CORES is FREERTOS related?

o-marshmallow:
Hello @FL0WL0W ,

Sorry for the delay of response, `Selected For Development` means that your PR is accepted and we are going to create an internal PR to merge it IDF. Once merged and synced to Github, it will be updated here.

o-marshmallow:
@FL0WL0W In order to preserve your as the commit author, would you mind squashing your first two commits, rebase on master and modify the description to something more explicit such as:
```
Use `CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE` configuration option instead of `CONFIG_FREERTOS_UNICORE` in components not related to FreeRTOS
```

If you don't mind losing the SHA and the commit author for your changes (due to ` git rebase`), I can perform these modifications locally after creating an internal MR.

Tell me what you prefer

FL0WL0W:
@o-marshmallow First time I have used rebase, but I think I did it correctly. If not, feel free to perform the modifications yourself.

o-marshmallow:
@FL0WL0W Thank you, it's much better!

o-marshmallow:
sha=c235f7b0a166b9633fe0d70398f94943dc6997c8

FL0WL0W:
I think this can be closed now that it is pulled in.

Issue #12493, opened 2023-10-31
open_memstream [used by esp-console] fails when CONFIG_HEAP_TASK_TRACKING: assert failed: insert_free_block tlsf.c:358 (current && "free list cannot have a null entry") (IDFGH-11345)


alonbl:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.2-dev-3775-gb4268c874a

### Espressif SoC revision.

ESP32-S3 (QFN56) (revision v0.1)

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32-S3-WROOM-1

### Power Supply used.

USB

### What is the expected behavior?

Provided the following program:

```c
#include <stdio.h>

void app_main(void)
{
        char *ptr = NULL;
        size_t size = 0;
        FILE *fp = open_memstream(&ptr, &size);
        if (fp != NULL) {
                fclose(fp);
        }
}
```

And the following config:

```ini
CONFIG_HEAP_TASK_TRACKING=y
```

Legit code should run properly and not crash.

### What is the actual behavior?

Crash as per the following traceback:

```
I (303) main_task: Started on CPU0
I (313) main_task: Calling app_main()

assert failed: insert_free_block tlsf.c:358 (current && "free list cannot have a null entry")
HINT: CORRUPT HEAP: heap metadata corrupted resulting in TLSF malfunction.
Make sure you are not making out of bound writing on the memory you allocate in your application.
Make sure you are not writing on freed memory.
For more information run 'idf.py docs -sp api-reference/system/heap_debug.html'.


[...]

PR #12507, opened 2023-11-01
change(console): drop the use of open_memstream() (IDFGH-11361)

alonbl:
the argtable3 provides string based output return, there is no need to go via stdio structures.

I found this usage during the review of #12493, we can implement this simpler, this code is the same as what runs in argparse3 when using `open_memstream()`.


igrr (approved):

alonbl:
Hi @0xjakob, can you please also review this one?

igrr:
sha=270112fdaf8c414364b73a9029206de76ae64c8f

PR #12514, opened 2023-11-02
Update ledc_basic_example_main.c the order of initializers should match the declaration (IDFGH-11370)

diplfranzhoepfinger:
the order of initializers should match the declaration, to be able to also use this code in C++


songruo:
sha=007c2b155e2506d432d92c540227fb470af12372

PR #12525, opened 2023-11-04
change(console): print sorted help (IDFGH-11382)

alonbl:
console commands may be registered in random order in application, for example each module registers its own commands.

the output of help is displayed to human, best to have consistent sorted output so that the implementation ordering will not affect the output and allow the user to see a list in some logic order.

### DIFF
--- a/components/console/commands.c
+++ b/components/console/commands.c
@@ -130,14 +130,17 @@ esp_err_t esp_console_cmd_register(const esp_console_cmd_t *cmd)
     }
     item->argtable = cmd->argtable;
     item->func = cmd->func;
-    cmd_item_t *last = SLIST_FIRST(&s_cmd_list);
+    cmd_item_t *last = NULL;
+    cmd_item_t *it;
+    SLIST_FOREACH(it, &s_cmd_list, next) {
+        if (strcmp(it->command, item->command) > 0) {
+            break;
+        }
+        last = it;
+    }
     if (last == NULL) {
         SLIST_INSERT_HEAD(&s_cmd_list, item, next);
     } else {
-        cmd_item_t *it;
-        while ((it = SLIST_NEXT(last, next)) != NULL) {
-            last = it;
-        }
         SLIST_INSERT_AFTER(last, item, next);
     }
     return ESP_OK;
--- a/components/console/test_apps/console/main/test_console.c
+++ b/components/console/test_apps/console/main/test_console.c
@@ -78,6 +78,21 @@ TEST_CASE("esp console help command", "[console][ignore]")
     TEST_ESP_OK(esp_console_cmd_register(&s_quit_cmd));
     TEST_ESP_OK(esp_console_register_help_command());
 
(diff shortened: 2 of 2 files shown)


0xjakob (approved):
@alonbl This code looks good! We'll handle this in our internal repository, where we will probably also add some unit tests. After merging there, the change will soon be published on master.

0xjakob:
sha=97251d324405e9296ccc753b93b3485e27e46656

alonbl:
Thanks, I can add unit test if required. I will work something out.

alonbl:
@0xjakob, I added a manual unit test that is per the way existing test works.

0xjakob:
sha=07a9137fd769f0da2eda5936cc3e3db3232e8166

chipweinberger:
personally, I initialize my console commands to get the precise order I want.

0xjakob:
@chipweinberger I'm sorry to hear that this change causes inconvenience for you. However, please keep in mind that before this change, the order of commands was implementation-specific, i.e., it was not defined by the API (see the documentation about [command registration](https://docs.espressif.com/projects/esp-idf/en/v5.2/esp32/api-reference/system/console.html#command-registration-and-dispatching) and the [help command](https://docs.espressif.com/projects/esp-idf/en/v5.2/esp32/api-reference/system/console.html#_CPPv433esp_console_register_help_commandv)). If we would not be allowed to change implementation defined behavior, this would put us into a really difficult position, because then any change, including bugfixes, might be a breaking change. After all, now the order is defined. Even though it might not be what you are used to and even though you might need to change some tools, now you can rely on the order without worrying that another change or minor bugfix can easily change it again.

chipweinberger:
of course i understand your sentiment 

but can you also have the old behavior as an option?

i.e. order depending on initialization order

0xjakob:
@chipweinberger @alonbl After consultation with colleagues, we concluded that there is a point with registering commands in a certain order, e.g., when there are more important commands which should appear first. We would like to add a Kconfig option for switching between sorting in registration order and sorting in alphabetical order then. Would that be OK for both of you?

chipweinberger:
yes thats great =)

alonbl:
Hi,

If custom order strategy is important, I would have added a callback
registration to perform the ordering and order hint when registering.
This will make sure that the API supports any strategy, we can provide out
of the box text sort and hint sort callbacks.
The hint is important as most (I hope) developers do not implement
monolithic code as implied above, this enables each module to register its
own commands, and achieve a specific order (high priority and low
priority). If you accept this I will workout a patch.

I also think that adding a selector should also be done using an API and
not Kconfig, as there is already initialization and configuration using API
for this module.

Regards,

0xjakob:
@alonbl This would work. However, the old API should still look and behave the same as before so users don't have a breaking change.

BTW., Kconfig options are not only used for pure configuration, but often also for reduction of code size. But so far, the code overhead seems negligible for the console component.

PR #12558, opened 2023-11-10
Fix: esptool_py incorrectly assumed target name equals binary name (IDFGH-11420)

itavero:
While integrating the ESP-IDF into our existing CMake structure, I've come across quite some hurdles.
Most I've been able to fix in our CMake files, however this one I could not.

Most of the targets created by the esptool_py component assume that the `EXECUTABLE` IDF build property (which contains the name of the CMake executable target) always equals the name of the created binary.

This is however not always  true. For instance, in our setup we use `CMAKE_EXECUTABLE_SUFFIX_C` and `CMAKE_EXECUTABLE_SUFFIX_CXX` in our toolchain file (both set to `.elf`).
If we do `add_executable(my_app)`, the target binary file would actually be `my_app.elf`.

In order to fix this, I've updated it to use the `TARGET_FILE` generated expression. That way we also no longer need the `EXECUTABLE_DIR` IDF build property here.

I've fixed this on v5.0.1 (as that's the ESP-IDF version I'm currently trying to integrate), but I assume it should be easy to apply the same fix to newer versions and the master branch as well.

Note that this problem might exist in multiple places where `EXECUTABLE` is being used.
While going through the ESP-IDF code base, I even noticed that a few places actually already seem to use the `TARGET_FILE` expression.
To be honest the property name might be somewhat confusing as well, as it is actually the **executable _target_**.

### DIFF
--- a/components/esptool_py/project_include.cmake
+++ b/components/esptool_py/project_include.cmake
@@ -122,7 +122,6 @@ idf_build_get_property(build_dir BUILD_DIR)
 
 idf_build_get_property(elf_name EXECUTABLE_NAME GENERATOR_EXPRESSION)
 idf_build_get_property(elf EXECUTABLE GENERATOR_EXPRESSION)
-idf_build_get_property(elf_dir EXECUTABLE_DIR GENERATOR_EXPRESSION)
 
 if(CONFIG_SECURE_BOOT_BUILD_SIGNED_BINARIES AND NOT BOOTLOADER_BUILD)
     set(unsigned_project_binary "${elf_name}-unsigned.bin")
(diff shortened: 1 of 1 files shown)


igrr:
Thanks for the fix @itavero!

We can't merge the PR as is, because we have to first make the fix in the `master` branch. We will rebase your commit to `master` and later will backport the change to the existing release branches (release/v5.2, release/v5.1, release/v5.0).

itavero:
@igrr I've rebased it on `master` and did a force push. Looking forward to a backport to v5.0.

fhrbata:
sha=5293eff4cdd8545266bafdec6fe98522bbaf9c28

PR #12559, opened 2023-11-10
[ESP32-C2] Correct REG_SPI_BASE(i) (IDFGH-11421)

TD-er:
Given these defines (verified against the technical reference guide):
```c++
#define DR_REG_SPI1_BASE                        0x60002000
#define DR_REG_SPI0_BASE                        0x60003000
#define DR_REG_SPI2_BASE                        0x60024000
```

The existing formula can never match these registers.


suda-morris:
sha=c55df6374df94160c04730869cacb7ec8d8f0308

PR #12566, opened 2023-11-11
add host midi example

Wunderbaeumchen99817:
This new example implements midi functionality with the ESP acting as a usb-host device. 

It connects to the connected usb-midi-controller & prints the received signals in the serial monitor.

### DIFF
--- a/examples/peripherals/usb/host/midi/CMakeLists.txt
+++ b/examples/peripherals/usb/host/midi/CMakeLists.txt
@@ -1,0 +1,6 @@
+# The following lines of boilerplate have to be in your project's
+# CMakeLists in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.5)
+
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+project(app-template)
--- a/examples/peripherals/usb/host/midi/README.md
+++ b/examples/peripherals/usb/host/midi/README.md
@@ -1,0 +1,136 @@
+| Supported Targets | ESP32-S2 | ESP32-S3 |
+| ----------------- | -------- | -------- |
+
+# USB MIDI Class Example
+
+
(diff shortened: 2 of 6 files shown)


chipweinberger:
@Wunderbaeumchen99817 , if they don't merge this, you should make a repo called "ESP32 Usb Midi Example".

It would be great for people to google and find it.

chaabanihoussem:
Hi, how does this works? 
so I flashed esp32s3 board and then I couldn't see it as a midi on ubuntu.

what I'm actually trying to do is to build a device that acts as a MIDI host when it's attached to a MIDI device and as MIDI device when it's attached to a MIDI host. The problem I'm trying to solve is that the esp32s3 flashed with MIDI device example now appears when attached to windows/macos/android but doesn't appear when attached to music instrument (korg keyboard).

Any idea will help
Thanks

roma-jam:
@chaabanihoussem,

In current PR there is a simple MIDI host. So if you are flash it and then attach it to the Ubuntu host - it is not going to work, as you connected two hosts together. 

For MIDI device example, which does not appear - which example did you try?

chaabanihoussem:
@roma-jam thank you getting back

I tried the tusb_midi under examples/peripherals/usb/device and that works fine when connected to a host.

I'm now trying to connect the esp32s3 to a MIDI device and send MIDI cmds to it as 
`tud_midi_stream_write(cable_num, note_off, 3);`
does in the midi device example.

Is there a way the host can send midi cmds to the device? 
maybe something like `tuh_midi_stream_write(cable_num, note_off, 3);` ?

Thanks

roma-jam:
@chaabanihoussem,

Unfortunately, we don't have MIDI class support. 
All supported classes could be found here:  https://github.com/espressif/esp-usb/tree/master/host/class

Right at the moment we don't have a plan to implement MIDI host class device, thus if you need MIDI class implementation, it is possible to design it using the class drivers as a reference by the link I have mentioned before.

PR #12570, opened 2023-11-13
Add check prov mgr is stopped (IDFGH-11433)

Diam2023:
When I called the wifi_prov_mgr_stop_provisioning function it didn't stop immediately,
So I added a check to see if the prov was stopped


laukik-hase:
Hello, @Diam2023!

Thanks for the contribution.

Rather than adding a new API, could you use an event handler to handle the `WIFI_PROV_END` which signifies the stopping of the provisioning service? See reference [here](https://github.com/espressif/esp-idf/blob/master/examples/provisioning/wifi_prov_mgr/main/app_main.c#L152-L155).

Diam2023:
> Hello, @Diam2023!
> 
> Thanks for the contribution.
> 
> Rather than adding a new API, could you use an event handler to handle the `WIFI_PROV_END` which signifies the stopping of the provisioning service? See reference [here](https://github.com/espressif/esp-idf/blob/master/examples/provisioning/wifi_prov_mgr/main/app_main.c#L152-L155).

Hello, @laukik-hase!

Thanks for your reply.

This is my idea, Consider the following situation:
In an inappropriate usage event, when the device is connected to Android, Android actively disconnects the Bluetooth connection, causing prov to fail. 
Now the status value obtained by `wifi_prov_mgr_get_wifi_state` is `WIFI_PROV_STA_DISCONNECTED`. This cannot indicate whether the prov_mgr prov is successful or not. The status is caused by prov failure, but the reason for the disconnection can be obtained by `wifi_prov_mgr_get_wifi_disconnect_reason`.
If I don't care about event, I need to call the stop network provisioning API and wait for it to completely stop before calling `wifi_prov_mgr_start_provisioning` to restart prov, But There is no way to know the status of `prov_ctx->prov_state`, so I want an API to get whether it is completely stopped or blocked until it is completely stopped.

For example:

```cpp

void wifi_prov_mgr_stop_provisioning_blocking()
{

    if (!prov_ctx_lock) {
        ESP_LOGE(TAG, "Provisioning manager not initialized");
        return;
    }

    ACQUIRE_LOCK(prov_ctx_lock);

    wifi_prov_mgr_stop_service(1);

    RELEASE_LOCK(prov_ctx_lock);
}

//  -------------- OR

bool wifi_prov_mgr_is_stopped()
{
    return prov_ctx->prov_state == WIFI_PROV_STATE_IDLE;
}

```

laukik-hase:
sha=65bc5668da702dead5424bc238bf548361a2ca1c

laukik-hase:
Merged with c93bce28ed52635065abfbee6cde107f2e13d8aa

Thanks for the contribution, @Diam2023!

PR #12618, opened 2023-11-17
example: Optimized parts of the code in the gatt_server_service_table example. (IDFGH-11491)

TritiumXs:
I found some redundancy in certain parts of the code, so I made modifications by removing redundant segments. These changes do not affect the functionality of the code.

### DIFF
--- a/examples/bluetooth/bluedroid/ble/gatt_server_service_table/main/gatts_table_creat_demo.c
+++ b/examples/bluetooth/bluedroid/ble/gatt_server_service_table/main/gatts_table_creat_demo.c
@@ -104,21 +104,8 @@ static esp_ble_adv_data_t adv_data = {
 };
 
 // scan response data
-static esp_ble_adv_data_t scan_rsp_data = {
-    .set_scan_rsp        = true,
-    .include_name        = true,
-    .include_txpower     = true,
-    .min_interval        = 0x0006,
-    .max_interval        = 0x0010,
-    .appearance          = 0x00,
-    .manufacturer_len    = 0, //TEST_MANUFACTURER_DATA_LEN,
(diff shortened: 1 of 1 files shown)


zhp0406:
Thank you for your feedback. The reason for writing it this way is primarily to emphasize the independence of the contents of adv_data and scan_rsp_data, aiming to prevent any potential misunderstanding on the part of the users.

TritiumXs:
> Thank you for your feedback. The reason for writing it this way is primarily to emphasize the independence of the contents of adv_data and scan_rsp_data, aiming to prevent any potential misunderstanding on the part of the users.

Okay, but the explanation for "scan_rsp_data.p_service_data" isn't the same as for "adv_data." This is a minor detail, but if this issue can be corrected, I believe it can be improved.

zhp0406:
Yes, we will further improve this area in the future. Do you find any other parts of the documentation unclear? We would greatly appreciate your feedback.

TritiumXs:
> Yes, we will further improve this area in the future. Do you find any other parts of the documentation unclear? We would greatly appreciate your feedback.

Oh, indeed, there are some suggestions. UUIDs are really hard to remember. I think we can define a macro, allowing users to initialize using this macro and then modify it according to their needs. Also, the explanation for "raw_adv_data" in this example isn't very clear and may not be very friendly for developers new to BLE. I hope it can be supplemented more comprehensively.

zhp0406:
Alright, I will explain the format of raw_scan_rsp_data in the document later to make it easier for beginners to understand.

TritiumXs:
> Alright, I will explain the format of raw_scan_rsp_data in the document later to make it easier for beginners to understand.

And I found that the macro for "appearance" has already been defined, but it hasn't been used in the example.

Issue #12389, opened 2023-10-11
Wrong SDK version read, when building binaries via docker image (IDFGH-11223)


bhuvanchandra:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.1.1

### Operating System used.

Linux

### How did you build your project?

using idf.py build from within esp-idf docker image

### If you are using Windows, please specify command line type.

None

### What is the expected behavior?

when using esp_get_idf_version expected version is "v5.1.1"

### What is the actual behavior?

when using esp_get_idf_version expected version is "HEAD-HASH-NOTFOUND"

### Steps to reproduce.

use esp-idf docker image for building the firmware binaries and call the esp_get_idf_version method.

### More Information.

When building the firmware binaries through the CI pipeline, the ESP-IDF version read via the API is incorrect. Instead of the expected "v5.1.1", it reads "HEAD-HASH-NOTFOUND". Upon verifying, the Docker image for the esp-idf doesn't explicitly check out to the release tag, which is necessary for the API to retrieve the correct version from the git tag.

I'm using [this](https://github.com/espressif/esp-idf-ci-action) action, v5.1.1 docker tag.

fhrbata:
Hello @bhuvanchandra , 

I've tried using the v5.1.1 docker image locally and also with the `esp-idf-ci-action`, but I cannot reproduce the behavior you are describing. My testing repository for this can be found [here](https://github.com/fhrbata/hello_world). It's the `hello_world` example from `esp-idf`  with dispatch action using `esp-idf-ci-action`.  I've downloaded the [artifacts](https://github.com/fhrbata/hello_world/actions/runs/6532320139) from the dispatch workflow and flash them into `esp32s3`.

1. Download the artifacts from https://github.com/fhrbata/hello_world/actions/runs/6532320139
2. Flashed with `esptool.py  -b 460800 --before default_reset --after hard_reset --chip esp32s3  write_flash --flash_mode dio --flash_size 2MB --flash_freq 80m 0x0 bootloader/bootloader.bin 0x8000 partition_table/partition-table.bin 0x10000 hello_world.bin`
[...]

PR #12636, opened 2023-11-21
Fix Dockerfile to create Images with no git safe directory restrictions

timoxd7:
## Problem
If a folder which is a repository is mounted on the docker container and then a child folder is build, the docker container is not able to see the git repo in the parent folder.

## Fix
Just set every folder as safe for git, as this implies no security risk in a docker container, solving the issue to not be able to access certain git repositories.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -38,6 +38,9 @@ RUN : \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3 10 \
   && :
 
+# Configure git to allow access to all git repositories in the Filesystem
+RUN git config --global --add safe.directory '*'
+
 # To build the image for a branch or a tag of IDF, pass --build-arg IDF_CLONE_BRANCH_OR_TAG=name.
 # To build the image with a specific commit ID of IDF, pass --build-arg IDF_CHECKOUT_REF=commit-id.
 # It is possibe to combine both, e.g.:


igrr:
@timoxd7 could you please give an example of a docker run command and project structure which results in this issue?

timoxd7:
@igrr I updated the comment on the pull request as i had it wrongly in my mind. The problem happens if the user inside the docker container does not match the one in the filesystem (.git parent folder). Might happen during WSL usage or in CI.

timoxd7:
@igrr Example Output:

```
$ docker run --rm -v "${PWD}":/project -w /project -it espressif/idf:latest
Detecting the Python interpreter
Checking "python3" ...
Python 3.10.12
"python3" has been detected
Checking Python compatibility
Checking other ESP-IDF version.
Adding ESP-IDF tools to PATH...
Checking if Python packages are up to date...
Requirement files:
 - /opt/esp/idf/tools/requirements/requirements.core.txt
Python being checked: /opt/esp/python_env/idf5.2_py3.10_env/bin/python
Python requirements are satisfied.
Added the following directories to PATH:
  /opt/esp/idf/components/espcoredump
  /opt/esp/idf/components/partition_table
  /opt/esp/idf/components/app_update
  /opt/esp/tools/xtensa-esp-elf-gdb/12.1_20221002/xtensa-esp-elf-gdb/bin
  /opt/esp/tools/riscv32-esp-elf-gdb/12.1_20221002/riscv32-esp-elf-gdb/bin
  /opt/esp/tools/xtensa-esp32-elf/esp-12.2.0_20230208/xtensa-esp32-elf/bin
  /opt/esp/tools/xtensa-esp32s2-elf/esp-12.2.0_20230208/xtensa-esp32s2-elf/bin
  /opt/esp/tools/xtensa-esp32s3-elf/esp-12.2.0_20230208/xtensa-esp32s3-elf/bin
  /opt/esp/tools/riscv32-esp-elf/esp-12.2.0_20230208/riscv32-esp-elf/bin
  /opt/esp/tools/esp32ulp-elf/2.35_20220830/esp32ulp-elf/bin
  /opt/esp/tools/cmake/3.24.0/bin
  /opt/esp/tools/openocd-esp32/v0.12.0-esp32-20230419/openocd-esp32/bin
  /opt/esp/python_env/idf5.2_py3.10_env/bin
  /opt/esp/idf/tools
Done! You can now compile ESP-IDF projects.
Go to the project directory and run:

  idf.py build

root@76277e23dee1:/project# git status
fatal: detected dubious ownership in repository at '/project'
To add an exception for this directory, call:

        git config --global --add safe.directory /project
root@76277e23dee1:/project#
```

dobairoland:
Related to https://github.com/espressif/esp-idf/issues/12389

fhrbata:
Hello @timoxd7 ,

first thank you for this contribution.

> However, i don't think configuring every folder inside the docker container as safe repository implies any security risks, as it is not on a local machine with other things going on, but only an isolated container. Also, everything going into the container is explicitly mounted, e.g. every repository brought into the container is already kind of "owned", making the safe directory redundant. What do you think about this?

IIUC the `safe.directory` git config option was added by [8959555cee7ec045958f9b6dd62e541affb7e7d9](https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9) to address [CVE-2022-24765](https://nvd.nist.gov/vuln/detail/cve-2022-24765).

I think that with this change we would make it possible to exploit this vulnerability again. For example user starting the container can bind mount some shared drive with plenty of git projects owned by different users. IMHO with `safe.directory '*'`  it would be still possible to start an arbitrary command if someone created a rogue .git directory in some parent directory and set e.g. `core.fsmonitor`. 

Thank you

fhrbata:
I would probably avoid setting `safe.directory '*' ` by default in our image, but I guess we can make it conditional with some additional info in our docs about the possible security impact, so users can easily build their custom image with setting. Thoughts?

dobairoland:
Good idea @fhrbata. Perhaps a variable which defaults to IDF_PATH which could be overwritten to `*`. The official images won't change and we wouldn't even need to explain the `*` case because it won't be in our dockerfiles.

timoxd7:
@fhrbata You are right, i mainly had a CI/CD in mind, but users could still use it in private systems and other setups this could potentially be a problem as you stated out.

@dobairoland I had this kind of variable also in mind, which will not solve the problem if the official image is used. Maybe a documentation entry for the docker image would be great, informing the user to add the project's repository to the docker internal git safe directories.
A variable defaulting to `IDF_PATH` would also be good, but i would suggest to have a variable like `IDF_GIT_SAFE_DIR` that will add all listed additional dirs to git's safe directory list. With this, the `IDF_PATH` will always be a safe directory for git.

If you both agree, i could make a new pull request with this variable while closing this one.

fhrbata:
Hello @timoxd7 ,

thank you for the additional input!

> @dobairoland I had this kind of variable also in mind, which will not solve the problem if the official image is used. Maybe a documentation entry for the docker image would be great, informing the user to add the project's repository to the docker internal git safe directories. A variable defaulting to `IDF_PATH` would also be good, but i would suggest to have a variable like `IDF_GIT_SAFE_DIR` that will add all listed additional dirs to git's safe directory list. With this, the `IDF_PATH` will always be a safe directory for git.
> 
> If you both agree, i could make a new pull request with this variable while closing this one.

This is just me thinking out loud, but maybe we can move setting the `safe.directory` into the `entrypoint.sh`, maybe based on an environmental variable passed to `docker run`. Or maybe passing it as argument to entrypoint via CMD. Meaning it would not be hardcoded in the image, but could be customized per container instance.

`docker run --rm -v "${PWD}":/project -w /project -it -e IDF_GIT_SAFE_DIR=<dirs> espressif/idf:latest`

and `entrypoint.sh` would set it. I haven't tested this yet, so it's possible that I'm completely missing something.

fhrbata:
Hello @timoxd7 , @dobairoland 

do you see any problem with the following approach?  I quickly tested it and it seems to be working.

Thank you

```diff
   diff --git a/tools/docker/entrypoint.sh b/tools/docker/entrypoint.sh                                                                 
index 7cf15f1917e1..513b3ba00d28 100755                                                                                              
--- a/tools/docker/entrypoint.sh                                                                                                     
+++ b/tools/docker/entrypoint.sh                                                                                                     
@@ -1,6 +1,20 @@                                                                                                                     
 #!/usr/bin/env bash                                                                                                                 
 set -e                                                                                                                              
                                                                                                                                     
+# IDF_GIT_SAFE_DIR has the same format as system PATH environment variable.                                                         
+# All path specified in IDF_GIT_SAFE_DIR will be added to user's                                                                    
+# global git config as safe.directory paths. For more information                                                                   
+# see git-config manual page.                                                                                                       
+if [ -n "${IDF_GIT_SAFE_DIR+x}" ]                                                                                                   
+then                                                                                                                                
+       echo "Adding following directories into git's safe.directory"                                                                
+       echo "$IDF_GIT_SAFE_DIR" | tr ':' '\n' | while read -r dir                                                                   
+       do                                                                                                                           
+               git config --global --add safe.directory "$dir"                                                                      
+               echo "  $dir"                                                                                                        
+       done                                                                                                                         
+fi                                                                                                                                  
+                                                                                                                                    
 . $IDF_PATH/export.sh                                                                                                               
                                                                                                                                     
 exec "$@"                                                                                                                           
```

usage
`docker run -it --rm -v /home/fhrbata/tmp/hello_world:/project -w /project -e IDF_GIT_SAFE_DIR='/project' idf-custom`
or
`docker run -it --rm -v /home/fhrbata/tmp/hello_world:/project -w /project -e IDF_GIT_SAFE_DIR='*' idf-custom`

dobairoland:
@fhrbata LGTM

timoxd7:
@fhrbata Looks good, if it is tested and works then i think it is the best solution. If it would be documented to be needed in the official documentation would make it perfect and easier to solve the versioning problem for someone first experiencing the problem.

fhrbata:
Hello @timoxd7 ,

thank you very much for your feedback.

> @fhrbata Looks good, if it is tested and works then i think it is the best solution. If it would be documented to be needed in the official documentation would make it perfect and easier to solve the versioning problem for someone first experiencing the problem.

I've already created a MR for this and once it's approved and merged, it will be synced into the github. It also contains a note about this possibility in the `Building a Project with CMake` section.

>  When the mounted directory, ``/project``, contains a git repository owned by a different user (``UID``) than the one running the Docker container, git commands executed within ``/project`` might fail, displaying an error message ``fatal: detected dubious ownership in repository at '/project'``. To resolve this issue, you can designate the ``/project`` directory as safe by setting the IDF_GIT_SAFE_DIR environment variable during the docker container startup. For instance, you can achieve this by including ``-e IDF_GIT_SAFE_DIR='/project'`` as a parameter. Additionally, multiple directories can be specified by using a ``:`` separator. To entirely disable this git security check, ``*`` can be used.

Hopefully it will be useful.

Thank you

PR #12637, opened 2023-11-21
Dockerfile with variable depth parameter (IDFGH-11513)

timoxd7:
## Problem
If a docker image is built from a branch that has a commit offset in respect to the last tag while using `IDF_CLONE_SHALLOW`, the `git describe` will fail returning "1".

## Fix
Add another parameter to specify the depth instead of falling back to "1". If set to the offset of the commit +1, the `git describe` works again, still without cloning the whole repository. For custom IDF and development builds, this helps to get a working build version number again, instead of just "1" being put into the build es ESP IDF Version.

## Backport
We use ESP-IDF v5.1.x, so a backport would be great.


timoxd7:
Fixed commit message as suggested

igrr (changes requested):
Thank you for the PR @timoxd7! Could you please add a line about this variable also to the docs over here:
https://github.com/espressif/esp-idf/blob/a7fbf452fa181452f266b16c49063e2ff069f119/docs/en/api-guides/tools/idf-docker-image.rst?plain=1#L142

timoxd7:
@igrr Forgot about that, i have added it to this pull request and the backport one.

igrr:
Thanks, looks good. Don't worry about the backport — since we will also need to update the Chinese version of the documentation (docs/zh_CN/api-guides/tools/idf-tools.rst), we will backport the final version together with the translation. We'll keep you as the commit author, of course.

igrr (approved):

igrr:
sha=7cfdec1649466ff5cb3363803f5f31424c986deb

PR #12639, opened 2023-11-21
Dockerfile with variable depth parameter [v5.1 backport] (IDFGH-11515)

timoxd7:
## Problem
If a docker image is built from a branch that has a commit offset in respect to the last tag while using `IDF_CLONE_SHALLOW`, the `git describe` will fail returning "1".

## Fix
Add another parameter to specify the depth instead of falling back to "1". If set to the offset of the commit +1, the `git describe` works again, still without cloning the whole repository. For custom IDF and development builds, this helps to get a working build version number again, instead of just "1" being put into the build es ESP IDF Version.

## Info
This is a Backport of #12637


dobairoland:
As I pointed out [here](https://github.com/espressif/esp-idf/pull/12640#issuecomment-1820912873), please point out in https://github.com/espressif/esp-idf/pull/12637 that you'd like to have it in v5.1.

timoxd7:
@dobairoland Did not know that, will keep it in mind for future PRs. Thank you, i will close this now.

PR #12640, opened 2023-11-21
Fix Dockerfile to create Images with no git safe directory restrictions [v5.1 backport] (IDFGH-11514)

timoxd7:
## Problem
If a folder is mounted to the docker container that is not owned by the running user, git disallows the direct access without explicit set of a safe directory. This can happen for example if using wsl on Windows while using a repository on the Windows Filesystem.

## Fix
Just set every folder as safe for git, as this implies no security risk in a docker container, solving the issue to not be able to access certain git repositories.

## Reference
The github action for ESP-IDF Builds solved this by adding the command before running the actual action command (See espressif/esp-idf-ci-action#27)

## Info
This is a backport of #12636

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -34,6 +34,9 @@ RUN : \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3 10 \
   && :
 
+# Configure git to allow access to all git repositories in the Filesystem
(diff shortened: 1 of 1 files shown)


dobairoland:
We only accept PRs for the master branch. It is enough to state in https://github.com/espressif/esp-idf/pull/12636 clearly that you'd like this to be backported to v5.1.

PR #12666, opened 2023-11-26
Update spi_master_example_main.c

FlyyingPiggy2020:
According to the definition of the structure, the number of bytes here should be 2


suda-morris:
sha=7317555d39b759c9c67e9fcd8ae433e8ea4fdcb7

Issue #12185, opened 2023-09-03
Undefined reference to `phy_bbpll_en_usb` when using USB CDC and WiFi at the same time on ESP32-S2 (IDFGH-10996)


Strat00s:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

V4.4.5, V5.1.1, master

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

None

### What is the expected behavior?

It should compile.

### What is the actual behavior?

Looks like S2 is missing `phy_bbpll_en_usb` implementation as all I get is `undefined reference to 'phy_bbpll_en_usb'`.

### Steps to reproduce.

0. Have ESP32-S2
1. Create wifi station or scan example (probably any wifi example will do, I tried just these 2).
2. Enable USB when phy init, set USB CDC for console output.
3. Try and build it.

### Build or installation Logs.

```plain
[1/111] Building C object esp-idf/bootloader_support/CMakeFiles/__idf_bootloader_support.dir/src/esp32s2/bootloader_soc.c.obj
[2/111] Building C object esp-idf/bootloader_support/CMakeFiles/__idf_bootloader_support.dir/src/bootloader_console_loader.c.obj
[3/111] Building C object esp-idf/bootloader_support/CMakeFiles/__idf_bootloader_support.dir/src/esp32s2/bootloader_sha.c.obj
[4/111] Building C object esp-idf/bootloader_support/CMakeFiles/__idf_bootloader_support.dir/src/bootloader_panic.c.obj
[5/111] Building C object esp-idf/efuse/CMakeFiles/__idf_efuse.dir/esp32s2/esp_efuse_fields.c.obj
[6/111] Building C object esp-idf/efuse/CMakeFiles/__idf_efuse.dir/esp32s2/esp_efuse_table.c.obj
[7/111] Building C object esp-idf/efuse/CMakeFiles/__idf_efuse.dir/esp32s2/esp_efuse_rtc_calib.c.obj
[8/111] Building C object esp-idf/efuse/CMakeFiles/__idf_efuse.dir/esp32s2/esp_efuse_rtc_table.c.obj
[...]

PR #12675, opened 2023-11-27
Fix phy_bbpll_en_usb compile error for ESP32-S2 (IDFGH-11554)

imoverclocked:
Closes https://github.com/espressif/esp-idf/issues/12185

Tested in VSCodium:
```
Version: 1.84.2
Release: 23314
```

Verified: compilation with esp32s2 target before commit fails with expected error and succeeds with this commit.

Verified: affected blocks still effective when target is esp32s3 or esp32c3

### DIFF
--- a/components/esp_phy/include/esp_private/phy.h
+++ b/components/esp_phy/include/esp_private/phy.h
@@ -114,9 +114,9 @@ uint8_t phy_dig_reg_backup(bool backup_en, uint32_t *mem_addr);
 void phy_freq_mem_backup(bool backup_en, uint32_t *mem);
 #endif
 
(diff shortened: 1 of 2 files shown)


Dazza0:
@imoverclocked Thanks for the PR. We're actually already working on a fix for this issue internally. However, this PR is missing a couple of targets where the `phy_bbpll_en_usb()` function is defined.

If you prefer for this this PR to be merged instead of closing it yourself, we can always squash our internal fix onto your PR commit so that contribution is still attributed to you.

imoverclocked:
@Dazza0 That’s great news!

Whatever is easiest, the PR is pretty trivial so I’m happy to close and let you carry on!

Issue #12682, opened 2023-11-28
W5500 Ethernet without interrupt (IDFGH-11561)


br101:
### Is your feature request related to a problem?

I have encountered a few different hardware designs which connect the W5500 to an ESP without the INT line connected (the Arduino Ethernet Shield 2 or M5Stack PoECAM for example) and it would be good to be able to support it with the W5500 driver in esp-idf.

### Describe the solution you'd like.

When the INT line is unavailable int_gpio_num could be set to -1 and the driver should work around it by using polling for events.

br101:
Something like:

diff --git a/components/esp_eth/src/esp_eth_mac_w5500.c b/components/esp_eth/src/esp_eth_mac_w5500.c
index 7a7708d0ca..58596cff4d 100644
--- a/components/esp_eth/src/esp_eth_mac_w5500.c
+++ b/components/esp_eth/src/esp_eth_mac_w5500.c
@@ -652,10 +652,14 @@ static void emac_w5500_task(void *arg)
     uint32_t frame_len = 0;
     uint32_t buf_len = 0;
     while (1) {
-        /* check if the task receives any notification */
-        if (ulTaskNotifyTake(pdTRUE, pdMS_TO_TICKS(1000)) == 0 &&    // if no notification ...
-            gpio_get_level(emac->int_gpio_num) != 0) {               // ...and no interrupt asserted
-            continue;                                                // -> just continue to check again
+        /* if using interrupts: check if the task receives any notification */
+        if (emac->int_gpio_num >= 0) {
+            if (ulTaskNotifyTake(pdTRUE, pdMS_TO_TICKS(1000)) == 0 &&    // if no notification ...
+                gpio_get_level(emac->int_gpio_num) != 0) {               // ...and no interrupt asserted
+                continue;                                                // -> just continue to check again
+            }
+        } else {
+            vTaskDelay(1);
         }
         /* read interrupt status */
         w5500_read(emac, W5500_REG_SOCK_IR(0), &status, sizeof(status));
@@ -702,12 +706,14 @@ static esp_err_t emac_w5500_init(esp_eth_mac_t *mac)
     esp_err_t ret = ESP_OK;
     emac_w5500_t *emac = __containerof(mac, emac_w5500_t, parent);
     esp_eth_mediator_t *eth = emac->eth;
-    esp_rom_gpio_pad_select_gpio(emac->int_gpio_num);
[...]

PR #12692, opened 2023-11-29
W5500: Add possibility to drive without IRQ line (IDFGH-11571)

br101:
For some available hardware where the the interrupt line of the W5500 is not connected to a GPIO it is necessary to use polling. While I think this is bad hardware design, this change makes it possible to use the driver in such situations by selecting CONFIG_ETH_SPI_ETHERNET_W5500_POLLING

### DIFF
--- a/components/esp_eth/Kconfig
+++ b/components/esp_eth/Kconfig
@@ -144,6 +144,14 @@ menu "Ethernet"
                 making it compatible with the software TCP/IP stack.
                 Say yes to enable W5500 driver.
 
+        if ETH_SPI_ETHERNET_W5500
+            config ETH_SPI_ETHERNET_W5500_POLLING
+                bool "Use W5500 in polling mode (without interrupt)"
+                help
+                    Running W5500 in polling mode is not ideal, but it allows this driver to be used on
+                    hardware where the interrupt line of the W5500 chip is not connected.
+        endif # ETH_SPI_ETHERNET_W5500
+
         config ETH_SPI_ETHERNET_KSZ8851SNL
             bool "Use KSZ8851SNL"
             help
--- a/components/esp_eth/src/esp_eth_mac_w5500.c
+++ b/components/esp_eth/src/esp_eth_mac_w5500.c
@@ -713,6 +713,7 @@ err:
     return ret;
 }
 
+#if !CONFIG_ETH_SPI_ETHERNET_W5500_POLLING
 IRAM_ATTR static void w5500_isr_handler(void *arg)
 {
     emac_w5500_t *emac = (emac_w5500_t *)arg;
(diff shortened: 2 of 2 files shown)


KaeLL:
LGTM.

dannybr:
it works but: (probably related to type of esp32 chip/MHZ)
if in the main_app esp_eth_mac_new(...) is called before esp_eth_phy_new(...)
in esp_eth_mac_w5500.c emac_w5500_task will run too early, and emac->eth is not initialized yet, and
a panic then follows. I added a check AFTER  the vtaskDelay(1) and if emac->eth == NULL continue.

br101:
> it works but: (probably related to type of esp32 chip/MHZ) if in the main_app esp_eth_mac_new(...) is called before esp_eth_phy_new(...) in esp_eth_mac_w5500.c emac_w5500_task will run too early, and emac->eth is not initialized yet, and a panic then follows. I added a check AFTER the vtaskDelay(1) and if emac->eth == NULL continue.

if i'm not mistaken, i think this is a separate issue and should be reported separately

dannybr:
hi and thanks for the prompt reply,

the problem is triggered by CONFIG_ETH_SPI_ETHERNET_W5500_POLLING
so IMHO it is the same issue.
then again, you might have a point, but I didn’t see this issue before with a w5500 and IRQ
….
static void emac_w5500_task(void *arg)
{
   emac_w5500_t *emac = (emac_w5500_t *)arg;
   uint8_t status = 0;
   uint8_t *buffer = NULL;
   uint32_t frame_len = 0;
   uint32_t buf_len = 0;
   while (1) {
#if CONFIG_ETH_SPI_ETHERNET_W5500_POLLING
		 vTaskDelay(1);
		 if(emac->eth == NULL) {
			  //printf("GUEVALT: no eth!!!\n");
			  continue;
		 }
#else
       /* check if the task receives any notification */
       if (ulTaskNotifyTake(pdTRUE, pdMS_TO_TICKS(1000)) == 0 &&    // if no notification ...
           gpio_get_level(emac->int_gpio_num) != 0) {               // ...and no interrupt asserted
           continue;                                                // -> just continue to check again
       }
#endif

kostaond:
This PR is not going to be accepted as I mentioned in #12682 since proposed solution has some limitations as discussed there. Solution using ESP timer is already in internal review and be merged soon. Thanks @br101 for the proposal and for rationale why not use interrupt.

Also note that the solution with ESP timer is planned to be backported to v5.1 and v5.2 so it could be available for Arduino users.

kostaond:
Just for reference, [here](https://github.com/espressif/esp-idf/commit/fd0a1dc53c4cff99f04dfe658047f7c2e3e6d8ad) is implementation using ESP timer.

dannybr:
Thanks!
Will try ASAP,
	Danny

dannybr:
it works!
i changed the poll_period_ms to 1
thanks,
	danny

PR #12698, opened 2023-11-30
Update HID Mouse example with scroll wheel status and middle mouse button click (IDFGH-11578)

jbilander:
This also requires that byte 3 (the fourth byte) is added in the hid_usage_mouse.h struct 

https://github.com/espressif/idf-extra-components/pull/273/commits/0c9d445b54fb3190664d85ca769c3310add51472


examples/peripherals/usb/host/hid/main/hid_host_example.c
```diff
@@ -329,16 +329,19 @@ static void hid_host_mouse_report_callback(const uint8_t *const data, const int
 
     static int x_pos = 0;
     static int y_pos = 0;
+    static int wheel_pos = 0;
 
     // Calculate absolute position from displacement
     x_pos += mouse_report->x_displacement;
     y_pos += mouse_report->y_displacement;
+    wheel_pos += mouse_report->scrollwheel;
```
roma-jam, line 337: Current version of HID Driver was made for devices which have a HID boot interface. In that case the mouse report is simple and always has the default structure. Structure definition can be found in Appendix B2 of [Device Class Definition for Human Interface Devices (HID) Version 1.11](https://www.usb.org/sites/default/files/documents/hid1_11.pdf).

That is why the the structure in the hid_usage_mouse.h file for hid_mouse_input_report_boot_t satisfy the specification. 
Sure, the scroll data in report for different devices pretty much always on the same place in buffer, but to be sure about the correct way is to handle the report device descriptor and handle the report data according to structure from descriptor.

roma-jam (commented):
Hi @jbilander , 
Thanks a lot for the contribution! 

We are planning https://github.com/espressif/idf-extra-components/pull/214 to add support not only for boot interface devices, but to the any other + make it possible to define handling manually (which means to work with report descriptor and with raw report data). 

Regarding the display `button3` status it seems that we can merge it right now and for the `scrollwheel` adding seems that it is better to return to that question after https://github.com/espressif/idf-extra-components/pull/214 being merged. 

Thanks.

PR #12720, opened 2023-12-04
Fix: Remove a sign-compare build warning (IDFGH-11602)

M-Bab:
Otherwise the compare in the return line of the function causes the following warning: `comparison of integer expressions of different signedness: 'int' and 'uint32_t' {aka 'unsigned int'}`

or even a build fail if Werror=sign-compare is configured. This behavior was introduced by https://github.com/espressif/esp-idf/commit/a2580b3f361a6531e3d4cc2ea89ccb1e3c9945ff and the tags 4.4.5 and 4.4.6 are affected.

Changing the uint32_t to int32_t matches the code from IDF 5.x (without any judgement if addresses should be signed or unsigned): https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/include/esp_memory_utils.h#L305C5-L305C12

### DIFF
--- a/components/soc/include/soc/soc_memory_types.h
+++ b/components/soc/include/soc/soc_memory_types.h
@@ -127,7 +127,7 @@ inline static bool IRAM_ATTR esp_ptr_in_iram(const void *p) {
 }
 
 inline static bool IRAM_ATTR esp_ptr_in_drom(const void *p) {
-    uint32_t drom_start_addr = SOC_DROM_LOW;
+    int32_t drom_start_addr = SOC_DROM_LOW;
 #if CONFIG_ESP32S3_DATA_CACHE_16KB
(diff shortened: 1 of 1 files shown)


KaeLL:
> without any judgement if addresses should be signed or unsigned

Let's open that can of worms, then. Why is `intptr_t` used instead of `uintptr_t`?

laukik-hase:
Hello, @M-Bab!

Thanks for your contribution.

We will be backporting the fix https://github.com/espressif/esp-idf/commit/b61e3777b803c764abda185436bd48103da6b722 to the branch `release/v4.4` which would close this PR when the fix sync to GitHub.

M-Bab:
Dear @laukik-hase - of course that is just as good as this MR. Thanks and keep up your good work!

laukik-hase:
Closed with 09089c3fdf18a22d211fdf131a16a0fc80627bf1

Issue #12779, opened 2023-12-12
Python path case sensitivity issue on Windows (IDFGH-11666)


dyarkovoy:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.3-dev-888-gcb66d6f6fa 

### Espressif SoC revision.

all

### Operating System used.

Windows

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

CMD

### Development Kit.

all

### Power Supply used.

USB

### What is the expected behavior?

idf.py does not complain about wrong python path when building project

### What is the actual behavior?

When the path of python.exe recorded in project's build config is case-different on a case-insensitive filesystem (Windows), this error occurs:

`'c:\work\esp32\.espressif\python_env\idf5.3_py3.11_env\Scripts\python.exe' is currently active in the environment while the project was configured with 'C:\work\esp32\.espressif\python_env\idf5.3_py3.11_env\Scripts\python.exe'. Run 'idf.py fullclean' to start again.`

the `C:` drive is identical to `c:` drive on Windows, but this is not taken into account

### Steps to reproduce.

1. Build any ESP-IDF project in VSCode with vscode-esp-idf-extension on Windows
2. Try to build the same project from command line using `idf.py build`
you should see the error as described above
...

### More Information.

I will submit PR correcting this behavior shortly

dobairoland:
@jakub-kocka PTAL

jakub-kocka:
@dyarkovoy, thanks for submitting also the [PR](https://github.com/espressif/esp-idf/pull/12780) which solves this.
I haven't been able to reproduce this issue but with the fix everything works as expected.

As you said, according to documentation and my tests the string path is changed only on Windows platform.

PR #12780, opened 2023-12-12
Fix for issue #12779 (IDFGH-11667)

dyarkovoy:
fix(esp32): This should fix python path issue on Windows
This fixes idf.py error

c:\work\esp32\.espressif\python_env\idf5.3_py3.11_env\Scripts\python.exe' is currently active in the environment while the project was configured with 'C:\work\esp32\.espressif\python_env\idf5.3_py3.11_env\Scripts\python.exe'. Run 'idf.py fullclean' to start again

on Windows

### DIFF
--- a/tools/idf_py_actions/tools.py
+++ b/tools/idf_py_actions/tools.py
@@ -626,7 +626,7 @@ def ensure_build_directory(args: 'PropertyDict', prog_name: str, always_run_cmak
 
     try:
         python = cache['PYTHON']
-        if python != sys.executable:
+        if os.path.normcase(python) != os.path.normcase(sys.executable):
             raise FatalError(
                 "'{}' is currently active in the environment while the project was configured with '{}'. "
                 "Run '{} fullclean' to start again.".format(sys.executable, python, prog_name))


dobairoland:
The linked issue is a valid requirement to be able to use the same project in Vscode and terminal as well but I don't think that we should sacrifice case-sensitive comparison on the ESP-IDF side in order to achieve that. Note that this fix is not limited to just the drive letter on Windows and case-insensitive filesystems. Case sensitive Windows, Linux and macOS case difference in all part of the path would be ignored as well.

dyarkovoy:
You're right, @dobairoland, we should not sacrifice case-sensitive comparison on case-sensitive filesystems. But the fix I suggest will only affect case-insensitive filesystem on Windows. If you look at the proposed commit, I use the `os.path.normcase()` Python function. Looking at Python [docs ](https://docs.python.org/3/library/os.path.html#os.path.normcase):
```
os.path.normcase(path)
Normalize the case of a pathname. On Windows, convert all characters in the pathname 
to lowercase, and also convert forward slashes to backward slashes. 
On other operating systems, return the path unchanged.
```

dobairoland:
Thanks @dyarkovoy. My mistake. You are right. Thanks!

jakub-kocka (approved):
Tested that the effect of this change is only on Windows platform and solves the issue https://github.com/espressif/esp-idf/issues/12779. Thanks @dyarkovoy.

dobairoland:
sha=392fc6c740a2065bb6155a859a7c167e2c46331e

Issue #12766, opened 2023-12-10
[ESP32-S3] ulp_adc_init(...) leaks ADC1 handle, preventing ADC1 reinitialization upon wakeup from Light Sleep (IDFGH-11653)


kutukvpavel:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.1.2

### Espressif SoC revision.

ESP32-S3 rev 0.1 (ESP32-S3-WROOM-1 MON4R2, P1N8R8)

### Operating System used.

Linux

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

None

### Development Kit.

Custom board, ESP32-S3-DevKitC-1 v1.6

### Power Supply used.

External 3.3V

### What is the expected behavior?

Let's say we use `ulp_adc_init(...)` API to initialize ADC for use with ULP-RISCV during Light Sleep. It should be possible to reconfigure ADC1 to work with main cores after wakeup from Light Sleep.

### What is the actual behavior?

After `ulp_adc_init()` is executed, it's no longer possible to reclaim ADC1 by any means other than a reset. It happens due to ADC handle leakage inside `ulp_adc_init()`:
```
esp_err_t ulp_adc_init(const ulp_adc_cfg_t *cfg)
{
    esp_err_t ret = ESP_OK;

    ESP_GOTO_ON_FALSE(cfg, ESP_ERR_INVALID_ARG, err, TAG, "cfg == NULL");
    ESP_GOTO_ON_FALSE(cfg->adc_n == ADC_UNIT_1, ESP_ERR_INVALID_ARG, err, TAG, "Only ADC_UNIT_1 is supported for now");

    //-------------ADC1 Init---------------//
    adc_oneshot_unit_handle_t adc1_handle;   <--- THIS

    adc_oneshot_unit_init_cfg_t init_config1 = {
        .unit_id = cfg->adc_n,
        .ulp_mode = cfg->ulp_mode,
    };

    if (init_config1.ulp_mode == ADC_ULP_MODE_DISABLE) {
        /* Default to RISCV for backward compatibility */
        ESP_LOGI(TAG, "No ulp mode specified in cfg struct, default to riscv");
[...]

PR #12785, opened 2023-12-12
Provide ulp_adc_deinit() API to fix ADC1 handle leakage (IDFGH-11673)

kutukvpavel:
Fix for #12766. Also restructured error handling in `ulp_adc_init(...)` so that it returns an error code instead of calling `abort()` from `ESP_ERROR_CHECK(...)`. 
Tested against `ulp_riscv_adc` example (deep sleep case) as well as our own custom app (light sleep case).

P.S. Right now `ulp_adc` driver is remarkably similar to `oneshot` driver itself, it mostly duplicates configuration API, introducing little new functionality. While I think this PR should be merged and backported to v5.1 branch ASAP to enable customers (like us) to use ULP ADC with Light Sleep, it's probably worth integrating ULP-related initialization code into `oneshot` driver itself in future releases. After all, `adc_oneshot_unit_init_cfg_t::ulp_mode` is already there.

### DIFF
--- a/components/ulp/ulp_common/include/ulp_adc.h
+++ b/components/ulp/ulp_common/include/ulp_adc.h
@@ -29,6 +29,13 @@ typedef struct {
  */
 esp_err_t ulp_adc_init(const ulp_adc_cfg_t *cfg);
 
+/**
+ * @brief Deinitialize ADC after use with ULP, allowing it to be reclaimed
+ *
+ * @return esp_err_t    ESP_OK for successful.
+*/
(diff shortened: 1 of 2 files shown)


ESP-Marius:
Thanks for adding this @kutukvpavel!

LGTM, i'll pull it into our internal gitlab for review, test and merge.

ESP-Marius:
sha=8596d46774d65171b4540288918037c43e3debb4

Issue #11830, opened 2023-07-07
SPIFFS use core1/CPU1 and blocks the thread (hangs forever) (IDFGH-10593)


creatormir:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.1

### Operating System used.

Windows

### How did you build your project?

Command line with CMake

### If you are using Windows, please specify command line type.

CMD

### Development Kit.

WROOM32E

### Power Supply used.

USB

### What is the expected behavior?

I expect SPIFFS will not use core1/CPU1

### What is the actual behavior?

Looks like SPIFFS uses interrupts on core1/CPU1 invisibly and this blocks the thread (hangs forever)

### Steps to reproduce.

1. Use the standard example [http_server/file_serving](https://github.com/espressif/esp-idf/blob/5c88c5996dbde6208e3bec05abc21ff6cd822d26/examples/protocols/http_server/file_serving/main/file_server.c)  

3. Add this code:
```
 void Core1_1( void* p) {
    portDISABLE_INTERRUPTS();   
    while (1)
    {
        /* code */
    }
    
}
// In app_main
    TaskHandle_t Task_1;  
    xTaskCreatePinnedToCore(
        Core1_1,       // Function that implements the task.
        "Core1_1",          // Text name for the task.
        STACK_SIZE *1,      // Stack size in bytes, not words.
        ( void * ) 1,    // Parameter passed into the task.
        25 -1, // tskIDLE_PRIORITY+2,
        // tskIDLE_PRIORITY +2, // tskIDLE_PRIORITY+2,
        // configMAX_PRIORITIES -1, // tskIDLE_PRIORITY+2,
        &Task_1,
        1 // Core 1
    );
```

4. After connecting to wifi-ap and trying to load a web page, the system will hang on this line  
[stat(filepath, &file_stat)](https://github.com/espressif/esp-idf/blob/5c88c5996dbde6208e3bec05abc21ff6cd822d26/examples/protocols/http_server/file_serving/main/file_server.c#L377)

igrr:
[...]

PR #12800, opened 2023-12-15
Experimental AMP Mode (IDFGH-11693)

FL0WL0W:
Changes needed to make AMP mode work. 

Because the ESP port needs to run on 2 cores and the FREERTOS on 1 core,  there became a conflict with the current usage of portNUM_PROCESSORS and configNUM_CORES. Because of this, I changed all areas looking at these definitions in relation to FREERTOS to use configNUM_CORES, and all areas of the ESP port to use portNUM_PROCESSORS. This is what i understand to be the correct ussage of these definitions.

The second core entry point is app_main1() and defaults into a delay loop. This code is found in app_startup.c. The check for making sure CONFIG_FREERTOS_UNICORE == CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE was removed since AMP is now experimentally supported.

The esp_ipc was modified in order to use crosscore_int to trigger the second core handling of ipc functions when in AMP mode. The first core still handles ipc functions through the RTOS task. Some code was rearranged to make this cohesive and readable. This consists of the bulk of the changes in this PR.

Some more CONFIG_FREERTOS_UNICORE checks were changed to CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE

I tested this by building my interrupt example and have been able to demonstrate a marked improvement on Interrupt Jitter. It is now within 1us.
https://github.com/FL0WL0W/ESP32InterruptExample/tree/AMP

Relevant issues
#10410 
#11830


mickeyl:
This is amazing. I'd really love to see this option publicly available in ESP-IDF. It would come in handy especially for situations where you have SPI-drivers and an INT line to indicate that _something_ did happen on the peripheral.

Since it's not really advisable to use SPI from an ISR, the traditional way to handle this seems to be a high priority task that does the SPI communication which then gets woken up using direct-to-task-notification like `xTaskNotifyFromISR`. This way though, it can take several hundred µsec until the task actually gets woken up ­– which might be a tad bit slow for high speed peripherals.

With a bare metal loop being woken up by the IRQ (and an assorted bit-banging SPI driver), I would hope to eliminate this delay. The question then is how to communicate the payloads between the FreeRTOS core and the bare metal core.

sudeep-mohanty:
@FL0WL0W Thank you for the contribution! The changes done for updating `portNUM_PROCESSORS` to `configNUM_CORES` LGTM. I shall be pulling it to our internal repository and running CI tests.

FL0WL0W:
Awesome! Let me know if you need anything from me.
Thank you @sudeep-mohanty for the update!

@mickeyl The built-in ESP-IDF driver for SPI should work on the bare metal second core, but I have not tested that. It would be great if you could test it from the bare metal core. You should also be able to notify tasks in FreeRTOS from the second core.

components/esp_system/Kconfig
```diff
@@ -55,10 +55,10 @@ menu "ESP System Settings"
             wait before the device reboots.
 
     config ESP_SYSTEM_SINGLE_CORE_MODE
-        bool
-        default n
+        bool "Only initialize first core."
+        default "y" if FREERTOS_UNICORE
```
igrr, line 59: I'm afraid there is one issue with this approach. If the project is initially configured with `CONFIG_FREERTOS_UNICORE=y`, this option will get set to `y` as well. However if the developer subsequently changes `CONFIG_FREERTOS_UNICORE` to `n` in menuconfig, this option will still be set to `y`. Kconfig doesn't update the options with already defined values if the default value of the option changes.

FL0WL0W: I think as long as the developer has not changed the default value of CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE, It should change to the default.

However, you do bring up a good point. ESP_SYSTEM_SINGLE_CORE_MODE=y and CONFIG_FREERTOS_UNICORE=n is an invalid configuration. Let me see if i can fix this.

FL0WL0W: I added a 
depends FREERTOS_UNICORE
which should fix the invalid configuration

igrr: > I think as long as the developer has not changed the default value of CONFIG_ESP_SYSTEM_SINGLE_CORE_MODE, It should change to the default.

Sadly, this is not how Kconfig works! (Would maybe be nice if it did.)

Thanks for the fix.

sudeep-mohanty (approved):
Hello @FL0WL0W, The changes for `configNUM_CORES` macro replacement LGTM and I shall be pulling it in to our internal repository for tests. Unfortunately though, we won't be able to incorporate the IPC related changes as currently we don't have a mandate of maintaining an AMP compatible IDF.

sudeep-mohanty:
sha=9cdb09c5e990ae45e0af1eaf887d3fa1a0d2dc16

FL0WL0W:
@sudeep-mohanty I guess having some of my changes pulled is better than nothing. It is still unfortunate my other changes to the ipc will not get pulled. I will open a new pull request for some other changes that can hopefully be pulled in.

sudeep-mohanty:
Hi @FL0WL0W,
After some more deliberations, we've decided to introduce a hidden Kconfig option, viz., `CONFIG_FREERTOS_NUMBER_OF_CORES` which should control the number of cores on which FreeRTOS is running. The intention is to use this new macro to iterate over SMP entities., instead of using `configNUM_CORES`. At the moment, this option will be tied to the value of `CONFIG_FREERTOS_UNICORE` but it will be made the default option for selecting the number of cores for FreeRTOS by the next major ESP-IDF release. This should not affect your efforts to enable AMP with ESP-IDF.

mickeyl:
@sudeep-mohanty That's very good news, thanks Espressif for reconsidering!

sudeep-mohanty:
PR updated and merged at 9605f9be3fc62607c69edbe37aeb3e7ba4b0c908. Closing this PR.

FL0WL0W:
@mickeyl careful, they did not add the AMP functionality. They only updated their C definitions to match closer to what they should be. All the changes needed to the cpu_start and the ipc_task were not accepted.

@sudeep-mohanty I will look over the changes and make another PR if I see anything wrong. Thanks.

mickeyl:
@FL0WL0W Oops, thanks for the warning. I'll stick with your fork then. The smaller the diff is the better though… and perhaps you can get more upstream eventually.

PR #12811, opened 2023-12-16
improve: add ethernet deinit to example (IDFGH-11704)

elton-a-vieira:
On the idf 5.1, the Ethernet driver has an internal counter that will be initialized at __esp_eth_driver_install__ and will be incremented at __esp_eth_new_netif_glue__.

To perform a successful ethernet deinit the counter must be cleared and it is done by __esp_eth_del_netif_glue__ which decrements the counter dealing to successful __esp_eth_driver_uninstall__.

This procedure was unclear in the previous code sample.


kostaond:
@elton-a-vieira thank you very much for your contribution! Unfortunately, the actual solution of described issue requires more complex update. The update will be provided soon. Please see https://github.com/espressif/esp-idf/issues/13443 for additional information.

PR #12852, opened 2023-12-21
cmake: correct the parameter annotation of idf_build_set_property (IDFGH-11752)

Augtons:
The value parameter should be marked as [in] instead of [out], as it is used to set the property value, not to get it.


igrr (approved):

igrr:
sha=a8ea638cdeeca22fffa917feab397c788a1d36d0

PR #12863, opened 2023-12-23
doc: qemu: link to qemu build instructions

KJ7LNW:
The build instructions for Espessif's qemu build are available but were not linked from this documentation.  Add the link and extend the explanation about building QEMU for Xtensa and esp32c3.

### DIFF
--- a/docs/en/api-guides/tools/qemu.rst
+++ b/docs/en/api-guides/tools/qemu.rst
@@ -16,7 +16,7 @@ To use QEMU with ``idf.py`` you first need to install the above-mentioned fork o
 
 After installing QEMU, make sure it is added to PATH by running ``. ./export.sh`` in the IDF directory.
 
-If you are using a different platform, you need to build QEMU from source. Refer to official QEMU documentation for instructions.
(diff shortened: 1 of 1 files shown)


KJ7LNW:
See also: https://github.com/espressif/esp-toolchain-docs/pull/4

docs/en/api-guides/tools/qemu.rst
```diff
@@ -16,7 +16,7 @@ To use QEMU with ``idf.py`` you first need to install the above-mentioned fork o
 
 After installing QEMU, make sure it is added to PATH by running ``. ./export.sh`` in the IDF directory.
 
-If you are using a different platform, you need to build QEMU from source. Refer to official QEMU documentation for instructions.
+If you are using a different platform, you need to build QEMU from source `for the Xtensa and RISC-V targets <https://github.com/espressif/esp-toolchain-docs/blob/main/qemu/esp32/README.md>`_ using the `Espressif QEMU work-in-progress fork <https://github.com/espressif/qemu>`_ containing ``xtensa-softmmu`` and ``riscv32-softmmu`` support.  Ultimately, the Xtensa and ``esp32c3`` RISC-V targets are intended to be mainlined into QEMU upstream. Refer to official QEMU documentation for additional build instructions.
```
igrr, line 19: Could you please make the reference to the architecture conditional on the current chip target? Otherwise it would be odd to mention Xtensa when building docs for C3, and vice versa. ([Docs](https://docs.espressif.com/projects/esp-docs/en/latest/writing-documentation/writing-for-multiple-targets.html#writing-for-multiple-targets))

(Or just let me know if you don't want to deal with this, I can fix up your commit before merging it.)

igrr (changes requested):
Thanks for the fix, just one extra request.

docs/en/api-guides/tools/qemu.rst
KJ7LNW: > (Or just let me know if you don't want to deal with this, I can fix up your commit before merging it.)

Thanks, please do!

PR #12895, opened 2023-12-29
feat: Add support for ESP-IDF versions below 5.2 (IDFGH-11800)

Augtons:
In ESP-IDF 5.2, **littlefs** became a valid [sub-type](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/partition-tables.html#subtype
) for the partition table, with the value of 0x83, and this example uses it. However, before that, this sub-type was not recognized by default. Therefore, I updated the `CMakeLists.txt` file to set it as a custom sub-type for ESP-IDF versions lower than 5.2, in order to make this example compatible with older ESP-IDF versions.

### DIFF
--- a/examples/storage/littlefs/CMakeLists.txt
+++ b/examples/storage/littlefs/CMakeLists.txt
@@ -1,6 +1,17 @@
 # The following lines of boilerplate have to be in your project's
 # CMakeLists in this exact order for cmake to work correctly
-cmake_minimum_required(VERSION 3.5)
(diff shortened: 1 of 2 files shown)


Augtons:
It looks like some checks didn't pass, I'm trying to install pre-commit and I'll start PR again when everything goes well!

Augtons:
It looks like some checks didn't pass, I'm trying to install pre-commit and I'll start PR again when everything goes well!

PR #12910, opened 2024-01-02
Renames wpa2_enterprise in README.md. (IDFGH-11816)

aadsm:
Renames `wpa2_enterprise` to `wifi_enterprise`.


aadsm:
Oh never mind. I don't think I can sign your CLA because of clause _"3.1 Patent License"_. Your CLA's scope is larger than just my _"Contribution"_, it's _"the Contribution and the Contribution in combination with the Material (and portions of such combination)"_. It's tricky because as stated in the CLA: _'"Material" means the software or documentation made available by Us to third parties.'_.
This leads me to believe that if in the future any part of your _"Material"_ (and not just my _"Contribution"_) infringes on my _"patents owned or controlled by [me], whether already acquired or hereafter acquired"_ I would have no means to protect myself.
If this is not the case could you clarify please?

Not a big deal though, as this is a super simple change anyone can implement in a couple of seconds.

kapilkedawat:
this change is not needed in latest master anymore.

PR #12944, opened 2024-01-07
Some improvement to adc_continuous and i2s_common

mcapdeville:
Hi,
   This is somme proposed little improvement to adc and i2s driver.

* In adc continuous driver, it(s not always a good idea to use ringbuffer. to avoid unnecessary copying of data. The first patch allow disabling ringbuff by setting max_store_buf_sieze member of adc_continuous_handle_cfg_t structure to 0.

* The second patch resolve some rounding problem in adc_hal_digi_sample_freq_config

* The third patch is for 
    - clearing i2s transmit dma buffer _before_ calling on_sent callback
    - directly setting the dma buffer address in event.data of callbacks instead of that of the finish_desc->buf member
 
--- 
Marc


components/hal/adc_hal.c
```diff
@@ -10,6 +10,7 @@
 #include "hal/assert.h"
 #include "soc/lldesc.h"
 #include "soc/soc_caps.h"
+#include "esp_log.h"
```
suda-morris, line 13: we don't allow to use "esp_log.h" in the hal component, instead, we suggest using "hal/log.h" in the hal.

L-KAYA:
Hi @mcapdeville , thanks for your contribution! I have one question, why we should clear the i2s tx buffer before callback? Some use cases would like to read back these sent data to check the data latency, clearing before callback will break these cases.

Another concern is that the pointer change of `event.data` might lead to breaking change, do you think this change is critical?

mcapdeville:
Hi,

> Hi @mcapdeville , thanks for your contribution! I have one question, why we should clear the i2s tx buffer before callback? Some use cases would like to read back these sent data to check the data latency, clearing before callback will break these cases.

In my application, I directly fill dma buffers in the callback function. I 'm trying to avoid unnecessary recopy of data by passing the buffer's pointer to the encoding function. If ther is no data to be sent, the output datas are allready zeroed.
- Zeroing the buffer after the callback will erase the encoded datas.
- The i2s_channel_write function add an unnecessary copy of data.

> 
> Another concern is that the pointer change of `event.data` might lead to breaking change, do you think this change is critical?

I think, it is more practical the the event struct contain the pointer to the dma buffer and size of the buffer instead of a pointer on the pointer to the dma buffer.

Regards.

--- 
Marc

L-KAYA:
sha=5e6a1e5aee8a446670f291d048ae24abf491f456

suda-morris:
sha=5e6a1e5aee8a446670f291d048ae24abf491f456

L-KAYA:
Hi @mcapdeville, the sequence of buffer clear and callback may need more discussion. 

If you want to copy data into the TX DMA buffer directly in the callback, you can just disable the `auto_clear` and operate the buffer in the `on_sent` callback. Or do you have any reason that `auto_clear` must be set?

As for the DMA buffer, indeed the first level pointer would be more practical, this should be worth a  breaking change

masterxq:
Just a user story:
I have very likely changes in my espressif, and I'm patching them in my pipelines to avoid extra copy in i2s... This is so unnecessary... I understand why this can be needed, but I really don't understand why this is the only possible way :/

In my implementation I have an additional function that's avoid copying and the Original is unpatches...

This could be a way around possible breaking changes, learn more user stories and avoid that users needs to implement it again and again...

L-KAYA:
I2S part has been merged in commit [fcd1f1c8](https://github.com/espressif/esp-idf/commit/fcd1f1c8086f5b75b1cac1a01ba8db9326c4003f) together with some refactor in the commit [fd27cef0](https://github.com/espressif/esp-idf/commit/fd27cef0451b6480247949c54db89a44a52ebbf7)

Issue #12957, opened 2024-01-10
uart deadlock while using automatic light sleep (IDFGH-11870)


tshcherban:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.3-dev-1353-gb3f7e2c8a4
v5.2-beta1-263-ge49823f10c
v4.4.6-374-gbb8dd9d35b

### Espressif SoC revision.

ESP32-S3 (QFN56) (rev v0.1)
ESP32-D0WD-V3 (rev v3.1)
ESP32-C3 (QFN32) (rev v0.3)

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

PowerShell

### Development Kit.

ESP32-S3/C3 DevKits
Wemos D1-Mini ESP32
ESP32/S3 custom boards

### Power Supply used.

External 3.3V, USB (irrelevant)

### What is the expected behavior?

Uart is working with automatic light sleep enabled

### What is the actual behavior?

Uart locks indefinitely with automatic light sleep enabled

### Steps to reproduce.

app main:

```
#include "driver/uart.h"
#include "esp_pm.h"
#include "esp_timer.h"
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "hal/uart_types.h"

const uart_port_t uart_nr = (uart_port_t)0;
const int rx_buffer_size = 256;
const int tx_buffer_size = 0;
const int queue_size = 20;
const int baud_rate = 115200;
const int rx_flow_ctrl_thresh = 112; // value taken from arduino core

QueueHandle_t uart_event_queue;

extern "C" void app_main()
{
    uart_config_t uart_config;
    uart_config.data_bits = uart_word_length_t::UART_DATA_8_BITS;
    uart_config.parity = uart_parity_t::UART_PARITY_DISABLE;
    uart_config.stop_bits = uart_stop_bits_t::UART_STOP_BITS_1;
    uart_config.flow_ctrl = uart_hw_flowcontrol_t::UART_HW_FLOWCTRL_DISABLE;
    uart_config.rx_flow_ctrl_thresh = rx_flow_ctrl_thresh;
[...]

PR #12960, opened 2024-01-11
fix(esp32): avoid repetitive uart semaphore take (IDFGH-11874)

tshcherban:
Avoid multiple xSemaphoreTake. Fixes [IDFGH-11870 (#12957)](https://github.com/espressif/esp-idf/issues/12957)


tshcherban:
Should i create a simmilar PRs for other active branches?

PR #12963, opened 2024-01-11
fix misformatted link (IDFGH-11877)

kisvegabor:
As the title says


suda-morris:
sha=d30b78aea739b50a439d9f667ecfa9fb2e17ed16

PR #12972, opened 2024-01-14
Fix APB_CLK_FREQ value for C6 (IDFGH-11886)

majkrzak:
This PR changes previously wrong value of `APB_CLK_FREQ`.  It solves the issue with RMT LED driver where pulse with where incorrectly calculated using 40 MHz instead of 80 MHz clock base.


suda-morris:
Hi @majkrzak 

The APB clock frequency is true to be 40MHz, but just on esp32c6, the APB clock can not work as a function clock for the RMT. Instead, the supported clock sources are listed [here](https://github.com/espressif/esp-idf/blob/master/components/soc/esp32c6/include/soc/clk_tree_defs.h#L187) It has a PLL clock source whose frequency is 80MHz, but it's not APB anymore.

If you wants to write some code to work cross the esp chips, you can set the RMT clock source to `RMT_CLK_SRC_DEFAULT`. Like [this](https://github.com/espressif/idf-extra-components/blob/master/led_strip/examples/led_strip_rmt_ws2812/main/led_strip_rmt_ws2812_main.c#L38)

majkrzak:
Ok, this mean ESP-idf documentation is wrong. It claims RMT is clocked with APB clock with default frequency of 80MHz.

PR #13010, opened 2024-01-18
Update documenting-code.rst (IDFGH-11930)

callpri:
Codelobster editor is added


ESP-Marius:
@callpri Thanks for your contribution, but release/v4.1 is already out of our [service period](https://github.com/espressif/esp-idf?tab=readme-ov-file#esp-idf-release-support-schedule) so we wont be backporting anymore fixes to this branch.

And I think this list is just meant as a suggestion, so we do not need to list all possible options here.

PR #13020, opened 2024-01-20
fix (esp_lcd): Don't assume panels are 16bit in VSYNC restart logic (IDFGH-11941)

bwhitman:
I'm using RGB332 for my RGB panel. 8 bits a pixel.

I tried using `CONFIG_LCD_RGB_RESTART_IN_VSYNC` and found that it permanently corrupted (shifted) the screen, instead of fixing shifts.

I found that `esp_lcd` assumes a 16-bit display during the transmission logic and doesn't check `fb_bits_per_pixel`.

This PR is a fix for that error so that restarting in VSYNC doesn't permanently shift the screen.

I also fixed what I believe to be a smaller error in prefilling the "second" bounce buffer `[1]`, was `[0]`.


suda-morris:
sha=25fba0a783701263bf1c576aac2c089fddddcc66

suda-morris:
merged in commit [ff840b6](https://github.com/espressif/esp-idf/commit/ff840b645d0c637771a8f44d70c167efb2495f63)

PR #13025, opened 2024-01-21
Missing include for rom/uart.h (IDFGH-11947)

ricardoquesada:
Without this include it fails to compile with this error:

```
esp/esp-idf-v5.2/components/esp_rom/include/esp32/rom/uart.h:262:32: error: implicit declaration of function 'UART_STATUS_REG' [-Werror=implicit-function-declaration]
  262 |         status = READ_PERI_REG(UART_STATUS_REG(uart_no));
```


suda-morris:
sha=c5251861dfa4e207398db3d22a84fd56d4e6f96b

PR #13039, opened 2024-01-23
Handle object files with `.*.o` patterns when running Linker Script Generator (IDFGH-11962)

valeros:
Currently, only `.o`, `.*.obj` and `.obj` patterns are taken into account. It would be great to have object files with the `.*.o` extension pattern (e.g. `file.cpp.o`) also processed as they're quite widespread in third-party integrations.


0xjakob:
sha=bd56ca48be81763872d1562822b94843eed12fdc

0xjakob:
@valeros We plan to merge this, thanks for the fix! After merging this internally, it'll appear on master, soon.

PR #13088, opened 2024-02-01
Add filename, linenumber and function name to log statements (IDFGH-12023)

0xFEEDC0DE64:
This merge request adds filenames, linenumbers and function names to all log statements.

Very often I am in a situation where a esp component logs some sort of warning or error and I need to check the exact conditions how they can appear, and I would like to jump to that exact code line, where this log statement occurs.
The old if logging style only contains the tag and the message, which is just not enough to pinpoint the log line for me.

In the past I have gotten the feedback that the binary size might increase significantly. If you want I will provide another commit that only turns on this additional log functionality when its not inside the bootloader + a Kconfig parameter. Let me know what you think.

### DIFF
--- a/components/app_trace/include/esp_app_trace_util.h
+++ b/components/app_trace/include/esp_app_trace_util.h
@@ -174,7 +174,7 @@ void esp_apptrace_log_unlock(void);
 #define ESP_APPTRACE_LOG_LEV( _L_, level, format, ... )   \
     do { \
         if (LOG_LOCAL_LEVEL >= level) { \
-            ESP_APPTRACE_LOG(LOG_FORMAT(_L_, format), esp_log_early_timestamp(), TAG, ##__VA_ARGS__); \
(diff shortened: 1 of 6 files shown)


jrahlf:
It is not just binary size. Currently all logging is synchronous which means the extra information makes logging slower and in turn has performance impacts on all firmwares.

KaeLL:
I wrote a variant of this for a project at work, but didn't even consider upstreaming because @jrahlf is right.

KonstantinKondrashov:
Hi! I am working on esp log v2 which will make the log more flexible and functional.
This particular feature I already implemented and will be available under a Kconfig option, keeping the general use case the same.
We gathered all requests for log features, and this log v2 improvement will cover all or at least most of them.
I have it in our internal repo, and this Q1 it will be gradually merged.

Sorry, we will not merge this PR, thank you for your contribution.

PR #13103, opened 2024-02-04
Remove validation of pins for LCD output. (IDFGH-12038)

Hello1024:
There are a bunch of cases you might want some pins not exposed.

Eg.

  * Reading say 8 bit data and outputting the top 5 bits, discarding the rest by not mapping those data pins to output pins

 * Not using hsync/vsync because sync data is embedded within the data bits for more timing flexibility (eg. interlacing).

 * Using the LCD module as a high speed parallel data output bus, with no need for sync/control pins.

Removing this validation makes these cases work.


suda-morris:
sha=a2e0b9891fe23a31aa364a74382c6ce09381e1c6

PR #13114, opened 2024-02-05
fix(esp32s3): Add config value for RTC_XTAL_CAL_RETRY for S3 hardware variant

DaveLegg:
Adding support for the RTC_XTAL_CAL_RETRY config setting to the ESP32S3.

This config setting is already supported on other chip variants. We have run into an issue recently with some production hardware where an additional retry was needed for the external crystal to be detected during a cold boot. Copying this behaviour from other chip variants has resolved the issue for us.

This change is such that it will not impact existing behaviour for users of the S3 variant, unless they specifically enable the config value


songruo:
sha=a8b4627705ee5814bedf0065f51373dd46e297ba

PR #13132, opened 2024-02-07
[SECURITY] [docs] Add warning about integrity of pre_encrypted_ota (IDFGH-12071)

redfast00:
Using the Encrypted Binary OTA provides confidentiality of the firmware, but does not provide integrity of the firmware: if an attacker is able to read out the firmware of one device (for example, via a fault-injection attack), they can also construct arbitrary valid encrypted OTA updates, that can then be installed on other devices that share the same private key.

This was not immediately obvious to me when I read what the pre_encrypted_ota functionality does, but became clear when I reviewed the cryptography used.

### DIFF
--- a/examples/system/ota/pre_encrypted_ota/README.md
+++ b/examples/system/ota/pre_encrypted_ota/README.md
@@ -9,6 +9,9 @@ Pre-encrypted firmware binary must be hosted on OTA update server.
 This firmware will be fetched and then decrypted on device before being flashed.
 This allows firmware to remain `confidential` on the OTA update channel irrespective of underlying transport (e.g., non-TLS).
 
+> [!CAUTION]
+> Using the Pre-encrypted Binary OTA provides confidentiality of the firmware, but it does not ensure authenticity of the firmware. For ensuring that the firmware is coming from trusted source, please consider enabling secure boot feature along with the Pre-encrypted binary OTA. Please refer to security guide in the ESP-IDF docs for more details.
+
 ## ESP Encrypted Image Abstraction Layer
 
 This example uses `esp_encrypted_img` component hosted at [idf-extra-components/esp_encrypted_img](https://github.com/espressif/idf-extra-components/blob/master/esp_encrypted_img) and available though the [IDF component manager](https://components.espressif.com/component/espressif/esp_encrypted_img).


mahavirj:
@redfast00 

> Using the Encrypted Binary OTA provides confidentiality of the firmware, but does not provide integrity of the firmware:

Do you mean authenticity of the firmware here? Because for integrity part, we already have several checks enabled during boot-up and OTA update stage.

> if an attacker is able to read out the firmware of one device (for example, via a fault-injection attack), they can also construct arbitrary valid encrypted OTA updates, that can then be installed on other devices that share the same private key.

It is recommended in the example README to protect the pre-encrypted OTA specific private key on the device. So I am not sure if I quite understand this point here, could you please elaborate? 

> This was not immediately obvious to me when I read what the pre_encrypted_ota functionality does, but became clear when I reviewed the cryptography used.

Secure boot and pre-encrypted OTA are 2 different features serving their own purposes. Ideally, we recommend that users refer to overall [security guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/security/security.html) while deciding the features that best suit their use-cases.

redfast00:
@mahavirj I indeed mean the authenticity (as in, 'can an attacker create a modified firmware update file that seems valid').

It is indeed recommended to protect the private key using Flash Encryption. However, on (I think currently all?) ESP32 chips, this flash encryption can be broken with fault injection attacks (see https://www.espressif.com/sites/default/files/advisory_downloads/AR2022-003%20Security%20Advisory%20Concerning%20Breaking%20the%20Hardware%20AES%20Core%20and%20Firmware%20Encryption%20of%20ESP32%20Chip%20Revision%20v3.0%20-%20V2.0%20EN.pdf): so an attacker can read out the private key on one device, and then use that private key to create updates for other devices that share the same keys.

Secure boot and pre-encrypted OTA are indeed two different things, so I don't think it hurts to link to secure boot in the README of pre-encrypted OTA.

examples/system/ota/pre_encrypted_ota/README.md
```diff
@@ -9,6 +9,9 @@ Pre-encrypted firmware binary must be hosted on OTA update server.
 This firmware will be fetched and then decrypted on device before being flashed.
 This allows firmware to remain `confidential` on the OTA update channel irrespective of underlying transport (e.g., non-TLS).
 
+> [!CAUTION]
+> Using the Encrypted Binary OTA provides confidentiality of the firmware, but does not provide integrity of the firmware: if an attacker is able to read out the firmware of one device, they can also construct arbitrary encrypted OTA updates. To defend against this, you'll additionally need to use [ESP32 secure boot V2](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/security/secure-boot-v2.html).
+
```
mahavirj, line 14:
```suggestion
> [!CAUTION]
> Using the Pre-encrypted Binary OTA provides confidentiality of the firmware, but it does not ensure authenticity of the firmware. For ensuring that the firmware is coming from trusted source, please consider enabling secure boot feature along with the Pre-encrypted binary OTA. Please refer to security guide in the ESP-IDF docs for more details.

```

mahavirj (changes requested):

mahavirj:
Changes, LGTM!

@redfast00 Could you please squash the commits?

redfast00:
@mahavirj done!

mahavirj (approved):

mahavirj:
sha=836516b4099deab22e9938d3660943884feef129

Issue #13136, opened 2024-02-08
i2c_master: Support I2C transaction failure notification because the new ESP32 I2C driver always returns ESP_OK (IDFGH-12075)


KJ7LNW:
### Is your feature request related to a problem?

Originally posted here: [Why does i2c_master_transmit_receive return ESP_OK when the device is unplugged?](https://esp32.com/viewtopic.php?f=13&t=38283)

I can successfully read from an I2C device (its an RTC clock) and it works. However, if the device is disconnected from the bus, the `i2c_master_transmit_receive()` function always returns `ESP_OK`.

Digging through the code, it appears that synchronous requests will set (or rather, pass-through) any of these in `i2c_master_bus_t->status`:
- I2C_STATUS_ACK_ERROR
- I2C_STATUS_TIMEOUT
- I2C_STATUS_DONE

as shown in [esp_driver_i2c/i2c_master.c](https://github.com/espressif/esp-idf/blob/master/components/esp_driver_i2c/i2c_master.c#L439):
```c
    if (i2c_master->status != I2C_STATUS_ACK_ERROR && i2c_master->status != I2C_STATUS_TIMEOUT) {
        atomic_store(&i2c_master->status, I2C_STATUS_DONE);
    }
```

### Describe the solution you'd like.

Provide an API accessor, something like this:

```c
i2c_master_status_t i2c_master_get_previous_io_status(i2c_master_bus_handle_t i2c_master)
{
  return i2c_master->status;
}
```

and expose [these enums](https://github.com/espressif/esp-idf/blob/master/components/esp_driver_i2c/include/driver/i2c_types.h#L26) publically (if they are not already):

```c
 */
typedef enum {
    I2C_STATUS_READ,      /*!< read status for current master command */
    I2C_STATUS_WRITE,     /*!< write status for current master command */
    I2C_STATUS_START,     /*!< Start status for current master command */
    I2C_STATUS_STOP,      /*!< stop status for current master command */
    I2C_STATUS_IDLE,      /*!< idle status for current master command */
    I2C_STATUS_ACK_ERROR, /*!< ack error status for current master command */
    I2C_STATUS_DONE,      /*!< I2C command done */
    I2C_STATUS_TIMEOUT,   /*!< I2C bus status error, and operation timeout */
} i2c_master_status_t;
```

### Describe alternatives you've considered.

[...]

PR #13142, opened 2024-02-08
i2c_master handle NACK (IDFGH-12083)

### LINKED ISSUE #13134: I²C NACK on write/address byte does not error using i2c_master (IDFGH-12073)
<sup>keywords: I²C NACK no error, i2c_master, i2c master, new driver<sup>

IDF version: 5454d37d496a8c58542eb450467471404c606501 (current master)

## Problem/Current behavior

When using the new I²C driver (`i2c_master`) functions, a NACK is perceived as an ACK on both slave
addresses and write bytes.

When the address byte results in a NACK (no device with this address is on the bus), the driver
continues as if an ACK was received.  
When a write byte results in a NACK, the driver continues as if an ACK was received.

This affects the following functions:
- `i2c_master_receive`
  - slave address ACK not verified
- `i2c_master_transmit`
  - slave address ACK not verified
  - write byte ACK not verified
- `i2c_master_transmit_receive`
  - slave address ACK not verified
  - write byte ACK not verified

A non-exhaustive list of things that could go wrong because of this:
- `i2c_master_transmit` on a missing slave appears to succeed.
- `i2c_master_transmit` on an existing but busy slave appears to succeed.
- `i2c_master_receive` and `i2c_master_transmit_receive` on a missing slave results in _fake_ `FF` bytes which is not distinguishable from an existing slave and valid `FF` bytes.

This is a regression from the old driver which verifies both the slave address ACK and the write byte's ACK, e.g.:
- `i2c_master_write_read_device`, address byte
  https://github.com/espressif/esp-idf/blob/5454d37d496a8c58542eb450467471404c606501/components/driver/i2c/i2c.c#L1077
- `i2c_master_write_read_device`, registry address byte
  https://github.com/espressif/esp-idf/blob/5454d37d496a8c58542eb450467471404c606501/components/driver/i2c/i2c.c#L1082

I²C _Read Byte_ example using `i2c_master_transmit_receive` resulting in `ESP_OK` and `FF` output.
i2C-nack-on-write-due-to-missing-slave

<details>
<summary>Code that generated this signal</summary>

```c
i2c_device_config_t dev_cfg = {
    .dev_addr_length = I2C_ADDR_BIT_LEN_7,
    .device_address = 0x2F,
    .scl_speed_hz = 50000,
};

i2c_master_dev_handle_t dev_handle;
ESP_ERROR_CHECK(i2c_master_bus_add_device(bus_handle, &dev_cfg, &dev_handle));

uint8_t readRegister = 0x20;
uint8_t receive[1]; // Will be 0xFF

 // Result will be 0, should be non-zero, e.g. 259
esp_err_t result = i2c_master_transmit_receive(dev_handle, &readRegister, 1, receive, 1, -1);
```

</details>

## Expected behavior
To be compliant with I²C and/or SMBus, the controller upon detection of the NACK condition must generate a STOP condition to abort the transfer.

## Solution
- Enable ACK verification on every write byte.
- Enable ACK verification on the address byte.

I will be making a PR that attempts to address both.

MatthiasKunnen:
Closes #13134 

In summary, this PR:
- Resolves a deadlock on `s_i2c_synchronous_transaction` failure
- Handles NACK received on the slave address when using `i2c_master_receive`
- Handles NACK received on the slave address and write bytes when using `i2c_master_transmit` and `i2c_master_transmit_receive`

#### `i2c_master_probe` reports incorrect _NOT FOUND_ if previous transaction failed
I noticed that after dealing with an unexpected NACK (on slave address or write byte) on `i2c_master_transmit`, the `i2c_master_probe` function will report NOT_FOUND on an existing address. This is seems to be due to some missing cleanup or resetting of initial state when starting `i2c_master_probe`. 

I was planning on continuing to develop this PR until the probe issue was resolved and was looking into writing tests. Since then, @mythbuster5 has voiced that they would add NACK checking in https://github.com/espressif/esp-idf/issues/13136#issuecomment-1933340595. Wanting to prevent duplication of effort, I'm not sure if I should continue further development on this PR. 

I look forward to feedback, both on the current approach and if/how to continue.

### DIFF
--- a/components/esp_driver_i2c/i2c_master.c
+++ b/components/esp_driver_i2c/i2c_master.c
@@ -974,7 +974,7 @@ esp_err_t i2c_master_transmit(i2c_master_dev_handle_t i2c_dev, const uint8_t *wr
 
     i2c_operation_t i2c_ops[] = {
         {.hw_cmd = I2C_TRANS_START_COMMAND},
-        {.hw_cmd = I2C_TRANS_WRITE_COMMAND(false), .data = (uint8_t *)write_buffer, .total_bytes = write_size},
+        {.hw_cmd = I2C_TRANS_WRITE_COMMAND, .data = (uint8_t *)write_buffer, .total_bytes = write_size},
         {.hw_cmd = I2C_TRANS_STOP_COMMAND},
     };
 
@@ -994,7 +994,7 @@ esp_err_t i2c_master_transmit_receive(i2c_master_dev_handle_t i2c_dev, const uin
 
     i2c_operation_t i2c_ops[] = {
         {.hw_cmd = I2C_TRANS_START_COMMAND},
-        {.hw_cmd = I2C_TRANS_WRITE_COMMAND(false), .data = (uint8_t *)write_buffer, .total_bytes = write_size},
+        {.hw_cmd = I2C_TRANS_WRITE_COMMAND, .data = (uint8_t *)write_buffer, .total_bytes = write_size},
         {.hw_cmd = I2C_TRANS_START_COMMAND},
         {.hw_cmd = I2C_TRANS_READ_COMMAND(ACK_VAL), .data = read_buffer, .total_bytes = read_size - 1},
         {.hw_cmd = I2C_TRANS_READ_COMMAND(NACK_VAL), .data = (read_buffer + read_size - 1), .total_bytes = 1},
--- a/components/esp_driver_i2c/i2c_master.c
+++ b/components/esp_driver_i2c/i2c_master.c
@@ -776,6 +776,7 @@ static esp_err_t s_i2c_asynchronous_transaction(i2c_master_dev_handle_t i2c_dev,
 
 static esp_err_t s_i2c_synchronous_transaction(i2c_master_dev_handle_t i2c_dev, i2c_operation_t *i2c_ops, size_t ops_dim, int timeout_ms)
 {
+    esp_err_t ret = ESP_OK;
     i2c_dev->master_bus->trans_done = false;
     TickType_t ticks_to_wait = (timeout_ms == -1) ? portMAX_DELAY : pdMS_TO_TICKS(timeout_ms);
     if (xSemaphoreTake(i2c_dev->master_bus->bus_lock_mux, ticks_to_wait) != pdTRUE) {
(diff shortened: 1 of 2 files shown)


KJ7LNW:
Hi @MatthiasKunnen, thanks for writing up this PR!

I just tested this PR with the i2c **a**sync callback as follows, and it doesn't seem to work as expected.  First, this is how we tested _without_ the PR applied:

**Without the PR**
1. Start querying an i2c RTC clock, it works.
2. Register an async i2c callback with `i2c_master_register_event_callbacks`
3. Unplug the clock from the i2c bus: the data returned becomes all **0xFF**'s, which is expected.
4. The callback from above never provides I2C_EVENT_NACK, it always claims I2C_EVENT_DONE.
5. Plug the i2c clock back into the bus (hot), and it continues to update: The 0xFF's return to proper time data.

**With the PR**

This is the same procedure with this PR applied:

1. Start querying an i2c RTC clock, it works.
2. Register an async i2c callback with `i2c_master_register_event_callbacks`
3. Unplug the clock from the i2c bus: the data returned becomes all **0x00**'s, which may be expected.
4. The callback from above never provides I2C_EVENT_NACK, it always claims I2C_EVENT_DONE.
5. Plug the i2c clock back into the bus (hot), and it no longer updates.  
       - N.B.: I know i2c hot-plug is not generally supported, but _without_ this PR the i2c bus _does_ recover from the disconnect/reconnect.

**TL;DR**
This PR doesn't fix NACK detection for async (I've not tested sync), and it introduces a hang when the bus is compromised.

Do you know what may have caused the hang, and how might the NACK detection be fixed for asynchronous requests?

This is our callback:

```c
bool esp32_i2c_dev_callback(i2c_master_dev_handle_t i2c_dev, const i2c_master_event_data_t *evt_data, void *arg)
{
       // i2c_req_t is our own i2c request tracking structure.  For the purpose of this example, 
       // its just a way of passing `evt_data->event` outside of the ISR:
       i2c_req_t *req = arg;

       req->esp_callback_event = evt_data->event;
       if (evt_data->event == I2C_EVENT_DONE)
               req->status = i2cTransferDone;
       else if (evt_data->event == I2C_EVENT_NACK) // <- never happens
               req->status = i2cTransferError;
       else
               req->status = -evt_data->event;

       return false;
}
```

MatthiasKunnen:
Since I read that the async API was still experimental, I hadn't tested it before pausing development.

I'll have a look and see what can be done.

KJ7LNW:
> I'll have a look and see what can be done.

Sounds good.  Here is a simple test routine that implements both sync and async side-by-side if it helps you with testing:

https://github.com/KJ7LNW/esp32-i2c-test

mythbuster5:
I merged https://github.com/espressif/esp-idf/commit/016877b7ce184e8e8e0cdb037c230a887bc8a5dc, so this is gonna fixed. Thanks for your contribution, I will close this one.

PR #13180, opened 2024-02-14
ulp_riscv_print: Add convenience function that supports different widths (IDFGH-12124)

mickeyl:
ULP's `print_hex` is pretty handy, but hardcoded to `uint32` with 8 digits. This patch adds a convenience function that supports different widths.

### DIFF
--- a/components/ulp/ulp_riscv/ulp_core/include/ulp_riscv_print.h
+++ b/components/ulp/ulp_riscv/ulp_core/include/ulp_riscv_print.h
@@ -38,6 +38,14 @@ void ulp_riscv_print_str(const char *str);
  */
 void ulp_riscv_print_hex(int h);
 
+/**
+ * @brief Prints a hex number with the specified number of digits. Does not print 0x, only the digits
+ *
+ * @param Hex number to print
+ * @param number_of_digits Number of digits to print.
+ */
+void ulp_riscv_print_hex_with_number_of_digits(int h, int number_of_digits);
+
 #ifdef __cplusplus
 }
 #endif
--- a/components/ulp/ulp_riscv/ulp_core/ulp_riscv_print.c
+++ b/components/ulp/ulp_riscv/ulp_core/ulp_riscv_print.c
@@ -49,3 +49,33 @@ void ulp_riscv_print_hex(int h)
         h <<= 4; // move the 2nd leftmost byte to the left, to be extracted next
     }
 }
+
+void ulp_riscv_print_hex_with_number_of_digits(int h, int number_of_digits)
+{
+    int x;
+    int c;
+
+    if (!s_print_ctx.putc) {
+        return;
+    }
+
+    if (number_of_digits < 1) {
+        return;
+    }
+
+    if (number_of_digits >= 8) {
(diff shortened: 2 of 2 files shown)


components/ulp/ulp_riscv/ulp_core/include/ulp_riscv_print.h
```diff
@@ -37,6 +37,7 @@ void ulp_riscv_print_str(const char *str);
  * @param Hex number to print
  */
 void ulp_riscv_print_hex(int h);
+void ulp_riscv_print_hex_with_number_of_digits(int h, int number_of_digits);
```
sudeep-mohanty, line 40: Could you also add a function description to explain the purpose and the arguments?

components/ulp/ulp_riscv/ulp_core/ulp_riscv_print.c
```diff
@@ -49,3 +49,24 @@ void ulp_riscv_print_hex(int h)
         h <<= 4; // move the 2nd leftmost byte to the left, to be extracted next
     }
 }
+
+void ulp_riscv_print_hex_with_number_of_digits(int h, int number_of_digits)
```
sudeep-mohanty, line 53: Since the input here is `int h`, the maximum number of hex digits would be 8. In this case, I assume the purpose of this function would be to print < 8 bytes. Is the understanding correct?

components/ulp/ulp_riscv/ulp_core/ulp_riscv_print.c
```diff
@@ -49,3 +49,24 @@ void ulp_riscv_print_hex(int h)
...
+{
+    int x;
+    int c;
+
```
sudeep-mohanty, line 57: Would be good to add a sanity check for `number_of_bytes < 0` which could return an error.

components/ulp/ulp_riscv/ulp_core/ulp_riscv_print.c
```diff
@@ -49,3 +49,24 @@ void ulp_riscv_print_hex(int h)
...
+    if (!s_print_ctx.putc) {
+        return;
+    }
+
+    // Does not print '0x', only the digits (8 digits to print)
```
sudeep-mohanty, line 62:
```suggestion
    // Does not print '0x', only the digits specified by the number_of_digits argument
```

sudeep-mohanty:
Thanks for the contribution @mickeyl! I've left some minor queries and comments. Once they are addressed, we should be good to merge. Thanks!

mickeyl:
@sudeep-mohanty Thanks for your comments. I have rebuild the pull request putting them into account. Please have another look.

mickeyl:
@sudeep-mohanty Indeed, my force-push didn't work for some reason. It should be visible now.

sudeep-mohanty:
Thanks @mickeyl. Changes LGTM. I shall do some more tests and then merge it.

sudeep-mohanty:
sha=89413eaab6501749f54072d80065c499b72c2c26

sudeep-mohanty (approved):

PR #13204, opened 2024-02-18
Certificate bundle upgrade (IDFGH-12148)

BitsForPeople:
Modified gen_crt_bundle.py and esp_crt_bundle.c to build/use a static 'index' of the certificates in a bundle and include this index data in the bundle data.
This obviates the need for esp_crt_bundle.c to allocate heap memory and build the index at runtime, saving ~120x4 bytes of heap for the full/'unfiltered' bundle.
Changed "mbedtls_x509_crt parent" in esp_crt_check_signature() to "mbedtls_pk_context pubkey" to save some 200 bytes of stack requirement.

### DIFF
--- a/components/mbedtls/esp_crt_bundle/esp_crt_bundle.c
+++ b/components/mbedtls/esp_crt_bundle/esp_crt_bundle.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2018-2022 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2018-2023 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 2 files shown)


Harshal5:
Hello @BitsForPeople,

It looks like you have made some changes in the `mbedtls` submodule too. Could you please confirm the changes that were needed in it? Also, it would be preferred if the submodule changes could be avoided as I think this certificate bundle format change would not need changes in the submodule itself.

Harshal5:
@BitsForPeople Also I think we could squash the three commits into one. You could follow the below steps:
1. Soft reset the three commits using the command: `git reset --soft c38b75fa31b2861ff8a037597c7abd2b05a5b7e1^`.
2. Then restore the staged `mbedtls` changes using the command: `git restore --staged components/mbedtls/mbedtls`
3. Now, the `mbedtls` changes need to be reset. Change directory to the `mbedtls` submodule and reset the changes using the following commands:
```
cd components/mbedtls/mbedtls
git reset --hard HEAD
```
4. Finally move back to the ESP-IDF git directory and commit the changes using `git commit` and maybe you could use the commit message as "feat(mbedtls/crt_bundle): Store cert indices to reduce memory usage".

Let me know in case you feel this is a bit complicated, I would be happy to help you by carrying this out myself.

BitsForPeople:
@Harshal5, thank you for providing the steps. Didn't work unfortunately, been having issues with git on my (Windows) machine for some time.
Would you be so kind and 'fix' the PR?

Harshal5:
sha=c38b75fa31b2861ff8a037597c7abd2b05a5b7e1

components/mbedtls/esp_crt_bundle/esp_crt_bundle.c
```diff
@@ -1,89 +1,244 @@
...
+        const size_t cert_name_len = esp_crt_get_name_len(cert);
+
+        // Issuers are in DER encoding, with lengths encoded in the content; if valid DER, differing lengths
+        // are reflected in differing content.
+        // Still, we won't try to memcmp beyond the given length:
+        int cmp_res = memcmp(issuer, esp_crt_get_name(cert), min(issuer_len, cert_name_len) );
+
+        if ( unlikely( cmp_res == 0 ) ) {
+            cmp_res = (int)issuer_len - cert_name_len;
+            if( likely( cmp_res >= 0 ) ) {
```
Harshal5, line 214: Hi @BitsForPeople, could you please explain why we allow the issuer name to be longer than the found certificate name? Do you have any example wherein this would be needed?

We think due to this, an incorrect certificate could get matched if the name of another certificate is a prefix of the name of the certificate we're looking for. What do you think?

BitsForPeople: Hi @Harshal5.
I'm actually not sure anymore why I did a >= 0 here, t.b.h. Specifically, I can't remember seeing a case where this would be needed.
Reviewing it now, I believe you are right: If one cert's name would be a prefix of another cert's name in a bundle, we might (in theory) pick the wrong cert.
I think changing it to == 0 should work.

Sorry for seemingly not getting back to this PR until now. I spent a few hours trying to fix the issue of the unrelated mbedtls files in the PR but could not for the life of me get my working tree back to a consistent state with a clean PR.
Is there something you can do with the PR on your end?

Issue #13249, opened 2024-02-24
Basic network examples should support all networks by default (IPv6-only, dual-stack, and IPv4-only) (IDFGH-12195)


sgryphon:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

The basic network examples are configured to only support IPv4 by default. Multiple steps are required to fully support IPv6, such as IPv6-only networks, which include having to turn off IPv4. 

Such complex re-configuration for the specific network type should not be required for a basic example.

The examples should by default support all network configurations simultaneously, without needing reconfiguration, i.e.
similar how a phone or laptop works, with all network types enabled and available, but optional.

If IPv6 is available then an IPv6 address should be acquired, and if IPv4 is available then an IPv4 address should be acquired,without either of them becoming mandatory. The appropriate network should then be used to simultaneously be able to connect to  IPv6, IPv4, and dual stack destinations, if possible. (Some combinations are not possible, e.g. an IPv4 only network can't reach an IPv6 only destination, see the table below) 

Supporting IPv6 does not mean making it mandatory, only available to used if needed.

Destination address selection, e.g. in DNS resolution, should be based on the source addresses available. (e.g. following RFC 6724)

Expected behaviour by network type and destination server (with default configuration):

| Network / Destination | IPv6 server | Dual-stack  | IPv4 only   |
| -- | -- | -- | -- |
| IPv6 with DNS64       | IPv6        | IPv6        | NAT64       |
| IPv6 only             | IPv6        | IPv6        | unreachable |
| Dual stack with DNS64 | IPv6        | IPv6        | NAT64       |
| Dual stack (1)        | IPv6        | IPv6        | IPv4 (3)    |
| IPv4 only (2)         | unreachable | IPv4 (3)    | IPv4 (3)    |

IPv6 (and IPv4) refer to (generally) public addresses (including IPv6 ULA)

(1) Typical configurations are dual stack without DNS64/NAT64
(2) In an IPv4 only network, an endpoint will still have a link-local IPv6 address
(3) For many networks IPv4 connectivity will be via NAT44.

[...]

PR #13250, opened 2024-02-24
Update protocol examples to support all network types #13249 (IDFGH-12196)

sgryphon:
This updates the protocol examples to support all network types, without any reconfiguration.

See related issue for details: https://github.com/espressif/esp-idf/issues/13249

The default configuration now supports IPv4-only, IPv6-only, and dual-stack networks (either with or without DNS64). This means it works without needing to know the target network (the examples could be run by anyone).

Both IPv4 and IPv6 are enabled, with IPv6 fully enabled for SLAAC, RDNSS, and DHCPv6, however neither IPv4 nor IPv6 are mandatory (the old example treated enable as meaning also mandatory, so required reconfiguration based on the network, so would not necessarily work out of the box).

The default behaviour for each network type is explained in the readme.

The change has been tested using ESP32 with WiFi for:
- Dual stack network -- connecting to IPv6 only, IPv4 only, and dual stack servers
  - Dual stack network without DNS64 -- connecting to IPv4 only server, showing DNS lookup fallback
- IPv6 only network with DNS64 -- connecting to dual stack and IPv4 (using DNS64)
- IPv4 only network -- connecting to dual stack server
  - IPv4 only network -- connecting to IPv6 server (fails as expected, as no IPv4 address available)

The change is broken up into the following commits:
1. Add an example function for DNS lookup that if there is a global scope IPv6 address then it tries IPv6 first, then falls back to IPv4 (for IPv4-only servers); if there is no global IPv6 then it only checks IPv4.
2. Turn on full IPv6 support by default (RA for address, RDNSS and DHCPv6 for DNS), and update wifi connect to wait for any preferred (global) address, not both addresses. This means it will work in IPv4-only, IPv6-only, and dual-stack.
3. Update the http_request example to use the example DNS function, so that both IPv6 and IPv4 DNS lookups are tried (depending on available addresses). Test destination addresses were also updated to use IPv6 test endpoints that provide IPv4-only, IPv6-only, and dual-stack addresses, and that return the observed address (useful to check NAT44 and NAT64).
4. Apply the changes to eth_connect (ppp_connect already support ANY address).
5. Fix to the PPP code that failed compilation if LWIP_IPV4 is turned off. There was a call to a macro that only exists if certain config was enabled, so I wrapped the call in the same flags.

**Assistance required**

I have made corresponding changes to the Ethernet connect, but I only have WiFi so have not been able to test. I have checked the code compiles, but can't deploy and run.  PPP did not require any changes, it was already configured to wait for ANY (optional), not ALL (mandatory), but still needs testing.

If there is someone who could assist with verifying the branch on Ethernet and PPP that would be appreciated.


sgryphon:
Based on feedback for my other PR, I used `sdkconfig.defaults` for setting defaults (instead of repeating configs in KConfig).

espressif-abhikroy:
@sgryphon Thanks, for your contribution. We saw the PR and will need some time to evaluate it.

examples/common_components/protocol_examples_common/Kconfig.projbuild
```diff
@@ -384,16 +384,29 @@ menu "Example Connection Configuration"
...
+            address prefixes from Router Advertisement (RA) messages, and both
+            RA RDNSS (Recursive DNS Server) and DHCPv6 Stateless mode to get IPv6 DNS
+            server details from whichever is available on the network.
 
     if EXAMPLE_CONNECT_IPV6
+        config EXAMPLE_CONNECT_PREF_ANY
+            bool "Obtain any preferred address"
+            default y
+            help
+                By default, examples will until any IP preferred address is obtained, either an IPv4 or
```
krzychb, line 400: ``
```suggestion
                By default, examples will wait until any IP preferred address is obtained, either an IPv4 or
```

examples/common_components/protocol_examples_common/Kconfig.projbuild
```diff
@@ -384,16 +384,29 @@ menu "Example Connection Configuration"
         select LWIP_IPV6
         select LWIP_PPP_ENABLE_IPV6 if EXAMPLE_CONNECT_PPP
         help
-            By default, examples will wait until IPv4 and IPv6 local link addresses are obtained.
-            Disable this option if the network does not support IPv6.
-            Choose the preferred IPv6 address type if the connection code should wait until other than
-            the local link address gets assigned.
-            Consider enabling IPv6 stateless address autoconfiguration (SLAAC) in the LWIP component.
+            By default, examples will until any IP preferred (usually global) address is obtained,
```
krzychb, line 387:
```suggestion
            By default, examples will wait until any IP preferred (usually global) address is obtained,
```

examples/protocols/http_request/README.md
```diff
@@ -38,77 +42,207 @@ See the Getting Started Guide for full steps to configure and use ESP-IDF to bui
...
-Age: 317271
-Cache-Control: max-age=604800
+
+### IPv4-only network connecting to dual-stack server
+
+This shows an IPv4 only network (Shadow) connecting to a dual-stack server (`v4v6.ipv6-test.com`).
+
+The connection attempt starts as soon as the IPv4 address is received, and as the device has no global IPv6 address, the DNS lookup and connection will only be IPv4 (`51.75.78.103`).
+
+Note that although the device has an IPv6 link-local address it is ignored, as it is not usable to connect to a public address returned from DNS.
```
krzychb, line 156:
```suggestion
Note that, although the device has an IPv6 link-local address, it is ignored because it cannot be used to connect to a public address returned from DNS
```

krzychb (commented):
@sgryphon, thank you for preparing detailed documentation!

examples/protocols/http_request/README.md
sgryphon: Done. Fix included and rebased to lastest head.

examples/common_components/protocol_examples_common/Kconfig.projbuild
sgryphon: Done. Fix included and rebased to lastest head.

examples/common_components/protocol_examples_common/Kconfig.projbuild
sgryphon: Done. Fix included and rebased to lastest head.

sgryphon:
Code / documentation has been updated based on feedback received.

AxelLin:
> @sgryphon Thanks, for your contribution. We saw the PR and will need some time to evaluate it.

@espressif-abhikroy 
It has been 4 Months since your last reply, any progress about this PR?

david-cermak:
Hi @sgryphon 

Thank you for the comprehensive work and the PR with detailed description. We appreciate the effort you've put into enhancing the protocol examples to demonstrate handling DNS in IPv6/only network types.

I apologize that it has taken this much time to process -- it fell behind the radar due to its complexity and the potential impact on other components and other code owners.

After careful consideration, I have some concerns regarding the default configurations, particularly with backward compatibility and the fact that not all users are on IPv6 networks, especially IPv6-only networks; and as the common connect component is widely used in our protocol tests (where we require the **mandatory** aspect of waiting for addresses) and, at the same time this component is not a good fit for connecting in real life applications:

https://github.com/espressif/esp-idf/blob/d7ca8b94c852052e3bc33292287ef4dd62c9eeb1/examples/protocols/README.md?plain=1#L13

To maintain the current stability and ensure a smoother experience for all users, I would suggest the following approach:

* squash these two commits: 7ffd2320a7fcdef699c12760a14bc4a6a3b5faef f4cf43525545a82fb1e65b784b05c68d58a89c35
* move `example_getaddrinfo()` from common connect component to the http request example
* rename `examples/protocols/http_request/sdkconfig.defaults` to `sdkconfig.ci.full_ipv6` and document it in the `README.md`
    - the defaults would be unchanged
    - users could still choose the preconfigured, full IPv6 stack with `idf.py ... -DSDKCONFIG_DEFAULTS="sdkconfig.ci.full_ipv6"  ...`
* and drop all other changes (you can create a followup PR adding support for the "ANY" address, but I'd suggest taking a similar approach to the PPP scheme and keeping the defaults as they are)

Such a change would make about 100 lines in one example only with no impact to other components/tests, thus much easier for us to accept. That would also cover 99.9% of the usecases (with the default common connect config, i.e. waiting for IPv4 and LL IPv6, so people with global address capabilities would first get an IPv4 and continue with the example, while getting the global address later -- exactly the same behavior as the 'ANY' option you've introduced).
The remaining 0.1% (correct me if I'm wrong) would represent IPv6 only networks, for which we'd require users to disable IPv4 (wouldn't be needed for the 'ANY' option you've proposed) -- which I think is acceptable.

---

PS: Here's my feedback to the commits:

* 7ffd232 feat(example/protocols): Add DNS fuction to check IPv6 first if relevant 

> Nice idea to demonstrate running `getaddrinfo()` with preferred IPv6 addresses, but I think it's better to create a separate example with that or move it outside of common-connect component

* 5752122 feat(example/protocols): Enable full IPv6 stack; default wait for any… 

> As mentioned above, we'd still like to keep the dualstack option with preferred IPv4, as it's simply (still) the most common scenario.
> Good idea to introduce this 'ANY' address config, just don't think it should be enabled by default.

* f4cf435 feat(example/protocols): Update http_request to support all networks

> :+1: Nice work, this could be accepted immediately, just suggest making the changes as outlined above.

* 3b8596c feat(example/protocols): Add default wait for any address to eth_connect

> Same as 5752122

* 55880ab fix(ppp): Only call peer DNS function if configured 

> :+1: Thanks for the fix, this issue has been already fixed on master by addressing unrelated PPP issue:

https://github.com/espressif/esp-idf/blob/d7ca8b94c852052e3bc33292287ef4dd62c9eeb1/components/esp_netif/lwip/esp_netif_lwip_ppp.c#L234-L236

espressif-abhikroy:
The `getaddrinfo()` system call in lwIP within ESP-IDF has a limitation when using `AF_UNSPEC`, as it defaults to returning only an IPv4 address in dual-stack mode.

This issue has been addressed in commit https://github.com/espressif/esp-idf/commit/e2ae81a101da18dc28fef1db3a76039cf4629d82.
When enabled, the behavior is now consistent with Linux, supporting both IPv4 and IPv6 resolutions as expected.

However, the feature is currently disabled by default.
To enable it, you can turn on the `CONFIG_LWIP_USE_ESP_GETADDRINFO` option. This can be configured in the menuconfig under:
`Component config -> LWIP -> DNS -> Enable esp_getaddrinfo() instead of lwip_getaddrinfo()`

david-cermak:
Closing as the problem's been addressed differently in https://github.com/espressif/esp-idf/pull/13250#issuecomment-2516063870

PR #13266, opened 2024-02-27
Updated Dockerfile for devcontainer usage (IDFGH-12213)

timoxd7:
## Info
Added pip and `get_idf` alias to .bashrc to use the docker container in a development container (e.g. github devcontainer).

Without pip, the docker image is not easily usable inside a vscode environment.

Without the alias, it is harder to get `idf.py` working (e.g. manually call export.sh) if not using entrypoint.sh or spawning additional terminals in a dev container environment.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -91,6 +91,9 @@ RUN : \
   && rm -rf $IDF_TOOLS_PATH/dist \
   && :
 
+# Add get_idf as alias
+RUN echo 'alias get_idf=". /opt/esp/idf/export.sh"' >> /root/.bashrc
+
 # The constraint file has been downloaded and the right Python package versions installed. No need to check and
 # download this at every invocation of the container.
 ENV IDF_PYTHON_CHECK_CONSTRAINTS=no


antmak (approved):
Thank you! It makes sense for me

tomassebestik:
Dockerfile modification LGTM, good idea!

@timoxd7 Can you please update your commit messages [as suggested by Danger bot](https://github.com/espressif/esp-idf/pull/13266#issuecomment-1966475332)?

tomassebestik (approved):

tools/docker/Dockerfile
```diff
@@ -28,6 +28,7 @@ RUN : \
     make \
     ninja-build \
     python3 \
+    python3-pip \
```
dobairoland, line 31: Can you please explain where is this necessary? 

All ESP-IDF related work will be done in the virtual environment created by `python3-venv`. `pip` is installed there so I think installing `python3-pip` is unnecessary. Moreover, it can be even misleading. If the user will install packages globally with `python3-pip` then it won't be available in the ESP-IDF environment.

dobairoland:
sha=626aa9f4ca752e68932187c6ef7cac49fcc37ce7

dobairoland (approved):

dobairoland:
Thank you for the contribution. This will probably merged soon.

Issue #13137, opened 2024-02-08
i2c: busy wait in `s_i2c_send_commands` is not very power efficient. (IDFGH-12076)


KJ7LNW:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

While researching #13136, I noticed that [s_i2c_send_commands](https://github.com/espressif/esp-idf/blob/master/components/esp_driver_i2c/i2c_master.c#L438) does a "busy wait" while waiting for the interrupt handler to (eventually) finish the i2c transfer.

Correct me if I'm wrong, but this isn't very power efficient; at low i2c clock rates, it isn't computationally efficient for other tasks, either, because it spins the CPU waiting for the transfer to complete:

```c
while (i2c_master->trans_done == false) {};
```

Wouldn't this be better for power consumption?

```c
while (i2c_master->trans_done == false) {
  esp_light_sleep_start();
};
```

or if FreeRTOS is involved, maybe this would be better to suspend the busy wait until the next systick?

```c
while (i2c_master->trans_done == false) {
  taskYIELD();
};
```

KJ7LNW:
If FreeRTOS is always expected to be in the build for this driver, then maybe the synchronous version should suspend the task with `task = xGetCurrentTaskHandle(); vTaskSuspend(NULL);` and the callback should resume it with `vTaskResume(task); return true;` in the callback handler.

This provides not just better CPU sharing and power-usage efficiency, but an 8x performance improvement over 1kHz systicks and ~75x over 100Hz systicks.  Here is a simple sync vs async benchmark: 

- async with task suspend/resume: 7506 samples/sec
- sync with `vTaskDelay(1);` is limited by the systick rate, so the best it can do is ~995 samples/sec, or 99 samples/sec with the default of systick at 100Hz.

Test project:
https://github.com/KJ7LNW/esp32-i2c-test/tree/master

n.b.: I'm not sure if `xGetCurrentTaskHandle()` is a thing, but I found a reference to it [here](https://www.freertos.org/FreeRTOS_Support_Forum_Archive/June_2005/freertos_Get_task_handle_1311096.html).

nebkat:
[...]

PR #13322, opened 2024-03-06
fix(i2c_master): Do not busy wait on sync transactions (IDFGH-12280)

KJ7LNW:
This commit introduces a private member `i2c_master_bus_t::in_progress_semphr` which allows `i2c_master_transmit` and related functions to give up their timeslice so other tasks can proceed.

Without this commit, one must delay at least one tick via `vTaskDelay(1)` to prevent trigging the idle watchdog, thus limiting the maximum I2C transaction rate to about the speed of `CONFIG_FREERTOS_HZ`.

With this commit, over 6600 samples/sec are possible and the tick delay is unnecessary.

CPU usage below was measured using `vTaskGetRunTimeStats()`.  The FreeRTOS %CPU numbers seem to increase over time and level out after a time, so CPU usage was sampled after 5 seconds across each test for consistency across measurements. Measurements were taken with `CONFIG_FREERTOS_HZ` set to 100Hz and 1000Hz:

Before:
-  100Hz: samples/sec=  99 with  1% CPU ( 99.0 samples/%cpu)
- 1000Hz: samples/sec= 995 with 10% CPU ( 99.5 samples/%cpu)

After:
-  100Hz: samples/sec=6637 with 23% CPU (288.6 samples/%cpu)
- 1000Hz: samples/sec=6610 with 24% CPU (275.4 samples/%cpu)

Closes: https://github.com/espressif/esp-idf/issues/13137

### DIFF
--- a/components/esp_driver_i2c/i2c_master.c
+++ b/components/esp_driver_i2c/i2c_master.c
@@ -432,10 +432,28 @@ static void s_i2c_send_commands(i2c_master_bus_handle_t i2c_master, TickType_t t
             s_i2c_start_end_command(i2c_master, i2c_operation, &address_fill, NULL);
         }
     }
+
+    // Flush the in_progress semaphore in case a previous request timed out but
+    // the request still completed later in the ISR. Such a case would cause
+    // the ISR to Give but would never have been Taken.  This zero-wait Take
+    // loop will reset to a zero-count state:
+    while (xSemaphoreTake(i2c_master->in_progress_semphr, 0) == pdTRUE) {}
+
+    // Start the transaction:
     i2c_hal_master_trans_start(hal);
 
-    // For blocking implementation, we must wait `done` interrupt to update the status.
-    while (i2c_master->trans_done == false) {};
+    // For blocking implementation, the `i2c_master_isr_handler_default` will
+    // unlock (Give) the `in_progress_semphr` when the transaction is complete.
+    // If the ISR is already done, then this xSemaphoreTake will fall through
+    // immediately without blocking:
+    if (xSemaphoreTake(i2c_master->in_progress_semphr, ticks_to_wait) == pdTRUE)
+        assert(i2c_master->trans_done == true);
+    else {
+        i2c_master->cmd_idx = 0;
+        i2c_master->trans_idx = 0;
+        i2c_master->status = I2C_STATUS_TIMEOUT;
+    }
+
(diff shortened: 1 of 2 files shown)


KJ7LNW:
We used this simple project to test the PR change:

- https://github.com/KJ7LNW/esp32-i2c-test/tree/master

Be sure to comment the `vTaskDelay(1)` line here, or the samples/sec will be limited to your `CONFIG_FREERTOS_HZ` rate:

https://github.com/KJ7LNW/esp32-i2c-test/blob/master/main/i2c-test.c#L152

KJ7LNW:
The force-push is a minor change: It simply sets `HPTaskAwoken` to true when `i2c_master->trans_done == true` to wake the caller.  

This reduces latency and increases the i2c sample rates to 6612 (1000Hz) and 6648 (100Hz).

mythbuster5:
Thanks a lot for this contribution. I am going to create an event queue instead of semaphore. Which can be used to store more information.

jrahlf:
@nebkat that does not make sense to me.
Interrupts do not run on any task, they interrupt tasks. If the I2C ISR yields at the end, the scheduler runs at the end of the ISR, instead of at the end of the current time slice. Of course FreeRTOS respects the priorities of the tasks here.

nebkat:
@jrahlf You're right my explanation was inaccurate, it would jump back to the higher priority task anyway. But we would have run the scheduler unnecessarily, so still better this way.

I have also learned `pxHigherPriorityTaskWoken` will only ever be set to `true` (it is or-ing), so there is no need for the `HPTaskAwokenInProgress` here

KJ7LNW:
Out of curiosity I wondered how task notifications perform, which FreeRTOS [claims to be up to 45% faster](https://www.freertos.org/2020/09/decrease-ram-footprint-and-accelerate-execution-with-freertos-notifications.html) than semaphores:
 
Notifications:
-   100Hz: samples/sec= 6654 with 18% CPU (efficiency: 369.7  samples/%cpu)
- 1000Hz: samples/sec= 6636 with 18% CPU (efficiency: 368.7 samples/%cpu)

Semaphores: (copied from above)
-  100Hz: samples/sec=6637 with 23% CPU (efficiency: 288.6 samples/%cpu)
- 1000Hz: samples/sec=6610 with 24% CPU (efficiency: 275.4 samples/%cpu)

As you can see, notifications are faster and CPU efficiency is _much_ higher with the task notification implementation.

I've pushed a commit converting to notifications on this PR so you can see the difference.

Issue #13350, opened 2024-03-11
fatfs: FF_USE_STRFUNC is not configurable, thus f_gets() is unavailable (IDFGH-12315)


KJ7LNW:
### Is your feature request related to a problem?

Legacy fatfs code being ported to ESP-IDF may use `f_gets()`, `f_puts()`, and `f_printf()` calls, but the define `FF_USE_STRFUNC` is set to 0 without kconfig options to enable it.

Others have had this problem, and we just ran into it, too: https://esp32.com/viewtopic.php?t=17148

### Describe the solution you'd like.

Enable kconfig options to appropriately set `FF_USE_STRFUNC` and related defines

### Describe alternatives you've considered.

Could bring in the `fatfs` component and set `FF_USE_STRFUNC` myself in `ffconf.h`, but managing an already packaged dependency just to set a config option seems silly.

PR #13351, opened 2024-03-11
feat(fatfs): add Kconfig settings for FF_USE_STRFUNC (IDFGH-12316)

KJ7LNW:
Legacy fatfs code being ported to ESP-IDF may use `f_gets()`, `f_puts()`, `f_putc()`, and `f_printf()` calls, but the define `FF_USE_STRFUNC` is set to 0 without Kconfig options to enable it.

This commit retains the existing default behavior of `FF_USE_STRFUNC=0` and adds Kconfig settings so users can configure `FF_USE_STRFUNC` and the related `FF_PRINT_LLI`, `FF_PRINT_FLOAT`, and `FF_STRF_ENCODE` if needed.

Closes: https://github.com/espressif/esp-idf/issues/13350

### DIFF
--- a/components/fatfs/Kconfig
+++ b/components/fatfs/Kconfig
@@ -205,6 +205,59 @@ menu "FAT Filesystem support"
             file is opened in write-mode, the seek mechanism will automatically fallback
             to the default implementation.
 
+    choice FATFS_USE_STRFUNC_CHOICE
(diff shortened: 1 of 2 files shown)


KJ7LNW:
This PR REPLACED BY #13352

Issue #13387, opened 2024-03-15
i2c_master can deadlock if `s_i2c_transaction_start()` fails (IDFGH-12356)


KJ7LNW:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

master (60a2bf6a68867fd468a250966230a067f8d97ea2)

### Espressif SoC revision.

n/a, affects all SoC's

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

n/a

### Power Supply used.

USB

### What is the expected behavior?

Deadlock when `s_i2c_transaction_start()` fails

### What is the actual behavior?

_No_ deadlock when `s_i2c_transaction_start()` fails

### Steps to reproduce.

This would be difficult to reproduce.  The problem is obvious based on static code review:

Very simply, a `Take` happens on line 781 but if line 795 fails, then the function will return without the matching `Give` on line 796:

https://github.com/espressif/esp-idf/blob/a7d1da94b9a70e9dfb8b48ebba0fdd0da5b2658b/components/esp_driver_i2c/i2c_master.c#L777-L797

### More Information.

This was mentioned in PR #13134 so this issue reports the specific bug to be addressed individually.

PR #13388, opened 2024-03-15
fix(i2c_master): fix deadlock on s_i2c_transaction_start failure (IDFGH-12357)

KJ7LNW:
As pointed out in PR #13134 by @MatthiasKunnen, there is a deadlock in `s_i2c_synchronous_transaction()` if `s_i2c_transaction_start()` should fail because, on error, s_i2c_synchronous_transaction() returns before releasing the lock.

This commit fixes the deadlock without any other changes.

Closes: #13387


suda-morris:
sha=8131d15eefe08b68095d5348efebe577e91fd441

Alvin1Zhang:
Thanks for contribution again, changes merged with https://github.com/espressif/esp-idf/commit/db857c362b3809253f2767e84d4873ec560d00f3

PR #13402, opened 2024-03-17
Allow to configure sleep RTC calibration

DarkZeros:
* The default behaviour is to run some calibration cycles to ensure RTC slow clock is vlaid during the sleep time.
* However, depending on the application this might be redundant or detrimental
* Added an option to configure it if required

### DIFF
--- a/components/esp_hw_support/port/esp32/Kconfig.rtc
+++ b/components/esp_hw_support/port/esp32/Kconfig.rtc
@@ -85,6 +85,24 @@ config RTC_CLK_CAL_CYCLES
             In case more value will help improve the definition of the launch of the crystal.
             If the crystal could not start, it will be switched to internal RC.
 
+config RTC_CLK_SLEEP_CAL_CYCLES
+    int "Number of cycles to calibrate RTC_SLOW_CLK before sleep"
+    default 10
+    range 0 1000
+    help
+        Just before entering light/deep sleep, a small calibration is done
+        on the RTC_SLOW_CLK. Having a calibrated RTC_SLOW_CLK is required
+        in order to proper measure time while in sleep modes and allow proper
+        wake up from timers.
+        This option defaults to 10 cycles (around 300us) but can be lowered
+        down to 0 to avoid calibration, and take the previous one as valid.
+        It can be particularly useful when the RTC_SLOW_CLK calibration is
+        done by the APP code using various external sources like NTP
+        and comparing the long term drift of the clock, or when using a very
+        stable external oscillator.
+        It can also be interesting for reducing power usage in APPs that are
+        constantly entering light sleep.
+
 config RTC_XTAL_CAL_RETRY
     int "Number of attempts to repeat 32k XTAL calibration"
     default 1
--- a/components/esp_hw_support/sleep_modes.c
+++ b/components/esp_hw_support/sleep_modes.c
@@ -125,7 +125,11 @@
 #define ESP_SLEEP_WAIT_FLASH_READY_DEFAULT_DELAY_US 700
 
 // Cycles for RTC Timer clock source (internal oscillator) calibrate
+#ifndef CONFIG_RTC_CLK_SLEEP_CAL_CYCLES
 #define RTC_CLK_SRC_CAL_CYCLES      (10)
+#else
+#define RTC_CLK_SRC_CAL_CYCLES CONFIG_RTC_CLK_SLEEP_CAL_CYCLES
+#endif
 #define FAST_CLK_SRC_CAL_CYCLES     (2048)  /* ~ 127.4 us */
 
 #ifdef CONFIG_IDF_TARGET_ESP32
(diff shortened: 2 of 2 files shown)


esp-wzh:
@DarkZeros 

Thank you for your contribution!!

But in fact we do not expect users to configure the number of calibration cycles each time they enter sleep. If the number of calibration cycles is too small, not all configuration values may be able to calibrate accurate frequency values. 

Of course, I have also noticed that too many calibrations in certain scenarios lead to too much overhead in and out of sleep.  You could try configuring `PM_LIGHTSLEEP_RTC_OSC_CAL_INTERVAL` this option to configure an interval to skip part of the calibration. Could this resolve the issue you are experiencing?

I think the best solution is to implement the global calibration service of rtc_timer. It is possible to automatically calibrate rtc_timer at regular intervals. I am trying to implement it (but the priority is not high).

DarkZeros:
@esp-wzh Thanks for the reply.

I understand that the current code has its value, this is why I left the defaults as before.

My use case is an ultra low power display update code. The time is handled by external RTC (+interrupt), therefore exact slow RTC is non needed, and the extra calibration cycles in/out sleep is causing 2ms total extra time (around 4% total power usage).
In my case a fixed call to esp_clk_slowclk_cal_set(16000000) makes more sense than a constant calibration in an out of sleep. Saves power and is also better because the external RTC is quite accurate.

I think this might also be useful for cases where time accuracy is not critical. So it may be interesting for other people to set this to 0, or even higher values for long sleep time (ie: 1day sleep). But feel freeto take it or not. I have forked the IDF project for my case, so it is ok if it is not taken upstream. :)

esp-wzh:
@DarkZeros 

Yes, I understand the problem you encountered, and this solution is reasonable in your case.

But what I am worried about is that when using the automatic sleep mode hosted by the esp_pm component (with FreeRTOS tickless idle enabled), there are certain requirements for the accuracy of the tick source provided by slow_clock. For example, freertos suppresses a 100ms ticks, butt if rtc_timer is particularly inaccurate, the chip sleeps for 101ms, so the events that need to be processed on the 100ms tick will be lost, and then FreeRTOS will crash.

DarkZeros:
> @DarkZeros
> 
> Yes, I understand the problem you encountered, and this solution is reasonable in your case.
> 
> But what I am worried about is that when using the automatic sleep mode hosted by the esp_pm component (with FreeRTOS tickless idle enabled), there are certain requirements for the accuracy of the tick source provided by slow_clock. For example, freertos suppresses a 100ms ticks, butt if rtc_timer is particularly inaccurate, the chip sleeps for 101ms, so the events that need to be processed on the 100ms tick will be lost, and then FreeRTOS will crash.

I won´t claim myself an expert in the ESP32 system. **So take everything I say with caution.**
But the calibration, as I see it, it is a way to estimate the slow RTC vs high frequency internal XTAL (40MHz tipically), assuming the slow RTC clock is quite bad (and tipically it is the case if RC 150kHz is used) (?), so it is better to calibrate it vs the FAST clock for less ppm drift.

Therefore, if the chip goes to sleep for ie: 100ms, needs to calibrate the RTC clock to know 100ms how many ticks it means of the RTC clock. To set it to as-closer as possible to 100ms.
When the ESP wakes up after RTC timer has finished, the ESP will "think" that 100ms have passed (maybe 101ms have passed), however it has no clue of knowing that, and the RTOS will function properly, I think.

So even if this is set to 0, a crash in RTOS I think is non possible, however it can cause a de-sync with other timers and external chips.

However, in my case the assumptions are the other way around, the RTC clock is very very accurate. So calibrating it makes no sense, it always gives 16000000, and a fast calibration of just 10 cycles tipically gives a bad calibration of 15999980, ...and so on.
Therefore is detrimental to perform it.

esp-wzh:
Yes, you're right, I didn't think clearly.

After discussion, we do not intend to solve this problem using the method in this PR.

We would like to add an menuconfig option to configure the calibration interval (count in rtc_timer time instead the number of calibration times) for calibrating slow_clock.

During the calibration request process, only when the time from the current moment to the last calibration time is greater than the configured threshold, the hardware will actually be called to calibrate, otherwise the previous calibration value will be returned.

How do you think? If it's ok, we will complete it in the near future.

DarkZeros:
That would definitely be a step forward, because it would not incur multiple calibrations that are redundant.
(Unless the calibration wants to take into account certain thermal drift)

I think in general, for RTC calibration it doesn´t really need to be a blocking process. 
2 samples can be performed (ie: 1 at startup, and 1 at a later time, and compute the total RTC/FAST cycles).
The only condition is that freqs cannot change, and device can´t sleep.

esp-wzh:
@DarkZeros  It's indeed a good suggestion. That's what I'm planning to do.

We can add an  new API `rtc_clk_cal_non_blocking_start`.

After starting the calibration, the CPU can continue to run the code, and the RTC clock calibration module is calibrating in parallel at this time.

When the program needs to use the calibration value, it calls the `rtc_clk_cal_non_blocking_end` API to block and wait for the calibration to be completed.

You are very thoughtful. The use of this pair of interfaces should have restrictions on the behavior between them. Actions that disrupt the calibration process are not allowed to occur. It should only be used in scenarios that have been strictly evaluated, but I think that at least in the current IDF sleep process, this solution can be used to improve the time overhead of entering and exiting sleep.

PR #13473, opened 2024-03-26
fix(openthread-core-esp32x-mtd-config.h): Enable MTD Joiner role

feherdave:
On ESP32 MTD devices Joiner role functionality was not working even if 'Joiner enable' was checked in SDK config because 0 was defined in OPENTHREAD_CONFIG_JOINER_ENABLE macro. Value modified to 1 to resolve the issue. Tested and working on ESP32 H2 MTD device.


zwx1995esp:
Hi, @feherdave , sorry for the delay, I was blocked by some other works, and I will start this process right now.

Issue #12325, opened 2023-09-28
Countdown GPTimer problem (IDFGH-11157)


KaeLL:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

Describe: v5.2-dev-2934-g3b748a6cb7

The following code works, and it's an attempt at testing one-shot functionality of the GPTimer.
```C
#include "driver/gpio.h"
#include "driver/gptimer.h"
#include "esp_attr.h"
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "soc/gpio_struct.h"

gptimer_handle_t gpt = NULL;

uint32_t ledp[] = { 0, 1, 4, 5, 16, 17, 20, 21 };
uint8_t idx = 0, idx1 = 7;

bool IRAM_ATTR gpt_cb( gptimer_handle_t timer, const gptimer_alarm_event_data_t *edata, void *user_ctx )
{
	assert( !gptimer_stop( gpt ) );
	GPIO.out_w1tc = ( uint32_t )ledp[ idx1 ];
	GPIO.out_w1ts = ( uint32_t )ledp[ idx ];

	idx1 = idx;
	idx = ( idx + 1 ) % ( sizeof ledp / sizeof *ledp );

	return false;
}

void gpt_run( uint16_t msec )
{
	ESP_ERROR_CHECK( gptimer_register_event_callbacks( gpt, &( const gptimer_event_callbacks_t ) { gpt_cb }, NULL ) );
	ESP_ERROR_CHECK( gptimer_enable( gpt ) );
	ESP_ERROR_CHECK( gptimer_set_alarm_action( gpt, &( const gptimer_alarm_config_t ) { 40000 * msec } ) );
	ESP_ERROR_CHECK( gptimer_start( gpt ) );
}

void gpt_init( void )
{
	ESP_ERROR_CHECK( gptimer_new_timer(
		&( const gptimer_config_t ) {
			.clk_src = GPTIMER_CLK_SRC_DEFAULT,
			.direction = GPTIMER_COUNT_UP,
			.intr_priority = 3,
			.resolution_hz = 40000000 },
		&gpt ) );
}

void app_main( void )
{
[...]

Issue #8745, opened 2022-04-07
detect if SPI bus has been initialized (IDFGH-7142)


morganrallen:
This is a followup on #8642 
If a method was provided to detect if the bus was already initialized a decision could be made on calling `spi_bus_free` before reinitializing. In my situation I have a driver that supports multiple devices that all might not be initialized at the same time. I need a way to see if the bus need initializing (for the first device) or not (for the rest of them).

**Describe the solution you'd like**
Some function call like `bool spi_bus_initialized(spi_host_device_t host_id, spi_bus_config_t *bus_config)` or `spi_bus_is_initialized` for better clarity.

KaeLL:
I think that's a trivial change, and there may even be an argument for creating an esp_err_t define specifically for denoting that something's already been initialized.  
Question, though: are you trying to have each peripheral's driver do all SPI initialization completely independent from one another?

rtborg:
I'd also appreciate such functionality - for the reason above, each peripheral driver does SPI initialization independent from the others.

abobija:
Bump 👀

MassimilianoMontagni:
+1

MassimilianoMontagni:
I'm using idf v 4.4.2 and I don't find `spi_bus_get_attr` do I import some header or it is no longer available?

morganrallen:
Also using ~4.4.2, you'll need to include `driver/spi_common_internal.h`

MassimilianoMontagni:
Thanks a lot @morganrallen but I still have the issue reported [here](https://github.com/espressif/esp-idf/issues/8642)

abobija:
Is this resolved? Is there a way to check if SPI is initialized or not?

abobija:
@KaeLL Sorry for the ping, but I see you are mentioning this issue frequently. Do you have any updates on this?

KaeLL:
No. @igrr final word, maybe?

morganrallen:
Something needs to get done, this is a needed feature, and my workaround doesn't even work anymore because `driver/spi_common_internal.h` isn't accessible anymore.

igrr:
Reading the original issue description, I think two solutions are possible:

- SPI bus is initialized once at the application level, and SPI devices don't take care of initializing the bus. For example, if packaging device drivers as components, they would receive an SPI bus handle as an argument.
[...]

PR #13493, opened 2024-03-28
fix(esp_event): esp_event_loop_create_default()=ESP_OK when existing (IDFGH-12482)

noahp:
Change the return value from `esp_event_loop_create_default()` to be `ESP_OK` instead of `ESP_ERR_INVALID_STATE` when the event loop already exists. This helps setups where there are multiple call sites with checks like this:

```c
ESP_ERROR_CHECK(esp_event_loop_create_default());
```

Which otherwise can result in a reboot loop.

### DIFF
--- a/components/esp_event/default_event_loop.c
+++ b/components/esp_event/default_event_loop.c
@@ -92,7 +92,7 @@ esp_err_t esp_event_isr_post(esp_event_base_t event_base, int32_t event_id,
 esp_err_t esp_event_loop_create_default(void)
 {
     if (s_default_loop) {
-        return ESP_ERR_INVALID_STATE;
+        return ESP_OK;
     }
 
     esp_event_loop_args_t loop_args = {
--- a/components/esp_event/include/esp_event.h
+++ b/components/esp_event/include/esp_event.h
@@ -61,9 +61,8 @@ esp_err_t esp_event_loop_delete(esp_event_loop_handle_t event_loop);
  * @brief Create default event loop
  *
  * @return
- *  - ESP_OK: Success
+ *  - ESP_OK: Success, or default event loop has already been created
  *  - ESP_ERR_NO_MEM: Cannot allocate memory for event loops list
- *  - ESP_ERR_INVALID_STATE: Default event loop has already been created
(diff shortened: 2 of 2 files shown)


ESP-Marius:
Hi @noahp, thanks for the suggestion! 

Since returning `ESP_ERR_INVALID_STATE` for an already initialized feature is something we do for multiple different APIs in esp-idf I'm not sure if we should change it.

I think it should be resolved by handling the errors, instead of using `ESP_ERROR_CHECK`. Part of the problem might be that we are a bit too liberal in our examples with  `ESP_ERROR_CHECK`, giving users the impression that using this everywhere is the correct way of doing it :sweat_smile: 

If you can point to any situation where returning `ESP_ERR_INVALID_STATE`  is causing issues that cannot easily be worked around then I'd be happy to reconsider this.

noahp:
> Hi @noahp, thanks for the suggestion!
> 
> Since returning `ESP_ERR_INVALID_STATE` for an already initialized feature is something we do for multiple different APIs in esp-idf I'm not sure if we should change it.

Thank you for reviewing! Makes perfect sense, "don't break user space" :sweat_smile: , closing :+1:

KaeLL:
> If you can point to any situation where returning ESP_ERR_INVALID_STATE is causing issues that cannot easily be worked around then I'd be happy to reconsider this.

#8745 
#12325

ESP-Marius:
@KaeLL Yeah, I agree that `INVALID_STATE` is a poor error condition for this, but at this point I dont think we can change it without it being considered a breaking change.

I'll create a ticket for our 6.0 release to track this, we could consider changing it then.

KaeLL:
@ESP-Marius pinging now that v6.0-dev is around

ESP-Marius:
@KaeLL Thanks for keeping an eye on us :smile: 

I've got an internal MR for adding a new `ESP_ERR_ALREADY_INITIALIZED` to at least encourage developers to use this one from now on.

Regarding if any specific drivers will be updated to use it will be up to the codeowner to decide though, they might not agree that it is a breaking change worth doing.

CC @suda-morris

PR #13523, opened 2024-04-01
feat(lwip): TCPIP_THREAD_NAME "tcpip"

nebkat:
"tiT" has been used since the original public version bd6ea4393c7d2f059fc4decc70f1ec3eb3597268, "tcpip" is less confusing.

### DIFF
--- a/components/lwip/port/include/lwipopts.h
+++ b/components/lwip/port/include/lwipopts.h
@@ -820,7 +820,7 @@ static inline uint32_t timeout_from_offered(uint32_t lease, uint32_t min)
 /**
  * TCPIP_THREAD_NAME: The name assigned to the main tcpip thread.
  */
-#define TCPIP_THREAD_NAME              "tiT"
+#define TCPIP_THREAD_NAME              "tcpip"
 
 /**
  * TCPIP_THREAD_STACKSIZE: The stack size used by the main tcpip thread.


KaeLL:
Why not "LwIP" or "lwip". I doubt this change will go through, but if it does, I hope they'd consider renaming the BLE and WiFi tasks as well.

nebkat:
> Why not "LwIP" or "lwip"

[`TCPIP_THREAD_NAME`](https://www.nongnu.org/lwip/2_0_x/group__lwip__opts__thread.html#ga405e604e4328e1feb878c6fe1798a587) in lwIP is "tcpip_thread", but I figured _thread was redundant. There appears to be a separate `DEFAULT_THREAD_NAME` which is set to "lwIP".

david-cermak:
Hi @nebkat 

Thanks for the PR!
I agree that "tiT" has become quite common, and while renaming it to `"tcpip"` might be less confusing in the long term, I suggest we hold off on this change until v6.0 of IDF; then we can proceed with the renaming. For now, let's put this pull request on hold.

nebkat:
Works for me, thanks!

nebkat:
@david-cermak Ping now that v6.0 is in development 🚀

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/31602cbe66c53a4a824e43933bda2c33fb893062

PR #13546, opened 2024-04-05
feat(esptool): merge_bin CMake target (IDFGH-12541)

nebkat:


nebkat:
@Honza0297 I haven't spent enough time around idf.py internals and the CMake system so I might not fully understand, but:

My main requirement is to be able to generate the merged binary without having to specify `@flash_args` and chip type manually, this will make it easier for both CI and developer using idf.py - I'm guessing both solutions achieve this.

It does feel like something that fits in to the CMake system since it is generating binaries, and this would allow it to be used from IDE (I am successfully using CMake in CLion). I don't have any such use case right now but it would also allow for other targets to depend on the merged binary file, maybe for some customized release system.

Honza0297:
I understand your main requirement - both approaches (CMake and Python) will solve it.  
We are trying to have these "aliases" for (not only) esptool commands  in one place in Python code and make CMake targets only if necessary.  
If there is no specific use case in the moment, we would stay with the current approach of implementing it in Python (specifically, `merge-bin` will be in the master branch in several days). 
We will close this PR for now, but we may reopen it if you or somebody else would need the target. 
Thank you for understanding.

dobairoland:
We will add a cmake target as well.

PR #13557, opened 2024-04-08
feat(espcoredump): Allow writing to both flash and UART

nebkat:
Writing core dump to UART is useful when actively developing in esp monitor and you want to immediately check the output.
Writing to flash is separately useful when a crash might occur when esp monitor is not connected.

This PR allows both options to be enabled at the same time.


erhankur:
@nebkat Thank you for your contribution. We will evaluate internally and decide whether to use both options at the same time.

PR #13559, opened 2024-04-08
esp_eth: DP83848: add list of supported models (IDFGH-12556)

karlp:
The existing dp83848.c is capable of supporting other PHYs, it just needs a list of supported models, instead of hardcoding a single valid model.

Tested with a DP83825i phy on a custom board.

Developed and tested against the v5.1 release branch, but patch is for master.

### DIFF
--- a/components/esp_eth/src/esp_eth_phy_dp83848.c
+++ b/components/esp_eth/src/esp_eth_phy_dp83848.c
@@ -10,9 +10,32 @@
 #include "esp_check.h"
 #include "esp_eth_phy_802_3.h"
 
-static const char *TAG = "dp83848";
+static const char *TAG = "dp838xx";
 
 /***************Vendor Specific Register***************/
+
+#ifndef ARRAY_LENGTH
+#define ARRAY_LENGTH(array) ( sizeof( array ) / sizeof( *( array ) ) )
+#endif
+
+struct model_support_s {
+	const char *name;
+	uint32_t oui;
+	uint8_t model;
+};
+
+const struct model_support_s models_supported[] = {
+	{
+		.name = "DP83848",
+		.oui = 0x80017,
+		.model = 0x09,
+	},
+	{
+		.name = "DP83825i",
+		.oui = 0x80028,
+		.model = 0x14,
+	},
+};
 
 /**
  * @brief PHYSTS(PHY Status Register)
@@ -163,8 +186,15 @@ static esp_err_t dp83848_init(esp_eth_phy_t *phy)
     uint8_t model;
     ESP_GOTO_ON_ERROR(esp_eth_phy_802_3_read_oui(phy_802_3, &oui), err, TAG, "read OUI failed");
     ESP_GOTO_ON_ERROR(esp_eth_phy_802_3_read_manufac_info(phy_802_3, &model, NULL), err, TAG, "read manufacturer's info failed");
-    ESP_GOTO_ON_FALSE(oui == 0x80017 && model == 0x09, ESP_FAIL, err, TAG, "wrong chip ID");
-
+    bool found = false;
+    for (int i = 0; i < ARRAY_LENGTH(models_supported); i++) {
+        if ((models_supported[i].oui == oui) && (models_supported[i].model == model)) {
+            ESP_LOGI(TAG, "detected phy %s", models_supported[i].name);
+            found = true;
+            break;
+        }
+    }
+    ESP_GOTO_ON_FALSE(found, ESP_FAIL, err, TAG, "unsupported oui/model: %x:%x", oui, model);
     return ESP_OK;
 err:
     return ret;


kostaond:
@karlp thank you for the contribution! Could you please run [`[ethernet]` target tests](https://github.com/espressif/esp-idf/tree/master/components/esp_eth/test_apps) with your PHY?

kostaond:
@karlp we should also modify `esp_eth_phy_new_dp83848` to `esp_eth_phy_new_dp838xx`. Note that both API need to coexist between major release. Therefore you can wrap `esp_eth_phy_new_dp838xx` by `esp_eth_phy_new_dp83848` and mark it deprecated.

karlp:
Hrm, I chose not to do that, exactly to avoid API changes.  From looking at other drivers, not all the 838xx are actually compatible, it's an odd mix of various parts.  but, sure, I can add the new one and a add a deprecated, that's totally fine and straightforward.  

For the test apps, that readme is ... a little slim, is there any other guidance on using that?

karlp:
Ok, this will need more work then.  autonegotiation and loopback fail, so it fails before it can go any further.  I'll try and find more time to figure out what it's upset about.

kostaond:
@karlp, OK. If you needed any assistance, just let me know and send me a log. Based on the description in datasheets, it should be the same type of loopback when enabled in **BMCR** register.

karlp:
yeah, it's reading autonegotiation as disabled after boot, which is.... weird, trying to diagnose :)

kostaond:
Check how "ANeg_Dis" bootstrap option is configured in your HW design.

karlp:
yeah, was just coming here to say, I've discovered that we're strapping that incorrectly, so it's hard disabled.  That's... good that we caught it now I guess :)

I've also discovered that the Loopback handling is different.  DP83848 uses _only_ bit 14 of BMCR (as does DP83822), but the DP83825i uses bit 14, and _also_ bit 2 of BICSR, (it supports more loopback modes)

Tests 4,5,6 work, and 7 will start to work if I use http, but fails on the redir to https, I'm not sure what's going on there, but I suspect that's a test problem, not a phy problem. (we've obviously been using this for a while now)  Out of the box test 7 will fail to connect to the server.

So I guess, what?  I can put special cases customizations into this file, similar to ksz80xx? or I can propose an entirely new file, that's 99% copy?

kostaond:
Yeah, DP83825i has more loopback options. However, I think MII loopback (enabled via BMCR at both chips) is sufficient for the most common debug purposes. If somebody wanted to use the other types of loopback, they can always do so via `ETH_CMD_READ_PHY_REG/ETH_CMD_WRITE_PHY_REG` commands.

Regarding test 7 ("download test"), it's OK if it's fail as I've already indicated in some of the comments above. It's not directly related to PHY functionality.

What about tests 1/2/3? Are they blocked by your current HW design? I'm mostly interested in those.

karlp:
test 1 passes, test 2 fails because autoneg is hard disabled by strapping.  test 3 fails because even plain MII loopback needs BMCR _plus_ this extra bit.  
image
I tried setting that as well, but I think I just started running into more speed/duplex/loopback special cases, and given that it was no longer going to work with the "standard" dp83848 phy driver, I didn't dig too much further into it.   We've never used the loopback modes at all, so never noticed it not working until we ran these tests for you :)


(test 1 output, KKK lines are debug for me)
```
Enter test for running.
1
Running ethernet io test...
I (2435) esp32_eth_test_common: DUT PHY: DP83848
I (2455) dp838xx: detected phy DP83825i
I (2455) KKK: autoget_g_stat returned 0, (bmcr: 0x2100)
I (2455) KKK: new driver installed with driver autoneg readback at 0
I (2455) esp32_eth_test: Ethernet MAC Address: 0c:8b:95:49:ef:3b
I (2465) esp32_eth_test: Ethernet MAC Address: 0c:8b:95:49:ef:d4
I (2475) esp32_eth_test: Ethernet PHY Address: 0
/IDF/components/esp_eth/test_apps/main/esp_eth_test_apps.c:49:ethernet io test:PASS
Test ran in 68ms
```

kostaond:
I see, it does not fully follow IEEE 802.3... OK, so add a new logic for DP83825 to set BISCR register based on selected speed as defined in "Table 13" of the datasheet. Keep everything in one src file. BTW, the src file needs to be renamed to `esp_eth_phy_dp838xx.c` as well.

kostaond:
@karlp are you planning to move forward with this PR and implement the changes we discussed above?

karlp:
It's on my list, the hardware is coming through production, I've just been working on other portions of the project.   If  someone  else wants to continue with it themselves though, I won't mind :)

kostaond:
@karlp did you have a chance to implement and test the changes we discussed above?

kostaond:
Note that PHY drivers with specific part number in it's name needs to be tested and they need to support all the features and possible non-comply to IEEE802.3. If users are OK with basic PHY functionality, I encourage to use `Generic PHY driver` https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_eth.html#_CPPv423esp_eth_phy_new_genericPK16eth_phy_config_t

karlp:
I got my tree updated to start re-testing this, but no, it didn't get any further.  kinda drowning in work sorry.  Generic phy sounds interesting, though, as we found above, it doesn't appear to be 802.3 basic anyway, so probably won't work.  I'm still employed on this though, so I hope to get to it unless someone else does it first :)

kostaond:
No problem, I just needed to check if I should keep this PR opened. Regarding generic PHY, you are right, it will not work just by one command. However, it is still should be possible to start it by `ETH_CMD_READ_PHY_REG/ETH_CMD_WRITE_PHY_REG` commands. It's not that convenient, it's more work and user need to check datasheet but it's because the DP83825i does not fully comply with standard. And in general, Generic PHY is intended for "generic" use cases, loopback is something special what 98% people does not use. It's used primarily for debugging and if you are debugging mode, you need to understand your chip anyway...

karlp:
well, exactly, I didn't need loopback at all, and was just trying to upstream that the existing driver _workd_ for the "most useful" cases but it was suggested that I would need to add proper loopback support to be able to upstream this :)

kostaond:
Yes, that's the point. I'll try to explain again:
- **Generic driver**: it is expected to support generic IEEE802.3 functions. A warning is given to users in documentation that some specific features of specific PHY may not work correctly.
- **Specific PHY drivers**: specific PHY drivers like DP83848, LAN87xx, etc. must correctly handle all supported features. In addition, we test the drivers on specific PHY. We have multiple CI runners with specific PHY's in the office.

Since you wanted to upstream the changes to the specific driver, I needed to request to add proper loopback support.

Maybe we can close this PR since now there is Generic PHY driver available in the ESP-IDF. What do you think?

karlp:
I don't want to close this until I can I verify that the new generic driver can be used to get a dp82825i working without any extra driver code. I'll take that internally.

PR #13560, opened 2024-04-08
esp_eth: DP83848: correct link detection to use BMSR (IDFGH-12557)

karlp:
Reading the link state via PHYSTS was incorrect, as it only reflects the link state bit from BMSR.  BMSR latches link down events, and are not cleared without being read.  (See 802.3-2008 section 2, section 22.2.4.2.13) This leads to the original DP828xx code only supporting link up, then a single link down event.

Switch to reading the link state via BMSR, but continuin to read the negotiation results via PHYSTS and ANLPAR.  This is inline with LAN8720x, RTL8201, KSZ80xx phy drivers, and other opensource drivers for the DP838xx family of devices.

Tested on a private board with a DP83825i PHY[1].  No publically available boards using the original DP83848 are known of for testing.

[1] https://github.com/espressif/esp-idf/pull/13559

### DIFF
--- a/components/esp_eth/src/esp_eth_phy_dp83848.c
+++ b/components/esp_eth/src/esp_eth_phy_dp83848.c
@@ -77,13 +77,15 @@ static esp_err_t dp83848_update_link_duplex_speed(phy_dp83848_t *dp83848)
     uint32_t peer_pause_ability = false;
     anlpar_reg_t anlpar;
     physts_reg_t physts;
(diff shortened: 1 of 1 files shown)


karlp:
the title is already `fix(esp_eth/dp83848): correct link detection to use BMSR` is that not sufficient?  I already amended it to suit the guidelines, the _pull request_ just still contains the original title.  (The JIRA issue check is just broken)

kostaond:
`dp83848` is not a component. `esp_eth` is the component. Therefore, ideally should be just `fix(esp_eth)`. I'm asking to fix it because I'm not sure if it is flagged in internal CI and so making it fail...

kostaond:
sha=e2945871b731b8e59ff718a9eec7962777fc6e6b

PR #13572, opened 2024-04-09
add ETH PHY JL1101 driver (IDFGH-12570)

Jason2866:
@me-no-dev this PR is needed before it can be added to Arduino.
Verified working. Tested with this board https://templates.blakadder.com/dingtian_DT-R008.html

### DIFF
--- a/components/esp_eth/CMakeLists.txt
+++ b/components/esp_eth/CMakeLists.txt
@@ -26,7 +26,8 @@ if(CONFIG_ETH_ENABLED)
                          "src/esp_eth_phy_ip101.c"
                          "src/esp_eth_phy_ksz80xx.c"
                          "src/esp_eth_phy_lan87xx.c"
-                         "src/esp_eth_phy_rtl8201.c")
+                         "src/esp_eth_phy_rtl8201.c"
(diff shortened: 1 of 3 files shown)


suda-morris:
Thanks for adding new Ethernet PHY driver! @Jason2866 

I suggest submitting the PHY driver to https://github.com/espressif/esp-eth-drivers

The benefit is, the PHY driver can have its own release cycle, which is independent to the release of esp-idf. So the user won't have to wait for the release of esp-idf v5.3 if he wants to use the JL1101 driver in the product.

Jason2866:
Since the ETH drivers in the component registry needs IDF 5.3 it makes no sense for this use case The driver should be available in Arduino too, which is based on IDF 5.1

kostaond:
@Jason2866 thanks for your contribution! We do not accept PR's to release branches. We have special backport process for it. In addition, the backports are intended for fixes and not for features.

There is no restriction for Ethernet drivers from component registry to need IDF v5.3.

Therefore, I'm closing this PR and let's discuss the rest at https://github.com/espressif/esp-eth-drivers/pull/28

PR #13578, opened 2024-04-10
Make FreeRTOS configUSE_TIMERS optional in KConfig (IDFGH-12576)

dvosully:
The FreeRTOS Software Timers may not be required in some user applications, as the esp_timer provides much of the same functionality.

This PR adds a KConfig option to allow the user to disable the FreeRTOS Software timers using the configUSE_TIMERS option. This allows saving the stack space that would have been used by the inactive FreeRTOS timer service task.

### DIFF
--- a/components/freertos/Kconfig
+++ b/components/freertos/Kconfig
@@ -161,6 +161,13 @@ menu "FreeRTOS"
                 Enable backward compatibility with APIs prior to FreeRTOS v8.0.0. (see
                 configENABLE_BACKWARD_COMPATIBILITY documentation for more details).
 
+        config FREERTOS_USE_TIMERS
+            bool "configUSE_TIMERS"
+            default y
+            depends on !IDF_TARGET_LINUX
+            help
+                Enable FreeRTOS Software Timers.
+
         config FREERTOS_TIMER_SERVICE_TASK_NAME
             string "configTIMER_SERVICE_TASK_NAME"
             default "Tmr Svc"
--- a/components/freertos/config/include/freertos/FreeRTOSConfig.h
+++ b/components/freertos/config/include/freertos/FreeRTOSConfig.h
@@ -183,7 +183,11 @@
 
 /* ------------------- Software Timer ---------------------- */
 
-#define configUSE_TIMERS                          1
+#if CONFIG_FREERTOS_USE_TIMERS
+    #define configUSE_TIMERS                      1
+#else
+    #define configUSE_TIMERS                      0
+#endif
 #define configTIMER_TASK_PRIORITY                 CONFIG_FREERTOS_TIMER_TASK_PRIORITY
 #define configTIMER_QUEUE_LENGTH                  CONFIG_FREERTOS_TIMER_QUEUE_LENGTH
 #define configTIMER_TASK_STACK_DEPTH              CONFIG_FREERTOS_TIMER_TASK_STACK_DEPTH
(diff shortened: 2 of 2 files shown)


KonstantinKondrashov:
Hi @dvosully!
This option is not needed for reducing the stack usage of the inactive FreeRTOS timer task. It is already implemented with weak func. If the application never calls xTimer* related functions, and a definition of xTimerCreateTimerTask was found in **tasks**.c.obj [ __attribute__( ( weak ) ) xTimerCreateTimerTask()](https://github.com/espressif/esp-idf/blob/master/components/freertos/esp_additions/freertos_tasks_c_additions.h#L381-L395), then timers.c.obj object file is not considered for linking, since it doesn't resolve any of the unresolved references.
It means this weak definition is used instead of xTimerCreateTimerTask() from timers.c.

If you now call some timer-related function from app_main, this will force the linker to consider **timers**.c.obj for linking. Therefore, the strong definition of xTimerCreateTimerTask will be found and will override the weak definition. In the map file, you'll see that xTimerCreateTimerTask will be found in libfreertos.a(**timers**.c.obj).

I close this as IDF already saves resources if xTimer functionality is not used.

dvosully:
Thanks @KonstantinKondrashov 
It looks like that is not working as it should.

If I create a brand new empty project for ESP32 which does nothing except list the tasks, there is a Tmr Svc task in the list. Nothing should be using any FreeRTOS timers, so why does this task get created?

igrr:
@dvosully If you can upload the .map file of the project, it could help figure out which part of the project causes FreeRTOS timers to be linked. You can also try to check it yourself using the "Cross-reference table" part of the map file, look for references to timers.c.obj (excluding tasks.c.obj).

igrr:
Thanks for the reproducer.

What seems to happen is that initially, idf_additions.c pulls in a dependency on event_groups.c by calling `xEventGroupCreateStatic` from `xEventGroupCreateWithCaps`. In turn, event_groups.c pulls in a dependency on timers.c by calling `xTimerPendFunctionCallFromISR`.

Later, at the stage when the linker removes unused sections, it figures out that `xEventGroupCreateWithCaps` isn't called anywhere, and removes it as well as all of the event_groups.c functions. But since the weak definition of `xTimerCreateTimerTask` has already been overridden, the timer task ends up being created.

(I am not sure why this all doesn't happen if  CONFIG_FREERTOS_USE_TRACE_FACILITY and CONFIG_FREERTOS_USE_STATS_FORMATTING_FUNCTIONS are disabled. Possibly, enabling these options forces creates dependencies which force the linker to visit the object files in a different order.)

I think we can split the idf_additions.c into multiple source files, in which case the dependency on `xTaskCreatePinnedToCoreWithCaps` (which does get used) won't cause the linker to also consider `xEventGroupCreateWithCaps` for linking (which doesn't get used in your case, but causes this chain of dependencies.)

Adding the Kconfig option to disable timers support is still possible, but if there is a way to avoid an additional compile-time configuration, such solution would be preferrable.

ESP-Marius:
> (I am not sure why this all doesn't happen if CONFIG_FREERTOS_USE_TRACE_FACILITY and CONFIG_FREERTOS_USE_STATS_FORMATTING_FUNCTIONS are disabled. Possibly, enabling these options forces creates dependencies which force the linker to visit the object files in a different order.)

Seems to happen because the function is only defined if  `configUSE_TRACE_FACILITY == 1 `, otherwise e.g.  `xEventGroupClearBitsFromISR` is simply a macro that defines to `xTimerPendFunctionCallFromISR`. So since it isnt used anywhere it never gets compiled and the dependency is never pulled in.

We'll fix this @dvosully, thanks for providing all the info to help us reproduce this!

ESP-Marius:
sha=fa79fead50bdff802ecccc5a8212b18ed965276a

SoucheSouche:
6cbee5c7f693d5daa380f46c8522d5424e1e50bd

Issue #6185, opened 2020-11-30
registered_heaps_write_lock will never be NULL from heap_caps_init.c (IDFGH-4347)


samuelbles07:
## Environment

- Development Kit:  ESP32-DevKitC
- Module or chip used:  ESP32-WROOM-32
- IDF version: v4.3-dev-1197-g8bc19ba89
- Build System: Make
- Compiler version: (crosstool-NG esp-2020r3) 8.4.0
- Operating System: Linux - ubuntu
- Using an IDE?: No
- Power Supply: USB

## Problem Description

I am trying to use this [repository](https://github.com/loboris/ESP32_TFT_library), i only change flash size to 4MB (expected by partition table total size) then directly `make` the project. An error happens that i think not from the TFT project but from idf  (CMIIW). 

### Expected Behavior

Build success

### Actual Behavior

```
/home/samuel/tools/esp/esp-idf/components/heap/heap_caps_init.c:230:44: error: the comparison will always evaluate as 'true' for the address of 'registered_heaps_write_lock' will never be NULL [-Werror=address]
     MULTI_HEAP_LOCK(&registered_heaps_write_lock);
                                            ^~
/home/samuel/tools/esp/esp-idf/components/heap/heap_caps_init.c:232:45: error: the comparison will always evaluate as 'true' for the address of 'registered_heaps_write_lock' will never be NULL [-Werror=address]
     MULTI_HEAP_UNLOCK(&registered_heaps_write_lock);

```

### Steps to reproduce

1. Clone [repository](https://github.com/loboris/ESP32_TFT_library
2. Change flash size to 4MB
3. Compile using `make`

igrr:
This seems to be due to this line: https://github.com/loboris/ESP32_TFT_library/blob/aa21772f54a71887ec08b2e8bbaef9e304009891/Makefile#L8
Could you try removing it?

Note that there are a couple other issues that need to be fixed in order to get the library to build: there is a missing `#include "driver/gpio.h"` in one file and a couple of misleading-indentation errors in another. That said, ESP32_TFT_library project hasn't been maintained since 2018, so even if it compiles, it might not work due to API changes. PR https://github.com/loboris/ESP32_TFT_library/pull/79 seems to have the changes to get the library to work with ESP-IDF v4.0, which is a bit more up to date.

samuelbles07:
Hi thank you for your response. Yes, i decided to use the PR that you mention. Thank you

PR #13595, opened 2024-04-12
-save-temps and error: the comparison will always evaluate as true (IDFGH-12596)

denizzzka:
I really need to start compiling with `-save-temps` to use obtained `*.i` files as bindings for D language

But I found what this is leads to compilation errors like:
```sh
esp-idf-v5.2.1/components/heap/heap_caps_init.c: In function 'heap_caps_add_region_with_caps':
esp-idf-v5.2.1/components/heap/heap_caps_init.c:245:44: error: the comparison will always evaluate as 'true' for the address of 'registered_heaps_write_lock' will never be NULL [-Werror=address]
```

I found all places affected when `-save-temps` flag is established:

* components/heap/multi_heap_platform.h - trivial fix
* lwip component - have multiple comparisions with NULL, don't know how to deal with it, dirty fixed in this PR by disabling throwing errors for such cases
* spiffs component - also have similar issue (tautological-compare), ditto: don't know how to deal with it, just disabled this error

I need ideas

Related:
* https://github.com/espressif/esp-idf/issues/6185

### DIFF
--- a/components/esp_common/include/esp_check.h
+++ b/components/esp_common/include/esp_check.h
@@ -7,6 +7,7 @@
 
 #include "esp_err.h"
 #include "esp_log.h"
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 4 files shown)


denizzzka:
Solved by menuconfig compiler options

PR #13599, opened 2024-04-12
Add user pointers to avoid using global variables in callbacks (IDFGH-12601)

carlosbar:
The motivation to propose this change is based on the callbacks used both by:

GAP:  (esp_ble_gap_register_callback, esp_ble_gap_get_callback)
GATTS: (esp_ble_gatts_register_callback, esp_ble_gatts_get_callback)

The proposed solution is to add 2 new more entry points for each of the profiles:

GAP: (esp_ble_gap_register_ptr, esp_ble_gap_get_ptr)
GATTS: (esp_ble_gatts_register_ptr, esp_ble_gatts_get_ptr)

With these new entry points, the registered callbacks now have to ability to retrieve a pre-set user pointer, avoiding, in this way, the use of global variables, which is consolidated as a bad practice and should be avoided when possible. This problem yet has another weight in multi-core environments.

### DIFF
--- a/components/bt/common/btc/core/btc_manage.c
+++ b/components/bt/common/btc/core/btc_manage.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2021 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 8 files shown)


zhp0406:
In practical usage, the callback functions for GAP and GATT are not frequently modified, and instances of competition or deadlock are rare.

What other issues do you think might arise from this approach?

carlosbar:
> In practical usage, the callback functions for GAP and GATT are not frequently modified, and instances of competition or deadlock are rare.
> 
> What other issues do you think might arise from this approach?

The issues that arise from globals I'm quite sure are possible race conditions and lack of easy tracking to whown performed some change. Besides that, not related with issues but with best practices and design, the use of global variables prevent isolation/encapsulation (as this is aware of parts of the code it is not related), prevents abstraction/compartmentalization because of almost the same reason (touching an external piece of code and not minding its own bits). Interfaces are created to comply with these, and the proposed change it is a form of interface. 

Of course this is going to be only (in most cases) a single variable outside of the callback context, but the simple usage of it breaks some of the above architecture patterns.

PR #13615, opened 2024-04-15
Fix: make esp_tls_wolfssl send entire client certificate chains (IDFGH-12619)

frankencode:
This change makes the wolfSSL backend sent the complete TLS client certificate chain. This align the wolfSSL backend with the behavior of the mbedtls backend. Some servers need the intermediate certificates to verify a client certificate. If the provided PEM file contains only a single certificate this change has no effect.

This impacts higher level APIs to function as someone would expect. E.g.: esp_websocket_client_config_t.client_cert: when passing here a pem file containing 2 certificates (the CA's and the client's) it would be expected that both are transmitted during TLS handshake.


frankencode:
OK, I reediting the commit message...

PR #13631, opened 2024-04-17
Eliminate compiler warning about 'BIT' macro.

kevlarh:


KaeLL:
What's that installation_notes.txt file?

PR #13635, opened 2024-04-17
feat(i2c): Add API to support sending register or command bytes separately from the data (IDFGH-12639)

andylinpersonal:
Adding this API can eliminate the need to allocate a temporary buffer to accommodate both memory address/command and data before the writing transaction, which is especially useful for large transactions, such as I2C EEPROM, I2C panels, and dataloggers. I2C EEPROM example has been modified accordingly. 
Tested on C3, S3, and C6 with a real AT24C256 EEPROM chip (LENGTH=pagesize=64B, can be seen on datasheet) in both sync and async mode.


nebkat:
Could you also add a `i2c_master_prefixed_transmit_receive`?

PR #13651, opened 2024-04-20
espcoredump: Ensure consistency between the written flash and checksum data (IDFGH-12659)

nomis:
This is a direct core dump of the current memory so it may change because of other interrupt handlers or the memory used by the core dump process itself. Always write the data via the cache buffer so that what's written matches the checksum.

Fixes part of espressif/esp-zigbee-sdk#326 (invalid CRC when writing a core dump to flash).


Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/1b71db8da38afc95ac7113e7482a9cb3e956cd13.

PR #13660, opened 2024-04-21
Add exec permission to `entrypoint.sh` script in Dockerfile. (IDFGH-12668)

xiahualiu:
Hello Espressif team! This is a minor change to the Dockerfile in the tools folder.

The Dockerfile did not set exec permission for the `entrypoint.sh` script.

I know the exec bit for `entrypoint.sh` is set in this git repository. But when user uses self-provided `entrypoint.sh` without exec permission, it will break the Dock container. *"unable to start container process: exec: "/opt/esp/entrypoint.sh": permission denied: unknown"*

Better do `chmod u+x /opt/esp/entrypoint.sh` after `COPY` to make sure `entrypoint.sh` is always executable.

Also, it is better than `COPY --chmod=0755`, for older docker versions don't support the `--chmod` option after `COPY`.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -102,5 +102,7 @@ ENV IDF_PYTHON_CHECK_CONSTRAINTS=no
 ENV IDF_CCACHE_ENABLE=1
 
 COPY entrypoint.sh /opt/esp/entrypoint.sh
+RUN chmod u+x /opt/esp/entrypoint.sh
+
 ENTRYPOINT [ "/opt/esp/entrypoint.sh" ]
 CMD [ "/bin/bash" ]


xiahualiu:
Changed the commit message and rebased the branch to pass the DangerJS pull request review check.

dobairoland:
@xiahualiu Thank you for your contribution.

Can you please install the pre-commit hooks (https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html) and make the commit while it is installed? It will point you to fix other issues as well.

xiahualiu:
Seems like I need to change the from branch name to pass the pre-commit checks. I will close this PR and create a new one with pre-commit shortly. Sorry for the trouble, this is my first PR to esp-idf and thanks you for the reviews.

dobairoland:
Merged by 248d25a235bc8e92471ec004ba138517bb8613eb

PR #13724, opened 2024-04-30
fix(ws_transport): fix server-key corruption when first fragment lost… (IDFGH-12742)

bryghtlabs-richard:
… during websocket connection

Fixes 2267d4b
Fixes https://github.com/espressif/esp-protocols/issues/396


bryghtlabs-richard:
This is ready for review.

PR #13746, opened 2024-05-05
feat(http_server): add 413 Payload Too Large response (IDFGH-12767)

mmalecki:
While not useful for the HTTP parser itself, this is very useful for memory-limited consumers of the HTTP server API.

### DIFF
--- a/components/esp_http_server/include/esp_http_server.h
+++ b/components/esp_http_server/include/esp_http_server.h
@@ -608,6 +608,9 @@ typedef enum {
      */
     HTTPD_411_LENGTH_REQUIRED,
 
+    /* Incoming payload is too large */
+    HTTPD_413_PAYLOAD_TOO_LARGE,
+
     /* URI length greater than CONFIG_HTTPD_MAX_URI_LEN */
     HTTPD_414_URI_TOO_LONG,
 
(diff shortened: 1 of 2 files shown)


components/esp_http_server/include/esp_http_server.h
```diff
@@ -608,6 +608,9 @@ typedef enum {
      */
     HTTPD_411_LENGTH_REQUIRED,
 
+    /* Incoming payload is too large */
+    HTTPD_413_PAYLOAD_TOO_LARGE,
```
mahavirj, line 612: Suggest changing to:

`HTTPD_413_CONTENT_TOO_LARGE`

same comment at other places too

hmalpani:
sha=2f8d39b68f729bbb743e912710818d772950c736

components/esp_http_server/include/esp_http_server.h
mmalecki: Happy to change this by all means, but all the other names appear to follow the literal HTTP status code names as defined in spec, so just curious as to the motivation?

hmalpani: I think by other places, he meant at other changes in this PR. Anyways, we have updated the commit to have that change. The MR is in internal review right now. The commit will be reflected on GitHub once the MR gets merged internally. Thanks for your contribution.

hmalpani:
Merged into master with e40b1402e44d512120e12e7ff97194adf5827a86

PR #13764, opened 2024-05-08
Document additional errs returned by esp_ota_write (IDFGH-12784)

lucic71:
Otherwise, as in the case of esp_partition_write, we can say: "or one of error codes from lower-level partitions driver".

### DIFF
--- a/components/app_update/include/esp_ota_ops.h
+++ b/components/app_update/include/esp_ota_ops.h
@@ -116,6 +116,8 @@ esp_err_t esp_ota_begin(const esp_partition_t* partition, size_t image_size, esp
  *    - ESP_ERR_OTA_VALIDATE_FAILED: First byte of image contains invalid app image magic byte.
  *    - ESP_ERR_FLASH_OP_TIMEOUT or ESP_ERR_FLASH_OP_FAIL: Flash write failed.
  *    - ESP_ERR_OTA_SELECT_INFO_INVALID: OTA data partition has invalid contents
+ *    - ESP_ERR_INVALID_SIZE: if write would go out of bounds of the partition
+ *    - or one of error codes from lower-level flash driver.
  */
 esp_err_t esp_ota_write(esp_ota_handle_t handle, const void* data, size_t size);
 


mahavirj (approved):

mahavirj:
sha=ea407fb82ec75a2e666713607e8892f0ffa91556

components/app_update/include/esp_ota_ops.h
```diff
@@ -116,6 +116,8 @@ esp_err_t esp_ota_begin(const esp_partition_t* partition, size_t image_size, esp
  *    - ESP_ERR_OTA_VALIDATE_FAILED: First byte of image contains invalid app image magic byte.
  *    - ESP_ERR_FLASH_OP_TIMEOUT or ESP_ERR_FLASH_OP_FAIL: Flash write failed.
  *    - ESP_ERR_OTA_SELECT_INFO_INVALID: OTA data partition has invalid contents
+ *    - ESP_ERR_INVALID_SIZE: if write would go out of bounds of the partition
+ *    - ESP_ERR_NOT_ALLOWED: if partition is read-only
```
hmalpani, line 120: `esp_ota_write()` will never return `ESP_ERR_NOT_ALLOWED` as the partitions of subtype `app` are never read-only. For reference: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/partition-tables.html#flags. I think we can remove this.

I think we can just add the change below as a placeholder for other flash related error codes:
```suggestion
 *    - or one of error codes from lower-level flash driver.
```

hmalpani (changes requested):
Hello @lucic71 
Thanks for updating the documentation. Just one comment otherwise LGTM. Can you please update the PR?
Thanks!

hmalpani:
Hello @lucic71 
Thanks for adding the suggested changes. Can you please squash the commits?

lucic71:
Sure, I can to that, @hmalpani.

hmalpani:
sha=6a8d3af89b9e2f056a48bb5a5fbee6f4ab9c6021

mahavirj:
Fixed by 37d9a3c183cc1900e797fac8b0dcbe8043ff54ae

Issue #13749, opened 2024-05-05
Persist DMA buffer in sdmmc_read_sectors/sdmmc_write_sectors


ducalex:
### Is your feature request related to a problem?

When internal memory runs out, SD card reads and writes start returning `ESP_ERR_NO_MEM` because this line fails:

`void* tmp_buf = heap_caps_malloc(block_size, MALLOC_CAP_DMA);`

In theory `SPIRAM_MALLOC_RESERVE_INTERNAL` should ensure that memory is always available but in practice it doesn't always work.

There are likely also performance implications to allocating/deallocating all the time (in some cases).

Having no DMA-capable memory left is obviously an undesirable situation in general, but it can happen and when it does we might still need SD card access.

### Describe the solution you'd like.

My understanding is that block_size is at most 512 bytes: `out_csd->sector_size = MIN(read_bl_size, 512);`.

Therefore there is no reason to not keep the buffer around after it's first allocated for future use.  In fact with that size it could be allocated preemptively when the SD card is initialized, or even placed in static BSS. Because if it's needed once, it will very likely be needed again!

### Describe alternatives you've considered.

My current solution is this:

1. Reserve a modest DMA-capable memory block on start
2. Wrap `sdmmc_host_t's do_transaction()` to detect errors
3. If an error occurs, free the reserved block and try the transaction again
4. If the transaction succeeds, try to grab the block back

But this is far from ideal and doesn't always work...

### Additional context.

#6596 suggested the same but was closed because of the emphasis on performance. I think the emphasis should be on stability improvements if this change is introduced.

huming2207:
> My understanding is that block_size is at most 512 bytes

Hmm I think that's not true? There are quite a lot of those big SD cards that comes with 1024 or even 4096 bytes sector.

rrtandler:
Hi @ducalex,
From what you have described so far, it looks like the rest of application is using DMA capable memory space so heavily, that there isn't even 512B block left at certain moment, when it comes to SD card access. I think keeping the buffer for SD allocated statically aka "forever" would just expose another place in the code, where the DMA capable memory won't be available when needed. Could you check the rest of your code for source of potential leak / fragmentation of DMA capable RAM ?

ducalex:
> Hmm I think that's not true? There are quite a lot of those big SD cards that comes with 1024 or even 4096 bytes sector.

[...]

PR #13767, opened 2024-05-09
sdmmc: fix miscalculated sector size

huming2207:
Previous implementation will restrict the SD card sector size to be no more than 512 bytes. However 512 bytes is the minimum size for some old generic SD cards, and some new ones may come with 1024 or 4096 bytes per sector. We need to support those as well.

This (partially) fixes: https://github.com/espressif/esp-idf/issues/13749#issuecomment-2101695716


igrr:
@huming2207 thanks for the PR! The fix looks correct to me, we'll merge it soon.

> However 512 bytes is the minimum size for some old generic SD cards, and some new ones may come with 1024 or 4096 bytes per sector.

Could you please mention the source for this statement? It doesn't seem to match SD specification. In SD v1 standard, the card provided read_bl_len and write_bl_len fields in the CSD register. These indicated _maximum_ supported read and write block sizes, respectively. The actual block size would default to 512 bytes and could be set using CMD16, up to the limit specified in the CSD.

In SD v2 standard, these fields in CSD are no longer variable. E.g.,

> READ_BL_LEN
> This field is fixed to 9h, which indicates READ_BL_LEN=512 Byte.

The description of CMD16 says the same:

> SET_BLOCKLEN
> In case of SDSC Card, block length is 
> set by this command.
> In case of SDHC and SDXC Cards, 
> block length of the memory access 
> commands are fixed to 512 bytes.

huming2207:
Hi @igrr 

> Could you please mention the source for this statement? It doesn't seem to match SD specification. 
 
Sorry that was straight out from my memory, I've double checked and I think I have misunderstood the standard and I thought all SD card sector size are fixed to whatever it was provided in the `READ_BL_LEN` in CSD. I should have read the spec first! 😅

Meanwhile indeed on some SD cards, the `READ_BL_LEN` can be 9 or 10 which means `1<<9` or `1<<10` and it means 512 or 1024 bytes for the maximum read block size. For example this 8Gbit SD NAND chip's  `READ_BL_LEN` is 1024 bytes:

image

Based on the spec, I guess the original implemetation (which fixed to 512) should still works, but it may affect the I/O performance?

Regards,
JAckson

igrr:
> Based on the spec, I guess the original implemetation (which fixed to 512) should still works, but it may affect the I/O performance?

I think I have this chip somewhere, I'll give it a try. I guess only the total size of the transfer should matter for performance, not the the block size.

The current implementation effectively always sets card->sector_size to 512, regardless of read_blk_len.
https://github.com/espressif/esp-idf/blob/d4cd437ede613fffacc06ac6d6c93a083829022f/components/sdmmc/sdmmc_sd.c#L387-L391

So we are not using the possibility to increase the read/write block lengths, always setting the effective block length to 512:
https://github.com/espressif/esp-idf/blob/d4cd437ede613fffacc06ac6d6c93a083829022f/components/sdmmc/sdmmc_cmd.c#L273-L278

Basically, the existing code should work even on cards with read_bl_len > 9. But anyway, your fix is logically correct, so we'll change this, even though this is effectively dead code...

igrr:
@huming2207 Somewhat unrelated, I find that XTSD01GLGEAG is marked as ["discontinued" at LCSC](https://www.lcsc.com/product-detail/FLASH_XTX-XTSD01GLGEAG_C558837.html). And XTX website mentions a different variant, XTSDG01GWSIGA, but it's [not in stock](https://www.lcsc.com/product-detail/NAND-FLASH_XTX-XTSDG01GWSIGA_C7429710.html).

huming2207:
> The current implementation effectively always sets card->sector_size to 512, regardless of read_blk_len.

Ah... this will definitely cause some undefined behaviours on one of our products, so it needs to be fixed. For that product we need to deal with some real-time data. We can't use any filesystems because they are often too slow, so we wrote our own block-level storage instead. We do rely on the `sector_size` to cache the data. But so far we are using 512-byte sector SD NAND chips only, like `CSNP1GCR01-AOW`, so we are still safe for now.

image

> Somewhat unrelated, I find that XTSD01GLGEAG is marked as ["discontinued" at LCSC](https://www.lcsc.com/product-detail/FLASH_XTX-XTSD01GLGEAG_C558837.html). And XTX website mentions a different variant, XTSDG01GWSIGA, but it's [not in stock](https://www.lcsc.com/product-detail/NAND-FLASH_XTX-XTSDG01GWSIGA_C7429710.html).

We haven't reached out to XTX for now, and we never tested their chips in a larger scale. We only have a few demo devices with their 1Gbit chip assembled but it seems working anyway. But I can't remember the part number and I'm not sure which one it is...😅 

We planned to use [Create Storage World](https://www.lcsc.com/products/NAND-FLASH_11266.html?brand=12587)'s chip for production later. The CSW's chip seems to always report 512 bytes on their `read/write_bl_len` no matter what capacity it is.

igrr:
> this will definitely cause some undefined behaviours on one of our products, so it needs to be fixed. For that product we need to deal with some real-time data. We can't use any filesystems because they are often too slow, so we wrote our own block-level storage instead. We do rely on the sector_size to cache the data. 

Sorry, could you explain why this will cause undefined behavior? To be clear, it is totally supported by SD standard to set the read/write block length to 512 bytes for SDSC cards, regardless of the _maxium supported_ read_blk_len/write_blk_len.  SDSC cards allow for the read/write block length to be larger than 512 bytes, but the host can still set the actual block length to 512. Linux and Zephyr SDMMC stacks to the same thing. In fact, neither of them even issue MMC_SET_BLOCKLEN command in the standard card initialization flow.

If you use `card->sector_size` value, then everything should work correctly, as that's the same value that SDMMC driver will use when sending data transfer commands.

huming2207:
> Sorry, could you explain why this will cause undefined behavior? 

Hmm nevermind, sorry I double checked our code against the SD spec and I think I might have misunderstood the spec again... It actually doesn't matter at all even with a "wrongly" reported `sector_size`. 😅

igrr:
I got one of those SD NAND chips with csd_ver=1 (i.e. SDSC) and read_bl_len == 10. On a quick look, the tests seem to pass, but I've reopened the internal issue to verify that everything works correctly.

PR #13770, opened 2024-05-09
feat(espcoredump): Include boot time and unix time in header

nebkat:
*Unfinished and untested but creating draft to gather feedback. Will create corresponding patch for esp-coredump python.*

Core dumps do not contain any form of timestamp to help identify when a crash occurred.

This patch adds two new fields to the core dump header - boot time and unix time:
- Boot time (microseconds since boot) is useful in determining whether a crash occurred immediately after boot or many hours in, which can aid in debugging. Observing the core dump data can indirectly provide this information but requires additional steps. In applications with known bugs (e.g. memory leaks) knowing this in advance can save time by reducing unnecessary debugging.
- Unix time (microseconds since epoch) may provide hints as to why the crash occurred in applications that depend on external inputs, and is useful when organizing files, both on local and remote storage. In my application we copy core dumps to an SD card, but we have no way of knowing whether a particular core dump occurred 1 week or 1 year ago.

### DIFF
--- a/components/espcoredump/include_core_dump/esp_core_dump_types.h
+++ b/components/espcoredump/include_core_dump/esp_core_dump_types.h
@@ -66,8 +66,8 @@ extern "C" {
                                                 (((_maj_)&0xFF) << 8) | \
                                                 (((_min_)&0xFF) << 0) \
                                             )
-#define COREDUMP_VERSION_BIN                0
-#define COREDUMP_VERSION_ELF                1
+#define COREDUMP_VERSION_BIN                1
+#define COREDUMP_VERSION_ELF                2
 
 /* legacy bin coredumps (before IDF v4.1) has version set to 1 */
 #define COREDUMP_VERSION_BIN_LEGACY         COREDUMP_VERSION_MAKE(COREDUMP_VERSION_BIN, 1) // -> 0x0001
@@ -146,6 +146,8 @@ typedef struct _core_dump_header_t {
     uint32_t tcb_sz;    /*!< Size of a TCB, in bytes */
     uint32_t mem_segs_num; /*!< Number of memory segments */
     uint32_t chip_rev; /*!< Chip revision */
+    uint64_t boot_time; /*!< Boot time, in microseconds */
+    uint64_t unix_time; /*!< Unix time, in microseconds */
 } core_dump_header_t;
 
 /**
--- a/components/espcoredump/src/core_dump_binary.c
+++ b/components/espcoredump/src/core_dump_binary.c
@@ -185,6 +185,11 @@ static esp_err_t esp_core_dump_write_binary(void)
     hdr.version   = COREDUMP_VERSION_BIN_CURRENT;
     hdr.tcb_sz    = tcb_sz;
     hdr.chip_rev  = efuse_hal_chip_revision();
+    dump_hdr.boot_time = esp_timer_get_time();
+
+    struct timeval tv;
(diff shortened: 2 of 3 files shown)


components/espcoredump/include_core_dump/esp_core_dump_types.h
```diff
@@ -146,6 +146,8 @@ typedef struct _core_dump_header_t {
     uint32_t tcb_sz;    /*!< Size of a TCB, in bytes */
     uint32_t mem_segs_num; /*!< Number of memory segments */
     uint32_t chip_rev; /*!< Chip revision */
+    uint64_t boot_time; /*!< Boot time, in microseconds */
+    uint64_t unix_time; /*!< Unix time, in microseconds */
```
igrr, line 150: To avoid changing the header format (which requires changing the COREDUMP_VERSION...), could we keep this information in an ELF info segment? That way, the change would not cause compatibility issues even if an older version of esp-coredump package is used to decode the new core dump binary.

(Of course, this feature would be specific to the ELF format, but BIN format is present for backwards compatibility only, so not adding the feature there is okay.)

components/espcoredump/include_core_dump/esp_core_dump_types.h
```diff
@@ -66,8 +66,8 @@ extern "C" {
                                                 (((_maj_)&0xFF) << 8) | \
                                                 (((_min_)&0xFF) << 0) \
                                             )
-#define COREDUMP_VERSION_BIN                0
-#define COREDUMP_VERSION_ELF                1
+#define COREDUMP_VERSION_BIN                1
+#define COREDUMP_VERSION_ELF                2
```
igrr, line 70: That's not correct, unfortunately — this changes the definition of COREDUMP_VERSION_BIN_LEGACY a few lines below, meaning that the coredumps in legacy format won't be encoded correctly, and older versions of esp-coredump won't be able to read them.

components/espcoredump/src/core_dump_binary.c
```diff
@@ -185,6 +185,11 @@ static esp_err_t esp_core_dump_write_binary(void)
     hdr.version   = COREDUMP_VERSION_BIN_CURRENT;
     hdr.tcb_sz    = tcb_sz;
     hdr.chip_rev  = efuse_hal_chip_revision();
+    dump_hdr.boot_time = esp_timer_get_time();
+
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
```
igrr, line 191: I'm not sure it is safe to call `gettimeofday` from the panic handler. Internally, `gettimeofday` may attempt to acquire a mutex (e.g. [here](https://github.com/espressif/esp-idf/blob/d4cd437ede613fffacc06ac6d6c93a083829022f/components/newlib/time.c#L92-L94) and a couple other places), which in general might not be possible in the panic handler. (Panic handler might be invoked from a high-priority ISR, such as an interrupt WDT ISR).

igrr: Perhaps the best alternative I can suggest is to handle this at the application level.

You can designate a static variable with `COREDUMP_DRAM_ATTR` to make sure it gets saved in the coredump. Then, set up a periodic timer to save the `gettimeofday` output into that variable. If a crash happens, you will get the latest value of that variable inside the dump.

igrr:
Thanks for the nice improvement idea @nebkat, I left some comments...

cc @erhankur

PR #13792, opened 2024-05-14
Air-222 - ESP OTA Fail Message Improvements (IDFGH-12818)

tleppien:
Improved the OTA failed messages.


mahavirj:
@tleppien Fix is not clear from the change-log here. Can you please check and fix the PR?

PR #13834, opened 2024-05-22
Nushell export script

portasynthinca3:
Allows [Nushell](https://www.nushell.sh/) users such as myself to use the ESP-IDF without the need to start a separate POSIX-compliant shell. `idf.py build` seems to be building my projects correctly after sourcing this script using `source-env /path/to/export.nu`.

 I'm quite new to this language and I might have missed some things. If I have, let me know!

### DIFF
--- a/export.nu
+++ b/export.nu
@@ -1,0 +1,97 @@
+# Tests whether a path is an IDF path
+def is_idf [] {
+    ($in | path join tools/idf.py | path exists) or ($in | path join tools/idf_tools.py | path exists)
+}
+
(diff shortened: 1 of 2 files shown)


dobairoland:
Hi @portasynthinca3. Thank you for the contribution. It looks pretty nice but we are hesitant to accept it because:

1. we are not sure how many of our customers would be interested in using it,
2. maintaining this scripts are hard as there couldn't be easily tested.

For this reason, I've opened https://github.com/espressif/esp-idf/issues/13868 as well for public survey.  We will get back to this PR later.....

Thank you for understanding!

PR #13856, opened 2024-05-25
Make spi_bus_free() abort early if something goes wrong (IDFGH-12895)

MathyV:
If for example there are still devices configured on the bus spi_bus_free() will report that with ESP_ERR_INVALID_STATE, but not before it frees all other things like IO and DMA. This means that after the function exits any other task that is still running and using those devices will all of a sudden stop working (and probably crash the device).

This commit prevents that situation by making the function return early so it can be called again safely later when all devices have been removed.


suda-morris:
sha=397aa282ce474823727d27dd19ba5b0c07d10f9a

Issue #3840, opened 2019-07-26
Insufficient TCP connections in ESP32-WROVER-B with latest SDK (IDFGH-1581)


surfskidude:
## Problem Description

I get the following error when I run a fairly heavy duty network application that is simultaneously acting as an HTTP server and HTTP client when connecting a browser.

W (338406) wifi: mem fail : 
W (338466) wifi: alloc eb len=24 type=3 fail, heap:4140676

After the above failure, I tried several configurations and they all failed.

I also expected the the ESP32-WROVER-B to be able to do many TCP connections, but I noticed that I can only configure up to 16 connections using make menuconfig. 

Do you have plans on supporting more connections? I would like to run up to 100 full duplex TCP connections.

## Environment

- Development Kit:   ESP32-WROVER-B
- Module or chip used:  WINGONEER T-Koala ESP32 
- IDF version: v4.0-dev-1191-g138c941fa
- Build System:         CMake
- Compiler version:  crosstool-NG esp32-2019r1) 8.2.0
- Operating System:     Linux
- Power Supply:         USB

negativekelvin:
https://github.com/espressif/esp-idf/issues/3592

freakyxue:
@surfskidude This problem has been fixed and synchronized to the hub, please test

surfskidude:
I noticed that max TCP connections are still 16 when using make menuconfig. Maybe the tool has not been updated?

I have not looked into manually crafting the config file so can you please send me an sdkconfig file configured for high end TCP use?

BTW, just to make sure, I did a new checkout as follows:
 git clone --recursive https://github.com/espressif/esp-idf.git

freakyxue:
@surfskidude 
sorry ，I didn't see what you need。
You want to configure 100 TCP sockets。
But IDF can only configure 16 maximum values through menuconfig. If too many settings are configured, RAM space bu'gou will be generated。

config LWIP_MAX_SOCKETS
        int "Max number of open sockets"
        range 1 16
        default 10

if  you  want to config 100 sockets, you can modify the value,"range 1 100"

jack0c:
@surfskidude are still working on this? I will close this issue, feel free to reopen it if necessary.

PR #13866, opened 2024-05-27
change: bump max sockets in lwip Kconfig (IDFGH-12906)

rdgoprslv:
Addresses #3840 

Tested under ESP32S3, running 50 tasks, each with a client, and a server listening to one client, totalizing 150 sockets/files open, forwarding data from PPP to Ethernet and vice versa.

Reasoning for 255 total sockets/files is VFS ``local_fd_t`` being a ``uint8_t``.

Since FD_SETSIZE is defined at tool level, a CMakeLists user definition is needed in cases above 64 (the default value).

### DIFF
--- a/components/lwip/Kconfig
+++ b/components/lwip/Kconfig
@@ -135,13 +135,19 @@ menu "LWIP"
 
     config LWIP_MAX_SOCKETS
         int "Max number of open sockets"
-        range 1 16
+        range 1 253
         default 10
         help
+            The practical maximum limit is
+            determined by available heap memory at runtime.
+
             Sockets take up a certain amount of memory, and allowing fewer
             sockets to be open at the same time conserves memory. Specify
             the maximum amount of sockets here. The valid value is from 1
-            to 16.
+            to 253. If using value above 61, update CMakeLists defining
+            FD_SETSIZE to the number of sockets used plus the
+            expected open files (minimum of +3 for stdout, stderr and stdin).
+
 
     config LWIP_USE_ONLY_LWIP_SELECT
         bool "Support LWIP socket select() only (DEPRECATED)"
@@ -669,7 +675,7 @@ menu "LWIP"
                 LWIP_TCP_RECCVMBOX_SIZE packets for each TCP socket, so the maximum possible cached TCP packets
                 for all TCP sockets is LWIP_TCP_RECCVMBOX_SIZE multiples the maximum TCP socket number. In other
                 words, the bigger LWIP_TCP_RECVMBOX_SIZE means more memory.
-                On the other hand, if the receiv mail box is too small, the mail box may be full. If the
+                On the other hand, if the receive mail box is too small, the mail box may be full. If the
                 mail box is full, the LWIP drops the packets. So generally we need to make sure the TCP
                 receive mail box is big enough to avoid packet drop between LWIP core and application.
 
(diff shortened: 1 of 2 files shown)


david-cermak (approved):
Thanks for this PR, this is very helpful, including the description and note about `FD_SETSIZE`!

Could you please squash the commits into one?

rdgoprslv:
Done.

atanisoft:
I'll agree this is a cool PR, but how do we handle `FD_SETSIZE`? Is it possible to define this generically when socket count exceeds the declared limit (61) before requiring an override?

Also what about the case of other VFS usages that are not sockets? Can that be confirmed as still working when using 61 sockets (as example)?

rdgoprslv:
I agree it would be ideal that Kconfig itself defined ``FD_SETSIZE``,  but I don't know enough about Kconfig to make it happen.

Also, I spotted a off-by-one error, ``local_fd_t`` supports 256 files, not 255 as previously stated. This add one to the theoretical maximum allowed ``FD_SETSIZE`` and ``CONFIG_LWIP_MAX_SOCKETS``.

## About testing

- Hardware:
  - Chip is ESP32-S3 (QFN56) (revision v0.2)
    - Features: WiFi, BLE, Embedded PSRAM 2MB (AP_3v3)
    - Crystal is 40MHz

PSRAM being used for configuration data only.

- Firmware:
  - Configurations used in every scenario:
    ``CONFIG_LWIP_MAX_ACTIVE_TCP=255``
    ``CONFIG_LWIP_MAX_LISTENING_TCP=255``
  - Running a Port Forwarder from PPP to Ethernet.
  - FS used for files is ``littlefs``.

### First scenario

``FD_SETSIZE=64`` (tool default)
``CONFIG_LWIP_MAX_SOCKETS=61``


Results:

- Simultaneously opened 61 listening sockets successfully, trying to accept a connection fails, as expected.
- Simultaneously opened 21 listening sockets, 20 accepted clients, and 20 sockets connected as a client successfully. Trying to accept the 21st client fails, as expected.
- Since ``FD_SETSIZE=64`` and ``CONFIG_LWIP_MAX_SOCKETS=61``, that leaves room for 3 "regular" FDs, used by ``stdout``, ``stdin`` and ``stderr``, trying to open a file via littlefs fails, as expected.

### Second scenario

``FD_SETSIZE=64``
``CONFIG_LWIP_MAX_SOCKETS=32``

Results:

- Simultaneously opened 16 files, 8 listening sockets, 8 accepted clients, and 8 sockets connected as a client successfully.

### Third scenario

``FD_SETSIZE=256``
``CONFIG_LWIP_MAX_SOCKETS=253``

Results:

- Simultaneously opened 50 listening sockets, 50 accepted clients, and 50 sockets connected as a client successfully. Hardware RAM limitations prevents opening more sockets.
- Since ``FD_SETSIZE=256`` and ``CONFIG_LWIP_MAX_SOCKETS=253``, that leaves room for 3 "regular" FDs, used by ``stdout``, ``stdin`` and ``stderr``, trying to open a file via littlefs fails, as expected.

### Fourth scenario

``FD_SETSIZE=256``
``CONFIG_LWIP_MAX_SOCKETS=200``

Results:

- Simultaneously opened 25 files, 25 listening sockets, 25 accepted clients, and 25 sockets connected as a client successfully. Hardware RAM is the limit.

Further optimizations on port forwarder firmware would allow bigger numbers, but I think this is enough for now.

euripedesrocha:
Hi @rdgoprslv thanks for the contribution. The changes were merged in 48295d3

atanisoft:
Any chance of a back port to 5.3?

PR #13879, opened 2024-05-29
esp_partition.h uses the esp_flash_t and should therefore include esp_flash.h (IDFGH-12919)

M-Bab:
Found by unit testing. esp_partition.h needs the type definition of esp_flash_t which is found in esp_flash.h.

This commit partially reverts c9c7573f715534edf19e8290bc10d2a5d134c4a8


M-Bab:
Okay that may break the regular build. Not good :thinking:

PR #13890, opened 2024-06-01
fix(bootloader_support): Add missing C linkage to some headers (IDFGH-12930)

andylinpersonal:
I'm developing some bootloader hooks in C++ and just found some headers lack ```extern "C" {}``` guard.
This issue can also be found in `release/v5.3` and `master`.


KonstantinKondrashov:
Hi @andylinpersonal!
Seems like we can take these changes to master. Backports will not be created though because it is not a critical issue.
I see that you made some changes. Is it the final version?
Have you achieved your goal with these changes? Ok, anyway I will check the rest bootloader headers if it is needed there as well.

andylinpersonal:
Oh it's the final version. I've just removed some accidentally re-formatted code.

andylinpersonal:
Fixed in 788455f3a80babac07d30fff55f31d3ed55028dc

PR #13911, opened 2024-06-04
Create jekyll-gh-pages.yml (IDFGH-12955)

LLotme:
..


LLotme (commented):
gh pr checkout 13911

dobairoland:
I'm sorry but I don't see the point behind this PR. The Jekyll part was already rejected by closing https://github.com/espressif/esp-idf/pull/13895. As for the Python package dependencies, versions should not be handled in those files. We use constraint files as documented at https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-tools.html.

LLotme (commented):
gh pr checkout 13911

Issue #9615, opened 2022-08-23
Why are you using int64_t for esp_timer_get_time output? (IDFGH-8116)


amirhb29:
Why are you using int64 for timer counter when we have 64 bits and we can uint64 to cover more time before overflowing the counter?

igrr:
@amirhb29 Good question! Since esp_timer counter is counting microseconds, the difference between a 63-bit and 64-bit counter is the difference between device uptime of ~292 thousand years and 584 thousand years. We thought that for most devices built with ESP-IDF, this difference wouldn't matter!

In fact, on ESP32-C3 and later chips, the underlying SYSTIMER peripheral has "just" a 52-bit counter, and it counts at 16 MHz frequency. This translates to approximately 9 years of _uninterrupted_ uptime, which corresponds to a 48-bit microsecond counter.

On the other hand, returning the int64_t makes it much easier to do time difference calculations: you can calculate `t_1 - t_0` without having to check whether `t_1` is greater than `t_0` and without potentially incorrect casts between uint64_t and int64_t.

ESP-Marius:
Closing this one as the original question has been answered. Feel free to reopen if there is anything that is still unclear.

PR #13931, opened 2024-06-07
fix(timer): change return type to uint64_t instead of int64_t (IDFGH-12980)

leandromattioli:
If we "trace" the implementation of the esp_timer_get_time() function, we eventually find out that it calls a tick_to_us_func_t, which is a function that returns an unsigned 64-bit wide integer.

Eventually, the uint64_t is cast to int64_t (apparently this is a mistake).

I have refactored the code replacing all int64_t related to calls to these functions to uint64_t.

It should be noted that with 64 bits we won't have overflow after about 70 minutes, so I think this is an improvement.

I have successfully built the example esp_timer after my changes, but an exhaustive test for all examples involved might be needed, if it is not done automatically with CI.

### DIFF
--- a/components/esp_timer/include/esp_timer.h
+++ b/components/esp_timer/include/esp_timer.h
@@ -220,7 +220,7 @@ esp_err_t esp_timer_delete(esp_timer_handle_t timer);
  * @brief Get time in microseconds since boot
  * @return Number of microseconds since the initialization of ESP Timer
  */
-int64_t esp_timer_get_time(void);
+uint64_t esp_timer_get_time(void);
 
 /**
  * @brief Get the timestamp of the next expected timeout
--- a/components/efuse/test_apps/main/test_efuse.c
+++ b/components/efuse/test_apps/main/test_efuse.c
@@ -670,9 +670,9 @@ static void task2(void* arg)
 {
     xSemaphoreTake(sema, portMAX_DELAY);
     uint8_t mac[6];
-    int64_t t1 = esp_timer_get_time();
+    uint64_t t1 = esp_timer_get_time();
     TEST_ESP_OK(esp_efuse_read_field_blob(ESP_EFUSE_MAC_FACTORY, &mac, sizeof(mac) * 8));
-    int64_t t2 = esp_timer_get_time();
+    uint64_t t2 = esp_timer_get_time();
     int diff_ms = (t2 - t1) / 1000;
     TEST_ASSERT_GREATER_THAN(diff_ms, delay_ms);
     ESP_LOGI(TAG, "read MAC address: %02x:%02x:%02x:%02x:%02x:%02x", mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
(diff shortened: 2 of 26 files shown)


components/esp_timer/include/esp_timer.h
```diff
@@ -220,7 +220,7 @@ esp_err_t esp_timer_delete(esp_timer_handle_t timer);
  * @brief Get time in microseconds since boot
  * @return Number of microseconds since the initialization of ESP Timer
  */
-int64_t esp_timer_get_time(void);
+uint64_t esp_timer_get_time(void);
```
0xjakob, line 223: Note that this would be a breaking change.

0xjakob (commented):
Hi @leandromattioli , thanks for your contribution. I took a brief look at the changes. Overall, they look reasonable to me. However, the change in `esp_timer.h` would be a breaking one. I'm afraid we can't accept breaking changes, except when changing the major release number (next would be version 6.0).

I'm also not sure if the choice of `int64_t` in the timer code was just an accident. We'll discuss this PR in more detail soon and come back here.

igrr:
https://github.com/espressif/esp-idf/issues/9615#issuecomment-1224181519

0xjakob:
@leandromattioli Regarding the signed vs. unsigned type choice, I think the commented posted by @igrr explains why we chose `int64_t` in purpose. Does this make sense to you?

leandromattioli:
In my opinion, there's still an inconsistency.

All different implementations of `esp_timer_impl_get_time` are returning int64_t, but they are merely calling functions which return uint64_t. So there's already a cast from uint64_t to int64_t happening.

# Implementation A (esp_timer_impl_lac.c)
`
int64_t IRAM_ATTR esp_timer_impl_get_time(void)
{
    return esp_timer_impl_get_counter_reg() / TICKS_PER_US;
}

// uint64_t IRAM_ATTR esp_timer_impl_get_counter_reg(void)
`

# Implementation B (esp_timer_impl_systimer.c)
`
int64_t IRAM_ATTR esp_timer_impl_get_time(void)
{
    // we hope the execution time of this function won't > 1us
    // thus, to save one function call, we didn't use the existing `systimer_hal_get_time`
    return systimer_hal.ticks_to_us(systimer_hal_get_counter_value(&systimer_hal, SYSTIMER_COUNTER_ESPTIMER));
}
// systimer_hal hat a pointer to a function having this prototype (note the unsigned values)
//typedef uint64_t (*us_to_ticks_func_t)(uint64_t us);    // prototype of function to convert microseconds to ticks
`

# Implementation C (esp_timer_impl_systimer.c; identical to implementation B)
int64_t IRAM_ATTR esp_timer_impl_get_time(void)
{
    // we hope the execution time of this function won't > 1us
    // thus, to save one function call, we didn't use the existing `systimer_hal_get_time`
    return systimer_hal.ticks_to_us(systimer_hal_get_counter_value(&systimer_hal, SYSTIMER_COUNTER_ESPTIMER));
}
// systimer_hal hat a pointer to a function having this prototype (note the unsigned values)
//typedef uint64_t (*us_to_ticks_func_t)(uint64_t us);    // prototype of function to convert microseconds to ticks
`

I understand, though, the advantages of signed integer for computing the difference between two instants, but I really think that should be a responsibility of the "client" function who calls `esp_timer_impl_get_time`. There are several places in the code with such calculations, but I guess a proper way would be explicit casts to int64_t, wherever negative results are expected.

leandromattioli:
After looking more carefully, I agree with #9615 . We already have 63 bits for the positive part, which is already enough for the time axis! The 70 minutes limit concerns only the `micros` function in arduino-esp32, and the workaround is actually calling `esp_timer_get_time()` directly.

I'm closing this PR now.

PR #13932, opened 2024-06-07
feat(ulp): RVfplib as alternative floating point library for ULP RISC-V and LP Core (IDFGH-12981)

andylinpersonal:
Add RVfplib as alternative soft-fp library, which might roughly reduce the code size by 1~3 KiB if only basic single precision arithmetics are used.
Though licensed under GPLv3 with library exeption, RVfplib has already been embedded into latest Cx, Hx and P4's rom code so it should not be an issue.

### DIFF
--- a/.gitlab/ci/rules.yml
+++ b/.gitlab/ci/rules.yml
@@ -144,6 +144,7 @@
   - "components/bt/controller/lib_esp32c6/esp32c6-bt-lib"
   - "components/bt/controller/lib_esp32c5/esp32c5-bt-lib"
   - "components/bt/esp_ble_mesh/lib/lib"
+  - "components/ulp/RVfplib"
   - ".gitmodules"
 
 .patterns-idf-pytest-plugin: &patterns-idf-pytest-plugin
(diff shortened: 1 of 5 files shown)


ESP-Marius:
@andylinpersonal Thanks for the contribution!

Since this use-case is a bit more niche, I'm not sure we should include it in the ULP component (will be difficult for us to support all custom behavior users want)

How do you think about this compromise:

We could make it possible for users to have more direct control of the ULP build system, e.g. by letting users also provide their own cmakefile. Then use-cases like this could be achieved without IDF modification directly in the user's project.

What do you think? (this would also allow setting `-msave-restore` from your project)

andylinpersonal:
@ESP-Marius 

> We could make it possible for users to have more direct control of the ULP build system, e.g. by letting users also provide their own cmakefile. Then use-cases like this could be achieved without IDF modification directly in the user's project.

Yes, a more flexible and less intrusive build system for ULP is what we demand. And it will be useful for the more complex use cases in the following ESP32-P4's LP subsystem.

ESP-Marius:
@andylinpersonal We've pushed a feature that should hopefully allow achieving stuff like this much easier for users.

You can see an example [here])https://github.com/espressif/esp-idf/blob/master/examples/system/ulp/lp_core/build_system/main/ulp/CMakeLists.txt)

I think with that we'll consider this ticket closed, feel free to reopen it if you still think RVfplib should be available by default for the lp-core.

Or open a new ticket if you are unhappy/have any requests with regards to the ULP build-system.

PR #13943, opened 2024-06-10
Do not allocate client->if_name twice in esp_http_client_init. (IDFGH-12992)

fbp2m:
When if_name is specified in config an CONFIG_ESP_HTTP_CLIENT_ENABLE_HTTPS is set, client->if_name is allocated twice, and first one is never free.


fbp2m:
New PR #13945 on master branch.

PR #13945, opened 2024-06-10
fix(esp_http_client): Do not allocate client->if_name twice in esp_http_client_init. (IDFGH-12994)

fbp2m:
When if_name is specified in esp_http_client_config_t like this :
```
.url = url,
.if_name = &ifr,
};
```
and CONFIG_ESP_HTTP_CLIENT_ENABLE_HTTPS is set.
"client->if_name" is allocated twice because "init_common_tcp_transport" is called twice.
First allocation is never free.


hmalpani (approved):
LGTM

hmalpani:
sha=f80749b2bf600bb0d2c99b2b1ed0c7190a1c838a

Issue #13152, opened 2024-02-09
error while loading shared libraries: libSDL2-2.0.so.0: cannot open shared object file: No such file or director (IDFGH-12094)


diplfranzhoepfinger:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

error while loading shared libraries: libSDL2-2.0.so.0: cannot open shared object file: No such file or director

```
franz@franz-Ubuntu:~/master/esp-idf$ . ./export.sh
Detecting the Python interpreter
Checking "python3" ...
Python 3.11.6
"python3" has been detected
Checking Python compatibility
Checking other ESP-IDF version.
Not using an unsupported version of tool cmake found in PATH: 3.27.4. To use it, run '/usr/bin/python3 /home/franz/master/esp-idf/tools/idf_tools.py export --prefer-system'
ERROR: tool qemu-xtensa version esp_develop_8.1.3_20231206 is installed, but returned non-zero exit code (127) with error message:
/home/franz/.espressif/tools/qemu-xtensa/esp_develop_8.1.3_20231206/qemu/bin/qemu-system-xtensa: error while loading shared libraries: libSDL2-2.0.so.0: cannot open shared object file: No such file or directory

ERROR: tool qemu-riscv32 version esp_develop_8.1.3_20231206 is installed, but returned non-zero exit code (127) with error message:
/home/franz/.espressif/tools/qemu-riscv32/esp_develop_8.1.3_20231206/qemu/bin/qemu-system-riscv32: error while loading shared libraries: libSDL2-2.0.so.0: cannot open shared object file: No such file or directory

franz@franz-Ubuntu:~/master/esp-idf$ 

```

same issue here: 
https://github.com/espressif/idf-eclipse-plugin/issues/898

not happening with esp-idf https://github.com/espressif/esp-idf/releases/tag/v5.1.2

only happens on master

GTCLive:
This is so odd. A full week went by and the linux & plugin ide development environment remain broken without a single follow-up, acknowledgement and/or neither a complain. Makes you wonder if anyone dev espressif products on the linux side, so odd and worrying...

GTCLive:
[...]

PR #13949, opened 2024-06-11
Update linux-macos-setup.rst (IDFGH-12999)

diplfranzhoepfinger:
libslirp0 libsdl2-dev added 

see https://github.com/espressif/esp-idf/issues/13152


diplfranzhoepfinger:
but how can i deselect it then ? 

if i follow your guide, it is installed if i want or not !

dobairoland:
The Qemu tools are not installed by default. One has to run the commands on https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/qemu.html. First the missing dependencies are installed and then the qemu tools themself.

So given into account the above, I'm closing this pull request.

PR #13963, opened 2024-06-12
fix(i2c): Don't print developer notes to ESP_LOG (IDFGH-13015)

nebkat:
`ESP_LOG` shouldn't be used for developer notes - especially for default configuration (external pullups). Users can figure out these details from reading documentation.

### DIFF
--- a/components/esp_driver_i2c/i2c_master.c
+++ b/components/esp_driver_i2c/i2c_master.c
@@ -918,9 +918,6 @@ esp_err_t i2c_new_master_bus(const i2c_master_bus_config_t *bus_config, i2c_mast
     i2c_master->base->sda_num = bus_config->sda_io_num;
     i2c_master->base->pull_up_enable = bus_config->flags.enable_internal_pullup;
 
-    if (i2c_master->base->pull_up_enable == false) {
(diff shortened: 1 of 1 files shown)


suda-morris:
sha=d7be7c45e191cdb4046ee5f2fda51ed682ed0d57

IanBurwell:
+1 for this change, I know as a developer that I have pull-up resistors on my hardware and don't want to see the warnings addressed in this PR. I can get around this for now by setting the `i2c.master` log level to ERROR, but this seems like a workaround. Hopefully espressif adopts this change.

italocjs:
+1. this may cause confusion when board already has external pullups

UncleGrumpy:
I agree, but it may be useful as a debug level message. Maybe this is a custom board, not designed by the firmware developer, that uses pull-resistors (or the developer forgot to use the resistors when breadboarding to test a custom design). It is spammy in most cases, but could be a helpful reminder if debugging hardware issues is necessary.

luftaquila:
+1 here. It makes me feel like the chip can sense the existence of the external pull-up resistor and warns me it is missing, even though it's there.

arc12:
Is this due for merge into master any time soon?

mythbuster5:
sha=d7be7c45e191cdb4046ee5f2fda51ed682ed0d57

PR #13992, opened 2024-06-16
fix(fatfs): _FFCONF_DEFINED added to ffconf.h (IDFGH-13046)

denizzzka:
I don’t know if this affects compilation, but preprocessed .i files parsing is broken because of this

### DIFF
--- a/components/fatfs/src/ffconf.h
+++ b/components/fatfs/src/ffconf.h
@@ -1,3 +1,6 @@
+#ifndef _FFCONF_DEFINED
+#define _FFCONF_DEFINED
+
 #include "sdkconfig.h"
 
 /*---------------------------------------------------------------------------/
@@ -362,3 +365,5 @@ void ff_memfree(void*);
 #define disk_read           ff_disk_read
 #define disk_write          ff_disk_write
 #define disk_ioctl          ff_disk_ioctl
+
+#endif /* _FFCONF_DEFINED */


denizzzka:
cla-assistant isn't works for me

haberturdeur:
Hi, thank you for contributing!

Could you elaborate a bit about what the issue was?

components/fatfs/src/ffconf.h
```diff
@@ -1,3 +1,6 @@
+#ifndef _FFCONF_DEFINED
```
haberturdeur, line 1: ESP-IDF uses `#pragma once` guards, instead of the old `#ifndef` ones

haberturdeur (changes requested):

denizzzka:
> Could you elaborate a bit further about what the issue was?

Sure! 

(This is only experiment)
To create D language language bindings I use preprocessed `.i` files which can be obtained by adding such code to the project root `CMakeLists.txt`:
```cmake
     # Preserve *.i files
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -save-temps")
```

After this, files are obtained, in some of which (in `diskio_rawflash.c.i`, for example) I found (because my `.i` parser failed on them) that the definitions like
```c
void* ff_memalloc (unsigned msize);
```
are scattered twice. These duplicated definitions was sourced only from `ffconf.h`.

components/fatfs/src/ffconf.h
denizzzka: I looked in the neighboring files of this component and did same

haberturdeur: Fair enough.
It is a bit of an edge case, since FATFS is basically an "external component".

haberturdeur:
LGTM!

haberturdeur (approved):

PR #14010, opened 2024-06-18
fix(ulp): Write pin's output mode to the correct register (IDFGH-13065)

LonerDan:
Fixes register mixup. 
According to the [ESP32-S3 TRM](https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf) (pages 515 - 516), the output pin's mode is set in the `RTC_GPIO_PINn_REG`,
bit `RTC_GPIO_PINn_PAD_DRIVER`, not the `RTC_IO_TOUCH_PADn_REG` field `RTC_IO_TOUCH_PADn_DRV`, 
which instead controls the drive output strength.


SoucheSouche:
Hi @LonerDan, thanks for the PR. We will be looking into it as soon as possible.

sudeep-mohanty (approved):

SoucheSouche:
e5d9d1dd607c4eb26deb0652d1b7d1f291c29bd9

PR #14033, opened 2024-06-20
idf_component_get_property returns a more intuitive error (e.g. wolfSSL for esp-tls) (IDFGH-13087)

gojimmypi:
When a requested component does not exist, the cmake `idf_component_get_property` does not currently return an intuitive error. Instead, there's just something about incorrect arguments.

For example, when enabling wolfSSL for the ESP-IDF `esp-tls` component without following the installation as described in the [documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/protocols/esp_tls.html#how-to-use-wolfssl-with-esp-idf), a message similar to this appears:

```
-- Adding linker script C:/SysGCC/esp32/esp-idf/v5.2/components/soc/esp32/ld/esp32.peripherals.ld
CMake Error at C:/SysGCC/esp32/esp-idf/v5.2/tools/cmake/component.cmake:384 (__component_get_property):
  __component_get_property Function invoked with incorrect arguments for
  function named: __component_get_property
Call Stack (most recent call first):
  C:/SysGCC/esp32/esp-idf/v5.2/components/esp-tls/CMakeLists.txt:26 (idf_component_get_property)

-- Configuring incomplete, errors occurred!
System.Exception: CMake exited with code 1
   at s24.e(c a, Object b)
```

This PR updates the cmake `idf_component_get_property` function with a more intuitive error and a suggestion how to resolve:  _component_target property not found for esp-wolfssl (does it need to be installed manually?)_.

```
-- Adding linker script C:/SysGCC/esp32/esp-idf/v5.2/components/soc/esp32/ld/esp32.peripherals.ld
-- Found esp-wolfssl
CMake Error at C:/SysGCC/esp32/esp-idf/v5.2/tools/cmake/component.cmake:385 (message):
  component_target property not found for esp-wolfssl (does it need to be
  installed manually?)
Call Stack (most recent call first):
  C:/SysGCC/esp32/esp-idf/v5.2/components/esp-tls/CMakeLists.txt:26 (idf_component_get_property)


-- Configuring incomplete, errors occurred!
System.Exception: CMake exited with code 1
   at s24.e(c a, Object b)
```


tools/cmake/component.cmake
```diff
@@ -379,7 +379,11 @@ function(idf_component_get_property var component property)
     if(__GENERATOR_EXPRESSION)
         set(val "$<TARGET_PROPERTY:${component_target},${property}>")
     else()
-        __component_get_property(val ${component_target} ${property})
+        if("${component_target}" STREQUAL "")
```
igrr, line 382: I think we have to do this before the `__GENERATOR_EXPRESSION` check, so that the correct error is also given for the `__GENERATOR_EXPRESSION` case.

tools/cmake/component.cmake
```diff
@@ -379,7 +379,11 @@ function(idf_component_get_property var component property)
...
+            message(FATAL_ERROR "component_target property not found for ${component} (does it need to be installed manually?)")
```
igrr, line 383: Perhaps more simply

```suggestion
            message(FATAL_ERROR "Component ${component} not found")
```

(You can give a more user-friendly message about installing the component for WolfSSL case, where you try adding WolfSSL as a dependency. "Installing manually" might not be a relevant suggestion for other components.)

PR #14060, opened 2024-06-24
fix(openthread): fix RCP build to pass time sync and CSL options (IDFGH-13113)

### LINKED ISSUE #14059: thread RCP build does not correctly pass through TIME_SYNC and CSL options (IDFGH-13112)
### Answers checklist.

### General issue report

When building a RCP, it appears that OPENTHREAD_CONFIG_TIME_SYNC_ENABLE (and the related CSL options) are never passed through because they are not set properly in `openthread-core-esp32x-radio-config.h`.

This causes a RCP build without TIME_SYNC and MAC IE header support, even though the menuconfig option is set.

 I will add a PR which fix this, however, I note that even with these options set it seems that the RCP crashes anyway whenever the host sends a frame with IE fields.

no2chem:
This PR fixes an issue where the menuconfig option `OPENTHREAD_TIME_SYNC` is set but `OPENTHREAD_CONFIG_TIME_SYNC_ENABLE` is never defined during a RCP build, resulting in an RCP build without time sync enabled. The same bug applies for CSL.

Fixes #14059


chshu:
@no2chem Thank you for your contribution!

It's indeed a feature has not been supported in current SDK. CSL was introduced in Thread 1.2, but it has not yet seen widespread adoption in products. Time Sync is still an experimental feature and is not defined in the Thread specification.

Both the features are well supported by single SoC, and we will evaluate the feasibility of supporting them in RCP mode. Could you provide more details about your product use case? Is it based on ESP Thread BR?

no2chem:
Thanks @chshu -

We would like to use it with ESP thread BR - but the library is not open source so we are using ot-br-posix instead. 

We've run into some stability issues with ESP thread BR so we're not as confident in using it because we can't fix it, but since the RCP should be agnostic to whether ESP thread BR or ot-br-posix is running on the NCP.

But both features work in RCP mode once the changes in this PR are applied, and #14089 is used to fix a bug in the 802154 state machine.

chshu:
@no2chem The change looks good with 2 minor comments, we will do some verification internally, and merge it then.

For any stability issue you encountered, please test again with the latest [release/v5.1](https://github.com/espressif/esp-idf/tree/release/v5.1) branch, and feel free to file issues here: https://github.com/espressif/esp-thread-br.

PR #14089, opened 2024-06-27
fix(802.15.4): fix races in 802.15.4 driver (IDFGH-13148)

### LINKED ISSUE #14068: [ieee802154] transmit with ie fields results in wrong state during RX abort (IDFGH-13124)
### Answers checklist.

### IDF version.

v5.4-dev-826-g8760e6d2a7

### Espressif SoC revision.

esp32h2-20221101

### Operating System used.

Linux

Command line with idf.py

None

### Development Kit.

esp32-h2-devboard-m

### Power Supply used.

USB

It should not assert and send the frame with ie fields properly

An assert in `isr_handle_rx_abort`, `IEEE802154_ASSERT(s_ieee802154_state == IEEE802154_STATE_RX);` is hit.

### Steps to reproduce.

1. Apply PR #14060 to allow time sync messages to work
2. build RCP with time sync enabled
3. run otbr-posix with time sync enabled
4. change state to leader
5 when time service sends first message with header ie, the assert will occur.

### Debug Logs.

```plain
assert failed: 0x4080a08a
0x4080a08a: isr_handle_rx_abort at esp-idf/components/ieee802154/driver/esp_ieee802154_dev.c:543 (discriminator 1)
 (inlined by) ieee802154_isr at esp-idf/components/ieee802154/driver/esp_ieee802154_dev.c:689 (discriminator 1)

Stack dump detected
Core  0 register dump:
MEPC    : 0x408007a0  RA      : 0x40805d8a  SP      : 0x40810c60  GP      : 0x4080edc4
0x408007a0: panic_abort at esp-idf/components/esp_system/panic.c:463
0x40805d8a: __ubsan_include at esp-idf/components/esp_system/ubsan.c:311

TP      : 0x4081d3b0  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130
S0/FP   : 0x4080a08e  S1      : 0x00000010  A0      : 0x40810c64  A1      : 0x00000061
0x4080a08e: ieee802154_isr at esp-idf/components/ieee802154/driver/esp_ieee802154_dev.c:631 (discriminator 2)

A2      : 0x00000003  A3      : 0x40810c78  A4      : 0x00000001  A5      : 0x40818000
A6      : 0x00000000  A7      : 0x76757473  S2      : 0x00000002  S3      : 0x600a3000
S4      : 0x00300020  S5      : 0x00000004  S6      : 0x40813ad8  S7      : 0x40814058
S8      : 0x00000081  S9      : 0x40814128  S10     : 0x00000001  S11     : 0x00000010
T3      : 0x6e6d6c6b  T4      : 0x6a696867  T5      : 0x66656463  T6      : 0x62613938
MSTATUS : 0x00001881  MTVEC   : 0x40800001  MCAUSE  : 0x00000007  MTVAL   : 0x00000000
0x40800001: _vector_table at esp-idf/components/riscv/vectors_intc.S:54

MHARTID : 0x00000000


Backtrace:


panic_abort (details=0x40810c64 <xIsrStack+1428> "assert failed: 0x4080a08a") at esp-idf/components/esp_system/panic.c:463
463	    *((volatile int *) 0) = 0; // NOLINT(clang-analyzer-core.NullDereference) should be an invalid operation on targets
#0  panic_abort (details=0x40810c64 <xIsrStack+1428> "assert failed: 0x4080a08a") at esp-idf/components/esp_system/panic.c:463
#1  0x40805d8a in esp_system_abort (details=<optimized out>) at esp-idf/components/esp_system/port/esp_system_chip.c:92
Backtrace stopped: frame did not save the PC
ELF file SHA256: 38afeeef0
```

### More Information.

_No response_

no2chem:
This PR fixes several races in the 802.15.4 driver

Specifically:
1) it is possible to be in `IEEE802154_STATE_TX_ACK` state when `IEEE802154_EVENT_RX_SFD_DONE` is fired

2) In case of an RX stop (i.e. `IEEE802154_RX_ABORT_BY_RX_STOP`) and RX_ON_IDLE is enabled, calling `next_operation()` would result in RX restarting because the state machine would think that it is going idle when the desired condition is that RX stop. So don't call `next_operation()` when a stop is requested.

3) if `ieee802154_set_cmd` is called before `ieee802154_set_state` outside of ISR it is possible that an 802.15.4 event interrupt would fire before the state was changed and the state machine would crash due to wrong state assertion

Overall, these fixes seem to fix the issues I had in #14068 and I don't observe the crashes anymore. 

Fixes #14068 and #14055

### DIFF
--- a/components/ieee802154/driver/esp_ieee802154_dev.c
+++ b/components/ieee802154/driver/esp_ieee802154_dev.c
@@ -343,10 +343,8 @@ static void enable_rx(void)
 {
     set_next_rx_buffer();
     IEEE802154_SET_TXRX_PTI(IEEE802154_SCENE_RX);
-
-    ieee802154_set_cmd(IEEE802154_CMD_RX_START);
-
     ieee802154_set_state(IEEE802154_STATE_RX);
+    ieee802154_set_cmd(IEEE802154_CMD_RX_START);
 }
 
 static IRAM_ATTR void next_operation(void)
@@ -477,7 +475,7 @@ static IRAM_ATTR void isr_handle_rx_abort(void)
     case IEEE802154_RX_ABORT_BY_TX_ACK_STOP:
     case IEEE802154_RX_ABORT_BY_ED_STOP:
         // do nothing
-        break;
+        return;
     case IEEE802154_RX_ABORT_BY_SFD_TIMEOUT:
     case IEEE802154_RX_ABORT_BY_CRC_ERROR:
     case IEEE802154_RX_ABORT_BY_INVALID_LEN:
@@ -601,8 +599,7 @@ static void ieee802154_isr(void *arg)
 
     if (events & IEEE802154_EVENT_RX_SFD_DONE) {
         // IEEE802154_STATE_TX && IEEE802154_STATE_TX_CCA && IEEE802154_STATE_TX_ENH_ACK for isr processing delay
-        IEEE802154_ASSERT(s_ieee802154_state == IEEE802154_STATE_RX || s_ieee802154_state == IEEE802154_STATE_RX_ACK || s_ieee802154_state == IEEE802154_STATE_TX || s_ieee802154_state == IEEE802154_STATE_TX_CCA || s_ieee802154_state == IEEE802154_STATE_TX_ENH_ACK);
-
+        IEEE802154_ASSERT(s_ieee802154_state == IEEE802154_STATE_RX || s_ieee802154_state == IEEE802154_STATE_RX_ACK || s_ieee802154_state == IEEE802154_STATE_TX || s_ieee802154_state == IEEE802154_STATE_TX_CCA || s_ieee802154_state == IEEE802154_STATE_TX_ENH_ACK || s_ieee802154_state == IEEE802154_STATE_TX_ACK);
         s_rx_frame_info[s_rx_index].timestamp = esp_timer_get_time();
         esp_ieee802154_receive_sfd_done();
 
@@ -803,11 +800,11 @@ static inline esp_err_t ieee802154_transmit_internal(const uint8_t *frame, bool 
     IEEE802154_SET_TXRX_PTI(IEEE802154_SCENE_TX);
 
     if (cca) {
-        ieee802154_set_cmd(IEEE802154_CMD_CCA_TX_START);
         ieee802154_set_state(IEEE802154_STATE_TX_CCA);
+        ieee802154_set_cmd(IEEE802154_CMD_CCA_TX_START);
     } else {
-        ieee802154_set_cmd(IEEE802154_CMD_TX_START);
         ieee802154_set_state(IEEE802154_STATE_TX);
+        ieee802154_set_cmd(IEEE802154_CMD_TX_START);
     }
 
     ieee802154_exit_critical();


components/ieee802154/driver/esp_ieee802154_dev.c
```diff
@@ -803,11 +800,11 @@ static inline esp_err_t ieee802154_transmit_internal(const uint8_t *frame, bool
     IEEE802154_SET_TXRX_PTI(IEEE802154_SCENE_TX);
 
     if (cca) {
-        ieee802154_set_cmd(IEEE802154_CMD_CCA_TX_START);
         ieee802154_set_state(IEEE802154_STATE_TX_CCA);
+        ieee802154_set_cmd(IEEE802154_CMD_CCA_TX_START);
     } else {
-        ieee802154_set_cmd(IEEE802154_CMD_TX_START);
         ieee802154_set_state(IEEE802154_STATE_TX);
+        ieee802154_set_cmd(IEEE802154_CMD_TX_START);
```
zwx1995esp, line 807: Due to we have already entering the critical, all isrs related to IEEE802154 are disabled. So it might be no need to consider the order of setting state and start an operation on HW.

components/ieee802154/driver/esp_ieee802154_dev.c
```diff
@@ -477,7 +475,7 @@ static IRAM_ATTR void isr_handle_rx_abort(void)
     case IEEE802154_RX_ABORT_BY_TX_ACK_STOP:
     case IEEE802154_RX_ABORT_BY_ED_STOP:
         // do nothing
-        break;
+        return;
```
zwx1995esp, line 478: Here we need to set next operation because of the feature `pending tx`. If we want to send a packet while the HW is receiving a packet. We will enable all rx abort at this time and pending the transmition. If any aborts arrived to HW. We should stop the Rx and transmitting the pending packet immidiately.

components/ieee802154/driver/esp_ieee802154_dev.c
```diff
@@ -343,10 +343,8 @@ static void enable_rx(void)
 {
     set_next_rx_buffer();
     IEEE802154_SET_TXRX_PTI(IEEE802154_SCENE_RX);
-
-    ieee802154_set_cmd(IEEE802154_CMD_RX_START);
-
     ieee802154_set_state(IEEE802154_STATE_RX);
+    ieee802154_set_cmd(IEEE802154_CMD_RX_START);
```
zwx1995esp, line 347: Here the enable_rx will be always running in critical context or ISR context. So there will be no 154 interrupt which can break the processing of this function. Also no need to consider the order of these two lines.

zwx1995esp (commented):
Hi @no2chem , thanks for your PR.

components/ieee802154/driver/esp_ieee802154_dev.c
zwx1995esp: And for your scenario: RX stop, the ISR IEEE802154_RX_ABORT_BY_RX_STOP should be disabled for almost time(only will be enable when there is a TX pending). So only risk is when stop RX while there is a pending tx.

components/ieee802154/driver/esp_ieee802154_dev.c
```diff
@@ -601,8 +599,7 @@ static void ieee802154_isr(void *arg)
 
     if (events & IEEE802154_EVENT_RX_SFD_DONE) {
         // IEEE802154_STATE_TX && IEEE802154_STATE_TX_CCA && IEEE802154_STATE_TX_ENH_ACK for isr processing delay
-        IEEE802154_ASSERT(s_ieee802154_state == IEEE802154_STATE_RX || s_ieee802154_state == IEEE802154_STATE_RX_ACK || s_ieee802154_state == IEEE802154_STATE_TX || s_ieee802154_state == IEEE802154_STATE_TX_CCA || s_ieee802154_state == IEEE802154_STATE_TX_ENH_ACK);
-
+        IEEE802154_ASSERT(s_ieee802154_state == IEEE802154_STATE_RX || s_ieee802154_state == IEEE802154_STATE_RX_ACK || s_ieee802154_state == IEEE802154_STATE_TX || s_ieee802154_state == IEEE802154_STATE_TX_CCA || s_ieee802154_state == IEEE802154_STATE_TX_ENH_ACK || s_ieee802154_state == IEEE802154_STATE_TX_ACK);
```
zwx1995esp, line 602: Hi, didi you see any crash here? I don't think this is a reasonable assert state here. Because when the SW state is `tx_ack`, the HW should only send the ack frame first and then go to another state to wait SW to set RX start. In the `tx_ack_done` isr process, SW will set `RX start` command to HW(in the function `next_operation()`). If SW do not set the RX start command, the HW can not receive any packets.

chshu:
@no2chem Thanks a lot for the contribution! We have refactored the next operation logic in this commit: https://github.com/espressif/esp-idf/commit/0353db2372284377968aea97e01b76bfc2198c7b, it should cover the race case you reported, please have a try.

PR #14110, opened 2024-06-29
fix(bt): Don't constrain local icon (appearance) values (IDFGH-13170)

nebkat:
[Bluetooth Assigned Numbers 2.6 Appearance Values](https://www.bluetooth.com/wp-content/uploads/Files/Specification/Assigned_Numbers.html#bookmark49) contains many values which are not listed and there is no real danger from an invalid value, so allow user to set any value.


nebkat:
@esp-zhp Could you maybe take a quick look at this one? There is no need for the list restriction here.

esp-zhp:
If no limitations are imposed, it could introduce other issues.

esp-zhp:
How does the current solution affect you?

nebkat:
The current function lists 58 possible values, but there are over 250 defined in https://www.bluetooth.com/wp-content/uploads/Files/Specification/Assigned_Numbers.html#bookmark49 (I would like to use one of these). I don't think it is practical to provide constants for all of these considering new ones get added occasionally.

The appearance value just determines the icon that is shown when searching for bluetooth devices so there really isn't any danger of allowing an "invalid" value.

esp-zhp:
We will evaluate it. In fact, our original code did not impose restrictions; the restrictions were added later to address some issues.

nebkat:
https://github.com/espressif/esp-idf/pull/15477#issuecomment-2890912898

PR #14132, opened 2024-07-04
feat(dhcp server): only hand out DNS if explicitly specified (IDFGH-13195)

showier-drastic:
By default there isn't any dns server running on the ESP32, and it does not make sense to hand out DNS config here. On the other hand, handing out DNS config will affect the client machine if we don't want DNS traffic to go through ESP32, and there's no way to disable it. So only hand it out if explicitly enabled.

### DIFF
--- a/components/lwip/apps/dhcpserver/dhcpserver.c
+++ b/components/lwip/apps/dhcpserver/dhcpserver.c
@@ -454,18 +454,13 @@ static u8_t *add_offer_options(dhcps_t *dhcps, u8_t *optptr)
         }
     }
 
-    *optptr++ = DHCP_OPTION_DNS_SERVER;
-    *optptr++ = 4;
     if (dhcps_dns_enabled(dhcps->dhcps_dns)) {
+        *optptr++ = DHCP_OPTION_DNS_SERVER;
+        *optptr++ = 4;
         *optptr++ = ip4_addr1(&dhcps->dns_server);
         *optptr++ = ip4_addr2(&dhcps->dns_server);
         *optptr++ = ip4_addr3(&dhcps->dns_server);
         *optptr++ = ip4_addr4(&dhcps->dns_server);
-    }else {
-        *optptr++ = ip4_addr1(&ipadd);
-        *optptr++ = ip4_addr2(&ipadd);
-        *optptr++ = ip4_addr3(&ipadd);
-        *optptr++ = ip4_addr4(&ipadd);
     }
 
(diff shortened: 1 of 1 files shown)


KisonHe:
Any update on this? I am also running into this problem. I plan to make my product communicate with PC using usb ncm. And I just find the dns server on esp is jamming net traffic on my client's PC. 

By applying changes here the problem is then gone.

espressif-abhikroy:
sha=c4e2274e058a82bf1ff9da29fe84cb66b1fc3442

david-cermak (approved):
Thanks for the PR, LGTM!

Could you please check that the [captive portal](https://github.com/espressif/esp-idf/tree/master/examples/protocols/http_server/captive_portal) example still works with this change? Just to be sure if we need a compatibility config option?

showier-drastic:
Hi @david-cermak , thanks for your reply!

This does break the captive portal example. How would you suggest adding a fix?

david-cermak:
> Hi @david-cermak , thanks for your reply!
> 
> This does break the captive portal example. How would you suggest adding a fix?

Thanks for checking. I suggest introducing a Kconfig option (enabled by default) which would keep the IP address as the DNS server, unless configured (current behaviour). If disabled, it would set the DNS option only if it's configured (behaviour after your change).
We'll remove this Kconfig (altering the behavior) in IDF `v6.0` (when we're allowed to make breaking changes).

PS: If you'd prefer, the ESP-IDF team can handle the backward compatibility changes on top of your commit. Let me know if you'd like us to take care of this!

showier-drastic:
> > Hi @david-cermak , thanks for your reply!
> > This does break the captive portal example. How would you suggest adding a fix?
> 
> Thanks for checking. I suggest introducing a Kconfig option (enabled by default) which would keep the IP address as the DNS server, unless configured (current behaviour). If disabled, it would set the DNS option only if it's configured (behaviour after your change). We'll remove this Kconfig (altering the behavior) in IDF `v6.0` (when we're allowed to make breaking changes).

Understood, thanks for explanations!
 
> PS: If you'd prefer, the ESP-IDF team can handle the backward compatibility changes on top of your commit. Let me know if you'd like us to take care of this!

Yes, that would be very nice.

david-cermak:
Great, we've accepted your commit internally -- please wait while we handle the backward compatibility layer and merge it at once.

tonyjackson-srt:
> Great, we've accepted your commit internally -- please wait while we handle the backward compatibility layer and merge it at once.

@david-cermak, will this be back-ported to ESP-IDF v4.4?

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/649816d9a104796ea451b2d2b1c6bc4900ffb80d.

PR #14150, opened 2024-07-08
fix(diskio): Remove const keyword from VolToPart (IDFGH-13213)

qwx:
Remove const keyword from VolToPart
So user can modify VolToPart while app is running
In Fatfs code source, no const on VolToPart
This is useful when physical drive has more than one partition


adokitkat:
Thank you for the PR. However we are preparing larger changes to the whole component, which includes this change, so we won't merge this PR. I hope it is OK.

PR #14159, opened 2024-07-10
fix(tool/mfg_gen) allow encodings u64 and i64 (IDFGH-13224)

liujiaqi:
downstream tool `nvs_partition_gen` support encodings u64 and i64, but `mfg_gen` verify it invalid.


hrushikesh430:
hi liujiaqi,
I was trying to reproduce same situation, using encoding methods like u64 and i64. But it is working fine. Can I get to know your idf version and other changes you have made?

hrushikesh430:
sha=a697ce0d3f2253ff011bf92020cb2fabec45b1d0

PR #14180, opened 2024-07-14
Create ssd1306_simple_example_main.c (IDFGH-13246)

sonictl:
for the new learner of esp-idf and ssd1306, add a simpler example code for them. 

This code is tested on .96 & .91 ssd1306 oled screen.


suda-morris:
we have simplified the oled example on the master branch to illustrate the usage of the `esp_lcd_panel_draw_bitmap` function: https://github.com/espressif/esp-idf/blob/master/examples/peripherals/lcd/i2c_oled/main/i2c_oled_example_main.c No more middleware to hide the usage of the LCD driver APIs.

So probably this PR is not needed any more. Thanks for your contribution again.

Issue #13490, opened 2024-03-28
sdkconfig changes when project is built on different machines (IDFGH-12478)


krupis:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

5.0.4

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

PowerShell

### What is the expected behavior?

Build the project without any sdkconfig conflicts

### What is the actual behavior?

We have multiple people working on the same esp-idf project. We have noticed that whenever someone makes changes and push to the repository, whenever other people try to pull all the changes and build the project on their own machine, during build process something changes on the sdkconfig.h. 

I want to understand why does the sdkconfig.h changes when project is built on different machines even if the same esp-idf version is used?

### Steps to reproduce.

1. Pull someones changes from the repository 
2. Build project

### Build or installation Logs.

For example, I have just pulled changes from my colleague and built the project. I can see that changes are automatically detected using the Source Control extension on the VSCode on the sdkconfig. 

It seems that the only changes are the order. Please see the example below:
![image](https://github.com/espressif/esp-idf/assets/22237034/e5ac7971-dfea-46a9-81da-6bdaf1620fbc)

![image](https://github.com/espressif/esp-idf/assets/22237034/31022bc3-78b2-45ac-9a69-7f0ab39a214f)

As you can see from 2 images above, on my machine, LittleFS is configured starting on the line 1675 and on my colleagues sdkconfig it is configured starting at line 1737. What is the reason for that and why that happens?

dobairoland:
Because items before writing those files are not sorted and their order is semi-random and could be different between systems just as you observe.

You are not supposed to commit `sdkconfig` or anything from the build directory, including `sdkconfig.h`, into source control management systems. Any changes different from the default ones should be stored in `sdkconfig.defaults` and that could be committed. See `idf.py save-defconfig` for how to generate that file. You can add items to that file manually as well. `sdkconfig` should be removed each time there is a change in `sdkconfig.defaults` or the changes will be ignored.

dobairoland:
[...]

PR #14201, opened 2024-07-16
fix(cmake): Rebuild on changes in sdkconfig.defaults

MathyV:
When you change sdkconfig.defaults, sdkconfig is not updated and a rebuild is not triggered. This commit fixes that by adding the files to the target.

### DIFF
--- a/tools/cmake/kconfig.cmake
+++ b/tools/cmake/kconfig.cmake
@@ -215,6 +215,17 @@ function(__kconfig_generate_config sdkconfig sdkconfig_defaults)
     # When sdkconfig file changes in the future, trigger a cmake run
     set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${sdkconfig}")
 
+    # Or if any of the sdkconfig.defaults files change
+    if(sdkconfig_defaults)
+        foreach(sdkconfig_default ${sdkconfig_defaults})
+            set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${sdkconfig_default}")
+
+            if(EXISTS "${sdkconfig_default}.${idf_target}")
+                set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${sdkconfig_default}.${idf_target}")
+            endif()
+        endforeach()
+    endif()
+
     # Ditto if either of the generated files are missing/modified (this is a bit irritating as it means
     # you can't edit these manually without them being regenerated, but I don't know of a better way...)
     set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${sdkconfig_header}")


igrr:
@MathyV Do I understand it right that the situation is:

1. The project is built once, sdkconfig file is generated
2. sdkconfig.defaults is modifed
3. The project is built again

Do you expect that sdkconfig file will be re-generated at step 3, applying new values from sdkconfig.defaults?

MathyV:
> Do you expect that sdkconfig file will be re-generated at step 3, applying new values from sdkconfig.defaults?

That is correct

igrr:
In that case, I think this doesn't match how we expect sdkconfig to work. Sdkconfig file should get values from the defaults files when the respective options are missing from sdkconfig. In other cases, it's okay for sdkconfig to contain a value different than sdkconfig.defaults has. For example, if you build the project once and then change some options in menuconfig, sdkconfig file will contain the value you have saved in menuconfig.

Otherwise, if sdkconfig.defaults would take precedence over sdkconfig (or would overwrite sdkconfig values) then it would not be possible to change the option in menuconfig to a value that doesn't match the one specified in sdkconfig.defaults.

nebkat:
I think this is something worth exploring further - IMO there should be an `sdkconfig.local` or similar file for explicit/manual/temporary `menuconfig` changes, but otherwise the use of `sdkconfig.defaults` and generated `sdkconfig` should be strongly encouraged and more closely integrated into the build process.

If a defaults file has been changed upstream it is usually with good reason - yet we explicitly favour the local file which may or may not even contain any local changes. 

I am also certain there are hundreds of projects out there using a git committed `sdkconfig` that was generated when the project was first created, using whatever the defaults were in an older version of ESP-IDF, and dragging around changes that were only intended to be temporary.

My team has successfully been using `set(SDKCONFIG "${CMAKE_BINARY_DIR}/sdkconfig")` and only working in terms of defaults files. The only remaining pitfall has been someone forgetting to do a fullclean before building and retaining the old config, at best having a compile error, or at worst losing hours debugging a runtime error.

KaeLL:
I don't think this change is necessary, although it also doesn't affect my workflow. My understanding is that IDF incentivizes working with sdkconfig.defaults rather than regular sdkconfig, both because the former is tidy and the latter is unstable (although after #13490, I don't know what to think anymore). Made changes in menuconfig? Save it as you would, which changes the sdkconfig and trigger rebuilds. Need to commit to/persist them? `idf.py save-defconfig` or menuconfig, press D (defconfig) and point it to your root sdkconfig.defaults and save it, which doesn't trigger rebuilds because it's not necessary at this stage.

MathyV:
Sorry for not chiming in earlier, lost track of this one a bit.

Like @nebkat , I consider `sdkconfig` to be a build artifact, rather than something that should be committed into the repository. I much prefer having upstream changes for things I don't know or care about over having to keep track of these things myself.

Another reason I do it like this is because I often target multiple platforms with the same code. I don't care about all the options that need to be set for example to have wifi work on the different devices, I just enable the main wifi setting in a shared `sdkconfig.defaults` file, know my code will work for all and don't have to worry much more about it.

I tend to structure my configuration over multiple `sdkconfig.defaults` files and then just link to the ones I need for a certain platform, much in the way that MicroPython does it. You can see an example of it it here:

https://github.com/MathyV/fri3d_firmware/blob/main/boards/fox/CMakeLists.txt

@nebkat what you are suggesting with a `sdkconfig.local`, I'm already doing in that code too. I use it to locally trigger extra logging output, to lower the risk of accidentally committing it in a production build. The file is in `.gitignore` so it doesn't end up in the repo. I don't feel this needs to be explicitly added to IDF though, as I think it much depends on developer preference and it's easy to add.

igrr:
> My team has successfully been using set(SDKCONFIG "${CMAKE_BINARY_DIR}/sdkconfig") and only working in terms of defaults files for the past 9 months. The only remaining pitfall has been someone forgetting to do a fullclean before building and retaining the old config, at best having a compile error, or at worst losing hours debugging a runtime error, so this change would be very welcome.

I think this could be achieved with a CMake recipe which monitors changes to sdkconfig.defaults and deletes sdkconfig whenever sdkconfig.defaults changes, forcing sdkconfig to be re-generated from sdkconfig.defaults.

In that case, you can use sdkconfig for local changes, and when you want to save them to sdkconfig.defaults you run `idf.py save-defconfig`. The only downside of this approach is that if you get sdkconfig.defaults updates after pulling a new version of your project from git, your local defaults get wiped. However other than having some interactive prompt asking to resolve the conflicts between local and remote changes, I can't think of any alternative which would retain both incoming and local changes. And that seems like something that could be handled using existing tools (e.g. merging your local sdkconfig.defaults with the remote one.)

I think we would be open to adding such a CMake recipe in the form of a function which can be called from the project CMakeLists.txt file. (E.g. `idf_rebuild_sdkconfig_on_defaults_changes(...)`). I don't think we can change the default behavior at this point.

We are also slowly working towards the new configuration system which will resolve some of Kconfig limitations, while retaining compatibility with Kconfig files. The most significant limitation we want to fix is that we can't distinguish sdkconfig options which have been explicitly set (but equal to the default value) from default values that the project developer doesn't care about. This will help us move closer to having an sdkconfig.defaults-like file as the single source of truth that tools like `menuconfig` will work with. However I don't have a timeline for this, so a small improvement in the form of a CMake helper function would be welcome in the meantime.

PR #14286, opened 2024-08-01
feat(bt) - Enable use of RESERVED bits in COD (IDFGH-13374)

WebDust21:
This is a very niche-case patch, one that most people should never have need to bother with.

But in my case, trying to make a replacement for an obsolete Bluetooth board, I ran into an issue with a nonstandard use of the BT COD field: the two low "reserved" bits were used!  Without access to the controlling firmware (and having to maintain compatibility with the thousands of "obsolete Bluetooth boards" in the field which will not function without said "reserved" bits being transmitted in the outgoing COD), this is the neatest solution to the problem.

I did a deep dive straight into the Bluedroid stack, and found exactly where (amongst the multiple unnecessary data type changes the provided COD went through) the two bits were getting lost.  The very simple changes here were successful in allowing the ESP32 to work as a drop-in replacement for the obsolete board.

### DIFF
--- a/components/bt/host/bluedroid/btc/profile/std/gap/btc_gap_bt.c
+++ b/components/bt/host/bluedroid/btc/profile/std/gap/btc_gap_bt.c
@@ -604,6 +604,7 @@ esp_err_t btc_gap_bt_get_cod(esp_bt_cod_t *cod)
         return ESP_BT_STATUS_FAIL;
     }
     cod->minor = p_cod.minor >> 2;
+    cod->reserved_2 = p_cod.minor & 0x03;
     cod->major = p_cod.major;
     cod->service = p_cod.service >> 5;
     return ESP_BT_STATUS_SUCCESS;
--- a/components/bt/host/bluedroid/btc/profile/std/gap/btc_gap_bt.c
+++ b/components/bt/host/bluedroid/btc/profile/std/gap/btc_gap_bt.c
@@ -586,6 +586,7 @@ static void btc_gap_bt_set_cod(btc_gap_bt_args_t *arg)
     tBTA_UTL_COD p_cod;
     esp_bt_cod_t *cod = &(arg->set_cod.cod);
     p_cod.minor = cod->minor << 2;
+    p_cod.minor |= cod->reserved_2;
     p_cod.major = cod->major;
     p_cod.service = cod->service << 5;
     bool ret = utl_set_device_class(&p_cod, arg->set_cod.mode);
--- a/components/bt/host/bluedroid/stack/include/stack/btm_api.h
+++ b/components/bt/host/bluedroid/stack/include/stack/btm_api.h
@@ -468,7 +468,7 @@ typedef enum {
 
 /* the COD masks */
 #define BTM_COD_FORMAT_TYPE_MASK      0x03
-#define BTM_COD_MINOR_CLASS_MASK      0xFC
+#define BTM_COD_MINOR_CLASS_MASK      0xFF
 #define BTM_COD_MAJOR_CLASS_MASK      0x1F
 #define BTM_COD_SERVICE_CLASS_LO_B    0x00E0
 #define BTM_COD_SERVICE_CLASS_MASK    0xFFE0


esp-qing:
Hi, @WebDust21 ,Thank you for your submission! We really appreciate it.  And there are two places that need to be modified:
1. Commit message: Suggest to modify it to `fix(bt): Enable use of ...... `
2. There are three commits, suggest to merge them into one commit.

WebDust21:
> Hi, @WebDust21 ,Thank you for your submission! We really appreciate it. And there are two places that need to be modified:
> 
>     1. Commit message: Suggest to modify it to `fix(bt): Enable use of ...... `
> 
>     2. There are three commits, suggest to merge them into one commit.

I'm sorry, I'm not a Github whiz...I don't know how to do either of these.  

from what I read, it should be easy for your side to "squash" the commits (I do not have the option).

I do not know how to change a commit message...although as this PR extends nonstandard functionality, I would consider this a "feature", and not necessarily a "fix".

esp-qing:
> from what I read, it should be easy for your side to "squash" the commits (I do not have the option).
> 
> I do not know how to change a commit message...although as this PR extends nonstandard functionality, I would consider this a "feature", and not necessarily a "fix".

1. I can't directly modify your commit message on GitHub. It needs to be modified on your side and changed to the correct format, otherwise it can't be merged.
2. According to the commit message format, this change should be considered a fix type.

You can modify it according to the following steps:

Run the command:  `git rebase -i 41dd1a351b4f8630a0b9dbdd87a6bf782094dd66`
After entering the following interface, change the last two `pick` to `s`
image
`:wq` Save and exit to enter the following interface
image
Then modify the commit information like this:
image

`:wq` Save and exit, then push it.

esp-qing:
Hi, There are two suggestions above. I‘m sorry that if you don't modify it to the correct format, it cannot be merged. And This PR will be closed after a week.

esp-qing:
> But in my case, trying to make a replacement for an obsolete Bluetooth board, I ran into an issue with a nonstandard use of the BT COD field: the two low "reserved" bits were used! ...

Hi, @WebDust21 In the latest Bluetooth protocol, the two least significant bits are fixed to 0b00. Could you please tell me what the model of  Bluetooth board are you using?  And which version of the Bluetooth protocol are you referring to? Thank you!

WebDust21:
> > from what I read, it should be easy for your side to "squash" the commits (I do not have the option).
> > I do not know how to change a commit message...although as this PR extends nonstandard functionality, I would consider this a "feature", and not necessarily a "fix".
> 
>     1. I can't directly modify your commit message on GitHub. It needs to be modified on your side and changed to the correct format, otherwise it can't be merged.
> 
>     2. According to the commit message format, this change should be considered a fix type.
> 
> 
> You can modify it according to the following steps:
> 
> Run the command: `git rebase -i 41dd1a351b4f8630a0b9dbdd87a6bf782094dd66` After entering the following interface, change the last two `pick` to `s` image `:wq` Save and exit to enter the following interface image Then modify the commit information like this: image
> 
> `:wq` Save and exit, then push it.

Thank you for the instructions; they would work if I had forked a local copy of the entire ESP-IDF to my computer.  However, I just made the edits from Github.com, with no local copy--and as such, there is no place to run the commands.



> > But in my case, trying to make a replacement for an obsolete Bluetooth board, I ran into an issue with a nonstandard use of the BT COD field: the two low "reserved" bits were used! ...
> 
> Hi, @WebDust21 In the latest Bluetooth protocol, the two least significant bits are fixed to 0b00. Could you please tell me what the model of Bluetooth board are you using? And which version of the Bluetooth protocol are you referring to? Thank you!

Like I mentioned in the very opening comment, this is NONSTANDARD use of the two low "reserved" bits, as required for a compatibility project.  This is for SPP mode (Bluetooth Classic) on an original ESP32; it is not BLE in any way, shape or form.
For whatever reason, the designers of the existing system set the two least significant bits (RESERVED bits) to 0b01--and if the ESP32 doesn't broadcast the COD with the two LSBs set to 0b01, it will be completely ignored by the existing system.

I've tested the modifications here as fully functional in getting the ESP32 to broadcast the necessary Bluetooth COD (although the "read COD" still incorrectly returns the LSBs as 0b00 for some reason.  Need to dig into that further.)

esp-qing:
Hi, @WebDust21 You don’t need to modify it anymore.. But in the original implementation, reserved_2 is not part of the minor class and is always set to 0. With your fix, reserved_2 may be non-zero, which changes the API behavior and affects users who don't want to overwrite reserved-2.

So after your commit, I made some changes to extend the functionality of the current API so that the previous API behavior does not change and allow users who want to set the reserved_2 to use the new esp_bt_cod_mode_t code.

WebDust21:
Can I see the changes to the code?

It's worth noting that the Bluetooth device scan does return the "reserved_2" bits--which is how I was able to determine that the units used a nonstandard Bluetooth COD.  It just didn't allow _setting_ a COD with those bits used, nor can I read back a hacked COD set (although it does work when set).

esp-qing:
@WebDust21 Yes, update to the latest version of master, you can see the changes.

Alvin1Zhang:
Thanks for contribution again, changes have been with https://github.com/espressif/esp-idf/commit/59066e3edfe22fc11466e0047909bd96d6785dc8. Thanks.

PR #14310, opened 2024-08-06
fix(gptimer): remove incorrect param from example (IDFGH-13403)

LonerDan:
Fixes call to `gptimer_start` in one of the examples which includes extra second parameter `&alarm_config`, when the function takes only one.


suda-morris:
sha=68b205d4c7f485914f9d426d2200055b35527b43

PR #14325, opened 2024-08-07
fix(rmt): set rx filter resolution_hz to rmt_sclk

LonerDan:
Fixes assignment of `filter_clock_resolution_hz` to `rx_channel->base.resolution_hz`, which is the clock frequency after the fractional divider `RMT_SCLK_DIV` (also referenced as `rmt_sclk` in the TRM), instead of the previously used value from `group->resolution_hz`, which represents the full undivided frequency from clock selected by `RMT_SCLK_SEL` (APB, RC_FAST or XTAL). 
Before this fix, this mixup resulted in an error `signal_range_min_ns too big` being returned by the `rmt_receive` function due to [this check](https://github.com/espressif/esp-idf/blob/8e4454b285ad39881c5bf3f440d8be617a2f18a8/components/esp_driver_rmt/src/rmt_rx.c#L410-L413) failing no matter the `resolution_hz` used in the configuration of the channel.

### DIFF
--- a/components/esp_driver_rmt/src/rmt_rx.c
+++ b/components/esp_driver_rmt/src/rmt_rx.c
@@ -283,7 +283,7 @@ esp_err_t rmt_new_rx_channel(const rmt_rx_channel_config_t *config, rmt_channel_
         ESP_LOGW(TAG, "channel resolution loss, real=%"PRIu32, rx_channel->base.resolution_hz);
     }
 
-    rx_channel->filter_clock_resolution_hz = group->resolution_hz;
+    rx_channel->filter_clock_resolution_hz = rx_channel->base.resolution_hz;
     // On esp32 and esp32s2, the counting clock used by the RX filter always comes from APB clock
     // no matter what the clock source is used by the RMT channel as the "core" clock
 #if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2


LonerDan:
NOTE: I couldn't verify this fix's correctness for the ESP32-C5 and ESP32-P4, since I could not find their TRMs (I suppose those are not available yet).

suda-morris:
Hi @LonerDan I'm afraid this is not a mistake. The RX channel can filter out glitches in the signals, so it needs a high frequency base clock. The hardware uses the group's clock as the filter's tick clock.

LonerDan:
Hi @suda-morris 
I was going off of the TRM (more specifically for the ESP32-S3) [figure 37-1](https://www.espressif.com/sites/default/files/documentation/esp32-s3_technical_reference_manual_en.pdf#page=1396), where the label `rmt_sclk` is put after the fractional divider and the same label is used with the arrow going into the filter block. Also, the text about the RX filter mentions that 

> The filter samples input signals continuously, and detects the signals which remain unchanged for a continuous RMT_RX_FILTER_THRES_CHm **_<ins>rmt_sclk cycles</ins>_** as valid, otherwise, the signals will be detected as invalid.

If that's not really the case, It could be beneficial to clarify this in the documentation.

LonerDan:
@suda-morris Ah, I see. Thank you for the correction. I looked closer into the source code and found the part where the `rmt_sclk` divider is always configured to no division:
https://github.com/espressif/esp-idf/blob/8e4454b285ad39881c5bf3f440d8be617a2f18a8/components/esp_driver_rmt/src/rmt_common.c#L210-L214
In my opinion, this should be configurable since the resolution of the filter is application-specific. In my case, the signal has shortest pulse length around 12 us and there is a lot of short spikes/noise mixed in, so I would like to filter out anything less than say 8 us to limit the number of data rmt rx channel generates for me to process.
I am willing to try to implement this, I just wanted to ask where you think is the best place to incorporate this parameter into? Not sure if `rmt_rx_channel_config_t` is the correct place since the `sclk` divider is common to the group.

PR #14326, opened 2024-08-07
fix(gptimer): remove incorrect param from example (IDFGH-13420)

LonerDan:
Fixes call to `gptimer_start` in one of the examples which includes extra second parameter `&alarm_config`, when the function takes only one.


LonerDan:
This is a resubmit of accidentally closed PR #14310

suda-morris:
FYI, the fix is already merged in https://github.com/espressif/esp-idf/commit/08fa1131ca13860fb4a3eb340fd37f76a1c651a6

PR #14328, opened 2024-08-07
change: .gitignore npp+nano temp extensions (IDFGH-13422)

bryghtlabs-richard:
Add notepad++ and nano editors temporary files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,12 @@ test_multi_heap_host
 *.sublime-project
 *.sublime-workspace
 
+# nano file
+*.tmp
+
+# Notepad++ file
+*.bak
+
 # Clion IDE CMake build & config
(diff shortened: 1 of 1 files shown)


dobairoland:
In my opinion editor-specific entries don't belong to the project's gitignore file because they are not related to the project. We already have gitignore entries which (in my opinion) don't belong there, but lets not extend this list. I hope you understand but everyone could request to add personal ignores into every project they are working on.

I think the best way is to have a personal global ignore file with editor- and OS-specific ignores. This can be defined in git config as follows:
```
[core]
    excludesfile = ~/.gitignore
```

bryghtlabs-richard:
Thanks Roland - I didn't know there was a config for that.

PR #14344, opened 2024-08-09
Update docker.yml (IDFGH-13438)

LLotme:
Thx


dobairoland:
This looks like a mistake. It doesn't have any meaningful changes.

PR #14347, opened 2024-08-10
Snyk fix 6937d5507373d8c5ccb18980348775b2 (IDFGH-13441)

LLotme:


dobairoland:
This is almost the same as https://github.com/espressif/esp-idf/pull/13911 and https://github.com/espressif/esp-idf/pull/13895.

I'm closing this because we haven't received any reply and this doesn't seems that something we want to accept.

PR #14367, opened 2024-08-14
.. (IDFGH-13465)

LLotme:


igrr:
Same note as in https://github.com/espressif/esp-idf/pull/14347#issuecomment-2283267991.

PR #14373, opened 2024-08-15
fix(esp32): delete the secondary judgment  in examples/mesh/internal_communication (IDFGH-13472)

TroyMitchell911:
Delete the secondary judgment of whether it is the root node in the sending task.


zhangyanjiaoesp:
@TroyMitchell911 thanks, will merge it ASAP.

PR #14376, opened 2024-08-15
feat(tools): Update cmake from 3.24.0 to 3.30.2 (IDFGH-13476)

nebkat:
[v3.24.0](https://github.com/Kitware/CMake/releases/tag/v3.24.0) is now 2 years old.

Have run into a dependency which needs v3.25.0+ :flushed:

This is however missing `linux-armel` and `linux-armhf`.

### DIFF
--- a/tools/tools.json
+++ b/tools/tools.json
@@ -497,6 +497,40 @@
       "versions": [
         {
           "linux-amd64": {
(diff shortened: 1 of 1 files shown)


mfialaf:
Hi @nebkat, thank you for noticing us. We'll update to 3.30.2 soon.

PR #14384, opened 2024-08-16
fix(pthread): Add esp_pthread function implementations for linux target (IDFGH-13485)

cristianfunes79:
This PR is to add functionality to pthread port for linux target. It extends changes done in #14339

## DESCRIPTION
As soon as pthread implementation for linux uses libpthread, it is not possible to use esp custom functions for setting pthread parameters. 
But this implementation at least can be use to validate arguments passed to the functions.

## DETAILED DESCRIPTION
When using pthread_create esp implementation, it uses freeRTOS task behind to create a new execution thread. This implementation uses `pthread_getspecifics` to get values set by `esp_pthread_set_cfg (const esp_pthread_cfg_t *cfg)` (see https://github.com/espressif/esp-idf/blob/master/components/pthread/pthread.c#L309).
When using linux target, that custom pthread implementation is no more used, so all `esp_pthread.h` functions doesn't affect thread creation.

### DIFF
--- a/components/pthread/port/linux/pthread.c
+++ b/components/pthread/port/linux/pthread.c
@@ -7,8 +7,25 @@
  * pthread port for Linux build
  */
 
+#include "sdkconfig.h"
 #include "esp_pthread.h"
+#include "esp_heap_caps.h"
+
+#include <pthread.h>
+
 #include <string.h>
+#include "freertos/FreeRTOS.h"
+
+static pthread_key_t s_pthread_cfg_key;
+static void esp_pthread_cfg_key_destructor(void *value)
+{
+    free(value);
+}
+
+static int get_default_pthread_core(void)
+{
+    return CONFIG_PTHREAD_TASK_CORE_DEFAULT == -1 ? tskNO_AFFINITY : CONFIG_PTHREAD_TASK_CORE_DEFAULT;
+}
 
 /**
  * @brief Creates a default pthread configuration based
@@ -24,8 +41,8 @@ esp_pthread_cfg_t esp_pthread_get_default_config(void)
         .prio = CONFIG_PTHREAD_TASK_PRIO_DEFAULT,
         .inherit_cfg = false,
         .thread_name = NULL,
-        .pin_to_core = 0,
-        .stack_alloc_caps = 0,
+        .pin_to_core = get_default_pthread_core(),
+        .stack_alloc_caps = MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT,
     };
 
     return cfg;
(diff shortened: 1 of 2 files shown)


cristianfunes79:
Closes #14387

ESP-Marius:
Thanks for adding this, we'll take a look!

components/pthread/port/linux/pthread.c
```diff
@@ -24,20 +41,62 @@ esp_pthread_cfg_t esp_pthread_get_default_config(void)
...
+    esp_pthread_cfg_t *p = pthread_getspecific(s_pthread_cfg_key);
+    if (!p) {
+        p = malloc(sizeof(esp_pthread_cfg_t));
+        if (!p) {
+            return ESP_ERR_NO_MEM;
+        }
+    }
+    *p = *cfg;
+    p->stack_alloc_caps = heap_caps;
+    pthread_setspecific(s_pthread_cfg_key, p);
```
0xjakob, line 80: Really minor thing: I'd suggest to add an assert here to make sure `pthread_setspecific()` succeeds, even though the chance is probably very little:

```suggestion
    int __attribute((unused)) res = pthread_setspecific(s_pthread_cfg_key, p);
    assert(res == 0);
```

You may need to `#include assert.h`.

0xjakob (approved):
@cristianfunes79 Thanks for this PR, it looks great. I just have one minor comment.

I think this is a good improvement as it keeps the basic behavior of the API on the Linux target aligned with the behavior of the API on the chip targets. This means less surprises for users.

components/pthread/port/linux/pthread.c
cristianfunes79: Ok I've tried it but it is not so straight forward, as soon as linux doesnt call esp_pthread_init, cfg_key is never created and setspecific assertion fails. I'll take a look on how to resolve that.

cristianfunes79: @0xjakob Actually I'm thinking about all functions marked as `ESP_SYSTEM_INIT_FN`. Does it make sense to create a ticket to implement a way to run such functions in linux?

0xjakob: > I've tried it but it is not so straight forward

Oh, sorry, I didn't think about that. I think you could try to use `__attribute__((constructor))` to create a constructor function to initialize the key. `__attribute__((constructor))` is documented in the [GCC function attributes doc page]( https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes)

`ESP_SYSTEM_INIT_FN` won't work on Linux. It's not clear yet, either, if we will implement it on Linux or not. We have a ticket for exactly this in our own ticket system already. Unless you have a concrete reason to implement it, there is no need to create a ticket here.

cristianfunes79: Ok I'll add constructor. The reason I can think to add `ESP_SYSTEM_INIT_FN` is to have more similar firmware behavior in linux than in xtensa. Maybe it could be as simple as creating a `#define ESP_SYSTEM_INIT_FN(foo) __attribute__((constructor)) foo`.

cristianfunes79: @0xjakob this is ready for review

0xjakob:
sha=92bfb26639e40149d02140a1f672254c4e161f15

0xjakob:
@cristianfunes79 We're merging this into our internal repository. Afterwards, it should soon be synchronized to master.

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/4c7d9f9859e13bfed9137ef59418a06b18e848c8.

PR #14392, opened 2024-08-18
Update uart_select_example_main.c (IDFGH-13497)

diplfranzhoepfinger:
UART_NUM_0 instead of 0 as constant for clearer readability.


suda-morris:
sha=18e69cab016046300c71ac4125bac6acec4f1986

PR #14393, opened 2024-08-18
strerror for clearer Error detection (IDFGH-13496)

diplfranzhoepfinger:
strerror brings error cause in clear - Text.


suda-morris:
sha=d836ea0107eb68752cf1570a4abd87a36b04449b

PR #14398, opened 2024-08-19
fix esp_secure_boot not compiling in C++ (IDFGH-13503)

0xFEEDC0DE64:
Dear espressif,

please also make sure your c headers compile fine in C++ compilers,

best regards :)

### DIFF
--- a/components/bootloader_support/include/esp_secure_boot.h
+++ b/components/bootloader_support/include/esp_secure_boot.h
@@ -87,7 +87,7 @@ typedef enum {
  *
  * @return key type for the selected secure boot scheme
  */
-static inline char* esp_secure_boot_get_scheme_name(esp_secure_boot_sig_scheme_t scheme)
+static inline const char* esp_secure_boot_get_scheme_name(esp_secure_boot_sig_scheme_t scheme)
 {
     switch (scheme) {
         case ESP_SECURE_BOOT_V2_RSA:


0xFEEDC0DE64:
Why do the specialists at espressif have a CI for commit messages but not for checking if the header compiles at all in a normal software project with C++

igrr:
Thank you for the PR @0xFEEDC0DE64!

> but not for checking if the header compiles at all in a normal software project with C++

We do, there is a [script](https://github.com/espressif/esp-idf/blob/master/tools/ci/check_public_headers.py) which checks that headers for common issues (missing pragmas, C++ guards, and so on). All the public headers except the ones listed [here](https://github.com/espressif/esp-idf/blob/master/tools/ci/check_public_headers_exceptions.txt) should pass this checks.

The problem is that in this specific case, detecting this issue also requires enabling one of 
`CONFIG_SECURE_BOOT` or `CONFIG_SECURE_SIGNED_APPS_NO_SECURE_BOOT` options, which the checker script doesn't know about.

So the problem becomes slightly harder: either "the checker needs to verify that headers build for every possible setting of Kconfig options" or "we shouldn't use Kconfig options in public headers".

AdityaHPatwardhan (approved):
Thanks for the fix @0xFEEDC0DE64,

AdityaHPatwardhan:
sha=82f0477dfddad77e5f72b01db5afe8bf40c327c0

PR #14402, opened 2024-08-19
fix(tools): python3.12 support: remove distutils use (IDFGH-13510)

nim65s:
Hi,

In python 3.12, the `distutils` module was removed from the standard library.
Here are simple ways to avoid it.

ref. https://docs.python.org/3/whatsnew/3.12.html#distutils

### DIFF
--- a/tools/mass_mfg/mfg_gen.py
+++ b/tools/mass_mfg/mfg_gen.py
@@ -7,8 +7,8 @@ import argparse
 import csv
 import os
 from itertools import zip_longest
+from pathlib import Path
 
-import distutils.dir_util
 import esp_idf_nvs_partition_gen.nvs_partition_gen as nvs_partition_gen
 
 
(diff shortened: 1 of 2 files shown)


tools/mass_mfg/mfg_gen.py
```diff
@@ -7,8 +7,8 @@
 import csv
 import os
 from itertools import zip_longest
+from shutil import Path
```
peterdragun, line 10: I believe you wanted to import Path from `pathlib` instead of `shutil`

peterdragun (changes requested):
Hi @nim65s, thank you for your contribution. Please take a look at the failing pre-commit ci. There are some changes requested like reordering imports. You can check how to install these hooks locally in [Contributions Guide](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/contribute/install-pre-commit-hook.html)

nim65s:
thanks @peterdragun, I think this is better now.

PR #14410, opened 2024-08-21
fix(http_server): Don't require LWIP_NETIF_LOOPBACK for Linux target (IDFGH-13519)

snake-4:
The commit https://github.com/espressif/esp-idf/commit/ef5e4ddee8c1667e32e1252e97c5a95604a003cc introduced a dependency on the LWIP_NETIF_LOOPBACK configuration option for all targets.

This makes it impossible to build esp_http_server without lwIP as the option can't be set. Also, the option is not necessary on Linux builds without lwIP.

### DIFF
--- a/components/esp_http_server/src/util/ctrl_sock.c
+++ b/components/esp_http_server/src/util/ctrl_sock.c
@@ -15,24 +15,22 @@
 #include "ctrl_sock.h"
 
 #if CONFIG_IDF_TARGET_LINUX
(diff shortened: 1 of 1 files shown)


mahavirj (approved):

nileshkale123 (approved):
LGTM.

nileshkale123:
sha=8bade3bf2cf0013cb8f71fbdc80a76779259cd18

Issue #12036, opened 2023-08-08
venv gerenation fails on python installed using the nix package manager (IDFGH-10828)


LukasACH:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.1

### Operating System used.

macOS

### How did you build your project?

Other (please specify in More Information)

### If you are using Windows, please specify command line type.

None

### What is the expected behavior?

The check for virtual environments should pass, and it should create the virtual environment and not fail with the error `ERROR: This script was called from a virtual environment, can not create a virtual environment again`.

### What is the actual behavior?

The check for a virtual environment in the file idf_tools.py on line 2035 (v5.1) fails when using python installed using the nix package manager, at least on macOS. This is the line that evalutates to `True` where it should probably be `False`, or at least have an additional check before outright failing to create the virtualenv.
```python
is_virtualenv = hasattr(sys, 'real_prefix') or (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix)
```

The paths sys.base_prefix and sys.prefix are not the same, even if no virtual environment is activated.
In a fresh python session, the REPL output looks like this:
```
Python 3.10.12 (main, Jul 18 2023, 17:17:22) [Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.base_prefix
'/nix/store/bq1q4gk52gsx4fg4pf07f2kxqgazkcls-python3-3.10.12'
>>> sys.prefix
'/nix/store/3f0019fqza36pp2f2znj5sbndhwddpi8-python3-3.10.12-env'
```

### Steps to reproduce.

How to reproduce the bug:
1. Install python using the nix package manager
2. Try to use esp-idf while the above installed python is the default

Workaround:
1. Replace the line with `is_virtualenv = False`
2. Everything works as it should.

### Build or installation Logs.

```plain
Compiling bindgen v0.63.0
   Compiling embuild v0.31.2
   Compiling esp-idf-sys v0.33.1
   Compiling esp-idf-hal v0.41.2
   Compiling esp-idf-svc v0.46.2
[...]

PR #14435, opened 2024-08-24
fix(tools): venv detection use VIRTUAL_ENV environment variable (IDFGH-13545)

nim65s:
Hi,

Here is a simple fix for #12036
This is required for Nix "venv"-like way.


DNedic:
Hello @nim65s , thank you for submitting this. It looks like this would fix using `nix` python derivations, will test.

`VIRTUAL_ENV` is not a perfect way to detect a virtual environment, as you could still run virtual environments scripts directly, without activating it, and it would not be set, but in this case, that should be irellevant.

Issue #10269, opened 2022-11-29
External 32.768 kHz oscillators (IDFGH-8843)


NoNullptr:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I need to evaluate ESP32-S3 with an external oscillator and ran into multiple issues.

1. The Programming Guide states:
`External 32 kHz oscillator at XTAL_32K_P pin: Allows using 32 kHz clock generated by an external circuit. The external clock signal must be connected to the XTAL_32K_P pin. The amplitude should be less than 1.2 V for sine wave signal and less than 1 V for square wave signal. Common mode voltage should be in the range of 0.1 < Vcm < 0.5xVamp, where Vamp stands for signal amplitude. In this case, the XTAL_32K_P pin cannot be used as a GPIO pin.`
- One minor issue is that it should say 32.768 kHz instead of 32 kHz, which is a standard clock oscillator frequency and all other documentation also refers to this frequency being the correct one.
- Let's call an oscillator's output high voltage `Voh` and an oscillator's output low voltage `Vol`. Then the amplitude is `Voh - Vol` and the common mode voltage is `0.5 * (Voh + Vol)`. The requirement `Vcm < 0.5xVamp` translates to `0.5 * (Voh + Vol) < 0.5 * (Voh - Vol)`, which means `Vol < 0`. Does the ESP32-S3 really require the low output voltage to be negative, below ground? That cannot be satisfied with any standard oscillator or clock signal and seems wrong. Standard oscillators require some headroom above ground, so the output low voltage is for example 10% of `Vdd`. Since no other documentation lists the oscillator requirements, it's impossible to find out if an oscillator swinging between, say, 300 mV and 900 mV are supported or not. One interpretation could be that the documentation writer used the terms amplitude and common mode without knowing their meaning.

2. The configuration options `RTC_CLK_SRC_EXT_CRYS` and `RTC_CLK_SRC_EXT_OSC` are only used to call either                 `rtc_clk_32k_enable(true)` or `rtc_clk_32k_enable_external()`.
- Both functions touch `RTC_IO_XTAL_32N_PAD_REG`, although only `XTAL_32K_P` should be used for an external oscillator.
[...]

PR #14442, opened 2024-08-26
fix: S3: configure external oscillator correctly (IDFGH-13553)

2opremio:
Addresses part of https://github.com/espressif/esp-idf/issues/10269

### DIFF
--- a/components/hal/esp32s3/include/hal/clk_tree_ll.h
+++ b/components/hal/esp32s3/include/hal/clk_tree_ll.h
@@ -90,16 +90,15 @@ static inline __attribute__((always_inline)) void clk_ll_xtal32k_enable(clk_ll_x
 {
     // Configure xtal32k
     clk_ll_xtal32k_config_t cfg = CLK_LL_XTAL32K_CONFIG_DEFAULT();
+    if (mode == CLK_LL_XTAL32K_ENABLE_MODE_EXTERNAL) {
+        cfg.dbuf = 0;
+    }
     REG_SET_FIELD(RTC_CNTL_EXT_XTL_CONF_REG, RTC_CNTL_DAC_XTAL_32K, cfg.dac);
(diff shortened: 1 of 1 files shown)


songruo:
@2opremio We have fixed the issue in b9c58c550c42f9a8f3efd821a9da49101637b682. Please check if you can use external oscillator on current master now. I will close the PR, if it is still a problem, feel free to reopen or create a new issue.

2opremio:
The PR doesn’t seem to set the right buf value for the external oscillator.

dvosully:
@songruo b9c58c5 does not fix the external oscillator problem for ESP32-S3. I am using IDF v5.3 which includes that patch. When I apply the above patch from 2opremio instead the external oscillator works.

I am using a custom board, with the 32.768kHz output from a DS3231 RTC capacitively coupled to GPIO15 of an ESP32-S3-WROOM module. The Vpk-pk is 640mV, Vrms is 304mV, square wave, 26% duty cycle. As above, it works when using the patch from 2opremio so a hardware issue seems unlikely.

AndreoBotelho:
> @songruo [b9c58c5](https://github.com/espressif/esp-idf/commit/b9c58c550c42f9a8f3efd821a9da49101637b682) does not fix the external oscillator problem for ESP32-S3. I am using IDF v5.3 which includes that patch. When I apply the above patch from 2opremio instead the external oscillator works.
> 
> I am using a custom board, with the 32.768kHz output from a DS3231 RTC capacitively coupled to GPIO15 of an ESP32-S3-WROOM module. The Vpk-pk is 640mV, Vrms is 304mV, square wave, 26% duty cycle. As above, it works when using the patch from 2opremio so a hardware issue seems unlikely.

Confirmed the same issue with ESP32-C3, the reference manual says that at reset RTC_CNTL_DBUF_XTAL_32K bit is set to ZERO but reading the register at runtime this bit was set to ONE even with external clock configuration, this leads to lots of  error on oscilator startup and calibration, aplying the path from @2opremio solves the issues and the RTC_CNTL_DBUF_XTAL_32K  is correctly set ZERO at runtime, I'm testing on idf 5.5.2

Issue #14441, opened 2024-08-26
RISCV ULP long wakeup time during deep sleep


2opremio:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

The RISC ULP in my S3 chip takes an unexpectedly long time (~8ms) to wake up between sleep periods while the main CPU is in deep sleep. 

While the Main CPU is awake, the wake up time is 16 times faster (around 0.5ms).

The wake up time during deep sleep seems to depend on the frequency of the RTC clock. In my case I am using a 32K external oscillator (it's linearly shorter when using the internal 136kHZ oscillator).

To measure the times I created a simple ULP program which goes to sleep immediately after toggling an output GPIO pin.

I measured the time between toggles with a logic analyzer (a Nordic PPK2) which also tells me when the system is in deep sleep.

Here are a couple of screenshots of a transition to deep sleep. The bottom line includes the GPIO toggles.

While the main CPU is awake, I measure 0.540ms between ULP wakeups:

<img width="971" alt="Screenshot 2024-08-26 at 02 26 48" src="https://github.com/user-attachments/assets/ce23bd20-0c24-4776-937d-046e155ae42d">

While the system is in deep sleep, I measure ~7.7ms:

<img width="979" alt="Screenshot 2024-08-26 at 02 27 16" src="https://github.com/user-attachments/assets/eb799e76-c69e-45f6-ba16-45052a028f2c">

Here is the full code:

`main.c`
```C
#include "esp_log.h"
#include "esp_sleep.h"
#include "ulp_riscv.h"
#include "driver/rtc_io.h"
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"



#define DEBUG_GPIO_NUM GPIO_NUM_5
extern const uint8_t ulp_main_bin_start[] asm("_binary_ulp_main_bin_start");
extern const uint8_t ulp_main_bin_end[] asm("_binary_ulp_main_bin_end");

void app_main(void)
{
    vTaskDelay(pdMS_TO_TICKS(1000));

    esp_sleep_wakeup_cause_t cause = esp_sleep_get_wakeup_cause();

    switch (cause) {
[...]

PR #14453, opened 2024-08-27
feat: S3: adjust default wait values to the slowclock frequency

2opremio:
The wait values depending on the slowclock should be chosen based on the frequency of the slowclock set.

Addresses part of https://github.com/espressif/esp-idf/issues/14441


ESP-Marius:
sha=44264e4466fa9449326e60d4ec669e3748639fb9

ESP-Marius:
Thanks @2opremio, we'll review this one internally

2opremio:
To be fair, since the slow clock can change dynamically it would be even better use a bit one value for the constants to set the wait based on the clock’s calibration.

The code already does this with XTAL wait.

ESP-Marius:
@2opremio just to let know you: we havent forgotten about this one, but any changes related to RTC parameters needs review and testing from the team responsible and they are unfortunately very busy :smile: 

They'd promised they will get to it eventually though! Sorry for the long delay.

PR #14468, opened 2024-08-30
fix(esp_clk_tree): SOC_MOD_CLK_REF_TICK now uses the corresponding de…

Andste82:
the function `esp_clk_tree_src_get_freq_hz` now uses the define `REF_CLK_FREQ` provided for this purpose for the clk_src `SOC_MOD_CLK_REF_TICK`.

developers who play around with the ref_tick to fulfill special requirements can now use the drivers for the peripherals that can use the ref_tick as source without any problems.


suda-morris:
sha=b7f95afe776b314efdd323ac60e46c35fdc3a5e1

PR #14469, opened 2024-08-30
fix(spinlock): comparison of integer expressions of different signedness (IDFGH-13581)

Andste82:
developers who do not use idf.py but prefer classic CMake currently receive a lot of warnings because a comparison in the header file `spinlock.h` is not correct.

`comparison of integer expressions of different signedness: 'esp_cpu_cycle_count_t' {aka 'long unsigned int'} and 'int32_t' {aka 'long int'} [-Wsign-compare]`

since it is a header, the gcc option `-Wall`, which is relevant for the application code, unfortunately also applies to all includes.

this fix casts the signed `timeout` to the corresponding unsigned type `esp_cpu_cycle_count_t` for a correct comparison.


KonstantinKondrashov:
sha=681b37266861b2f3d35e50d43b146691cdfe7281

PR #14481, opened 2024-09-02
Support ESP32 TinyS3 (IDFGH-13594)

zoonman:
In ESP32 TinyS3 LED is powered by pin IO17 and controlled by IO18.

Current example doesn't work out of the box for those boards because IO17 is low. We need to supply power in order to control WS2812B.

See schematic https://github.com/UnexpectedMaker/esp32s3/blob/main/schematics/schematic-tinys3.pdf

This board is very popular amongst beginners and proposed change will significantly ease learning curve.

### DIFF
--- a/examples/get-started/blink/main/blink_example_main.c
+++ b/examples/get-started/blink/main/blink_example_main.c
@@ -91,6 +91,19 @@ static void configure_led(void)
 void app_main(void)
 {
 
+#ifdef CONFIG_IDF_TARGET_ESP32S3
+    /* In ESP32TinyS3 LED is powered by pin IO17 and controlled by IO18.
+     *
+     * Don't forget to run idf.py menuconfig and set
+     * Blink LED type: strip, periferal to RMT
+     * Blink GPIO Number: 18
+     *
+     * */
+    gpio_set_direction(GPIO_NUM_17, GPIO_MODE_OUTPUT);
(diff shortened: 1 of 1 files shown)


suda-morris:
@zoonman Thanks for your contribution to esp-idf. It's nice to see a :new:  ESP32-S3 board.

> In ESP32 TinyS3 LED is powered by pin IO17

This looks dangerous, as usual, we don't recommend connecting a GPIO to a device's VDD, becuase GPIO has limited current output ability.

> Current example doesn't work out of the box for those boards because IO17 is low.

Unfortunately, this is a common "issue" for esp-idf examples, that it can't work out of the box on many boards, because it doesn't have the idea of "what board is running with". "Set the GPIO17" may benefit for your board, but may cause problem for other boards. 

Since this is a famous board, I would suggest making a BSP package for it. You can find many existing BSP packages [here](https://github.com/espressif/esp-bsp/tree/master/bsp).

zoonman:
Hi @suda-morris . I think making BSP package is slightly above my commitment plan.

I think it would be worth for @UnexpectedMaker to look into something like that.

zoonman:
I am closing this one. I hope @UnexpectedMaker will look into BSP package option. I honestly don't have time for that.

@suda-morris thank you for replying and explanation. I truly appreciate that.

PR #14488, opened 2024-09-03
Provisioning more than ssid and password (Discussion) (IDFGH-13601)

eug-ev:
Hi!
I've recently done a little change to wifi_provisioning library to support custom data provisioning along with wifi ssid and password.
Briefly, how it works:
If we pass this to password:
```!@CDATA_%DATA_LEN%%DATA%%WifiPassword%```
It would parse this string and place DATA in nvs as blob.
As example:
```!@CDATA_QwAAAA==tok=1234567890:AIHyuaZFjK6qegPCojGJI0_0l7GPUNV8_PD,userid=123456789asdfzxcvb```
where %DATA_LEN% is QwAAAA== (0x43),
wifi password - asdfzxcvb

Not ideal solution but it works!
Maybe some ideas for improvement or redesign to have this in idf?
Rationale: I needed to provision telegram bot token and my telegram user ID, apart from wifi credentials

### DIFF
--- a/components/wifi_provisioning/CMakeLists.txt
+++ b/components/wifi_provisioning/CMakeLists.txt
@@ -29,5 +29,5 @@ endif()
 idf_component_register(SRCS "${srcs}"
                     INCLUDE_DIRS include
                     PRIV_INCLUDE_DIRS src proto-c
-                    REQUIRES lwip protocomm
+                    REQUIRES lwip protocomm mbedtls nvs_flash
                     PRIV_REQUIRES protobuf-c bt json esp_timer esp_wifi)
--- a/components/wifi_provisioning/src/wifi_config.c
+++ b/components/wifi_provisioning/src/wifi_config.c
@@ -17,6 +17,9 @@
 #include <esp_err.h>
 #include <esp_log.h>
 
+#include "mbedtls/base64.h"
+#include "nvs.h"
+
 #include "wifi_constants.pb-c.h"
 #include "wifi_config.pb-c.h"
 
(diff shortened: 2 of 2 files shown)


DaveLegg:
This functionality already exists, you can use calls to wifi_prov_mgr_endpoint_create/wifi_prov_mgr_endpoint_register to add additional endpoints to receive your custom data, so in your case likely one for the telegram bot token, and one for the user ID

[wifi_prov_mgr_endpoint_create](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/provisioning/wifi_provisioning.html#_CPPv429wifi_prov_mgr_endpoint_createPKc)

eug-ev:
Thanks for the answers and sorry I hadn't researched more for this.

Does Android app supports some extra endpoints for custom data?

hrushikesh430:
> This functionality already exists, you can use calls to wifi_prov_mgr_endpoint_create/wifi_prov_mgr_endpoint_register to add additional endpoints to receive your custom data, so in your case likely one for the telegram bot token, and one for the user ID
> 
> [wifi_prov_mgr_endpoint_create](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/provisioning/wifi_provisioning.html#_CPPv429wifi_prov_mgr_endpoint_createPKc)

@eug-ev  check this [example](https://github.com/espressif/esp-idf/blob/master/examples/provisioning/wifi_prov_mgr/main/app_main.c#L460) for reference.

hrushikesh430:
> Thanks for the answers and sorry I hadn't researched more for this.
> 
> Does Android app supports some extra endpoints for custom data?
> 
> I've found only one related function in Android repo, sendDataToCustomEndpoint. However its not used anywhere and probably supposed to be implemented in custom Android app.
> 
> My point was to add some support in already existing application. Thanks for attention!



> Thanks for the answers and sorry I hadn't researched more for this.
> 
> Does Android app supports some extra endpoints for custom data?
> 
> I've found only one related function in Android repo, sendDataToCustomEndpoint. However its not used anywhere and probably supposed to be implemented in custom Android app.
> 
> My point was to add some support in already existing application. Thanks for attention!

Yes this Api will help. If you need example or some issue you can file an issue regarding this api on this [repo](https://github.com/espressif/esp-idf-provisioning-android)

hrushikesh430:
As API is already available for custom data as mentioned above, so closing this PR.

PR #14493, opened 2024-09-03
Make esp_mbedtls_server_session_create async compatible (IDFGH-13606)

thetek42:
This makes `esp_mbedtls_server_session_create` async-friendly by splitting up the function into two parts. This change was originally proposed here in order to make an asynchronous TLS server possible: https://github.com/esp-rs/esp-idf-svc/pull/368#discussion_r1512376482

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -73,6 +73,8 @@ static const char *TAG = "esp-tls";
 #define _esp_tls_free_client_session        esp_mbedtls_free_client_session
 #define _esp_tls_get_ssl_context            esp_mbedtls_get_ssl_context
 #define _esp_tls_server_session_create      esp_mbedtls_server_session_create
+#define _esp_tls_server_session_init        esp_mbedtls_server_session_init
+#define _esp_tls_server_session_continue_async     esp_mbedtls_server_session_continue_async
 #define _esp_tls_server_session_delete      esp_mbedtls_server_session_delete
 #define _esp_tls_server_session_ticket_ctx_init    esp_mbedtls_server_session_ticket_ctx_init
 #define _esp_tls_server_session_ticket_ctx_free    esp_mbedtls_server_session_ticket_ctx_free
@@ -90,6 +92,8 @@ static const char *TAG = "esp-tls";
 #define _esp_tls_conn_delete                esp_wolfssl_conn_delete
 #define _esp_tls_net_init                   esp_wolfssl_net_init
 #define _esp_tls_server_session_create      esp_wolfssl_server_session_create
+#define _esp_tls_server_session_init        esp_wolfssl_server_session_init
+#define _esp_tls_server_session_continue_async     esp_wolfssl_server_session_continue_async
 #define _esp_tls_server_session_delete      esp_wolfssl_server_session_delete
 #define _esp_tls_get_bytes_avail            esp_wolfssl_get_bytes_avail
 #define _esp_tls_init_global_ca_store       esp_wolfssl_init_global_ca_store
@@ -702,6 +706,22 @@ void esp_tls_cfg_server_session_tickets_free(esp_tls_cfg_server_t *cfg)
 int esp_tls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
 {
     return _esp_tls_server_session_create(cfg, sockfd, tls);
+}
+/**
+ * @brief      Initialization part of esp_tls_server_session_create
+ */
+int esp_tls_server_session_init(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
+{
+    return _esp_tls_server_session_init(cfg, sockfd, tls);
+}
+/**
+ * @brief      Asynchronous continue of esp_tls_server_session_create, to be
+ *             called in a loop by the user until it returns 0,
+ *             ESP_TLS_ERR_SSL_WANT_READ or ESP_TLS_ERR_SSL_WANT_WRITE
+ */
+int esp_tls_server_session_continue_async(esp_tls_t *tls)
+{
+    return _esp_tls_server_session_continue_async(tls);
 }
 /**
  * @brief      Close the server side TLS/SSL connection and free any allocated resources.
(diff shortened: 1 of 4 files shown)


components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -914,6 +914,21 @@ esp_err_t set_client_config(const char *hostname, size_t hostlen, esp_tls_cfg_t
  * @brief      Create TLS/SSL server session
  */
 int esp_mbedtls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
+{
+    int ret = 0;
+    if ((ret = esp_mbedtls_server_session_create_start(cfg, sockfd, tls)) != 0) return ret;
+    while ((ret = esp_mbedtls_server_session_create_continue_async(tls)) != 0) {
+        if (ret != ESP_TLS_ERR_SSL_WANT_READ && ret != ESP_TLS_ERR_SSL_WANT_WRITE) {
```
AdityaHPatwardhan, line 921:
```suggestion
    if ((ret = esp_mbedtls_server_session_create_start(cfg, sockfd, tls)) != 0) {
        return ret;
    }
    while ((ret = esp_mbedtls_server_session_create_continue_async(tls)) != 0) {
        if (ret != ESP_TLS_ERR_SSL_WANT_READ && ret != ESP_TLS_ERR_SSL_WANT_WRITE) {
```

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -914,6 +914,21 @@ esp_err_t set_client_config(const char *hostname, size_t hostlen, esp_tls_cfg_t
...
+            return ret;
+        }
+    }
+    return ret;
+}
+
+/**
+ * @brief      Initialization part of esp_mbedtls_server_session_create
+ */
+int esp_mbedtls_server_session_create_start(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
```
AdityaHPatwardhan, line 931:
```suggestion
int esp_mbedtls_server_session_init(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
```

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -932,19 +947,28 @@ int esp_mbedtls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp
...
-    }
     return 0;
 }
+
+/**
+ * @brief      Asyncronous continue of esp_mbedtls_server_session_create, to be
+ *             called in a loop by the user until it returns 0, ESP_TLS_ERR_SSL_WANT_READ
+ *             or ESP_TLS_ERR_SSL_WANT_WRITE
+ */
+int esp_mbedtls_server_session_create_continue_async(esp_tls_t *tls)
```
AdityaHPatwardhan, line 958:
```suggestion
int esp_mbedtls_server_session_continue_async(esp_tls_t *tls)
```

AdityaHPatwardhan: Similar change in Header file

AdityaHPatwardhan (commented):
Thanks for the PR @thetek42,
I agree that an async version would be helpful.
So the mbedtls is an internal layer for esp-tls.
It is not supposed to be directly used publicly (outside of esp-tls component)
Instead I think we should create appropriate APIs for the esp-tls layer and use that layer publicly.

thetek42:
@AdityaHPatwardhan I updated the function names and also ported them to both WolfSSL and esp-tls.

components/esp-tls/esp_tls.c
```diff
@@ -703,6 +707,22 @@ int esp_tls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls
 {
     return _esp_tls_server_session_create(cfg, sockfd, tls);
 }
+/**
+ * @brief      Initialization part of esp_tls_server_session_create
+ */
+int esp_tls_server_session_init(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls_t *tls)
```
AdityaHPatwardhan, line 713: Regarding the new API.
Is it possible to keep the return type as esp_err_t in this case.
I think this would go well with the other APIs added in `esp_tls.h`.
Sorry for the late suggestion, I can also do this in a supplementary commit.

components/esp-tls/esp_tls.c
```diff
@@ -703,6 +707,22 @@ int esp_tls_server_session_create(esp_tls_cfg_server_t *cfg, int sockfd, esp_tls
...
+{
+    return _esp_tls_server_session_init(cfg, sockfd, tls);
+}
+/**
+ * @brief      Asynchronous continue of esp_tls_server_session_create, to be
```
AdityaHPatwardhan, line 718:
```suggestion
 * @brief      Asynchronous continue of esp_tls_server_session_create, to be
 *             called in a loop by the user until it returns 0. If this functions returns something other than 0, ESP_TLS_ERR_SSL_WANT_READ or ESP_TLS_ERR_SSL_WANT_WRITE, the esp-tls context must not be used and should be freed using esp_tls_conn_destroy();
 
```

AdityaHPatwardhan (commented):
Thanks for updating the PR.
Just a couple of suggestions, otherwise looks good to me

components/esp-tls/esp_tls.c
```diff
@@ -90,6 +92,8 @@ static const char *TAG = "esp-tls";
 #define _esp_tls_conn_delete                esp_wolfssl_conn_delete
 #define _esp_tls_net_init                   esp_wolfssl_net_init
 #define _esp_tls_server_session_create      esp_wolfssl_server_session_create
+#define _esp_tls_server_session_init        esp_wolfssl_server_session_init
```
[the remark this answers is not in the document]
thetek42: Sorry for the misunderstanding, but can you please clarify what your suggestion here is?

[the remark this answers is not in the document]
AdityaHPatwardhan: I meant that the support for wolfSSL need not be added. Since I am not sure if you have tested the wolfSSL support yourself and we dont require wolfSSL support to be added so we can just mark the `_esp_tls_server_session_create` as `NULL`.

and in the `esp_tls_server_session_create` API where you use `_esp_tls_server_session_create`
you can just handle this case by doing something like
```
if (_esp_tls_server_session_create == NULL) {
    return ESP_ERR_NOT_SUPPORTED.
}
```

AdityaHPatwardhan:
@thetek42 Actually, if it is fine with you, I can take over the PR. Please merge your changes in one commit. I will update my changes on a separate commit so your contribution would still be visible.

thetek42:
PR updated. Thanks!

AdityaHPatwardhan:
sha=083b39640524dc96d5bccffc396667f5f3a616f3

PR #14503, opened 2024-09-04
change(esp-tls): add option to enable/disable the full set of OCSP checks for wolfSSL (IDFGH-13619)

frankencode:
I've added an option to enable/disable WOLFSSL_OCSP_CHECKALL when enabling OCSP in esp_tls_wolfssl.c .

### DIFF
--- a/components/esp-tls/Kconfig
+++ b/components/esp-tls/Kconfig
@@ -115,4 +115,12 @@ menu "ESP-TLS"
         help
             Enable detailed debug prints for wolfSSL SSL library.
 
+    config ESP_TLS_OCSP_CHECKALL
+        bool "Enabled full OCSP checks for ESP-TLS"
+        depends on ESP_TLS_USING_WOLFSSL
+        default y
+        help
+            Enable a fuller set of OCSP checks: checking revocation status of intermediate certificates,
+            optional fallbacks to CRLs, etc.
+
 endmenu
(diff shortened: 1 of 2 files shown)


components/esp-tls/Kconfig
```diff
@@ -115,4 +115,12 @@ menu "ESP-TLS"
         help
             Enable detailed debug prints for wolfSSL SSL library.
 
+    config ESP_WOLFSSL_OCSP_CHECKALL
```
AdityaHPatwardhan, line 118: Can you please move this config option to `esp-wolfssl` component?

AdityaHPatwardhan: Yes, I was first confused with the name being ESP_WOLFSSL.
Maybe it would be better to change this config option as  `ESP_TLS_OCSP_CHECKALL`
Other things can remain the same

AdityaHPatwardhan:
@frankencode Thanks for the change.
Just one thing as mentioned above but otherwise the PR looks good to me.

frankencode:
OK, I've fixed that Kconfig to proper "ESP_TLS_" prefix.

components/esp-tls/Kconfig
```diff
@@ -115,4 +115,12 @@ menu "ESP-TLS"
         help
             Enable detailed debug prints for wolfSSL SSL library.
 
+    config ESP_TLS_OCSP_CHECKALL
+        bool "Enabled full OCSP checks for wolfSSL"
```
AdityaHPatwardhan, line 119: Im sorry, is it possible to change this wolfSSL to ESP-TLS as well?
I missed this earlier

frankencode:
Done.

AdityaHPatwardhan (approved):

AdityaHPatwardhan:
sha=ba27281c3ed7c3ee174db5bd75f4dc7cd698c241

PR #14512, opened 2024-09-05
feat(esp_http_client): Add function to delete all headers set by 'esp_http_client_set_header' (IDFGH-13628)

DCSBL:
Related to https://esp32.com/viewtopic.php?f=13&t=41695. TL;DR: Headers by persistent sessions created by `esp_http_client_init` are kept, and it is not easy to remove all headers unless you keep track of which one you set.

This PR adds a function that wraps the already existing `http_header_clean` function to help with that.

We still mis cleanup functions for other `esp_http_client_set_*` functions, but lets discuss this idea first :)

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -369,6 +369,11 @@ esp_err_t esp_http_client_delete_header(esp_http_client_handle_t client, const c
     return http_header_delete(client->request->headers, key);
 }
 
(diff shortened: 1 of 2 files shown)


DCSBL:
Note: I was unable to find unit tests for `esp_http_client_delete_header` so I did not add a unit test for this new feature.

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/72f3ef8469a617779a81df8099935f9b4da8329c.

PR #14641, opened 2024-09-28
fix(ble/bluedroid): Prevent esp_ble_gatts_send_indicate() with value_len > 512 (IDFGH-13782)

nebkat:
## Description

BLE indications are limited by max attribute length (512) and also by MTU size - 3 (see https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/attribute-protocol--att-.html#UUID-fd156898-28db-9cd8-e6e1-94d0278d0d1c).

Checking for MTU size is the user's responsibility, however an easy mistake is to chunk data by MTU size - 3 only. At the maximum MTU size of 517 results in 514, which is greater than 512, and causes heap corruption due to a buffer overflow.

## Checklist

### DIFF
--- a/components/bt/host/bluedroid/api/esp_gatts_api.c
+++ b/components/bt/host/bluedroid/api/esp_gatts_api.c
@@ -260,6 +260,11 @@ esp_err_t esp_ble_gatts_stop_service(uint16_t service_handle)
 esp_err_t esp_ble_gatts_send_indicate(esp_gatt_if_t gatts_if, uint16_t conn_id, uint16_t attr_handle,
                                       uint16_t value_len, uint8_t *value, bool need_confirm)
 {
+    if (value_len > ESP_GATT_MAX_ATTR_LEN) {
+        LOG_ERROR("%s, value_len > ESP_GATT_MAX_ATTR_LEN.", __func__);
+        return ESP_ERR_INVALID_SIZE;
+    }
+
     btc_msg_t msg = {0};
     btc_ble_gatts_args_t arg;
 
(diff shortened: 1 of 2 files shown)


nebkat:
@esp-zhp

nebkat:
Yes - not good!

https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bt/host/bluedroid/api/esp_gatts_api.c#L260-L292

Goes to

https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bt/host/bluedroid/btc/profile/std/gatt/btc_gatts.c#L677-L680

Goes to

https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bt/host/bluedroid/bta/gatt/bta_gatts_api.c#L411-L436

Line 427 then overflows `tBTA_GATTS_API_INDICATION::value[BTA_GATT_MAX_ATTR_LEN == 512]`

https://github.com/espressif/esp-idf/blob/1c468f68259065ef51afd114605d9122f13d9d72/components/bt/host/bluedroid/bta/gatt/include/bta_gatts_int.h#L112-L118

And since that buffer is at the end of `tBTA_GATTS_API_INDICATION` it goes into another heap object, causing unrelated subsystems to crash, so definitely one to fix ASAP.

nebkat:
For extra context: this would have started happening in `v5.3` as a result of this fix https://github.com/espressif/esp-idf/commit/20a780da8cccfe4b5af14c55e223a12af0432cd7, where previously MTU was always < 600.

nebkat:
Thanks @esp-zhp!

nebkat:
@esp-zhp I see also the documentation change was removed - once [this](https://github.com/espressif/esp-idf/commit/70f19793f8154463fefa9ebbc10836e91e52a4f3#commitcomment-158930205) is fixed it should ideally be added back. The max notification length remains `min(512, MTU - 3)`.

PR #14706, opened 2024-10-10
fix(ws_transport): correct split header bytes

### LINKED ISSUE #14704: Websocket parser fails when header/size/mask not received contiguously. (IDFGH-13857)
### Answers checklist.

### IDF version.

v5.1.2

### Espressif SoC revision.

ESP32-S3

### Operating System used.

Windows

Command line with idf.py

CMD

### Development Kit.

Custom board with esp32s3-wroom-1

### Power Supply used.

USB

I expect esp_transport_read_internal() to wait up to timeout_ms milliseconds to read data.

When reading a websocket header, and the buffer layer has 1B inside it, esp_transport_read_internal() returns 1B, and ws_read_header() does not check for this, so the stream becomes misaligned by 1B. Then WS parsing is corrupt and parser cannot be realigned.

### Steps to reproduce.

1. Decrease MTU to minimum
2. Setup server to send random-length WS frames
3. Hope that somehow we align on 1B buffer boundary
4. add assert(rlen==header) to ws_read_header() after esp_transport_read_internal()

### Debug Logs.

```plain
I (11887) SIOC: unhandled text msg:~

^This is because "~" is the next byte of the websocket header.
```

### More Information.

_No response_

bryghtlabs-richard:
Not ready to submit, prototype fix only. Posting for discussion.

## Description

Prototype workaround for WS failing to parse when WS frames get split across a very specific timing and spacing pattern.

This fix is functional, but may rarely exceed the expected timeout.

I'm not sure if this is the right layer to fix it - perhaps the underlying TCP and TLS transports should be fixed instead to always wait up to the max timeout before returning failed bytes?

## Related

Fixes #14704 

## Testing

To test, a printout was added when fallback is triggered. When we tried to make this failure more consistent, we could only get it to around 10% of the time. It's network-timing and traffic-burst specific.

---

## Checklist

### DIFF
--- a/components/tcp_transport/transport_ws.c
+++ b/components/tcp_transport/transport_ws.c
@@ -133,6 +133,33 @@ static int esp_transport_read_internal(transport_ws_t *ws, char *buffer, int len
     return to_read;
 }
 
+static int esp_transport_read_blocking(transport_ws_t *ws, char *buffer, int len, int timeout_ms)
+{
+    int orig_len = len;
+    int did_read = 0;
+
+    //Sometimes we get 1B/read, sometimes we need 8B?
+    //Never seen this need >2 reads, better be safe though - need those headers.
+    const unsigned MAX_ATTEMPT = 8;
+    //TODO: change this(or the underlying transports) to retry only until timeout_ms expires.
+    unsigned attempt = 0;
+    for(attempt = 0; attempt < MAX_ATTEMPT && len > 0; ++attempt){
+        int tmp = esp_transport_read_internal(ws, buffer, len, timeout_ms);
+        if(tmp < 0){
+            return tmp;
+        }
+
+        buffer += tmp;
+        len -= tmp;
+        did_read += tmp;
+    }
+
+    if(attempt > 1){
+        ESP_LOGI(TAG, "tried %u attempts to read %i bytes. did read %i bytes", attempt, orig_len, did_read);
+    }
+    return did_read;
+}
+
 static char *trimwhitespace(char *str)
 {
     char *end;
@@ -486,7 +513,7 @@ static int ws_read_header(esp_transport_handle_t t, char *buffer, int len, int t
     // Receive and process header first (based on header size)
     int header = 2;
     int mask_len = 4;
-    if ((rlen = esp_transport_read_internal(ws, data_ptr, header, timeout_ms)) <= 0) {
+    if ((rlen = esp_transport_read_blocking(ws, data_ptr, header, timeout_ms)) < header) {
         ESP_LOGE(TAG, "Error read data");
         return rlen;
     }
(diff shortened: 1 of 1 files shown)


filzek:
I occasionally encounter an issue with the websocket where a semaphore is halted for 10 seconds. It seems that the underlying `transport_ws` layer is experiencing performance problems and sometimes gets stuck. Additionally, there are other issues in the transport layer that close the connection with errno 118/119. We haven't been able to solve these problems yet.

We need to conduct more tests. We will try to check and print the payload sizes to determine what is causing the issue.

bryghtlabs-richard:
Could be related? I'm not familiar with when that semaphore is held. What file is that in?

The problem I've run into is whenevr the next layer down's read function returns early(before the timeout with fewer bytes than expected). When that happens, transport_ws's parser gets off a few bytes, then may misinterpret the payload as header. Worst-case, the following payload starts with byte 126 or 127, which are RFC6455 codes for "extended payload length". When that happens, transport_ws may try to read a large(125B - 4GB) frame - if the other side isn't sending a lot of data, this may take approximately far too long for the read to complete.

bryghtlabs-richard:
Are you using esp_websocket_client? It looks like that lock would be held across the call to esp_websocket_client_recv(), which can hang while parsing garbage. Are you using a binary payload by chance? That's likely to be worse - we're sending EngineIO, so our misinterpreted payload-as-header bytes are usually ASCII numbers.

filzek:
> Are you using esp_websocket_client? It looks like that lock would be held across the call to esp_websocket_client_recv(), which can hang while parsing garbage. Are you using a binary payload by chance? That's likely to be worse - we're sending EngineIO, so our misinterpreted payload-as-header bytes are usually ASCII numbers.

We are using text for both sending and receiving. The problem sometimes occurs during both receiving and sending, and we haven't been able to identify the source to reproduce it consistently. We can reproduce it by dropping the Wi-Fi connection and through other external actions, but this doesn't reveal the origin of the problem.

components/tcp_transport/transport_ws.c
```diff
@@ -486,7 +514,7 @@ static int ws_read_header(esp_transport_handle_t t, char *buffer, int len, int t
     // Receive and process header first (based on header size)
     int header = 2;
     int mask_len = 4;
-    if ((rlen = esp_transport_read_internal(ws, data_ptr, header, timeout_ms)) <= 0) {
+    if ((rlen = esp_transport_read_exact_size(ws, data_ptr, header, timeout_ms)) < header) {
```
david-cermak, line 517:
```suggestion
    if ((rlen = esp_transport_read_exact_size(ws, data_ptr, header, timeout_ms)) <= 0) {
```
Now, we're reading the *exact size* which reports an error if we'd get less bytes

david-cermak: (same in other locations)

david-cermak (commented):
LGTM, otherwise. Thanks for the fixes!

david-cermak (approved):
Thanks for the updates, LGTM!

PR #14789, opened 2024-10-26
Add files via upload (IDFGH-13958)

xsmilexxxl:
# bmx280 for esp-idf


suda-morris:
Hey @xsmilexxxl We're glad to see this sensor driver been supported on the IDF platform. But we think the esp-idf repository is not the best place to host these sensor drivers. Alternatively, we highly recommend users to use the [component registry](https://components.espressif.com/). Yes, to maintain the idf component under your namespace and upload it to the component registry (the way like the npm in the node js ecosystem). The benefit is, you can release a new version of this driver at anytime, it doesn't have to be coupled with the esp-idf's release cycle.

For how to use the component manager, I suggest you watching this [video](https://www.bilibili.com/video/BV1Ym4y1w7AA/?spm_id_from=333.999.0.0&vd_source=5facc390324ce87bd14041128ce34a77). The document of the component manger can be found [here](https://docs.espressif.com/projects/idf-component-manager/en/latest/).

components/esp-idf-bmx280/include/bmx280.h
```diff
@@ -0,0 +1,125 @@
...
+
+#include "driver/i2c.h"
+#include "sdkconfig.h"
+#include <limits.h>
+#include <stdint.h>
+
+#define BMXAPI extern
+
+/**
+ * 用于驱动设置的匿名结构体。
```
suda-morris, line 25: Suggest not using Chinese in the API comment.

Alvin1Zhang:
Thanks for contribution again, please help check suggestions here https://github.com/espressif/esp-idf/pull/14789#issuecomment-2440426387, feel free to reopen.

PR #14813, opened 2024-11-01
support for rfc7692 message deflate

ariken74:
## Description

Adds support for querying rsv1 flag of websocket header. Needed to implement per message deflate as RFC7692.

https://datatracker.ietf.org/doc/html/rfc7692#section-9.2

### DIFF
--- a/components/tcp_transport/include/esp_transport_ws.h
+++ b/components/tcp_transport/include/esp_transport_ws.h
@@ -150,6 +150,16 @@ int esp_transport_ws_send_raw(esp_transport_handle_t t, ws_transport_opcodes_t o
 bool esp_transport_ws_get_fin_flag(esp_transport_handle_t t);
 
 /**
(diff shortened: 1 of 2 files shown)


euripedesrocha:
Hi @ariken74, thanks for your contribution. Could you please squash the commits rewrite your contribution's commit using conventional commits so I can follow with the processing of this PR?

huming2207:
I'm also curious about this RFC7692 support but I think it's very incomplete? It seems like this PR only adds the `RSV1` compression indication bit's setter and getter, but not the actual compression.

ariken74:
> I'm also curious about this RFC7692 support but I think it's very incomplete? It seems like this PR only adds the `RSV1` compression indication bit's setter and getter, but not the actual compression.

The idea was to be able to read this flag from websocket_client and handle the RFC7692 there. It took so long to consider this PR, in the meantime I switched to the mqtt library.

PR #14834, opened 2024-11-05
[Examples] [Async Handlers] simplify code by splitting long_async_handler into two functions (IDFGH-14011)

chipweinberger:
I originally wrote this example, but made it too complicated.

I don't know what I was thinking. I was calling `long_async_handler` twice, and used a `is_on_async_worker_thread` to do different behavior depending on which thread you are on.

This is needlessly complicated.

It's much simpler to have two functions:
`long_handler` - called on http thread
`long_async` - called on async thread.

This leads to much more straight forward code.

I also simplified some variable names.

### DIFF
--- a/examples/protocols/http_server/async_handlers/main/main.c
+++ b/examples/protocols/http_server/async_handlers/main/main.c
@@ -38,7 +38,7 @@ static const char *TAG = "example";
 
 // Async requests are queued here while they wait to
 // be processed by the workers
-static QueueHandle_t async_req_queue;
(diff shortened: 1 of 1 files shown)


nileshkale123:
sha=a89a954a850bdb425edfef920ce7c05ab77188c9

chipweinberger:
please merge this

nileshkale123:
This is already merged in the commit https://github.com/espressif/esp-idf/commit/ac3559574676346c638c591e479eec5c74995499

PR #14868, opened 2024-11-11
updates the btm_rrm_t task to use the same core as in the wifi (IDFGH-14049)

filzek:
This Push updates the btm_rrm_t task to use the same core as in the wifi, so its become binded to the same core wifi is running, this is the best approach to the system.


AxelLin:
@filzek Does this PR fix any issue? (Just to check if this is a bug fix)

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/950da105b3d6c1f94f7f1dd8611e563f5be24882.

PR #14909, opened 2024-11-19
fix(esp-tls): Remove useless const from size paramter (IDFGH-14099)

0xFEEDC0DE64:
Get rid of useless const from size parameter as it does not allow for easy setup of the struct

### DIFF
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -49,7 +49,7 @@ typedef enum esp_tls_role {
  */
 typedef struct psk_key_hint {
     const uint8_t* key;                     /*!< key in PSK authentication mode in binary format */
-    const size_t   key_size;                /*!< length of the key */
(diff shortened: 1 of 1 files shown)


AdityaHPatwardhan:
Hi @0xFEEDC0DE64, The parameter has been marked as `const` to indicate that the function does not intend to update the value of the parameter in any case. Also, it would be considered as a breaking change now if we are to modify the type.
May I know what problem is it causing at your side?
Is it possible for you to `typecast` it to `const` at the time of initialising the config structure if your code is not defining it as a const variable?

0xFEEDC0DE64:
I just need to be able to build that struct from c++ code without having all the data already available on construction.

Please never make any class members const as this breaks anything like copy constructors and so on

AdityaHPatwardhan:
sha=4bbff34ac22449f7fd12bccad0382bacb5b99538

PR #14929, opened 2024-11-24
fix(esp_common): fix C++ variant of ESP_RETURN_VOID_ON_FALSE

chschu:
## Description

This PR fixes the `ESP_RETURN_VOID_ON_FALSE` macro used for newer versions of C++.

The macro parameter `err_code` is removed. It doesn't exist for other variants of the macro, and it isn't used in the changed macro definition.

This could be considered a breaking change, because existing C++ code might rely on the current definition of the macro.

## Checklist

### DIFF
--- a/components/esp_common/include/esp_check.h
+++ b/components/esp_common/include/esp_check.h
@@ -254,7 +254,7 @@ extern "C" {
  * Macro which can be used to check the condition. If the condition is not 'true', it prints the message
  * and returns without a value.
  */
-#define ESP_RETURN_VOID_ON_FALSE(a, err_code, log_tag, format, ...) do {                                   \
+#define ESP_RETURN_VOID_ON_FALSE(a, log_tag, format, ...) do {                                             \
         if (unlikely(!(a))) {                                                                              \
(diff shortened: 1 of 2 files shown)


safocl:
Can't https://github.com/espressif/esp-idf-cxx/blob/7532e372082dce481c1435bb82e880aca96b08c7/include/esp_exception.hpp#L49 be used for this?

chschu:
> Can't https://github.com/espressif/esp-idf-cxx/blob/7532e372082dce481c1435bb82e880aca96b08c7/include/esp_exception.hpp#L49 be used for this?

Even though `CHECK_THROW` and `ESP_RETURN_VOID_ON_FALSE` both interrupt control flow, they are fundamentally different. In particular, `ESP_RETURN_VOID_ON_FALSE` can be used with exceptions disabled.

The PR is not about using a different macro. It simply aligns the signatures of the three different definitions of `ESP_RETURN_VOID_ON_FALSE`.

ESP-Marius:
@chschu sorry, this one seems to have slipped through the cracks.

Changes seems reasonable to me!

ESP-Marius:
sha=f94862243973ab8d10055dce7b8f7de1441cc893

PR #14951, opened 2024-11-28
fix(esp_ulp): Add support for multiple ULP program embedding without name collision (IDFGH-14151)

### LINKED ISSUE #14945: Using multiple ULP program in a larger project fails silently at runtime (IDFGH-14143)
### Answers checklist.

### General issue report

It's a bit similar to #14921 in the concept. I have multiple different program for the ULP's LP core on ESP32 C6 CPU. I've declared them like this in the `main`'s CMakelists.txt:
```cmake
set(ulp_app_name ulp_${COMPONENT_NAME})
set(ulp_sources "ulp/ulp_main.c")
set(ulp_exp_dep_srcs ${srcs})

ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")

set(ulp_app_name2 ulp_ds_cal)
set(ulp_sources2 "ulp/ulp_ds_cal.c")

ulp_embed_binary(${ulp_app_name2} "${ulp_sources2}" "${ulp_exp_dep_srcs}")
```

The first issue happens if you have the same variable name in both file. In that case it's not possible, in C to include the automatically generated header in your main function, since both refer to the same variable, but variables are located at different address. So if I have, let's say a variable called `uint32_t state` in both file, the generated header `ulp_main.h`  and `ulp_ds_cal.h` both have an `extern uint32_t ulp_state;`. Yet, the first ULP program will have this linker script: `PROVIDE ( ulp_state = 0x50000ab8 );` and the second will have ` PROVIDE ( ulp_state = 0x50000640 );`.
Yet, it'll link fine but obviously fails at runtime, since the variable will just refer to a random memory location when using the second program.

So, since it's clear that both variables can't be referred with a single name, I've used C++ like this to move each declaration in its own namespace:
```cpp
namespace ULPMain 
{
  #include "ulp_main.h"
}
namespace ULPDSCal
{
  #include "ulp_ds_cal.h"
}

// Now, ULPDSCal::ulp_state should refer to ulp_ds_cal.c's version of state
```
This strangely links fine (I would have expected that the linker choked on an undefined reference on `extern uint32_t ULPDSCal::ulp_state` but it's perfectly happy with that), but obviously it doesn't work at runtime. If I print the address of the `ULPDSCal::ulp_state` variable in HP core, I get 0x50000ab8, which is the address of the first program, not the second one. 

I think this is a bug in the linker script that's not using the expected mangled name for linking (it should have failed to link a symbol with a different name, the mangled name for the `ULPDSCal::ulp_state` variable is not `ulp_state`).

It can only be solved by either changing the symbol processing script to allow changing the prefix for the variable (so the first script could use `ulp_` and the second would use, for example, `ulp_2_`) or, better, to allow using a C++ namespace instead (so instead of `ulp_VARIABLE`, simply have a `NAMESPACE::VARIABLE` declaration).

Another option would be to have a macro for variable declaration allowing to position the variable at a specific address. Notice that using `uint32_t variable = *(volatile uint32_t*)0x50000210;` is not possible since the linker can place whatever it wants at this address later on. On ARM, there's the [attribute `__attribute(at(address))__`](https://developer.arm.com/documentation/dui0376/c/Compiler-specific-Features/Variable-attributes/--attribute----at-address----variable-attribute) that can be used for this exact feature.

There also the hacky:
```cpp
extern const uint32_t state;
asm(".equ state, 0x50000210");
```
that seems to work but this become completely invisible to the linker's script and variable renaming so it's also very fragile since the linker doesn't not know that a variable exists at 0x50000210 and can put whatever it wants here, breaking the code.

X-Ryl669:
## Description

Currently, the ULP's auto variable name embedding script has multiple issues:
1. It names variable in the global namespace (in extern "C" mode) so multiple ULP program using the same variable name results in impossible-to-solve access and memory corruption (see linked issue)
2. It doesn't support generating arrays (so one has to trick/hack around a declared variable, rendering the whole process useless)
3. It create a useless linker script in the build process
4. It creates a lots of useless symbols declaration (like all functions and global included objects), which, if used in the HP core, crashes the HP core

This PR fixes it:

1. It changes from using `nm` cross compilation tool to `readelf` tool for extracting symbols. This allows keeping the symbol's scope (global or local), the symbol type (object or function or ...) and the symbol size (for arrays).
2. It only extract the meaningful symbols and doesn't pollute the global scope with unusable symbols like `ulp_lp_core_lp_ana_peri_intr_handler`
3. It allows to specify the prefix to use for generating the ULP's variable name (defaults to the current `ulp_` prefix, but now, you can specify it)
4. You can use C++ prefix too (like a namespace `ULPmain::` and it'll mangle the name and work without `extern "C"`): perfect to move the shorter variables in their own namespace and avoid collisions without dumb long names
5. It removes the generated linker script (useless to specify the symbol location, it can be done in the header file directly). 
6. It add support for array `uint32_t myVariable[32]` is correctly exported so can be used like a native variable in HP core, no more hacky `((uint32_t*)myVariable)[0]` like specified in the documentation.

From a user perspective, nothing changes (the useless symbol will be removed, but they couldn't be used anyway previously).
However, she can now replace the line:
```cmake
set(ulp_app_name ulp_${COMPONENT_NAME})
set(ulp_sources "ulp/ulp_main.c")
set(ulp_exp_dep_srcs ${srcs})

ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")

set(ulp_app_name2 ulp_ds_cal)
set(ulp_sources2 "ulp/ulp_ds_cal.c")

ulp_embed_binary_prefix(${ulp_app_name2} "${ulp_sources2}" "${ulp_exp_dep_srcs}")
```
to this:
```cmake
set(ulp_app_name ulp_${COMPONENT_NAME})
set(ulp_sources "ulp/ulp_main.c")
set(ulp_exp_dep_srcs ${srcs})

ulp_embed_binary_prefix(${ulp_app_name} "ULPmain::" "${ulp_sources}" "${ulp_exp_dep_srcs}")

set(ulp_app_name2 ulp_ds_cal)
set(ulp_sources2 "ulp/ulp_ds_cal.c")

ulp_embed_binary_prefix(${ulp_app_name2} "ULPds_cal::" "${ulp_sources2}" "${ulp_exp_dep_srcs}")
```
And build collision-free and working multiple ULP binary. 

## Related

Fix #14945 and other issues too related to misusage of ULP symbols in HP core.

## Testing

I've tested on my system with ESP32C6. I haven't tested with other CPU types. However, the tools used in this PR are common cross compilation tools, it shouldn't depend on the CPU architecture.
---

## Checklist

- [*] 🚨 This PR does not introduce breaking changes.

### DIFF
--- a/components/ulp/cmake/CMakeLists.txt
+++ b/components/ulp/cmake/CMakeLists.txt
@@ -8,4 +8,4 @@ include(IDFULPProject)
 
 ulp_apply_default_options(${ULP_APP_NAME})
 ulp_apply_default_sources(${ULP_APP_NAME})
-ulp_add_build_binary_targets(${ULP_APP_NAME})
+ulp_add_build_binary_targets(${ULP_APP_NAME} ${ULP_VAR_PREFIX})
--- a/components/ulp/cmake/IDFULPProject.cmake
+++ b/components/ulp/cmake/IDFULPProject.cmake
@@ -178,7 +178,7 @@ function(ulp_apply_default_sources ulp_app_name)
     endif()
 endfunction()
 
-function(ulp_add_build_binary_targets ulp_app_name)
+function(ulp_add_build_binary_targets ulp_app_name prefix)
 
     if(CONFIG_ULP_COPROC_TYPE_LP_CORE)
         set(ULP_BASE_ADDR "0x0")
@@ -190,7 +190,7 @@ function(ulp_add_build_binary_targets ulp_app_name)
 
     # Dump the list of global symbols in a convenient format
     add_custom_command(OUTPUT ${ULP_APP_NAME}.sym
-                    COMMAND ${CMAKE_NM} -f posix -g $<TARGET_FILE:${ulp_app_name}> > ${ulp_app_name}.sym
+                    COMMAND ${CMAKE_READELF} -sW $<TARGET_FILE:${ulp_app_name}> > ${ulp_app_name}.sym
                     DEPENDS ${ulp_app_name}
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
@@ -201,7 +201,7 @@ function(ulp_add_build_binary_targets ulp_app_name)
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
     add_custom_command(OUTPUT ${ulp_app_name}.ld ${ulp_app_name}.h
-                    COMMAND ${ULP_MAP_GEN} -s ${ulp_app_name}.sym -o ${ulp_app_name} --base ${ULP_BASE_ADDR}
+                    COMMAND ${ULP_MAP_GEN} -s ${ulp_app_name}.sym -o ${ulp_app_name} --base ${ULP_BASE_ADDR} --prefix ${prefix}
                     DEPENDS ${ulp_app_name}.sym
                     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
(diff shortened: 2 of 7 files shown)


KonstantinKondrashov:
Orignal https://github.com/espressif/esp-idf/pull/14954

PR #14954, opened 2024-11-28
fix(esp_ulp): Add support for multiple ULP program embedding without name collision (IDFGH-14152)

### LINKED ISSUE #14945: Using multiple ULP program in a larger project fails silently at runtime (IDFGH-14143)

X-Ryl669:
## Description

Currently, the ULP's auto variable name embedding script has multiple issues:
1. It names variable in the global namespace (in extern "C" mode) so multiple ULP program using the same variable name results in impossible-to-solve access and memory corruption (see linked issue)
2. It doesn't support generating arrays (so one has to trick/hack around a declared variable, rendering the whole process useless)
3. It create a useless linker script in the build process
4. It creates a lots of useless symbols declaration (like all functions and global included objects), which, if used in the HP core, crashes the HP core

This PR fixes it:

1. It changes from using `nm` cross compilation tool to `readelf` tool for extracting symbols. This allows keeping the symbol's scope (global or local), the symbol type (object or function or ...) and the symbol size (for arrays).
2. It only extract the meaningful symbols and doesn't pollute the global scope with unusable symbols like `ulp_lp_core_lp_ana_peri_intr_handler`
3. It allows to specify the prefix to use for generating the ULP's variable name (defaults to the current `ulp_` prefix, but now, you can specify it)
4. You can use C++ prefix too (like a namespace `ULPmain::` and it'll mangle the name and work without `extern "C"`): perfect to move the shorter variables in their own namespace and avoid collisions without dumb long names
5. It removes the generated linker script (useless to specify the symbol location, it can be done in the header file directly). 
6. It add support for array `uint32_t myVariable[32]` is correctly exported so can be used like a native variable in HP core, no more hacky `((uint32_t*)myVariable)[0]` like specified in the documentation.

From a user perspective, nothing changes (the useless symbol will be removed, but they couldn't be used anyway previously).
However, she can now replace the line:
```cmake
set(ulp_app_name ulp_${COMPONENT_NAME})
set(ulp_sources "ulp/ulp_main.c")
set(ulp_exp_dep_srcs ${srcs})

ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")

set(ulp_app_name2 ulp_ds_cal)
set(ulp_sources2 "ulp/ulp_ds_cal.c")

ulp_embed_binary_prefix(${ulp_app_name2} "${ulp_sources2}" "${ulp_exp_dep_srcs}")
```
to this:
```cmake
set(ulp_app_name ulp_${COMPONENT_NAME})
set(ulp_sources "ulp/ulp_main.c")
set(ulp_exp_dep_srcs ${srcs})

ulp_embed_binary_prefix(${ulp_app_name} "ULPmain::" "${ulp_sources}" "${ulp_exp_dep_srcs}")

set(ulp_app_name2 ulp_ds_cal)
set(ulp_sources2 "ulp/ulp_ds_cal.c")

ulp_embed_binary_prefix(${ulp_app_name2} "ULPds_cal::" "${ulp_sources2}" "${ulp_exp_dep_srcs}")
```
And build collision-free and working multiple ULP binary. 

## Related

Fix #14945 and other issues too related to misusage of ULP symbols in HP core.

## Testing

I've tested on my system with ESP32C6. I haven't tested with other CPU types. 
However, the tools used in this PR are common cross compilation tools, it shouldn't depend on the CPU architecture.

## Checklist

### DIFF
--- a/components/ulp/esp32ulp_mapgen.py
+++ b/components/ulp/esp32ulp_mapgen.py
@@ -2,71 +2,105 @@
 # esp32ulp_mapgen utility converts a symbol list provided by nm into an export script
 # for the linker and a header file.
 #
-# SPDX-FileCopyrightText: 2016-2021 Espressif Systems (Shanghai) CO LTD
+# SPDX-FileCopyrightText: 2016-2024 Espressif Systems (Shanghai) CO LTD
 # SPDX-License-Identifier: Apache-2.0
-
-from __future__ import print_function
-
 import argparse
 import os
+import re
 import textwrap
 import typing
 
 UTIL = os.path.basename(__file__)
 
 
-def gen_ld_h_from_sym(f_sym: typing.TextIO, f_ld: typing.TextIO, f_h: typing.TextIO, base_addr: int) -> None:
+def name_mangling(name: str) -> str:
+    # Simple and dumb name mangling for namespaced name following GCC algorithm
+    ns, n = name.split('::')
+    return '_ZN{0}{1}{2}{3}E'.format(len(ns), ns, len(n), n)
+
+
+def gen_ld_h_from_sym(f_sym: typing.TextIO, f_ld: typing.TextIO, f_h: typing.TextIO, base_addr: int, prefix: str) -> None:
     f_ld.write(textwrap.dedent(
         f"""
         /* ULP variable definitions for the linker.
          * This file is generated automatically by {UTIL} utility.
          */
-        """
+        """  # noqa: E222
     ))
-    f_h.write(textwrap.dedent(
-        f"""
-        /* ULP variable definitions for the compiler.
-         * This file is generated automatically by {UTIL} utility.
-         */
-        #pragma once
-        #ifdef __cplusplus
-        extern "C" {{
-        #endif
-        """
-    ))
+    cpp_mode = False
+    var_prefix = prefix
+    namespace = ''
+    if '::' in prefix:
+        # C++ mode, let's avoid the extern "C" type and instead use namespace
+        f_h.write(textwrap.dedent(
+            f"""
+            /* ULP variable definitions for the compiler.
+             * This file is generated automatically by {UTIL} utility.
+             */
+            #pragma once
+            """  # noqa: E222
+        ))
+        tmp = prefix.split('::')
+        namespace = tmp[0]
+        var_prefix = '_'.join(tmp[1:])  # Limit to a single namespace here to avoid complex mangling rules
+        f_h.write('namespace {0} {{\n'.format(namespace))
+        cpp_mode = True
+    else:
+        f_h.write(textwrap.dedent(
+            f"""
+            /* ULP variable definitions for the compiler.
+             * This file is generated automatically by {UTIL} utility.
+             */
+            #pragma once
+            #ifdef __cplusplus
+            extern "C" {{
+            #endif
+            """  # noqa: E222
+        ))
 
+    expr = re.compile('^\\s*\\d+: ([a-f0-9]{8})\\s+(\\d+) OBJECT\\s+GLOBAL\\s+DEFAULT\\s+[^ ]+ (.*)$')
+    already_defined = 'this_symbol_is_already_defined_please_use_prefix_in_ulp_embed_binary'
     for line in f_sym:
-        # NM "posix" format output has the following structure:
-        # symbol_name symbol_type addr_hex [size_hex]
-        parts = line.split()
-        name = parts[0]
-        addr = int(parts[2], 16) + base_addr
-        f_h.write('extern uint32_t ulp_{0};\n'.format(name))
-        f_ld.write('PROVIDE ( ulp_{0} = 0x{1:08x} );\n'.format(name, addr))
+        # readelf format output has the following structure:
+        #   index: addr_hex     size TYPE  SCOPE DEFAULT    junk symbol_name
+        # So match the line with a regular expression to parse it first
+        groups = expr.match(line)
+        if groups is None:  # Ignore non global or non object
+            continue
+        addr = int(groups.group(1), 16) + base_addr
+        size = int(groups.group(2))
+        name = var_prefix + groups.group(3)
+        f_h.write('extern uint32_t {0}{1};\n'.format(name, '[{0}]'.format(int(size / 4)) if size > 4 else ''))
+        f_ld.write('{0} = DEFINED({0}) ? {2} : 0x{1:08x};\n'.format(
+            name_mangling(namespace + '::' + name) if cpp_mode else name, addr, already_defined))
 
-    f_h.write(textwrap.dedent(
-        """
-        #ifdef __cplusplus
-        }
-        #endif
-        """
-    ))
+    if cpp_mode:
+        f_h.write('}\n')
+    else:
+        f_h.write(textwrap.dedent(
+            """
+            #ifdef __cplusplus
+            }
+            #endif
+            """
+        ))
 
 
 def main() -> None:
     description = ('This application generates .h and .ld files for symbols defined in input file. '
-                   'The input symbols file can be generated using nm utility like this: '
-                   '<PREFIX>nm -g -f posix <elf_file> > <symbols_file>')
+                   'The input symbols file can be generated using readelf utility like this: '
+                   '<PREFIX>readelf -sW <elf_file> > <symbols_file>')
 
     parser = argparse.ArgumentParser(description=description)
     parser.add_argument('-s', '--symfile', required=True, help='symbols file name', metavar='SYMFILE', type=argparse.FileType('r'))
     parser.add_argument('-o', '--outputfile', required=True, help='destination .h and .ld files name prefix', metavar='OUTFILE')
     parser.add_argument('--base-addr', required=True, help='base address of the ULP memory, to be added to each symbol')
+    parser.add_argument('-p', '--prefix', required=False, help='prefix for generated header file', default='ulp_')
 
     args = parser.parse_args()
 
     with open(args.outputfile + '.h', 'w') as f_h, open(args.outputfile + '.ld', 'w') as f_ld:
-        gen_ld_h_from_sym(args.symfile, f_ld, f_h, int(args.base_addr, 0))
+        gen_ld_h_from_sym(args.symfile, f_ld, f_h, int(args.base_addr, 0), args.prefix)
 
 
 if __name__ == '__main__':
--- a/components/ulp/project_include.cmake
+++ b/components/ulp/project_include.cmake
@@ -2,7 +2,7 @@
 #
 # Create ULP binary and embed into the application.
 
-function(__setup_ulp_project app_name project_path s_sources exp_dep_srcs)
+function(__setup_ulp_project app_name project_path prefix s_sources exp_dep_srcs)
 
     if(NOT CMAKE_BUILD_EARLY_EXPANSION)
         spaces2list(s_sources)
@@ -60,6 +60,7 @@ function(__setup_ulp_project app_name project_path s_sources exp_dep_srcs)
                             -DCMAKE_TOOLCHAIN_FILE=${TOOLCHAIN_FLAG}
                             -DULP_S_SOURCES=$<TARGET_PROPERTY:${app_name},ULP_SOURCES>
                             -DULP_APP_NAME=${app_name}
+                            -DULP_VAR_PREFIX=${prefix}
                             -DCOMPONENT_DIR=${COMPONENT_DIR}
                             -DCOMPONENT_INCLUDES=$<TARGET_PROPERTY:${COMPONENT_TARGET},INTERFACE_INCLUDE_DIRECTORIES>
                             -DIDF_TARGET=${idf_target}
@@ -92,9 +93,14 @@ function(__setup_ulp_project app_name project_path s_sources exp_dep_srcs)
 endfunction()
 
 function(ulp_embed_binary app_name s_sources exp_dep_srcs)
-    __setup_ulp_project("${app_name}" "${idf_path}/components/ulp/cmake" "${s_sources}" "${exp_dep_srcs}")
+    cmake_parse_arguments(ULP "" "PREFIX" "" ${ARGN})
+    if(NOT ULP_PREFIX)
+        set(ULP_PREFIX "ulp_")
+    endif()
+    __setup_ulp_project("${app_name}" "${idf_path}/components/ulp/cmake"
+                        "${ULP_PREFIX}" "${s_sources}" "${exp_dep_srcs}")
 endfunction()
 
 function(ulp_add_project app_name project_path)
-    __setup_ulp_project("${app_name}" "${project_path}" "" "")
+    __setup_ulp_project("${app_name}" "${project_path}" "ulp_"  "" "")
 endfunction()
--- a/docs/en/api-reference/system/ulp-lp-core.rst
+++ b/docs/en/api-reference/system/ulp-lp-core.rst
@@ -36,7 +36,21 @@ Using ``ulp_embed_binary``
     ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")
 
 The first argument to ``ulp_embed_binary`` specifies the ULP binary name. The name specified here is also used by other generated artifacts such as the ELF file, map file, header file, and linker export file. The second argument specifies the ULP source files. Finally, the third argument specifies the list of component source files which include the header file to be generated. This list is needed to build the dependencies correctly and ensure that the generated header file is created before any of these files are compiled. See the section below for the concept of generated header files for ULP applications.
+Variables in the ULP code will be prefixed with ``ulp_`` (default value) in this generated header file.
 
+If you need to embed multiple ULP programs, you may add a custom prefix in order to avoid conflicting variable names like this:
+
+.. code-block:: cmake
+
+    idf_component_register()
+
+    set(ulp_app_name ulp_${COMPONENT_NAME})
+    set(ulp_sources "ulp/ulp_c_source_file.c" "ulp/ulp_assembly_source_file.S")
+    set(ulp_exp_dep_srcs "ulp_c_source_file.c")
+
+    ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}" PREFIX "ULP::")
+
+The additional argument can be a C style prefix (like ``ulp2_``) or a C++ style prefix (like ``ULP::``).
 
 Using a Custom CMake Project
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -140,13 +154,7 @@ The header file contains the declaration of the symbol:
 
     extern uint32_t ulp_measurement_count;
 
-Note that all symbols (variables, arrays, functions) are declared as ``uint32_t``. For functions and arrays, take the address of the symbol and cast it to the appropriate type.
-
-The generated linker script file defines the locations of symbols in LP_MEM:
-
-.. code-block:: none
-
-    PROVIDE ( ulp_measurement_count = 0x50000060 );
+Note that all symbols (variables, functions) are declared as ``uint32_t``. Arrays are declared as ``uint32_t [SIZE]``. For functions, take the address of the symbol and cast it to the appropriate type.
 
 To access the ULP LP core program variables from the main program, the generated header file should be included using an ``include`` statement. This allows the ULP LP core program variables to be accessed as regular variables.
 
@@ -161,7 +169,7 @@ To access the ULP LP core program variables from the main program, the generated
 .. note::
 
     Variables declared in the global scope of the LP core program reside in either the ``.bss`` or ``.data`` section of the binary. These sections are initialized when the LP core binary is loaded and executed. Accessing these variables from the main program on the HP-Core before the first LP core run may result in undefined behavior.
-
+    The ``ulp_`` prefix is the default value. You can specify the prefix to use with ``ulp_embed_binary`` to avoid name collisions for multiple ULP programs.
 
 Starting the ULP LP Core Program
 --------------------------------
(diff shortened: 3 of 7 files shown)


components/ulp/project_include.cmake
```diff
@@ -86,15 +87,18 @@ function(__setup_ulp_project app_name project_path s_sources exp_dep_srcs)
...
         target_add_binary_data(${COMPONENT_LIB} ${CMAKE_CURRENT_BINARY_DIR}/${app_name}/${app_name}.bin BINARY)
     endif()
 endfunction()
 
 function(ulp_embed_binary app_name s_sources exp_dep_srcs)
-    __setup_ulp_project("${app_name}" "${idf_path}/components/ulp/cmake" "${s_sources}" "${exp_dep_srcs}")
+    __setup_ulp_project("${app_name}" "${idf_path}/components/ulp/cmake" "ulp_" "${s_sources}" "${exp_dep_srcs}")
+endfunction()
+
+function(ulp_embed_binary_prefix app_name prefix s_sources exp_dep_srcs)
```
igrr, line 98: Nitpick, I think it would be nicer to add and optional PREFIX keyword argument to ulp_embed_binary, so that the usage would be 
`ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}" PREFIX "::ulp")`

components/ulp/esp32ulp_mapgen.py
```diff
@@ -2,71 +2,95 @@
...
+        if groups is None:  # Ignore non global or non object
+            continue
+        addr = int(groups.group(1), 16) + base_addr
+        size = int(groups.group(2))
+        name = groups.group(3)
+        if not cpp_mode:
+            name = prefix + name
+        f_h.write('extern uint32_t {0}{1};'.format(name, '[{0}]'.format(int(size / 4)) if size > 4 else ''))
+        # Then set the address to find the variable
+        f_h.write(' asm(".equ {0}, 0x{1:08X}");\n'.format(name_mangling(prefix + name) if cpp_mode else name, addr))
```
igrr, line 65: Compared to generation of the linker script, seems like this approach might prevent generation of bindings for programs written in other languages, e.g. Rust.

components/ulp/esp32ulp_mapgen.py
```diff
@@ -2,71 +2,95 @@
...
     parser.add_argument('-s', '--symfile', required=True, help='symbols file name', metavar='SYMFILE', type=argparse.FileType('r'))
     parser.add_argument('-o', '--outputfile', required=True, help='destination .h and .ld files name prefix', metavar='OUTFILE')
     parser.add_argument('--base-addr', required=True, help='base address of the ULP memory, to be added to each symbol')
+    parser.add_argument('-p', '--prefix', required=False, help='prefix for generated header file', default='ulp_')
 
     args = parser.parse_args()
 
-    with open(args.outputfile + '.h', 'w') as f_h, open(args.outputfile + '.ld', 'w') as f_ld:
-        gen_ld_h_from_sym(args.symfile, f_ld, f_h, int(args.base_addr, 0))
+    with open(args.outputfile + '.h', 'w') as f_h:
```
igrr, line 92: Should the file name be `.hpp` for C++ case, making it clear that the file shall not be usable from C code?

igrr:
@X-Ryl669 Thank you for the PR. We can definitely accept the changes related to customization the prefix (`ulp_`). The part related to replacement of the linker script with inline assembly looks more risky for the reason mentioned above.

components/ulp/project_include.cmake
X-Ryl669: Ok, I need to learn how to do that and will do.

components/ulp/esp32ulp_mapgen.py
X-Ryl669: > The asm block will only have effect if it is included into a C or C++ source file and processed by a C or C++ compiler. If a bindings generator parses the header file, instead, then the asm block won't have any effect.

That's true but I don't see how it would be different from the linker's script. If the binding generators parses the C header string as it is now, it doesn't know the address for the symbol either. The only way for the binding generator to know the address for a symbol right now is to parse the linker's script, which is a PITA to get right. It's probably a lot easier to parse the output of `nm` or `readelf`

The issue with the linker's script is that it doesn't choke on error. The linker will silently accept 2 variables with the same name and different addresses and use whatever comes first or last in the linking order, IDK. I've lost a complete day figuring out why the ULP didn't worked the way I expected thinking that the ULP code or HP code was wrong, but the error was in the tool not the code. 

When the address is set in the C header file, if you have 2 similar declarations with different address in the same translation unit (that will happen in your main.cpp file for example), then you'll get a compile time error, IMHO it's much better.

X-Ryl669: Nevermind, I've just tested it and it didn't choke as I thought it would. The asm directive just defines a variable/constant and the code that's using the variable get replaced by the constant. It accepts that the constant is changed, unlike what's written in the documentation. So there's no reason to use the asm directive in the code, I'll reinstantiate the linker's script then. I need to find a way to get a linker error if one symbol is declared twice. I think it can be done by dropping the "PROVIDE" part of the line, so that multiple definition will emit an error. I need to test to make sure it works as intended.

X-Ryl669: Dropped that part and finally used a linker script with some additional protection to prevent linking in case the symbol is already defined.

components/ulp/esp32ulp_mapgen.py
X-Ryl669: Tried to do that but it's a PITA since changing the extension here means also making the CMake code conditional to the prefix type (make all dependency tracking & artifacts as variable, ...). I don't think it matters anyway, the only way to get a C++ header is to specify a prefix with a namespace declaration, and in that case, if the `.h` header is included in a C source file, the compiler will just choke on the "namespace" part so it's pretty obvious.

X-Ryl669:
@igrr Can you review the change to see if they fit your requirements, please?

components/ulp/esp32ulp_mapgen.py
```diff
@@ -2,71 +2,103 @@
...
+        # C++ mode, let's avoid the extern "C" type and instead use namespace
+        f_h.write(textwrap.dedent(
+            f"""
+            /* ULP variable definitions for the compiler.
+             * This file is generated automatically by {UTIL} utility.
+             */
+            #pragma once
+            """  # noqa: E222
+        ))
+        namespace, var_prefix = prefix.split('::')
```
sudeep-mohanty, line 43: This could fail for nested namespaces. Suggest using `prefix.rsplit('::', 1)`.

components/ulp/esp32ulp_mapgen.py
```diff
@@ -2,71 +2,103 @@
...
+            #ifdef __cplusplus
+            }
+            #endif
+            """
+        ))
 
 
 def main() -> None:
     description = ('This application generates .h and .ld files for symbols defined in input file. '
                    'The input symbols file can be generated using nm utility like this: '
```
sudeep-mohanty, line 89: Nitpick:
```suggestion
                   'The input symbols file can be generated using readelf utility like this: '
```

components/ulp/esp32ulp_mapgen.py
```diff
@@ -2,71 +2,103 @@
...
-                   '<PREFIX>nm -g -f posix <elf_file> > <symbols_file>')
+                   '<PREFIX>readelf -sW <elf_file> > <symbols_file>')
 
     parser = argparse.ArgumentParser(description=description)
     parser.add_argument('-s', '--symfile', required=True, help='symbols file name', metavar='SYMFILE', type=argparse.FileType('r'))
     parser.add_argument('-o', '--outputfile', required=True, help='destination .h and .ld files name prefix', metavar='OUTFILE')
     parser.add_argument('--base-addr', required=True, help='base address of the ULP memory, to be added to each symbol')
+    parser.add_argument('-p', '--prefix', required=False, help='prefix for generated header file', default='ulp_')
```
sudeep-mohanty, line 96: Nitpick: 

```suggestion
    parser.add_argument('-p', '--prefix', required=False, help='prefix for variables defined in the generated header file, default='ulp_')
```

docs/en/api-reference/system/ulp-lp-core.rst
```diff
@@ -36,7 +36,21 @@ Using ``ulp_embed_binary``
     ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")
 
 The first argument to ``ulp_embed_binary`` specifies the ULP binary name. The name specified here is also used by other generated artifacts such as the ELF file, map file, header file, and linker export file. The second argument specifies the ULP source files. Finally, the third argument specifies the list of component source files which include the header file to be generated. This list is needed to build the dependencies correctly and ensure that the generated header file is created before any of these files are compiled. See the section below for the concept of generated header files for ULP applications.
+Variables in the ULP code will be prefixed with ``ulp_`` (default value) in this generated header file.
 
+If you need to embed multiple ULP program, in order to avoid conflicting variable names, it's possible to modify the prefix like this:
```
sudeep-mohanty, line 41: Nitpick:

```suggestion
If you need to embed multiple ULP programs, you may add a custom prefix in order to avoid conflicting variable names like this:
```

docs/en/api-reference/system/ulp-lp-core.rst
```diff
@@ -140,13 +154,7 @@ The header file contains the declaration of the symbol:
...
     extern uint32_t ulp_measurement_count;
 
-Note that all symbols (variables, arrays, functions) are declared as ``uint32_t``. For functions and arrays, take the address of the symbol and cast it to the appropriate type.
-
-The generated linker script file defines the locations of symbols in LP_MEM:
-
-.. code-block:: none
-
-    PROVIDE ( ulp_measurement_count = 0x50000060 );
+Note that all symbols (variables, functions) are declared as ``uint32_t``. Array are declared as ``uint32_t [SIZE]``. For functions, take the address of the symbol and cast it to the appropriate type.
```
sudeep-mohanty, line 157: Nitpick:

```suggestion
Note that all symbols (variables, functions) are declared as ``uint32_t``. Arrays are declared as ``uint32_t [SIZE]``. For functions, take the address of the symbol and cast it to the appropriate type.
```

docs/en/api-reference/system/ulp-lp-core.rst
```diff
@@ -161,7 +169,7 @@ To access the ULP LP core program variables from the main program, the generated
 .. note::
 
     Variables declared in the global scope of the LP core program reside in either the ``.bss`` or ``.data`` section of the binary. These sections are initialized when the LP core binary is loaded and executed. Accessing these variables from the main program on the HP-Core before the first LP core run may result in undefined behavior.
-
+    The ``ulp_`` prefix is the default value. You can specify the prefix to use with ``ulp_embed_binary`` to avoid name collusions for multiple ULP program.
```
sudeep-mohanty, line 172: Nitpick:

```suggestion
    The ``ulp_`` prefix is the default value. You can specify the prefix to use with ``ulp_embed_binary`` to avoid name collisions for multiple ULP programs.
```

docs/en/api-reference/system/ulp-risc-v.rst
```diff
@@ -39,7 +39,21 @@ Using ``ulp_embed_binary``
     ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}")
 
 The first argument to ``ulp_embed_binary`` specifies the ULP binary name. The name specified here is also used by other generated artifacts such as the ELF file, map file, header file, and linker export file. The second argument specifies the ULP source files. Finally, the third argument specifies the list of component source files which include the header file to be generated. This list is needed to build the dependencies correctly and ensure that the generated header file is created before any of these files are compiled. See the section below for the concept of generated header files for ULP applications.
+Variables in the ULP code will be prefixed with ``ulp_`` (default value) in this generated header file.
 
+If you need to embed multiple ULP program, in order to avoid conflicting variable names, it's possible to modify the prefix like this:
```
sudeep-mohanty, line 44: Nitpick:

```suggestion
If you need to embed multiple ULP programs, you may add a custom prefix in order to avoid conflicting variable names like this:
```

docs/en/api-reference/system/ulp-risc-v.rst
```diff
@@ -143,11 +157,7 @@ The header file contains the declaration of the symbol:
 
     extern uint32_t ulp_measurement_count;
 
-Note that all symbols (variables, arrays, functions) are declared as ``uint32_t``. For functions and arrays, take the address of the symbol and cast it to the appropriate type.
-
-The generated linker script file defines the locations of symbols in RTC_SLOW_MEM::
-
-    PROVIDE ( ulp_measurement_count = 0x50000060 );
+Note that all symbols (variables, functions) are declared as ``uint32_t``. Array are declared as ``uint32_t [SIZE]``. For functions, take the address of the symbol and cast it to the appropriate type.
```
sudeep-mohanty, line 160: Nitpick:

```suggestion
Note that all symbols (variables, functions) are declared as ``uint32_t``. Arrays are declared as ``uint32_t [SIZE]``. For functions, take the address of the symbol and cast it to the appropriate type.
```

docs/en/api-reference/system/ulp-risc-v.rst
```diff
@@ -162,7 +172,7 @@ To access the ULP RISC-V program variables from the main program, the generated
 .. note::
 
     Variables declared in the global scope of the ULP RISC-V program reside in either the ``.bss`` or ``.data`` section of the binary. These sections are initialized when the ULP RISC-V binary is loaded and executed. Accessing these variables from the main program on the main CPU before the first ULP RISC-V run may result in undefined behavior.
-
+    The ``ulp_`` prefix is the default value. You can specify the prefix to use with ``ulp_embed_binary`` to avoid name collusions for multiple ULP program.
```
sudeep-mohanty, line 175: Nitpick:

```suggestion
    The ``ulp_`` prefix is the default value. You can specify the prefix to use with ``ulp_embed_binary`` to avoid name collisions for multiple ULP programs.
```

docs/en/api-reference/system/ulp.rst
```diff
@@ -50,6 +50,21 @@ To compile the ULP FSM code as part of the component, the following steps must b
     ulp_embed_binary(${ulp_app_name} "${ulp_s_sources}" "${ulp_exp_dep_srcs}")
 
 The first argument to ``ulp_embed_binary`` specifies the ULP FSM binary name. The name specified here will also be used by other generated artifacts such as the ELF file, map file, header file and linker export file. The second argument specifies the ULP FSM assembly source files. Finally, the third argument specifies the list of component source files which include the header file to be generated. This list is needed to build the dependencies correctly and ensure that the generated header file will be created before any of these files are compiled. See the section below for the concept of generated header files for ULP applications.
+Variables in the ULP code will be prefixed with ``ulp_`` (default value) in this generated header file.
+
+If you need to embed multiple ULP program, in order to avoid conflicting variable names, it's possible to modify the prefix like this:
```
sudeep-mohanty, line 55: Nitpick:

```suggestion
If you need to embed multiple ULP programs, you may add a custom prefix in order to avoid conflicting variable names like this:
```

docs/en/api-reference/system/ulp.rst
```diff
@@ -50,6 +50,21 @@ To compile the ULP FSM code as part of the component, the following steps must b
...
+
+    idf_component_register()
+
+    set(ulp_app_name ulp_${COMPONENT_NAME})
+    set(ulp_sources "ulp/ulp_c_source_file.c" "ulp/ulp_assembly_source_file.S")
+    set(ulp_exp_dep_srcs "ulp_c_source_file.c")
+
+    ulp_embed_binary(${ulp_app_name} "${ulp_sources}" "${ulp_exp_dep_srcs}" PREFIX "ULP::")
+
+The additional argument can be a C style prefix (like ``ulp2_``) or a C++ style prefix (like ``ULP::``).
```
sudeep-mohanty, line 67: This is not applicable for the ULP-FSM type of co-processor which does not support C/C++.

components/ulp/esp32ulp_mapgen.py
X-Ryl669: I think nested namespace is not supported and I don't see high interest in trying to support them. It has implications for the name mangling algorithm used in the file to generate the linker's symbol name and also means a recursive algorithm to write `namespace X { namespace Y { ...`, because the "namespace X::Y" writing requires C++17. I'll add a test to return an error in case there are more than one "::", since it's not intended to be handled.

X-Ryl669: Ok, make it more like this: if prefix is `A::B::C_`, the code will generate `A::B_C_` as prefix. This saves from implementing complex name mangling rules with nested namespace and avoid breaking the tool like it would have done before. I think it's ok to do our best here, I don't think anyone will try to use more than 1 level namespace anyway.

sudeep-mohanty (approved):

sudeep-mohanty:
sha=a71e63e266e25309d96d89e9555f3b6739fd4de6

X-Ryl669:
Should I resolve the new conflict with master branch, and if yes, via merging or rebasing?

sudeep-mohanty:
> Should I resolve the new conflict with master branch, and if yes, via merging or rebasing?

I can take care of it :)

sudeep-mohanty:
Hi @X-Ryl669,
To keep you updated on this work, I had to make a few changes before I could merge the changes -
- Removed the link time symbol name collision detection since it flagged all globals being included from common files across ULP binaries. Primarily, globals defined in the ULP component library.
- Added the extraction of `NOTYPE` symbols to not break global variables defined in the assembly code for ULP-FSM type of the coprocessor (available in the esp32, esp32s2, esp32s3)
- Some cmake changes and general clean up to avoid any breaking changes and make the code a bit more readable.
- Added a unit test for the feature.

Thank you for the contribution.

X-Ryl669:
Ok, thank you very much.

About: "Removed the link time symbol name collision detection since it flagged all globals being included from common files across ULP binaries. Primarily, globals defined in the ULP component library."

I think that was the point of this PR. I'm probably misunderstanding here what you've done. 

I have a global log library I'm using in my ULP program. It defines an array of logs (`u32 logBuffer[64]`). Without symbol collision detection, when it's included and built in program 1, it'll be mapped to, let's say 0x50000100. In the second program 2, it'll be mapped to 0x50000200. 

If included in the main HP core, only one address will be used (likely the first linked, since the linker will just ignore the second one). If your code is trying to read from the buffer, it'll read from the wrong address. If the data read is used to read elsewhere, (like, in my case, when I'm storing a number of argument from the buffer), it'll make your main program crash.

So, yes, moving the symbols in their own namespace is one way to prevent this, but since it isn't forced to do so, it'll still cause issues for developers that don't know about the problem.

Using the linker to flag that it has already found this symbol makes sure that it can't happen. I've tried this on the multiple ULP program I'm using and it never conflicted wrongly.

If it's causing issues, maybe it would be a good idea to warn the user that symbol collision will happen (I think it's possible in CMake to actually check a variable (the last prefix name?) and if set, warn the user about potential collisions, else set it).

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/5dea6adbef36144924a3ed225c737e25fe5ecef9.

PR #14975, opened 2024-12-04
fix(ble/bluedroid): Don't log error on 16/128-bit UUID mixed descriptors (IDFGH-14175)

nebkat:
## Description

If multiple descriptors are defined for a characteristic where there is a mix of 16-bit and 128-bit UUIDs, an error log was printed during the discovery process.

```
E (27577) BT_GATT: format mismatch
```

As per `BLUETOOTH CORE SPECIFICATION Version 5.4 | Vol 3, Part F` pages 1419/1420:

> If sequential attributes have differing UUID sizes, the ATT_FIND_INFORMATION_RSP PDU shall end with the first attribute of the pair even though this may mean that it is not filled with the maximum possible amount of (handle, UUID) pairs. This is because it is not possible to include attributes with differing UUID sizes into a single response packet. In this situation, the client must use another ATT_FIND_INFORMATION_REQ PDU with its starting handle updated to fetch the second attribute of the pair and any further ones in its original request.

In other words, if we start returning 16-bit UUID attributes and we encounter 128-bit (or vice-versa) we should simply return what we have so far, and the client should send a new `ATT_FIND_INFORMATION_REQ`. There is no need for the error log.

Also, interestingly, the return code is ignored anyway:

https://github.com/espressif/esp-idf/blob/030c9957dbc5c5a627cf1bb4b2b856f1b6c1343c/components/bt/host/bluedroid/stack/gatt/gatt_sr.c#L1016-L1020

## Checklist

### DIFF
--- a/components/bt/host/bluedroid/stack/gatt/gatt_sr.c
+++ b/components/bt/host/bluedroid/stack/gatt/gatt_sr.c
@@ -788,7 +788,7 @@ static tGATT_STATUS gatt_build_primary_service_rsp (BT_HDR *p_msg, tGATT_TCB *p_
 **                  buffer.
 **
 ** Returns          TRUE: if data filled successfully.
-**                  FALSE: packet full, or format mismatch.
+**                  FALSE: packet full.
 **
 *******************************************************************************/
 static tGATT_STATUS gatt_build_find_info_rsp(tGATT_SR_REG *p_rcb, BT_HDR *p_msg, UINT16 *p_len,
@@ -831,10 +831,9 @@ static tGATT_STATUS gatt_build_find_info_rsp(tGATT_SR_REG *p_rcb, BT_HDR *p_msg,
                     gatt_convert_uuid32_to_uuid128(p, ((tGATT_ATTR32 *) p_attr)->uuid);
                     p += LEN_UUID_128;
                 } else {
-                    GATT_TRACE_ERROR("format mismatch");
-                    status = GATT_NO_RESOURCES;
+                    // UUID format mismatch in sequential attributes
+                    // A new request will be sent with the starting handle of the next attribute
                     break;
-                    /* format mismatch */
                 }
                 p_msg->len += info_pair_len[p_msg->offset - 1];
                 len -= info_pair_len[p_msg->offset - 1];
--- a/components/bt/host/bluedroid/stack/gatt/gatt_sr.c
+++ b/components/bt/host/bluedroid/stack/gatt/gatt_sr.c
@@ -200,7 +200,7 @@ static BOOLEAN process_read_multi_rsp (tGATT_SR_CMD *p_cmd, tGATT_STATUS status,
             *p++ = GATT_RSP_READ_MULTI;
             p_buf->len = 1;
 
-            /* Now walk through the buffers puting the data into the response in order */
+            /* Now walk through the buffers putting the data into the response in order */
             list_t *list = NULL;
             const list_node_t *node = NULL;
             if (! fixed_queue_is_empty(p_cmd->multi_rsp_q)) {
@@ -321,7 +321,7 @@ static BOOLEAN process_read_multi_var_rsp (tGATT_SR_CMD *p_cmd, tGATT_STATUS sta
             *p++ = GATT_RSP_READ_MULTI_VAR;
             p_buf->len = 1;
 
-            /* Now walk through the buffers puting the data into the response in order */
+            /* Now walk through the buffers putting the data into the response in order */
             list_t *list = NULL;
             const list_node_t *node = NULL;
             if (! fixed_queue_is_empty(p_cmd->multi_rsp_q)) {
@@ -735,7 +735,7 @@ static tGATT_STATUS gatt_build_primary_service_rsp (BT_HDR *p_msg, tGATT_TCB *p_
                     handle_len = 4 + p_uuid->len;
                 }
 
-                /* get the length byte in the repsonse */
+                /* get the length byte in the response */
                 if (p_msg->offset == 0) {
                     *p ++ = op_code + 1;
                     p_msg->len ++;
@@ -889,7 +889,7 @@ static tGATT_STATUS gatts_validate_packet_format(UINT8 op_code, UINT16 *p_len,
                 /* parse uuid now */
                 if (gatt_parse_uuid_from_cmd (p_uuid_filter, uuid_len, &p) == FALSE ||
                         p_uuid_filter->len == 0) {
-                    GATT_TRACE_DEBUG("UUID filter does not exsit");
+                    GATT_TRACE_DEBUG("UUID filter does not exist");
                     reason = GATT_INVALID_PDU;
                 } else {
                     len -= p_uuid_filter->len;
@@ -1042,7 +1042,7 @@ static void gatts_process_find_info(tGATT_TCB *p_tcb, UINT8 op_code, UINT16 len,
 **
 ** Function         gatts_process_mtu_req
 **
-** Description      This function is called to process excahnge MTU request.
+** Description      This function is called to process exchange MTU request.
 **                  Only used on LE.
 **
 ** Returns          void
@@ -1055,7 +1055,7 @@ static void gatts_process_mtu_req (tGATT_TCB *p_tcb, UINT16 len, UINT8 *p_data)
     BT_HDR        *p_buf;
     UINT16   conn_id;
 
-    /* BR/EDR conenction, send error response */
+    /* BR/EDR connection, send error response */
     if (p_tcb->att_lcid != L2CAP_ATT_CID) {
         gatt_send_error_rsp (p_tcb, GATT_REQ_NOT_SUPPORTED, GATT_REQ_MTU, 0, FALSE);
     } else if (len < GATT_MTU_REQ_MIN_LEN) {
@@ -1081,7 +1081,7 @@ static void gatts_process_mtu_req (tGATT_TCB *p_tcb, UINT16 len, UINT8 *p_data)
             attp_send_sr_msg (p_tcb, p_buf);
 
             /* Notify all registered application with new MTU size. Us a transaction ID */
-            /* of 0, as no response is allowed from applcations                    */
+            /* of 0, as no response is allowed from applications                        */
 
             for (i = 0; i < GATT_MAX_APPS; i ++) {
                 if (gatt_cb.cl_rcb[i].in_use ) {
@@ -1448,7 +1448,7 @@ void gatt_attr_process_prepare_write (tGATT_TCB *p_tcb, UINT8 i_rcb, UINT16 hand
     }
 
     if ((prepare_record->error_code_app == GATT_SUCCESS)
-        // update prepare write status for excute write request
+        // update prepare write status for execute write request
         && (status == GATT_INVALID_OFFSET || status == GATT_INVALID_ATTR_LEN || status == GATT_REQ_NOT_SUPPORTED)) {
         prepare_record->error_code_app = status;
     }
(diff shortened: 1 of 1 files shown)


esp-zhp:
Your commit has been merged into the master branch, but it will take some time to sync to GitHub.  

Additionally, your commit will also be synced to the release branch, which will also take some time.  

Please note that the final changes may not be exactly the same as your original modifications, as I have made some optimizations to conform to coding standards. However, your commit information has been retained.  

Here are the details:  
- **master:** fed1d41aa7170d8418dd126dd8e3b47a977b4aca
- **v5.4:**  7790c95832a3df80b518a38fbb83cb76774b6924
- **v5.3:**  ff6425d301e87f6d7b8c8486e24227c6c58be9e7
- **v5.2:**  72527f85ea6d3016b55e0da9bb59d199a4d4f742
- **v5.1:**  c8adc1a9ff85253d846088835014330531a873ca
- **v5.0:** 3e48f50e4ec3bc8a3f55cb12f887f53b500f5b98

esp-zhp:
@nebkat 
Hi Nebojša,

We are following up on your ESP-IDF contribution from December 2024:

**Commit:** `fed1d41aa71` — *fix(ble/bluedroid): Don't log error on 16/128-bit UUID mixed descriptors*  
**File:** `components/bt/host/bluedroid/stack/gatt/gatt_sr.c` — `gatt_build_find_info_rsp()`

We agree with your original rationale: when sequential attributes in a Find Information response have different UUID sizes, the server should return a partial response and the client should continue with a new `ATT_FIND_INFORMATION_REQ`. That behaviour is consistent with Bluetooth Core Specification Vol 3, Part F (Find Information Response), and removing the `format mismatch` error log was reasonable, since this is expected protocol behaviour rather than a fault.

However, we have received an external report that suggests the fix may have introduced a regression by also removing `status = GATT_NO_RESOURCES;` in the format-mismatch branch. We would like to confirm with you whether that removal was intentional, or an unintended side effect of suppressing the log.

---

### Report from the others(AAA)

**Environment**
- Device: ESP32-based appliance module (GATT server / peripheral)
- Affected: ESP-IDF v5.x and v6.x (confirmed on master)
- Working: ESP-IDF v4.4
- Client: Nordic nRF52840 central (blatann / pc_ble_driver_py), MTU 247
- Component: `components/bt/host/bluedroid/stack/gatt/gatt_sr.c`

**Summary**

After migrating from IDF v4 to v6 with no changes to the GATT database, Find Information responses omit valid attributes inside the requested handle range, while including attributes from a later service with higher handles. Because the ATT continuation algorithm resumes from `last_handle + 1`, the omitted attributes can never be discovered again. There is no ATT error and no way for the client to detect the gap.

**Observed behaviour**

Identical request in both firmware versions: **Find Information Request (0x0004, 0xFFFF)**

| Firmware | Handles returned |
|----------|------------------|
| IDF v4.4 | 4, 20, 21, 22, 23, 24, 25, 26, 40, 41 |
| IDF v6   | 4, 20, 21, 22, 23, 24, 25, 26, 40, 41, **76, 77** |

The v6 response jumps from handle 41 to 76, omitting handles 42–75. Those attributes exist — handle 46 (a CCCD, UUID 0x2902) can still be read directly and returns `0x0000`.

Handles 5–19 and 27–39 are not present in the database (expected gaps from handle allocation). Handles 42–75 are real attributes whose omission violates the contiguity requirement in Vol 3, Part F, §3.4.3.2.

**Consequence:** all eight CCCDs of their vendor service (handles 46, 50, 54, 58, 62, 66, 70, 74) become invisible to the client, so notifications cannot be enabled. Associated User Description descriptors (43, 47, 51, …) are also lost.

**Relevant GATT layout (simplified)**

| Service UUID | Handle range | Notes |
|--------------|--------------|-------|
| 0x1801       | 1–5          | GATT service |
| 0x1800       | 20–30        | GAP service |
| 7eed67d5-…   | 40–75        | Vendor service (128-bit characteristics, 16-bit descriptors) |
| 10faa7a7-…   | 76–…         | Second vendor service |

**Root cause analysis (from AAA)**

In `gatt_build_find_info_rsp()`, when a UUID format mismatch occurs:

**v4.4 (working):**
```c
} else {
    GATT_TRACE_ERROR("format mismatch");
    status = GATT_NO_RESOURCES;
    break;
}
```

**v5.0 onwards (including master, after your change):**
```c
} else {
    // UUID format mismatch in sequential attributes
    // A new request will be sent with the starting handle of the next attribute
    break;
}
```

Without `status = GATT_NO_RESOURCES`, `status` remains `GATT_SUCCESS` from previously appended attributes. In `gatts_process_find_info()`, the service loop only breaks early on `GATT_NO_RESOURCES`:

```c
tGATT_STATUS build_status = gatt_build_find_info_rsp(p_rcb, p_msg, &buf_len, s_hdl, e_hdl);
if (build_status == GATT_SUCCESS) {
    reason = GATT_SUCCESS;
} else if (build_status == GATT_NO_RESOURCES) {
    reason = GATT_SUCCESS;
    break;
}
```

So on format mismatch, the caller treats the current service as complete, advances to the next in-range service, and appends its attributes after the gap.

**Trace for request (0x0004, 0xFFFF):**

| Service | Range | Emitted | Then |
|---------|-------|---------|------|
| 0x1801 | 1–5 | 4 | list ends → SUCCESS → next service |
| 0x1800 | 20–30 | 20…26 | list ends → SUCCESS → next service |
| 7eed67d5 | 40–75 | 40, 41 | handle 42 is 128-bit → format break |
| 10faa7a7 | 76–… | 76, 77 | handle 78 is 128-bit → break |

- **v4.4:** format break sets `GATT_NO_RESOURCES` → caller `break`s → response `[4, 20…26, 40, 41]`
- **v5.5+:** format break returns `GATT_SUCCESS` → caller continues → response `[4, 20…26, 40, 41, 76, 77]`

This matches their captured logs handle-for-handle.

**Trigger conditions (all three required):**
1. Requested range overlaps two or more registered services
2. A UUID-format mismatch stops the builder inside a non-final in-range service
3. A later in-range service still contributes at least one attribute

Clients that discover descriptors per characteristic (Vol 3, Part G, §4.7.1), as iOS CoreBluetooth and Android typically do, may not hit this path. A range spanning multiple services is nevertheless valid ATT.

**Suggested fix:** restore `status = GATT_NO_RESOURCES;` in the format-mismatch branch (while keeping the error log removed).

---

### Our understanding of your original change

In your PR description, you noted that on format mismatch the server should simply return what it has so far, and that the return code is “ignored anyway” because the caller converts `GATT_NO_RESOURCES` to `GATT_SUCCESS` before transmission.

Our reading is slightly different:

- **Toward the client:** `GATT_NO_RESOURCES` is indeed not exposed as an error; it becomes a normal partial Find Information Response. That aligns with the spec.
- **Inside the stack:** `GATT_NO_RESOURCES` is still meaningful — it triggers `break` in the per-service loop in `gatts_process_find_info()`, preventing attributes from subsequent services from being appended into the same PDU after a format break.

So it appears that `GATT_NO_RESOURCES` was serving two roles: “MTU buffer full” and “format break — end this PDU”. Both require the same control-flow outcome (stop building the response), even though neither is an ATT error toward the client.

Your added comment — *“A new request will be sent with the starting handle of the next attribute”* — is correct **only if** the format break terminates the entire response PDU. After removing `status = GATT_NO_RESOURCES`, that is no longer guaranteed when multiple services fall within the requested range; the client resumes from the last handle in the PDU (e.g. 77), not from where the break occurred (e.g. 42).

We also note that your scenario (mixed 16-bit / 128-bit descriptor UUIDs within a single characteristic) likely involves only one service in the request range. In that case, the bug may not manifest even without `GATT_NO_RESOURCES`, which could explain why the regression was not caught during your testing.

From our git history check, the behavioural change appears to have landed in **v5.5** via your commit.

---

### Questions for you

1. Was removing `status = GATT_NO_RESOURCES;` intentional, or was it a side effect of removing the error trace?
2. Did you test Find Information requests whose handle range spans **multiple services** (not only per-characteristic descriptor discovery)?
3. Would you agree that the correct fix is to **keep the log removed** but **restore `status = GATT_NO_RESOURCES;`** (or an equivalent internal signal that causes the service loop to `break`)?
4. Are you aware of any reason we should **not** restore that assignment?

If you are able to reproduce AAA’s scenario or confirm our analysis, that would be very helpful. We are happy to share any additional logs or test results if useful.

Thank you for your contributions to ESP-IDF, and for taking the time to look into this.

nebkat:
Hi, thanks for the thorough analysis — you're right, and the fix you propose is correct.

Removing `status = GATT_NO_RESOURCES;` was a mistake on my part. I only considered the client-facing status (which the caller maps back to `GATT_SUCCESS`) and missed that it's also the signal that breaks the per-service loop in `gatts_process_find_info()`. Without it the builder falls through to the next in-range service and appends its attributes after the gap, which is exactly the behaviour you captured. I only tested per-characteristic descriptor discovery within a single service, so the multi-service case never came up.

To your questions: (1) unintentional side effect, (2) no, (3) yes, and (4) no reason not to.

So: restore `status = GATT_NO_RESOURCES;`, but please keep the error log removed — a UUID format break is expected protocol behaviour, not a fault, and logging it as an error was the actual bug in the original code. A short comment at the assignment explaining that it means "end this PDU" rather than a real error would help avoid a repeat of my mistake.

shiftreg:
> Hi, thanks for the thorough analysis — you're right, and the fix you propose is correct.
> 
> Removing `status = GATT_NO_RESOURCES;` was a mistake on my part. I only considered the client-facing status (which the caller maps back to `GATT_SUCCESS`) and missed that it's also the signal that breaks the per-service loop in `gatts_process_find_info()`. Without it the builder falls through to the next in-range service and appends its attributes after the gap, which is exactly the behaviour you captured. I only tested per-characteristic descriptor discovery within a single service, so the multi-service case never came up.
> 
> To your questions: (1) unintentional side effect, (2) no, (3) yes, and (4) no reason not to.
> 
> So: restore `status = GATT_NO_RESOURCES;`, but please keep the error log removed — a UUID format break is expected protocol behaviour, not a fault, and logging it as an error was the actual bug in the original code. A short comment at the assignment explaining that it means "end this PDU" rather than a real error would help avoid a repeat of my mistake.

> Hi, thanks for the thorough analysis — you're right, and the fix you propose is correct.
> 
> Removing `status = GATT_NO_RESOURCES;` was a mistake on my part. I only considered the client-facing status (which the caller maps back to `GATT_SUCCESS`) and missed that it's also the signal that breaks the per-service loop in `gatts_process_find_info()`. Without it the builder falls through to the next in-range service and appends its attributes after the gap, which is exactly the behaviour you captured. I only tested per-characteristic descriptor discovery within a single service, so the multi-service case never came up.
> 
> To your questions: (1) unintentional side effect, (2) no, (3) yes, and (4) no reason not to.
> 
> So: restore `status = GATT_NO_RESOURCES;`, but please keep the error log removed — a UUID format break is expected protocol behaviour, not a fault, and logging it as an error was the actual bug in the original code. A short comment at the assignment explaining that it means "end this PDU" rather than a real error would help avoid a repeat of my mistake.

Hi everyone,
I’m the original external reporter of this issue. I agree with @nebkat that adding a short comment like “end this PDU” would help clarify the intended use of the status and avoid similar regressions in the future.

Thanks for looking into this and for preparing the fix/backports.

esp-zhp:
Hi everyone,

I’ve fixed this issue in the master branch. The fix is in commit `196cb395452108b4084e8fee868efaee7eac4b0e`, which will be synced to GitHub shortly.

I’ll also comment here as soon as the backports to the other release branches are completed.

Thanks again for your contribution and for the thorough analysis and investigation!

PR #14997, opened 2024-12-07
Update spi_lcd_touch_example_main.c (IDFGH-14199)

ostifede02:
## Description
Set uint32_t to int32_t. If lv_timer_handler() is called too fast, returns -1. If time_till_next_ms is unsigned, it becomes 0xFFFFFF... (very large). Therefore it overrides the threshold and code locks in usleep().


suda-morris:
sha=663b432caeec3d12c00667abfb21bc64cba8762b

Kainarx:
@ostifede02  Hi, I'm sure the function returns the uin32 type. https://github.com/lvgl/lvgl/blob/c9d35e1038060d07555903707da91e3ce49672f7/src/misc/lv_timer.h#L53  
Can you describe in detail the problem you're having?

PR #15110, opened 2024-12-30
Added requires libs to the code example so that it now works

b0tFilip:
Added libs so that the code example works.
It didn't work because of missing declarations:
    struct addrinfo 
    getaddrinfo()

It's just a small detail but it cost me an hour of debuging to find these libs that fix it.


david-cermak (commented):
Thanks for your help and contribution!
could you please address the formal issues with
* you commit message as mentioned [here](https://github.com/espressif/esp-idf/pull/15110#issuecomment-2565524250)?
* make the same changes also the CN docs in `docs/zh_CN/api-reference/protocols/icmp_echo.rst`?

david-cermak:
Closing for inactivity, feel free to reopen and address https://github.com/espressif/esp-idf/pull/15110#pullrequestreview-2531813529

PR #15129, opened 2025-01-03
docs: Add chip target to direct CMake example code snippet  in build system docs

Dazza0:
## Description

The Build System documentation provides and example code snippet on how to use CMake and Ninja directly, bypassing the `idf.py` frontend.

However,  the snippet does not demonstrate how to set the chip target using CMake (equivalent to `idf.py set-target`). This PR updates the example code snippets to set the chip target.

## Related

- Build System [documentation](https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32/api-guides/build-system.html#using-cmake-directly)

## Checklist

### DIFF
--- a/docs/en/api-guides/build-system.rst
+++ b/docs/en/api-guides/build-system.rst
@@ -70,7 +70,7 @@ When ``idf.py`` does something, it prints each command that it runs for easy ref
 
   mkdir -p build
   cd build
-  cmake .. -G Ninja   # or 'Unix Makefiles'
+  cmake .. -DIDF_TARGET={IDF_TARGET_PATH_NAME} -G Ninja   # or 'Unix Makefiles'
   ninja
 
 In the above list, the ``cmake`` command configures the project and generates build files for use with the final build tool. In this case, the final build tool is Ninja_: running ``ninja`` actually builds the project.
--- a/docs/zh_CN/api-guides/build-system.rst
+++ b/docs/zh_CN/api-guides/build-system.rst
@@ -70,7 +70,7 @@ idf.py
 
     mkdir -p build
     cd build
-    cmake .. -G Ninja   # 或者 'Unix Makefiles'
+    cmake .. -DIDF_TARGET={IDF_TARGET_PATH_NAME} -G Ninja   # 或者 'Unix Makefiles'
     ninja
 
 在上面的命令列表中，``cmake`` 命令对项目进行配置，并生成用于最终构建工具的构建文件。在这个例子中，最终构建工具是 Ninja_: 运行 ``ninja`` 来构建项目。


dobairoland:
Thanks @Dazza0 for the contribution. You are right that this could be explained. But the comment for that code snippet states that it is equivalent to running `idf.py build`. So I think this should be explained separately, if necessary. 

On the other hand, the sentence just before it states that "When idf.py does something, it prints each command that it runs for easy reference.". I see it as a guide how to find the right cmake commands to run if someone wants to bypass `idf.py`. If we go into details then we will have to expand this short section with a complete guide for cmake underneath `idf.py`.

It is definitely worth considering but probably only after the related build system improvements we plan to do soon.

Dazza0:
@dobairoland 

> So I think this should be explained separately, if necessary.

Would this information be better placed in the [Using ESP-IDF in Custom CMake Projects](https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32/api-guides/build-system.html#using-esp-idf-in-custom-cmake-projects).  I think it's likely the main use case for users needing to call CMake directly is if they are trying to integrate IDF into an existing CMake.

For context, this is what I'm trying to achieve. I have an existing CMake project that needs to build on multiple MCU families (e.g., STM32, PIC32, ESP32).

> I see it as a guide how to find the right cmake commands to run if someone wants to bypass idf.py. If we go into details then we will have to expand this short section with a complete guide for cmake underneath idf.py.

Would it be possible to provide a short list of what targets are exposed to the underlying build system. Comparing the target list between `ninja help` and `idf.py --help`, it seems like `ninja` has a subset of `idf.py` targets (e.g., has `ninja menuconfig|flash|monitor` but not `ninja gdb|openocd`. Thus it would be helpful to know what targets are officially exposed.

dobairoland:
> Would this information be better placed in the [Using ESP-IDF in Custom CMake Projects](https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32/api-guides/build-system.html#using-esp-idf-in-custom-cmake-projects). I think it's likely the main use case for users needing to call CMake directly is if they are trying to integrate IDF into an existing CMake.

I think it could have a separate section much earlier, e.g. between https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32/api-guides/build-system.html#using-cmake-directly and https://docs.espressif.com/projects/esp-idf/en/v5.3.2/esp32/api-guides/build-system.html#flashing-with-ninja-or-make.

> Would it be possible to provide a short list of what targets are exposed to the underlying build system. Comparing the target list between ninja help and idf.py --help, it seems like ninja has a subset of idf.py targets (e.g., has ninja menuconfig|flash|monitor but not ninja gdb|openocd. Thus it would be helpful to know what targets are officially exposed.

Ah, you are right. We used to have all idf.py targets available in cmake as well but we deviated from this when we added the debugging targets. This certainly needs to be addressed or, as you suggest, describe the limitation in the documentation. There is one issue with having a list there. I don't think anyone would remember to update it over time when we add/remove a new target.

Thanks for the suggestions!

PR #15132, opened 2025-01-03
fix(newlib): usleep returning early (IDFGH-14342)

stevenoonan:
## Description

The current implementation of `usleep()` can return in the less the specified time, which breaks assumptions callers of `usleep()` have. This proposed change ensures `usleep()` never returns in less than the specified time.

From `man 3 usleep`:
> The  usleep()  function suspends execution of the calling thread for (at least) usec microseconds.

Returning before the specified time can cause great inefficiencies when using C++'s `this::thread::sleep_for()`. 

I stumbled upon this when I upgraded a project from IDF 4.4 to IDF 5.3.1 (GCC got updated to 13.2). After updating, my project was running sluggish. I profiled the code and found a thread was using way more processing time than before. It was a simple loop with a `this:thread::sleep_for(10ms)` call.

Specifically, the changes of this commit https://github.com/gcc-mirror/gcc/commit/cfef4c324ac300c0ad120f0fcee376159de84a0c in libstdc++'s `__sleep_for()` is what causes behavior to change. But the problem is in `usleep()`. In the older version of libstdc++ [`__sleep_for()`](https://github.com/gcc-mirror/gcc/blob/releases/gcc-8.4.0/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L186), `usleep()` is always called once. The newer version added a [while loop](https://github.com/gcc-mirror/gcc/blob/1c0ecf06c0c9f3268ba52d4027e9b78aae4acb51/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L235) and a check to the monotonic clock to verify the sleep did not come back early, and if it does, it keeps calling `usleep()`. 

The only documented reason a sleep should return early is due to a signal, but of course that is not the case on the ESP32. 

The reason `usleep()` returns early is because it does not compensate for a worst case systick ISR timing. Let's play out an example. If we wanted to sleep for 15ms where the tick period is 10ms, the calculations would give us `vTaskDelay(2)`:
1) `portTICK_PERIOD_MS` is defined as `10`. So `us_per_tick` is `10,000`.
2) `us` is `15000`
3) `(15000 + 10000 - 1) / 10000)` is `2.4999`, so `2` ticks

Two ticks has a minimum of 10ms, because the first tick ISR can happen immediately, so you can only count on n-1 tick periods of time. So of course 10ms is less than 15ms. The current `(us + us_per_tick - 1) / us_per_tick` compensation attempt is broken. If you do the math for a 10ms wait you'll get `1` tick, which has a minimum of 0 tick periods. But even if it does get half (or 90%) of a period of time slept, we still haven't slept the minimum time we should have.

This means `usleep()` will return back to libstdc++ `__sleep_for()`, the monotonic clock check fails, and `usleep()` is now called again, with the remainder of time. This will always been less than a tick period, so `esp_rom_delay_us()` is now called. I don't have the source code for `esp_rom_delay_us()` but I assume it is a blocking busy wait.

Any time spent in `esp_rom_delay_us()` is quite inefficient for scheduling. If the thread blocking here is the highest priority thread on the system, it will prevent FreeRTOS from time slicing to any other ready state thread during that time. That was exactly what was happening in my project.

My fix is to double check the monotonic clock in `usleep()`, and it will `vTaskDelay()` an additional time when needed. This guarantees `usleep()` will never return early, but sometimes later. Which is to spec, and also, it means libstdc++ won't call `usleep()` in a way that will cause `esp_rom_delay_us()` to be called unexpectedly.

For 15ms sleep with a 10ms systick period, the absolute minimal tick count is `2`. My change does a simple division of equivalent of `(15000/10000)+1` which gives `2.5` or int `2`. In the worst case, another tick is needed because as stated above, a tick count of 2 means a worst case of only 10ms waited, so the loop will check for that and call `vTaskDelay()` again.

## Related
* GCC 8.4 [`__sleep_for()`](https://github.com/gcc-mirror/gcc/blob/releases/gcc-8.4.0/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L186)
* GCC 13.2 [`__sleep_for()`](https://github.com/gcc-mirror/gcc/blob/1c0ecf06c0c9f3268ba52d4027e9b78aae4acb51/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L235)

## Testing

On my own project, as stated above.

### DIFF
--- a/components/newlib/src/time.c
+++ b/components/newlib/src/time.c
@@ -207,10 +207,19 @@ int usleep(useconds_t us)
     if (us < us_per_tick) {
         esp_rom_delay_us((uint32_t) us);
     } else {
-        /* since vTaskDelay(1) blocks for anywhere between 0 and portTICK_PERIOD_MS,
-         * round up to compensate.
-         */
-        vTaskDelay((us + us_per_tick - 1) / us_per_tick);
+        /* vTaskDelay may return up to (n-1) tick periods due to the tick ISR
+           being asynchronous to the call. We must sleep at least the specified
+           time, or longer. Checking the monotonic clock allows making an
+           additional call to vTaskDelay when needed to ensure minimal time is
+           actually slept. Adding `us_per_tick - 1` prevents ever passing 0 to
+           vTaskDelay(). 
+        */
+        uint64_t now_us = esp_time_impl_get_time();
+        uint64_t target_us = now_us + us;
+        do {
+            vTaskDelay((((target_us - now_us) + us_per_tick - 1) / us_per_tick));
+            now_us = esp_time_impl_get_time();
+        } while (now_us < target_us);
     }
     return 0;
 }


safocl:
Aren't these changes the same as the described changes to C++'s `this::thread::sleep_for()` ?

stevenoonan:
> Aren't these changes the same as the described changes to C++'s `this::thread::sleep_for()` ?

The change I'm proposing with this PR is the same method `this::thread::sleep_for()` (`__sleep_for()`) is already using. But this is the best option we have, I believe.

Let me play out some examples to clarify what is going on.

With a system tick period of 10ms, calling `this::thread::sleep_for(10ms)` is causes to a single call to `usleep(10000)`, which calls `vTaskDelay(1)` in IDF 4. The app thread will yield up until the next system tick. Therefore the thread will sleep will be between 0 and 10ms.

In IDF 5, calling `this::thread::sleep_for(10ms)` (almost always) causes two calls to `usleep()`. First will be `usleep(10000)`, which calls `vTaskDelay(1)`. `vTaskDelay(1)` will sleep between 0 and 10ms. The newer `__sleep_for()` function will almost always determine that  less than 10ms was actually slept, so it will call `usleep()` with the remainder of time. This remainder is always less than the system tick period. This means `esp_rom_delay_us()` will be called, for up to 99,000+ microseconds. That will prevent any thread of the same priority to be switched to during that time, which I think is a huge problem for anyone using C++ with IDF, and definitely was the problem in my project when I upgraded to IDF5.

So what options do we have? If we make `usleep()` guarantee it will always sleep at least the time specified, `__sleep_for()` will not call `usleep()` a second time, and `esp_rom_delay_us()` will not be called. I suppose we could make a custom build of libstdc++ to modify `__sleep_for()` but I doubt that would be viewed as worth it.

So if we must modify `usleep()`, what is the best way to do it? These are the options I came up with:
1.  Don't call `esp_rom_delay_us()` in `usleep()` for lengths less than a system tick period.
2.  _Always_ sleep the extra tick necessary to ensure the minimum time is slept. 
3.  Check an independent clock to see if an extra is needed, and only sleep the extra tick if needed.
4.  Use a hardware timer ISR to sleep the precise amount of time.
5.  Implement a portable API to check when the next system tick ISR will happen.

Option (1) is a non-starter for me. We need a supported way of sleeping accurately for small time periods. Plus, it would break a lot of code currently in the field. Option (2) works, but will often sleep longer than we needed to. With a 10ms system tick, a 15ms sleep sometimes only needs 2 ticks, sometimes three. So why not use option (3) and just check to see if we really need the extra tick?

Option (4) is a non-starter because we don't have an arbitrary number hardware timers, but we do have an arbitrary number of threads that might sleep.

Option (5) is interesting, and may be possible. However it requires an API to be developed that will read the hardware timer registers being employed for the system tick, which I am sure is different across all the different architectures supported. If we know exactly when the next system tick is about to occur, we can correctly calculate how many ticks we need to sleep without subsequently double checking the monotonic clock. This would be awesome, but a lot more work. It might not even be possible for some reason I am unaware of.

The fact that is the same approach used in `__sleep_for()` is what it is. I don't see a better option. And even if we were using the old version of IDF when the monotonic clock is not checked in `__sleep_for()`, we still have the reality that calling `std::this_thread::sleep_for(10ms)` in IDF does _not_ follow [spec](https://en.cppreference.com/w/cpp/thread/sleep_for): 

> Blocks the execution of the current thread for _at least_ the specified sleep_duration.

As I already mentioned, calling `std::this_thread::sleep_for(10ms)` in IDF4 will almost always sleep less than 10ms. It was always broken, because `usleep()` was and still is not following spec. In IDF5  `std::this_thread::sleep_for(10ms)` does sleep _exactly_ 10ms, at the cost of a lot of wasted cycles.

safocl:
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/pthread.html?highlight=usleep#rtos-integration
> When calling a standard libc or C++ sleep function, such as usleep defined in unistd.h, the task will only block and yield the core if the sleep time is longer than [one FreeRTOS tick period](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig-reference.html#config-freertos-hz). If the time is shorter, the thread will busy-wait instead of yielding to another RTOS task.

behavior varies for different sleep durations

stevenoonan:
> https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/pthread.html?highlight=usleep#rtos-integration
> 
> > When calling a standard libc or C++ sleep function, such as usleep defined in unistd.h, the task will only block and yield the core if the sleep time is longer than [one FreeRTOS tick period](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig-reference.html#config-freertos-hz). If the time is shorter, the thread will busy-wait instead of yielding to another RTOS task.
> 
> 
> 
> behavior varies for different sleep durations

Thanks, that covers my addendum edit. The original issue remains where calling sleep_for with a time larger than the system tick will cause a portion of the time to block wait in IDF5.

safocl:
@stevenoonan but the guaranty is not broken with `std::this_thread::sleep_for()`...

stevenoonan:
> @stevenoonan but the guaranty is not broken with `std::this_thread::sleep_for()`...

@safocl The main point is that the guaranty of `usleep()` is broken. This results in unnecessary **blocking** whenever  `std::this_thread::sleep_for()` is called with a time _more_ than a tick period in IDF5. With a 10ms tick period, calling `std::this_thread::sleep_for(15ms)` can result in 5 milliseconds of blocking busy-wait (with the other 10ms yielded). There is no good reason for this blocking to occur - it appears to be an unintended consequence of the interaction between the newer `__sleep_for()` and `usleep()` returning before the specified time. It can cause a huge amount of wasted cycles, and _that_ is the problem. This PR ensures _all_ 15ms will be yielded. It also brings `usleep()` into conformance with the spec.

stevenoonan:
[c060a85](https://github.com/espressif/esp-idf/commit/c060a852e842e4b5af2374b45cafc267d2866a36) Improves the implementation and comments.

stevenoonan:
For more info, I wrote a blog about this issue: https://interrupt.memfault.com/blog/why-sleep-for-is-broken-on-esp32

What can be done to get some movement or feedback on this? Anyone using sleep_for() on ESP32 is wasting cycles until this is fixed.

feature-engineer:
@stevenoonan Maybe opening an issue and relating it to this PR could induce movement or feedback? I'm totally with you on this (but I'm not a developer of esp-idf, just using the library). Energy efficiency is much more important than exact timing when using threading sleep.

sudeep-mohanty:
Hi @stevenoonan,
Apologies for getting to this PR late. This has been in our plans but we haven't had the time to get it through (no pun intended :P). The changes LGTM! I shall pull it in to our internal repository for further tests before merging it. Thank you for the contribution.

sudeep-mohanty (approved):

sudeep-mohanty:
sha=c060a852e842e4b5af2374b45cafc267d2866a36

stevenoonan:
> Hi @stevenoonan, Apologies for getting to this PR late. This has been in our plans but we haven't had the time to get it through (no pun intended :P). The changes LGTM! I shall pull it in to our internal repository for further tests before merging it. Thank you for the contribution.

No worries on the delay, I assumed you guys are busy :). Glad it looks good and hopefully it passes testing!

stevenoonan:
@sudeep-mohanty I'm looking at https://github.com/espressif/esp-idf/commit/d5d7bb61e59949871feb732d2e8117dd4b321d6f and I noticed the fix is different than what I proposed:

```
        uint64_t now_us = esp_time_impl_get_time();
         uint64_t target_us = now_us + us;
         do {
             vTaskDelay((((target_us - now_us) + us_per_tick - 1) / us_per_tick));
             now_us = esp_time_impl_get_time();
             /* If the time left until the target is less than 1 tick, then we use ROM delay to fill the gap */
             uint64_t time_left = target_us - now_us;
             if (time_left != 0 && time_left < us_per_tick) {
                 esp_rom_delay_us(time_left); // block up to 1 tick period?
                 break;
             }
         } while (now_us < target_us);
```

Calling `esp_rom_delay_us(time_left);` will busy wait for up to 10 ms on a standard 10m tick rate. Correct?

This fixes the "return too early" problem with arguably a bigger problem. If you busy wait on a high priority thread, the FreeRTOS scheduler will not time slice to any other threads, effectively blocking an entire core for 10ms. That doesn't seem like a good solution.

Is there something I'm missing? I get that yielding an extra tick would often mean `usleep()` would sleep too long, but I believe that is the better tradeoff than wasting cycles.

The reason I came to this problem and made a PR was because my use of `std::this_thread::sleep_for()` caused [99% use of one of the ESP32 cores](https://interrupt.memfault.com/blog/why-sleep-for-is-broken-on-esp32#why-std-this_thread-sleep_for-is-broken-on-esp32) when I upgraded from IDF 4 to IDF 5. The new code merged would produce the same exact problem when using C++'s `sleep_for()`. Nothing is fixed for my use case. Moreover, using `usleep()` for longer than tick delays also has this problem when it didn't before.

sudeep-mohanty:
@stevenoonan I understand what the concern is here. For multi-tick `usleep()` or `sleep_for()` calls, at some point the sleep could busy wait if the sleep time left is less than a tick period, instead of yielding the CPU time. If that period is 10 ms, the busy-wait time could be anywhere between a few ms and 10 ms. This arguably makes `usleep()` more accurate but does occupy the CPU resources for the period where we busy-wait. So, as I see it, the trade-off here is more accurate sleep timing versus a less resource hungry one. Having said that, I do agree that it would better to be to let `usleep()` potentially sleep for a tick longer than expected but let other tasks to run. Thank you for pointing out this use-case. I shall make further updates. Re-opening the ticket to track the progress.

PR #15206, opened 2025-01-14
Better export.ps1 (IDFGH-14426)

SaintFTS:
## Description
Now, if you want to add this ps1 file to the $PATH without adding other scripts, like install.ps1, you can now just drag it in some empty directory and add it to the $PATH. Just define $IDF_PATH.  

Suggestion to define $IDF_PATH was already in the script, but шt didn't work because $IDF_PATH was just set to the root of the script ("$PSScriptRoot"). 

I also added more "clean" exit out of a script - it won't just scrap $IDF_PATH enviroment variable if it didn't find python scripts.

## Checklist

🚨 This PR does not introduce breaking changes.

### DIFF
--- a/export.ps1
+++ b/export.ps1
@@ -7,17 +7,22 @@ if ($env:ESP_IDF_LEGACY_EXPORT) {
     exit $LASTEXITCODE
 }
 
-$idf_path = "$PSScriptRoot"
+if ("$env:IDF_PATH" -ne "") {
+    $idf_path = "$env:IDF_PATH"
+} else {
+    Write-Output "No IDF_PATH enviroment variable detected. Using the root directory as IDF_PATH"
+    $idf_path = "$PSScriptRoot"
+}
 
 if (-not (Test-Path "$idf_path/tools/idf.py") -or
     -not (Test-Path "$idf_path/tools/idf_tools.py") -or
     -not (Test-Path "$idf_path/tools/activate.py")) {
-
-    Write-Output "Could not detect IDF_PATH. Please navigate to your ESP-IDF directory and run:"
-    Write-Output ".\export.ps1"
-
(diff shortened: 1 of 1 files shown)


export.ps1
```diff
@@ -7,17 +7,22 @@ if ($env:ESP_IDF_LEGACY_EXPORT) {
     exit $LASTEXITCODE
 }
 
-$idf_path = "$PSScriptRoot"
+if ("$env:IDF_PATH" -ne "") {
+    $idf_path = "$env:IDF_PATH"
```
fhrbata, line 11: Hello @SaintFTS, 

thank you for your contribution and suggestion. The reason previous `IDF_PATH` settings aren't taken into account is that it's up to the export script to set it to the correct path. In my opinion, the script shouldn't take previous `IDF_PATH` settings into consideration, as this could cause unexpected behavior. For example, you might have two directories with different esp-idf versions, such as `esp-idf-v5.5` and `esp-idf-v5.4`. Suppose you run `. esp-idf-v5.5/export.ps1`, which activates the esp-idf environment for version 5.5, setting `IDF_PATH` to `esp-idf-v5.5`. If you then try to activate version 5.4 using `. esp-idf-v5.4/export.ps1`, the `IDF_PATH` remains set to `esp-idf-v5.5`, so the environment for version 5.5 will be activated instead of the intended version 5.4.

Thank you

SaintFTS: Yeah, it's my bad, sorry for confusion. Git clone somehow cloned an older repo (i think?), where export.ps1 stated that i "have to have IDF_PATH enviroment variable before starting". 

Again, sorry for wasting your time. Should've checked it before making a PR.

fhrbata: Hello @SaintFTS , no worries and thank you!

PR #15225, opened 2025-01-16
change(rmt): change rx buffer prints from error to debug (IDFGH-14449)

swoboda1337:
## Description

When using RF it is possible there is a lot of noise. The developer cannot control this noise. Noise could be very long and cause these prints to occur very often. If CONFIG_RMT_ISR_IRAM_SAFE is enabled these prints will be sent over the UART in ISR context. Each print will take a few milliseconds to send at 115200. The result can be system instability because ISRs are blocked for long periods of time. Currently the only work around is to disable all logging or disable UART logging.

Since these are not real errors and occur all the time in noisy environments it would be better to use ESP_DRAM_LOGD instead of ESP_DRAM_LOGE.

## Testing

Verified prints are debug now and no longer cause instability.

## Checklist


suda-morris:
sha=374f6f1f2add598a2a84f07e8cd4c7ff9dd0aa92

PR #15243, opened 2025-01-21
Set default check register for shared interrupts

udoudou:
## Description

The CPU interrupt signal of the ESP chip is very limited, and some basic peripherals in IDF have registered and used some interrupt signals. Usually in more complex application scenarios, it is easy to encounter the problem of being unable to register interrupts. For this situation, there are generally two ways to deal with it: 
1. The peripheral driver registered by the application layer sets the registration of shared interrupts; 
2. If it is a multi-core chip, some peripheral driver interrupts can also be registered on other cores.

However, the current shared interrupt logic design has defects. If the driver does not explicitly set statusreg when registering a shared interrupt, different peripherals assigned to the same interrupt signal, as long as one peripheral generates an interrupt, all peripheral interrupt programs will be executed once, which will cause some peripheral driver processing logic to be abnormal (such as SPI master driver).

There are two ideas to solve this problem: 

1. Adjust the peripheral driver code itself to adapt it to the shared interrupt situation; 
2. Set a default statusreg for the shared interrupt and only call the peripheral interrupt program triggered by the interrupt.

This modification corresponds to the second point above.

## Testing

Based on the examples/peripherals/lcd/spi_lcd_touch routine, I modified the initialization of spi to a shared interrupt, and added the initialization of the i2c peripheral to a shared interrupt, and periodically executed i2c transmission to cause a crash.
I tested this modification on the ESP32/ESP32S2/ESP32S3/ESP32C2/ESP32C3/ESP32C6 platform and no crash occurred

---

## Checklist


SoucheSouche:
Hi @udoudou, thanks for submitting the PR. We will take a look at it shortly.

ESP-Marius:
As discussed internally, fixing this in the interrupt allocator would be a breaking change in behavior as well as coupling the SW implementation hardware to HW details.

Issue will be resolved by fixing this "missuse" in the individual drivers.

PR #15262, opened 2025-01-22
Fix s2 and s3 Cache_Count_Flash_Pages rom function wrapper (IDFGH-14493)

EliteTK:
## Description

The `Cache_Count_Flash_Pages` rom function on the s2 and s3 only counts one page for any pages which are mapped to page 0 of flash as the `Cache_Flash_To_SPIRAM_Copy` function attempts to map all flash page 0 mapped pages to one PSRAM page.

As this function can be called for multiple regions, it needs to track if a page mapped to page 0 has previously been accounted for by a previous call. It does this using the `page0_mapped` in-out parameter. This logic contains an error[0]:

```c
// This code was reverse engineered from the ROM with the help of Ghidra and
// then cleaned up
uint32_t Cache_Count_Flash_Pages(uint32_t bus, uint32_t *page0_mapped)
{
    volatile uint32_t *mmu_table = (volatile uint32_t *)DR_REG_MMU_TABLE;
    uint32_t start, end;
    if (bus == CACHE_IBUS) {
        start = CACHE_IROM_MMU_START;
        end = CACHE_IROM_MMU_END;
    } else {
        start = CACHE_DROM_MMU_START;
        end = CACHE_DROM_MMU_END;
    }

    uint32_t count, page0_count = 0, valid_flash_count = 0;
    for (uint32_t i = start >> 2; i < end >> 2; i++) {
        uint32_t mapping = mmu_table[i];
        if ((mapping & (SOC_MMU_INVALID | SOC_MMU_TYPE)) == (SOC_MMU_VALID | SOC_MMU_ACCESS_FLASH)) {
            if ((mapping & SOC_MMU_VALID_VAL_MASK) == 0) page0_count++;
            valid_flash_count++;
        }
    }
    // BUG: If page0_count is 0, 1 is still added
    // Should be `if (page0_count != 0 && *page0_mapped == 0) {`
    if (*page0_mapped == 0) {
        count = valid_flash_count + 1 - page0_count;
    } else {
        count = valid_flash_count - page0_count;
    }
    *page0_mapped += page0_count;
    return count;
}
```

The current `Cache_Count_Flash_Pages` wrapper in the idf attempts to compensate for this bug by checking if the `page0_mapped` parameter was changed by a call to the function and reducing the count if it has not.

This, however, will incorrectly over-compensate in situations where the initial value of `page0_mapped` was not zero as the code above only miscounts when it was zero.

This patch addresses the issue in this wrapper function by correctly compensating for the bug only in cases where the final `page0_mapped` value is 0.

## Testing

I used a ESP32-S3 based LCD board to run code which made use of `CONFIG_SPIRAM_FETCH_INSTRUCTIONS` and `CONFIG_SPIRAM_RODATA`, I added some additional verbose logging in `mmu_config_psram_text_segment` and `mmu_config_psram_rodata_segment` to check the calculated page count and compare it to the change in `page_id` after the SPIRAM copy is performed.

Since my code did not result in page0 getting mapped (I'm not sure how to arrange for that organically), I manually triggered the issue by initialising `page0_mapped` to 1 which resulted in the original code under-counting the total number of required pages by 1 in each instance of the function being called.

With the fix applied, the count matched the jump in page number that subsequent calls to `Cache_Flash_To_SPIRAM_Copy` resulted in. This was tested both in cases which did and cases which did not trigger the bug.

## Checklist

      NOTE: This change would only break code which would already break due to a lack of PSRAM space.

### DIFF
--- a/components/esp_rom/patches/esp_rom_cache_esp32s2_esp32s3.c
+++ b/components/esp_rom/patches/esp_rom_cache_esp32s2_esp32s3.c
@@ -21,14 +21,13 @@
 extern uint32_t rom_Cache_Count_Flash_Pages(uint32_t bus, uint32_t * page0_mapped);
 uint32_t Cache_Count_Flash_Pages(uint32_t bus, uint32_t * page0_mapped)
 {
-    uint32_t page0_before_count = *page0_mapped;
     uint32_t flash_pages = 0;
     flash_pages = rom_Cache_Count_Flash_Pages(bus, page0_mapped);
 
-/* No page mapped to page0, in this condition, the rom api will return
+/* No page mapped to page0 yet, in this condition, the rom api will return
  * unexpected value + 1.
  */
-    if (page0_before_count == *page0_mapped) {
+    if (*page0_mapped == 0) {
         flash_pages--;
     }
     return flash_pages;


EliteTK:
I've resubmitted this as the CI complained about the branch name.

weiyuhannnn:
Hello @EliteTK，

Thank you for finding this bug in esp-idf. We have reviewed this commit and found there are still some issues.
Firstly, we get to know that rom function `Cache_Count_flash_Pages` will return the wrong result only when:

1. the *page0_mapped parameter is 0
2. the real pages mapped to page0 is 0

So we can conclude all the possible results into 4 scenarios (page0 = 1 means there is at least 1 page mapped to page0) and compare the patch currently IDF used and your commits:

- **Current IDF patch**
1.  `*page0_mapped = 0` and `page0 = 0`: `Cache_Count_flash_Pages` will return + 1, and the patch will be -1, the final result is right
2.  `*page0_mapped = 0` and `page0 = 1`:  `Cache_Count_flash_Pages` will return right, and the patch will not -1, the final result is right
3.  `*page0_mapped = 1` and `page0 = 0 `:  `Cache_Count_flash_Pages` will return right, and the patch will be -1, **the final result is wrong**
4. `*page0_mapped = 1 `and `page0 = 1`:  `Cache_Count_flash_Pages` will return right,  and the patch will not -1, the final result is right

- **your commit**

1. `*page0_mapped = 0` and `page0 = 0`: `Cache_Count_flash_Pages` will return + 1, and your commit will be -1, the final result is right
2. `*page0_mapped = 0` and `page0 = 1 `: `Cache_Count_flash_Pages` will return right, and your commit will be -1, **the final result is wrong**
3. `*page0_mapped = 1` and `page0 = 0`: `Cache_Count_flash_Pages` will return right, and your commit will not -1, the final result is right
4. `*page0_mapped = 1` and `page0 = 1`: `Cache_Count_flash_Pages` will return right, and your commit will not -1, the final result is right


We think we need combine the two conditions into:
```
uint32_t page0_before_count = *page0_mapped;
//...
if (page0_before_count == *page0_mapped && *page0_mapped == 0) {
    flash_pages--;
}
```
Then all the scenarios will be correct!

EliteTK:
Regarding your suggested change:

```c
uint32_t page0_before_count = *page0_mapped;
//...
if (page0_before_count == *page0_mapped && *page0_mapped == 0) {
    flash_pages--;
}
```

I can't think of a normal scenario where `page0_before_count == *page0_mapped` would be false when `*page0_mapped == 0` is true. In that conditional `*page0_mapped` should always be greater or equal to `page0_before_count` and page0_before_count must be greater to equal to 0. It follows, then, that if `*page0_mapped` is zero, then `page0_before_count` is zero.

The only (abnormal) scenario where this can turn out to be wrong is if `*page0_mapped` wraps during the call to `Cache_Count_Flash_Pages` which certainly could happen with enough calls but would itself be a bug, there's no meaningful recovery from such a situation. That's why I had modified the code to remove `page0_before_count` entirely.

I can modify my patch if you still feel it is appropriate. I would argue, however, that maybe in this case the function should report an error (although there's no obvious way to do that given its current interface, so maybe it could abort?).

Let me know if I understood correctly and if you still want me to make this adjustment.

weiyuhannnn:
Hello @EliteTK,
After reviewing the changes, we agree that your approach is correct. We will keep only your commit and discard the others. **We will update once we merge your commit.**
Let us know if you have any further suggestions!

weiyuhannnn:
Hello @EliteTK, 

The following is the commit of this fix. If this fix does not meet expectations, please feel free to reopen the ticket. Thanks!
- master:[af7b6b71ea681fb24c9e3c5bc6468978e257f481](https://github.com/espressif/esp-idf/commit/af7b6b71ea681fb24c9e3c5bc6468978e257f481)
- v5.4: [452e20d90bc8e16be4798273b2dc364f63b966e6](https://github.com/espressif/esp-idf/commit/452e20d90bc8e16be4798273b2dc364f63b966e6)
- v5.3: [664b7821f74f8e9e88d281a699d99f236ed1d4f3](https://github.com/espressif/esp-idf/commit/664b7821f74f8e9e88d281a699d99f236ed1d4f3)
- v5.2: [84a162aba759f7d450bdf05b3f391b77b4e460b4](https://github.com/espressif/esp-idf/commit/84a162aba759f7d450bdf05b3f391b77b4e460b4)
- v5.1: [aceebcccfe8c25ca975fcac610b7dac90d229952](https://github.com/espressif/esp-idf/commit/aceebcccfe8c25ca975fcac610b7dac90d229952)
- v5.0: [bf5ce28592a5ae28789f90251e5d1e58cbaffea1](https://github.com/espressif/esp-idf/commit/bf5ce28592a5ae28789f90251e5d1e58cbaffea1)

PR #15277, opened 2025-01-24
fix: disable Werror=unused-values

KaeLL:


dobairoland:
@antmak @Lapshin Can you please take a look?

igrr:
@KaeLL perhaps you could add, why do you think that the default behavior should be changed? If you want to allow this warning in your project, you can already do so, it doesn't seem to be necessary to change the default.

KaeLL:
I believe having dead code or unused variables is reason for warnings, which is what this change is all about, but not build-breaking errors. For regular development, it breaks the usual "edit-compile-test-debug" cycle in exchange for absolutely nothing of positive that a warning wouldn't suffice. For production builds, the code will likely end up eliminated by the compiler anyways.
I guess the question is: why shouldn't this default be changed (back?)? I could be wrong or missing the whole picture.

dobairoland:
It is just my opinion but these errors are generated in order to ensure code cleanness of ESP-IDF releases. If you don't want it in your project and want to allow unused values then you can achieve it at the project level, as it was pointed out earlier. 

So if there is no objection from any side then I will close this PR.

KaeLL:
Might as well, then.

dobairoland:
Alright, it looks there are no more comments/opinions. I'm sorry but I'm closing it.

PR #15288, opened 2025-01-26
Improve httpd string value fetching efficiency

Adam5Wu:
## Description

- Avoid using `strlcpy()` when source is not null-terminated;
- Also avoid duplicate `strlen()` when `strlcpy()` is used.

`strlcpy()` will implicitly perform a `strlen()` and report the size of source string as return value [1].
This means calling `strlcpy()` when source string is no null-terminated (for the intended scope) will incur a high penalty.

For example, query key-values are delimited by `&`, but calling `strlcpy()` will always have to traverse to the very end of all query key-values. Similarly for cookies.

[1] https://linux.die.net/man/3/strlcpy

## Testing

Just performance improvements, no externally obverable behavioral differences.

---

## Checklist

- [n/a] Documentation is updated as needed.
- [n/a] Tests are updated or added as necessary.


hrushikesh430:
sha=16d9b94c563dda4c35c96393c58d53f75a2edc1c

PR #15291, opened 2025-01-27
fix(esp_http_client): Fix invalid content length header (IDFGH-14528)

chris-durand:
In case a request with no content (e.g GET) is sent after one containing data the incorrect content length header of the previous request is sent with the subsequent one.

For instance, an empty GET request after a PUT request will still have the non-zero content length header set.
This is fixed by clearing the header in that case.

The bug was introduced as a side effect of #14459


nileshkale123:
Hello @chris-durand ,

Could you please let me know the steps you have followed to reproduce the issue.

mahavirj (approved):

mahavirj:
sha=c0590f5b50556ae4decf237aa8715375096a2402

mahavirj:
> Hello @chris-durand ,
> 
> Could you please let me know the steps you have followed to reproduce the issue.

Using same HTTP connection to send a request with `PUT` followed by `GET` with no body (0 content length) should result in this issue. We can consider adding a test for this scenario.

PR #15300, opened 2025-01-29
fix(transport_ws): remove unnecessary noise from error message (IDFGH-14538)

erkia:
A message reporting an error about `esp_transport_read()` prints contents of the buffer that was passed to this function. The contents of this buffer are irrelevant in the context of this message.

1) it can contain random, not NUL-terminated data from possibly partial read
2) it can contain data from previous `esp_transport_write()` call that is using the same buffer.

In either case this is just a noise:

Screenshot_20250129_151900


euripedesrocha:
Hi @erkia thanks for the contribution. I'll follow with the process to merge.

euripedesrocha:
sha=b1c3b69391ffb72463f48f596555f7abb9123269

PR #15323, opened 2025-02-03
fix(esp_hid): Fix Device ID for BT Classic

mitchellcairns:
Right now, the VID/PID/VID Source are unused for BT Classic HID. This resolves that particular issue. This also Implements VID Source for BLE.

### DIFF
--- a/components/bt/host/bluedroid/api/include/api/esp_hidd_api.h
+++ b/components/bt/host/bluedroid/api/include/api/esp_hidd_api.h
@@ -98,6 +98,10 @@ typedef struct {
     uint8_t subclass;                                /*!< HID device subclass */
     uint8_t *desc_list;                              /*!< HID descriptor list */
     int desc_list_len;                               /*!< size in bytes of HID descriptor list */
+    uint16_t vendor_id;                              /*!< HID Vendor ID */
+    uint16_t product_id;                             /*!< HID Product ID */
+    uint16_t version;                                /*!< HID Product Version */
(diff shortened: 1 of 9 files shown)


boblane1:
Hi @mitchellcairns,

Thank you for submitting the PR. Would you mind making a slight modification to the code? Instead of directly calling `BTA_DmSetLocalDiRecord`, could you create the DIP record by invoking `esp_sdp_create_record`? This way, there would be no need to change esp_hidd_app_param_t.  `esp-idf/examples/bluetooth/esp_hid_device/main/esp_hid_device_main.c` has shown how the function `esp_sdp_create_record` works.

mitchellcairns:
@boblane1 Are you proposing implementing a SDP callback within the esp_hid component? I want to make sure I understand what you're getting at. I think this would complicate this a bit as what would then happen later if the user wants to use SDP callbacks for other behavior?

boblane1:
>  I think this would complicate this a bit as what would then happen later if the user wants to use SDP callbacks for other behavior

This does seem to be a problem!

However, the primary DIP record created by the user using the SDP API may also overwrite the DIP Record already created in the HID.  What's your opinion about the following suggestion?

- Create a non-primary DIP record in HID
- Create a primary DIP record in HID but try not to change the `esp_hidd_api.h`.
- Just let users manage DIP themselves

If you decide to create a DIP record in HID, please care about the resource release.

mitchellcairns:
> Just let users manage DIP themselves

In the end I think this is probably the more wise solution. I think I have a better understanding of the situation. Thank you for reviewing!

PR #15337, opened 2025-02-05
fix(esp_hid): Enable notifications/Indications for HID Devices

mitchellcairns:
When HID devices reconnect to a host such as a WIndows device, Windows will not subscribe to notifications/indications, resulting in a state where no reports may be sent.

Open to another fix, but this ought to be looked into and resolved. It results in HID devices only being able to report on initial pairing, and not again with reconnections thereafter.

NOTE: This does not always happen on Windows devices due to some caching behavior apparently. But, where it does occur, it is consistent behavior.

### DIFF
--- a/components/esp_hid/src/ble_hidd.c
+++ b/components/esp_hid/src/ble_hidd.c
@@ -476,6 +476,17 @@ static void hid_event_handler(esp_ble_hidd_dev_t *dev, int device_index, esp_gat
             dev->conn_id   = param->connect.conn_id;
             memcpy(dev->remote_bda, param->connect.remote_bda, ESP_BD_ADDR_LEN);
 
+            // ADD THIS BLOCK: Initialize CCC for all reports
+            for (int i = 0; i < dev->devices[device_index].reports_len; i++) {
+                hidd_le_report_item_t *map = &dev->devices[device_index].reports[i];
+                if (map->ccc_handle && !map->ccc.notify_enable && !map->ccc.indicate_enable) {
+                    ESP_LOGW(TAG, "Initializing CCC for report %d", map->report_id);
+                    map->ccc.notify_enable = true;
+                    map->ccc.indicate_enable = true;
+                    map->ccc.value = 0x03; // Enable notifications and indications
+                }
+            }
+
             esp_ble_set_encryption(param->connect.remote_bda, ESP_BLE_SEC_ENCRYPT_NO_MITM);
 
             esp_hidd_event_data_t cb_param = {


esp-zhp:
Hello,  

Thank you for your feedback!  

The root cause of this issue is that after the ESP device powers off, the CCC (Client Characteristic Configuration) information used by the `esp_ble_hidd_dev_input_set` function is not restored. Additionally, the peer device (such as a smartphone acting as an HID host) does not rewrite the CCC value, resulting in an undefined CCC state.  

Different smartphones exhibit different behaviors: some will attempt to set the CCC upon reconnection, while others will not, leading to compatibility issues. The best approach to resolving this issue is to restore the CCC information from NVS, rather than modifying the implementation in the way you suggested.  

Please allow me some time to work on this.

mitchellcairns:
> Hello,
> 
> Thank you for your feedback!
> 
> The root cause of this issue is that after the ESP device powers off, the CCC (Client Characteristic Configuration) information used by the `esp_ble_hidd_dev_input_set` function is not restored. Additionally, the peer device (such as a smartphone acting as an HID host) does not rewrite the CCC value, resulting in an undefined CCC state.
> 
> Different smartphones exhibit different behaviors: some will attempt to set the CCC upon reconnection, while others will not, leading to compatibility issues. The best approach to resolving this issue is to restore the CCC information from NVS, rather than modifying the implementation in the way you suggested.
> 
> Please allow me some time to work on this.

Thank you. Looking forward to a fix.

esp-zhp:
Does this issue also occur when using the **NimBLE host**?

mitchellcairns:
> Does this issue also occur when using the **NimBLE host**?

I have not tested NimBLE as it's out of my project scope for the time being. I have to switch back and forth between BLE and BR/EDR

esp-zhp:
Your commit has been merged into the master branch, but it will take some time to sync to GitHub.  

Additionally, your commit will also be synced to the release branch, which will also take some time.  

Please note that the final changes may not be exactly the same as your original modifications, as I have made some optimizations to conform to coding standards. However, your commit information has been retained.  

Here are the details:  
- **master:** d4b3a7e99d29150d9718415384db7ac5ae7673d5  
- **v5.4:** ddff846e4ef4ffc5e5203ec7e872236bc6ac61b5  
- **v5.3:** d30c46cf98acc1a01587991d67f293208b619a7b  
- **v5.2:** 72c86ba35c0e083049c1629fd2930a281aaa28e1  
- **v5.1:** 6de88b82042739b7683cae941f4234af256f9120  
- **v5.0:** bd504372b60278807eba3251e0d279c851af6a5e

esp-zhp:
image

PR #15459, opened 2025-02-24
Improve PPP netif server support  (IDFGH-14720)

gonzzor:
## Description

The previous introduction of server support was a nice addition, improve this by making it easier to run both a PPP server and PPP client on different interfaces.

Note that if CONFIG_LWIP_PPP_SERVER_SUPPORT=y and ppp_passive isn't set the code will call ppp_connect() instead of ppp_listen(). However, ppp_connect() and ppp_listen() are the same for pppos so it shouldn't matter.

For PPPoE ppp_listen() isn't really supported.

## Testing

Tested with PPP over serial as server and esp-modem for uplink.

---

## Checklist

### DIFF
--- a/components/esp_netif/include/esp_netif_ppp.h
+++ b/components/esp_netif/include/esp_netif_ppp.h
@@ -33,6 +33,7 @@ typedef struct esp_netif_ppp_config {
     esp_ip4_addr_t ppp_their_ip4_addr; /**< Set our preferred address, typically used when we're the PPP server */
     esp_ip4_addr_t ppp_dns1_addr;      /**< DNS to provide if peer asks for it, typically used when we're the PPP server */
     esp_ip4_addr_t ppp_dns2_addr;      /**< DNS to provide if peer asks for it, typically used when we're the PPP server */
+    bool ppp_passive;                  /**< Try once to initiate connection, stay silent if it fails, typically used when we're the PPP server */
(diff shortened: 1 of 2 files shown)


david-cermak (commented):
Thanks for this update!

LGTM, but please remove the new events from `ip_event_t`
* IP_EVENT_PPPD_GOT_IP
* IP_EVENT_PPPD_LOST_IP
The netif pointer is passed to the event so app code can tell, which is which, not getting `IP_EVENT_PPP_...`

components/esp_netif/include/esp_netif_types.h
```diff
@@ -105,6 +105,8 @@ typedef enum {
     IP_EVENT_PPP_GOT_IP,               /*!< PPP interface got IP */
     IP_EVENT_PPP_LOST_IP,              /*!< PPP interface lost IP */
     IP_EVENT_TX_RX,                    /*!< transmitting/receiving data packet */
+    IP_EVENT_PPPD_GOT_IP,              /*!< PPP server interface got IP */
+    IP_EVENT_PPPD_LOST_IP,             /*!< PPP server interface lost IP */
```
david-cermak, line 109: Please remove

gonzzor: You are correct, they aren't needed. I've removed them and updated the branch.

PR #15473, opened 2025-02-26
fix docker build

dobairoland:


dobairoland:
None of the suggestions from https://askubuntu.com/questions/804997/dpkg-error-processing-package-libc-bin-configure helped

Issue #15327, opened 2025-02-03
BLE Appearance Characteristic only supports an incomplete hard-coded list of icons


qwertymodo:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

idf-release_v5.3-cfea4f7c-v1

### Espressif SoC revision.

ESP32

### Operating System used.

Windows

### How did you build your project?

Other (please specify in More Information)

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32-DevKitC

### Power Supply used.

USB

### What is the expected behavior?

When I call esp_ble_gap_config_local_icon() with a valid Appearance value, I expect that value to be accepted and the Appearance to be updated accordingly.

### What is the actual behavior?

esp_ble_gap_config_local_icon() seems to have a hard-coded list of Appearance values [here](https://github.com/espressif/esp-idf/blob/master/components/bt/host/bluedroid/api/esp_gap_ble_api.c#L304) that it will accept, and any other values are rejected with an ESP_ERR_INVALID_ARG error. Unfortunately, this hard-coded list is both woefully incomplete and also unmaintained, as the git blame seems to indicate only a single value has been added to the list since it was originally created 7 years ago, so many completely valid Appearance values are rejected.

### Steps to reproduce.

Attempt to call esp_ble_gap_config_local_icon with any Appearance value not explicitly defined in the switch case found at https://github.com/espressif/esp-idf/blob/master/components/bt/host/bluedroid/api/esp_gap_ble_api.c#L304

e.g.:
0x03CA (Presentation Remote)
0x04C1 (Switch)
0x0540 (Generic Sensor)
etc.

Check the return value, which will be 0x102 ESP_ERR_INVALID_ARG and the Appearance will remain 0x0000 Unknown when connecting to the device.

### Debug Logs.

```plain

```

### More Information.

I recommend simply removing the switch statement entirely.  It would be one thing if the list was actively maintained and up-to-date with all current valid Appearance definitions, but as that is clearly not the case, the "validation" is not itself valid and is instead rejecting perfectly valid inputs for no reason.

qwertymodo:
[...]

PR #15477, opened 2025-02-26
Add a bunch of missing appearance definitions (IDFGH-14739)

qwertymodo:
## Description

The list of appearance characteristics defined in this library is woefully incomplete (less than 20% of all appearance values listed in the official spec are represented).  To make matters worse, because the API validates any attempt to change the appearance characteristic against this incomplete list and rejects all other values, it is impossible to set the appearance characteristic of a device to over 250 perfectly valid appearance values.  This PR updates the list of definitions using [the most up-to-date parseable source I could find](https://github.com/NordicSemiconductor/bluetooth-numbers-database/blob/master/v1/gap_appearance.json).  These changes were automated using a Python script to parse the list and generate the proper defines/case statements, because I didn't feel like manually typing out over 300 names multiple times.  If there are any objections or modifications requested for things like naming style on the defines, let me know and I can make changes as necessary.  I have attached the python script that I used in case it's of any use in maintaining this list in the future.

[gap_appearance_defs.py.txt](https://github.com/user-attachments/files/18996709/gap_appearance_defs.py.txt)

## Related

Fixes: #15327

As mentioned in #15327 I think validating against this hard-coded list is unmaintainable and improper, but in the interest of trying not to break anything, I have simply expanded the case statement.

## Testing

This PR has not been exhaustively tested.  However, as it is merely a long list of preprocessor defines without any conflicts elsewhere in the codebase, there is little to no risk of this causing any issues.

---

## Checklist

- [✅] 🚨 This PR does not introduce breaking changes.
- [✅] All CI checks (GH Actions) pass.
- [N/A] Documentation is updated as needed.
- [N/A] Tests are updated or added as necessary.
- [✅] Code is well-commented, especially in complex areas.
- [✅] Git history is clean — commits are squashed to the minimum necessary.


nebkat:
https://github.com/espressif/esp-idf/pull/14110 There is really no need for the API to constrain the value, but hopefully one of these can get merged...

esp-zhp:
Is there a way to represent it using a range? Though it might be difficult to define the range clearly.

nebkat:
The Bluetooth spec has over the years added new values and existing devices are expected to continue working, so it would seem to me that allowing any value is most sensible.

The [definition of the characteristic](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/CSS_v11/out/en/supplement-to-the-bluetooth-core-specification/data-types-specification.html#UUID-ccfc3325-626d-a1cf-3083-1d5a9112023a) in the spec only states that:

> «Appearance» uint16 The Appearance value shall be the enumerated value as defined by [Assigned Numbers](https://www.bluetooth.com/specifications/assigned-numbers).

We could define a range based on the [current values](https://www.bluetooth.com/wp-content/uploads/Files/Specification/Assigned_Numbers.html#bookmark49), which are `0x0000` to `0x0DFF` and `0x1440` to `0x147F` but we may end up with the same problem every time the list is updated. That said it would be much better than the switch statement now which is probably generating a lot of machine code.

qwertymodo:
The issue with a range for validation is that the values are actually bit
fields containing a category and sub-category, so the defined values are
non-contiguous, and you actually have something like 30+ separate ranges to
check. Also, outside of the giant switch statement you still want every
value to be defined so that you have access to semantically meaningful
names, so even if you ditch the validation method you still end up with
those giant lists elsewhere.

esp-zhp:
Thank you for your feedback. We understand that this issue can be frustrating, especially when working with the BLE Appearance characteristic and encountering limitations due to the current implementation.

At the moment, we perform strict validation against a predefined list of Appearance values to prevent the use of invalid entries. While we don’t intend to remove this validation — as it helps catch incorrect usage — we also recognize that maintaining a hard-coded list becomes increasingly difficult as the Bluetooth specification evolves.

You’re absolutely right that including all defined Appearance values would significantly increase the code size, which goes against our principle of keeping the codebase lean and efficient.

As for how other BLE hosts handle this: from our understanding, most major BLE stacks (such as Nordic’s SDK or BlueZ) **do not strictly validate the Appearance value**. They typically allow setting any 16-bit value and treat the defined values as optional enums to assist developers, not as a hard constraint. This approach provides more flexibility and reduces maintenance overhead.

We are actively evaluating alternative approaches as well — for example, relaxing the validation, or using range checks to reduce the complexity of the current `switch-case` structure — so that we can stay compatible with the specification while also improving performance and maintainability.

Thanks again for raising this concern and for your thoughtful suggestions. Please feel free to share any additional feedback — we truly appreciate your input!

KaeLL:
@esp-zhp 
You can't restrict correct options in the name of correctness if you can't guarantee the definition of correct options are updated promptly. The switch statement has got to go.

If it were up to me, I'd replace it with an enum, preferably with a fixed underlying type (C23 feature but supported by GCC as an extension). You get enough correctness through the type check, unless client code intentionally gets around it to use values not yet added to it, or custom ones, either way, not prohibiting the practice, just providing a good enough default. It's also better than the way nimBLE does it (although it doesn't provide such a thing), with just plain preprocessor constants scattered within header files that you ultimately need to guess whether they are actually related to the elements you're assigning them to.

qwertymodo:
Based on how hesitant you are to include this PR, I am skeptical that this list will be kept up-to-date in the future.  That coupled with the fact that the current code rejects over 80% of all valid inputs is causing me to lean toward suggesting that the validation check should be informational-only, and the code should attempt to set the appearance value regardless of if it's known or not.  The spec is not final, so unless there are any values explicitly called out as reserved or invalid (there aren't in this case), there are **_no_** values which are actually known to be invalid, just currently undefined.  If you don't actually know a value to be invalid, then you should not be rejecting it as such, and you should leave it up to the user code to ensure that they are setting the correct value that they intend to.

qwertymodo:
Thanks for that.  With the restriction removed, all the new defines should be helpful to users so they can easily pick valid values by name instead of needing to look up numbers in the spec, so that should help keep undefined values to a minimum.  If there aren't any further concerns on your end, that looks good to me, hopefully this can get merged soon.

esp-zhp:
Hi all
We have fixed the issue in our codebase. The fix is included in the master branch at commit 6552854cb5b0a9dc5c68f89e6fec438bcbd764c5. Other branches will be synchronized to GitHub gradually.

Please let us know if you need any further assistance.thanks.

PR #15499, opened 2025-03-01
Add missing break statements to usb_serial_jtag_ll_phy_select (IDFGH-14760)

renzenicolai:
This commits adds two missing "break" statements to the usb_serial_jtag_ll_phy_select HAL function. The missing statements prevent users from switching back the two PHYs to the default configuration as the statement will always fall through to the second case in the switch case.


suda-morris:
sha=c47349272527ffd3e967f5ac2c228bda4812c86d

Issue #15496, opened 2025-02-28
PSRAM heap start calculation bug with CONFIG_SPIRAM_ALLOW_NOINIT_SEG_EXTERNAL_MEMORY enabled (IDFGH-14757)


erkia:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.3.2

### Espressif SoC revision.

ESP32-S3

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

Custom Board

### Power Supply used.

External 5V

### What is the expected behavior?

I expect heap structures not to reside inside psram .noinit segment.

### What is the actual behavior?

Heap structures end up inside psram .noinit segment.

### Steps to reproduce.

1. Enable `CONFIG_SPIRAM_ALLOW_NOINIT_SEG_EXTERNAL_MEMORY=y`
2. Allocate aligned buffer in external .noinit RAM, i.e.:
`EXT_RAM_NOINIT_ATTR __attribute__((aligned(32))) uint8_t pucRingbufferStorage[32768];`

### Debug Logs.

```plain
Linker output in map file:

 ...
 .bss.gEapSm    0x3c2038e4        0x4 esp-idf/wpa_supplicant/libwpa_supplicant.a(esp_eap_client.c.obj)
 *libwpa_supplicant.a:(COMMON)
                0x3c2038e8                        . = ALIGN (0x4)
                0x3c2038e8                        _ext_ram_bss_end = ABSOLUTE (.)

.ext_ram_noinit
                0x3c203900     0x8000
                0x3c203900                        _ext_ram_noinit_start = ABSOLUTE (.)
 *(.ext_ram_noinit*)
 .ext_ram_noinit.2
                0x3c203900     0x8000 esp-idf/main/libmain.a(logger.cpp.obj)
                0x3c203900                pucRingbufferStorage
                0x3c20b900                        . = ALIGN (0x4)
                0x3c20b900                        _ext_ram_noinit_end = ABSOLUTE (.)
...


Output from manually added printf() on esp_psram.c:324:

vaddr_start is 0x3c20b8e8


Crashes in random places:

[...]

PR #15513, opened 2025-03-04
fix(psram): correct heap vaddr calculation (IDFGH-14775)

erkia:
## Description

In case both `.ext_ram_bss` and `.ext_ram_noinit` segments exist, the heap start calculation may be incorrect when there is a gap between two sections. This is because the heap start calculation only considers segment sizes and not a possible gap between them. This gap may be introduced when using `__attribute__((aligned(...)))` on the first buffer in `.ext_ram_noinit` segment, i.e.:
```
...
 .bss.gEapSm    0x3c2038e4        0x4 esp-idf/wpa_supplicant/libwpa_supplicant.a(esp_eap_client.c.obj)
 *libwpa_supplicant.a:(COMMON)
                0x3c2038e8                        . = ALIGN (0x4)
                0x3c2038e8                        _ext_ram_bss_end = ABSOLUTE (.)

.ext_ram_noinit
                0x3c203900     0x8000
                0x3c203900                        _ext_ram_noinit_start = ABSOLUTE (.)
 *(.ext_ram_noinit*)
 .ext_ram_noinit.2
                0x3c203900     0x8000 esp-idf/main/libmain.a(logger.cpp.obj)
                0x3c203900                pucRingbufferStorage
                0x3c20b900                        . = ALIGN (0x4)
                0x3c20b900                        _ext_ram_noinit_end = ABSOLUTE (.)
...
```
In this case, `s_psram_ctx.regions_to_heap[PSRAM_MEM_8BIT_ALIGNED].vaddr_start` is incorrectly calculated to be `0x3c20b8e8`, while this address obviously resides inside the allocated buffer. 

Correct address would be: `0x3c203900 + 0x8000 = 0x3c20b900`

This PR changes the calculation so that the beginning of the first segment and the end of the last segment is used to calculate the true allocated size of the both regions. It keeps the (theoretical?) possibility of these segments being in random order (don't know if this was the intention of the previous implementation).

## Related

Fixes https://github.com/espressif/esp-idf/issues/15496

### DIFF
--- a/components/esp_psram/esp_psram.c
+++ b/components/esp_psram/esp_psram.c
@@ -307,21 +307,31 @@ static void s_psram_mapping(uint32_t psram_available_size, uint32_t start_page)
     * After mapping, we DON'T care about the PSRAM PHYSICAL ADDRESS ANYMORE!
     *----------------------------------------------------------------------------*/
 
-    //------------------------------------Configure .bss in PSRAM-------------------------------------//
+    //------------------------------------Configure heap in PSRAM-------------------------------------//
+    uintptr_t ext_segment_start = UINTPTR_MAX;
+    uintptr_t ext_segment_end = 0;
+
 #if CONFIG_SPIRAM_ALLOW_BSS_SEG_EXTERNAL_MEMORY
-    //should never be negative number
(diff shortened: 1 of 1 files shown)


erkia:
Thanks!

PR #15527, opened 2025-03-06
python-pip is also a required package on Arch Linux (IDFGH-14789)

redfast00:
`python-pip` is also a required package; but most people probably don't run into this documentation issue, since a lot of people already have this package installed for other projects/programs


dobairoland:
Hi @redfast00. Thank you for the contribution. Could you please fix the commit title based on the above suggestion from github-actions? Also could you please make the same change in the Chinese mirror documentation page? The file name is the same and the location can be easily found.

dobairoland:
Thanks!

dobairoland:
SHA=08b6f6018f251f5adf5b425cf3fda38302293b28

PR #15564, opened 2025-03-12
fix: race condition when mmapping or munmapping in esp_partition (IDFGH-14841)

smjothen:
We have multiple tasks that do reads from encrypted partitions. This incurs quite a few calls to `esp_partition_mmap`, which then calls `spi_flash_mmap` and then `esp_mmu_map`. `esp_mmu_map` is not thread safe, however, and thus with lots of reads to encrypted partitions, some corruption can occur to the internal `esp_mmu` list of mapped regions.

This usually results in a panic of some sort, usually either due to trying to map an invalid address, or trying to unmap an address that has already been unmapped.

## Description

To fix this there is a new mutex added in esp_partition to ensure there is no concurrent changes to mapped regions.

## Testing

Created an example application that has two tasks that repeatedly call `esp_partition_mmap` and `esp_partition_munmap` very frequently. The issue usually triggers within a few seconds, causing a panic of some sort.

With the fix applied, the application no longer panics.

---

## Checklist

### DIFF
--- a/components/esp_partition/partition_target.c
+++ b/components/esp_partition/partition_target.c
@@ -23,6 +23,12 @@
 
 #define HASH_LEN 32 /* SHA-256 digest length */
 
+/*
+ * Due to esp_mmu_map and spi_flash_mmap APIs not being thread-safe, to avoid
+ * race conditions, we need to lock any map/unmap operations.
+ */
+static _lock_t s_mmap_lock;
+
 esp_err_t esp_partition_read(const esp_partition_t *partition,
                              size_t src_offset, void *dst, size_t size)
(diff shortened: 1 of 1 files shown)


smjothen:
Here's an example application that triggers panics for me:

[mmap_race.tar.gz](https://github.com/user-attachments/files/19209245/mmap_race.tar.gz)

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/b45b29ae163b2ac57c5e8b54c6b06ce4be2673a0.

PR #15569, opened 2025-03-14
Remove problematic and unuseful logging statement from gpio.c (IDFGH-14846)

skibu:
Removed inappropriate and misleading logging statement. The logging statement was at INFO level, which means that it was inappropriately logging cryptic info. And the context wasn't explained, which means that it took quite a long time to find out exactly why the logging statement wrongly implied that I wasn't successfully setting up input and output bits. Best to simply remove it.

### DIFF
--- a/components/esp_driver_gpio/src/gpio.c
+++ b/components/esp_driver_gpio/src/gpio.c
@@ -339,11 +339,6 @@ esp_err_t gpio_config(const gpio_config_t *pGPIOConfig)
 {
     uint64_t gpio_pin_mask = (pGPIOConfig->pin_bit_mask);
     uint32_t io_num = 0;
-    uint8_t input_en = 0;
-    uint8_t output_en = 0;
-    uint8_t od_en = 0;
-    uint8_t pu_en = 0;
-    uint8_t pd_en = 0;
 
     if (pGPIOConfig->pin_bit_mask == 0 ||
             pGPIOConfig->pin_bit_mask & ~SOC_GPIO_VALID_GPIO_MASK) {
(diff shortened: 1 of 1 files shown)


skibu:
I have now signed the contributor agreement.

Do the commit titles need to be changed in order for this PR to be processed? I ask because this cannot be done via GitHub website. If the titles truly need to be changed then I have to download my fork and modify the commits manually.

wanckl:
@skibu Don't think so, the log you removed is usefull for my debug, and can be bypassed by level.

may I know how it misleads you, we can update it more understandable

KaeLL:
No offense, but seems like you're overreacting because you don't know what you're doing.
> The logging statement was at INFO level, which means that it was inappropriately logging cryptic info.

Non-sequitur
> And the context wasn't explained

Care to elaborate on how that could be fixed?

> it took quite a long time to find out exactly why the logging statement wrongly implied that I wasn't successfully setting up input and output bits.

How so?

skibu:
> may I know how it misleads you, we can update it more understandable

@wanckl , thank you for starting the dialog. I'm certain we can come up with a mutually satisfactory solution.

**What happened:** 
I was using new to me software (esp-idf-cxx) to configure GPIO bits on a new ESP32 board. With informational logging enabled I noticed a peculiar log message when I was setting up IO bits to be either inputs or outputs. The log message was something like:
```
GPIO[17]| InputEn: 0| OutputEn: 0| OpenDrain: 0| Pullup: 0| Pulldown: 0| Intr:0
```
The reason this logging message was misleading is because it implied that after I setup my GPIO bit as an input or an output that **InputEn:0** and **OutputEn: 0**. The message was implying that the bit was not actually successfully configured. 

The reason for this problem is that the logging message was part of a reset that was done before the bit was actually configured as an input or an output. But the message did not indicate that it was just part of the reset.

**Possible Solutions:**
* Could simply delete the logging statement as I proposed, but this would limit your debugging ability. So likely not the best solution.
* Could modify the logging statement:
   1) change to debug logging. ESP_LOGI() => ESP_LOGD(). This way the message of course wouldn't show up unless debugging.
   1) add context to show that the debug logging statement is for a reset command. Could simply add to the beginning of the logging statement so that "GPIO[17]..." => "Reseting bit GPIO[17]..."

Let me know if I can further explain what happened. Also, let me know if you want to close this PR and have me submit a new one with a different solution.

igrr:
> change to debug logging. ESP_LOGI() => ESP_LOGD(). This way the message of course wouldn't show up unless debugging

+1 for this option.

I think the GPIO driver differs from other peripheral drivers in that it logs its configuration at info level. The reason why it differs is probably that GPIO driver was one of the first drivers we added in IDF, and there were no established conventions for logging from IDF components at that point.

Calling `esp_log_level_set("gpio", ESP_LOG_NONE)` is of course possible, but so is `esp_log_level_set("gpio", ESP_LOG_DEBUG)`, provided that CONFIG_LOG_MAXIMUM_LEVEL_DEBUG is set. So I think the _debugging_ use case should still be satisfied if we move this log to _debug_ level.

PR #15622, opened 2025-03-21
Add support for partial websocket frame payload reads

mrsobakin:
## Description

Added support for partial websocket frame payload reads. This was implemented as a separate function `httpd_ws_recv_frame_part` to keep the backwards compatibility with subtle `httpd_ws_recv_frame` behavior, that being that it returns an undocumented error when `max_len < pkt->len`.

`httpd_ws_frame_t` now also contains new `left_len` field, which is used to keep track of the yet to be received data and calculating the mask offset. In the future, it shall also be used for implementing partial frame sends.

## Testing

This feature was tested by me on my project and on some additional edge cases, including trying to over-receive (reading more data than there is in a frame payload) and to under-receive (leaving data unprocessed and unread and completing the request). No unexpected errors were identified.

The change itself is not very complex, however in my opinion it is imperative that someone with more knowledge about the ESP HTTP stack verifies that there is no deeper issues with this implementation (and the idea itself, for that matter), which I may have missed.

---

## Checklist

### DIFF
--- a/components/esp_http_server/include/esp_http_server.h
+++ b/components/esp_http_server/include/esp_http_server.h
@@ -1708,6 +1708,8 @@ typedef struct httpd_ws_frame {
     httpd_ws_type_t type;       /*!< WebSocket frame type */
     uint8_t *payload;           /*!< Pre-allocated data buffer */
     size_t len;                 /*!< Length of the WebSocket data */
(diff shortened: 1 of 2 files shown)


nileshkale123:
sha=6518fb938e06558c95820aeb90f75a3515e482e9

PR #15623, opened 2025-03-21
change(ble_mesh): Allow setting the default pb-gatt device name

s0be:
## Description

Currently, even if you call esp_ble_mesh_set_unprovisioned_device_name immediately after esp_ble_mesh_init, you can end up sending out an advertisement using the default 'ESP-BLE-MESH' name.  Unfortunately, due to iOS limitations, this can be problematic as there's no way for a user app to flush the cached names.

Resolve this by allowing the default name to be set before calling esp_ble_mesh_init.

## Testing

Modified our esp_ble_mesh test hardware and confirmed iOS never saw ESP-BLE-MESH advertised.  Backported to esp-idf 5.2 for testing.
---

## Checklist

### DIFF
--- a/components/bt/esp_ble_mesh/api/core/esp_ble_mesh_provisioning_api.c
+++ b/components/bt/esp_ble_mesh/api/core/esp_ble_mesh_provisioning_api.c
@@ -170,6 +170,11 @@ esp_err_t esp_ble_mesh_set_unprovisioned_device_name(const char *name)
             == BT_STATUS_SUCCESS ? ESP_OK : ESP_FAIL);
 }
 
+esp_err_t esp_ble_mesh_set_default_unprovisioned_device_name(const char *name)
+{
+    return bt_mesh_set_default_device_name(name) ? ESP_FAIL : ESP_OK;
+}
+
 #if CONFIG_BLE_MESH_PROVISIONER
 esp_err_t esp_ble_mesh_provisioner_read_oob_pub_key(uint8_t link_idx, uint8_t pub_key_x[32],
                                                     uint8_t pub_key_y[32])
--- a/components/bt/esp_ble_mesh/api/core/include/esp_ble_mesh_provisioning_api.h
+++ b/components/bt/esp_ble_mesh/api/core/include/esp_ble_mesh_provisioning_api.h
@@ -117,6 +117,19 @@ esp_err_t esp_ble_mesh_node_input_string(const char *string);
 esp_err_t esp_ble_mesh_set_unprovisioned_device_name(const char *name);
 
 /**
+ * @brief        Using this function, you can set the default name set during init.
+ *
+ * @param[in]    name: Unprovisioned device name
+ *
+ * @note         This API applicable to PB-GATT mode only by setting the name to the scan response data,
+ *               it doesn't apply to PB-ADV mode.
+ *
+ * @return       ESP_OK on success or error code otherwise.
+ *
+ */
+esp_err_t esp_ble_mesh_set_default_unprovisioned_device_name(const char *name);
+
+/**
  * @brief        Provisioner inputs unprovisioned device's oob public key.
  *
  * @note         In order to avoid suffering brute-forcing attack (CVE-2020-26559).
--- a/components/bt/esp_ble_mesh/core/proxy_server.c
+++ b/components/bt/esp_ble_mesh/core/proxy_server.c
@@ -88,6 +88,7 @@ static enum {
     MESH_GATT_PROXY,
 } gatt_svc = MESH_GATT_NONE;
 
+static char default_device_name[DEVICE_NAME_SIZE + 1] = "ESP-BLE-MESH";
 static char device_name[DEVICE_NAME_SIZE + 1];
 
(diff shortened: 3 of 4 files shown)


forx157:
Currently this part of the functionality is already available, you can learn about the API: `esp_ble_mesh_set_unprovisioned_device_name`，Thank you for your contribution.

s0be:
@forx157 Unfortunately `esp_ble_mesh_set_unprovisioned_device_name`, is insufficient.  When you call `esp_ble_mesh_init` it leads to a call to `bt_mesh_proxy_server_init`, which explicitly sets the name to "ESP-BLE-MESH", and starts advertising.  If an iOS devices sees "ESP-BLE-MESH", it caches that name, so calling `esp_ble_mesh_set_unprovisioned_device_name` will not show up on the iOS device.  The goal of this patch is to give the esp-idf APP the opportunity to set what the device name is before the first advertisement can be sent.  iOS apps do not have the ability to flush the cache, so the only recovery path if you show the user the advertising name is to reboot the phone.

s0be:
For context, here's the change we made to our app when introducing this patch and it resolved seeing 'ESP-BLE-MESH' as the name:

```
+    err = esp_ble_mesh_set_default_unprovisioned_device_name(deako_provisioning_get_serial());
+    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
+
     err = esp_ble_mesh_init(get_provisioning_data(), get_composition_data());
     RETURN_ON_ESP_ERROR(err, TAG, "Failed to initialize mesh stack");
 
-    err = esp_ble_mesh_set_unprovisioned_device_name(deako_provisioning_get_serial());
-    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
-
```

s0be:
> Currently this part of the functionality is already available, you can learn about the API: `esp_ble_mesh_set_unprovisioned_device_name`，Thank you for your contribution.

Any chance you can comment on if there's a better way for us to solve the problem this patch solves?  We're fine with carrying a patch to cover deficiencies in esp-idf, but I'm sure we're not the only customer who has or will see this problem.

s0be:
@forx157 Friendly ping.  Can you review my comments here?  The existing API is not sufficient for our needs for the reasons outlined.

forx157:
Hi, @s0be 
Sorry for the delayed reply. Would it meet your needs if we made the DEFAULT_DEVICE_NAME settable via Kconfig?

s0be:
Sadly no, we use our device's serial number as the BLE Gatt Name, and need that as part of our pairing flow.

forx157:
Hi, @s0be 

The following code sequence for mesh initialization may meet your needs:
```
esp_ble_mesh_init
esp_ble_mesh_set_unprovisioned_device_name
esp_ble_mesh_prov_enable
```
The node does not start advertising immediately after esp_ble_mesh_init is called, but only after esp_ble_mesh_prov_enable is invoked. Thus, changing the device name before advertising starts can prevent iOS cache issues.

s0be:
> For context, here's the change we made to our app when introducing this patch and it resolved seeing 'ESP-BLE-MESH' as the name:
> 
> ```
> +    err = esp_ble_mesh_set_default_unprovisioned_device_name(deako_provisioning_get_serial());
> +    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
> +
>      err = esp_ble_mesh_init(get_provisioning_data(), get_composition_data());
>      RETURN_ON_ESP_ERROR(err, TAG, "Failed to initialize mesh stack");
>  
> -    err = esp_ble_mesh_set_unprovisioned_device_name(deako_provisioning_get_serial());
> -    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
> -
> ```

Sadly, that calling path was not sufficient, as I called out above.  A small percentage of the time, the BLE stack would send out 1 advertisement with 'ESP-BLE-MESH' which would then get seen by iOS and cached.

forx157:
> > For context, here's the change we made to our app when introducing this patch and it resolved seeing 'ESP-BLE-MESH' as the name:
> > ```
> > +    err = esp_ble_mesh_set_default_unprovisioned_device_name(deako_provisioning_get_serial());
> > +    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
> > +
> >      err = esp_ble_mesh_init(get_provisioning_data(), get_composition_data());
> >      RETURN_ON_ESP_ERROR(err, TAG, "Failed to initialize mesh stack");
> >  
> > -    err = esp_ble_mesh_set_unprovisioned_device_name(deako_provisioning_get_serial());
> > -    RETURN_ON_ESP_ERROR(err, TAG, "esp_ble_mesh_set_unprovisioned_device_name failed");
> > -
> > ```
> 
> Sadly, that calling path was not sufficient, as I called out above. A small percentage of the time, the BLE stack would send out 1 advertisement with 'ESP-BLE-MESH' which would then get seen by iOS and cached.

Theoretically, this situation shouldn't occur. I haven't encountered this issue in my local tests. Would you mind sharing your testing method and logs?

PR #15661, opened 2025-03-26
feat(bod): allow overriding brownout handling (IDFGH-14954)

bryghtlabs-richard:
## Description

Weakly-link esp_brownout_init() so that a project may override it with their own implementation installed at the same point in startup.

Sometimes a device needs a brownout handler to trigger some quick hardware maintenance like setting or clearing a pin, and needs the handler setup early.

## Testing

Tested on ESP32-S3 by replacing `esp_brownout_init` and `rtc_brownout_isr_handler` with my own, then setting BOD level high so CPU still runs ok but triggers quickly, then verifying the new handler is hit.

---

## Checklist


suda-morris:
sha=7511742ecd6b64e325159bffb15008f7f257db6f

PR #15692, opened 2025-03-31
change(esp_wifi): Improve handling group parameter A in H2E (IDFGH-14995)

### LINKED ISSUE #15691: wpa_supplicant SAE H2E related porting could be improved (IDFGH-14994)
### Answers checklist.

### General issue report

The current SAE H2E sswu() function is using hard coded ECC domain parameter A for P-256. This is neither generic nor elegant.
However, we could improve it by adding a new API for getting A. The API handles Mbed TLS not storing A in it's ECP group structure correctly.
The newly added API will also help supporting other ECP groups(e.g. group 20,21) in the future.
The only drawback I can think of is that it introduces a slight increase in code size.
Plus, other issues were spotted when I was working on a fix. The most significant concern is the absence of validation checks for the loading of A.

### Proposed fix
See #15692 .

ivq:
## Description

Fixes #15691 

## Testing

It was tested that ESP32-C6 could associate to a WPA3 H2E-only AP successfully.

---

## Checklist

### DIFF
--- a/components/wpa_supplicant/esp_supplicant/src/crypto/crypto_mbedtls-ec.c
+++ b/components/wpa_supplicant/esp_supplicant/src/crypto/crypto_mbedtls-ec.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -147,6 +147,42 @@ const struct crypto_bignum *crypto_ec_get_prime(struct crypto_ec *e)
 const struct crypto_bignum *crypto_ec_get_order(struct crypto_ec *e)
 {
     return (const struct crypto_bignum *) & ((mbedtls_ecp_group *)e)->N;
+}
+
+struct crypto_bignum * crypto_ec_get_a(struct crypto_ec *e)
+{
+    int ret = -1;
+    struct crypto_bignum *a;
+    mbedtls_mpi *m_a;
+    mbedtls_ecp_group *grp = (mbedtls_ecp_group *)e;
+    if (mbedtls_ecp_get_type(grp) != MBEDTLS_ECP_TYPE_SHORT_WEIERSTRASS) {
+        return NULL;
+    }
+    a = crypto_bignum_init();
+    if (!a) {
+        return NULL;
+    }
+    m_a = (mbedtls_mpi *)a;
+    /* Handle Mbed TLS quirk.
+     *
+     * Mbed TLS default ECP implementation is using grp->A = NULL to represent A = -3 for
+     * Short Weierstrass curves(e.g. P-256) thus accessing A needs some tweaking.
+     *
+     * See mbedtls/ecp.h for details. */
+#ifdef MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED
+    if (mbedtls_ecp_group_a_is_minus_3(grp)) {
+        MBEDTLS_MPI_CHK(mbedtls_mpi_sub_int(m_a, &grp->P, 3));
+    } else {
+        MBEDTLS_MPI_CHK(mbedtls_mpi_copy(m_a, &grp->A));
+    }
+#else
+    goto cleanup;
+#endif
+    return a;
+
+cleanup:
+    crypto_bignum_deinit(a, 0);
+    return NULL;
 }
 
 const struct crypto_bignum * crypto_ec_get_b(struct crypto_ec *e)
--- a/components/wpa_supplicant/src/common/sae.c
+++ b/components/wpa_supplicant/src/common/sae.c
@@ -594,9 +594,7 @@ static struct crypto_ec_point * sswu(struct crypto_ec *ec, int group,
 
 	prime = crypto_ec_get_prime(ec);
 	prime_len = crypto_ec_prime_len(ec);
-	/* Value of 'a' defined for curve secp256r1 in 'y^2 = x^3 + ax + b' */
-	uint8_t buf[32] = {0xff,0xff,0xff,0xff,0x00,0x00,0x00,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xfc};
-	a = crypto_bignum_init_set(buf, 32);
+	a = crypto_ec_get_a(ec);
 	b = crypto_ec_get_b(ec);
 
 	u2 = crypto_bignum_init();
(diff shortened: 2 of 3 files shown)


components/wpa_supplicant/esp_supplicant/src/crypto/crypto_mbedtls-ec.c
```diff
@@ -149,6 +149,46 @@ const struct crypto_bignum *crypto_ec_get_order(struct crypto_ec *e)
...
+        goto cleanup;
+#endif
+    }
+    return a;
+
+cleanup:
+    crypto_bignum_deinit(a, 0);
+    return NULL;
+}
+
```
kapilkedawat, line 191:
```suggestion
struct crypto_bignum * crypto_ec_get_a(struct crypto_ec *e)
{
    int ret = -1;
    struct crypto_bignum *a;
    mbedtls_mpi *m_a;
    mbedtls_ecp_group *grp = (mbedtls_ecp_group *)e;

    if (mbedtls_ecp_get_type(grp) != MBEDTLS_ECP_TYPE_SHORT_WEIERSTRASS) {
        return NULL;
    }
    a = crypto_bignum_init();
    if (!a) {
        return NULL;
    }
    m_a = (mbedtls_mpi *)a;

    /* Handle Mbed TLS quirk.
     *
     * Mbed TLS default ECP implementation is using grp->A = NULL to represent A = -3 for
     * Short Weierstrass curves(e.g. P-256) thus accessing A needs some tweaking.
     *
     * See mbedtls/ecp.h for details. */
#ifdef MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED
        if (mbedtls_ecp_group_a_is_minus_3(grp)) {
            MBEDTLS_MPI_CHK(mbedtls_mpi_sub_int(m_a, &grp->P, 3));
        } else {
            MBEDTLS_MPI_CHK(mbedtls_mpi_copy(m_a, &grp->A));
        }
#endif
    }
    return a;

cleanup:
    crypto_bignum_deinit(a, 0);
    return NULL;
}

```

components/wpa_supplicant/src/common/sae.c
```diff
@@ -594,9 +594,7 @@ static struct crypto_ec_point * sswu(struct crypto_ec *ec, int group,
 
 	prime = crypto_ec_get_prime(ec);
 	prime_len = crypto_ec_prime_len(ec);
-	/* Value of 'a' defined for curve secp256r1 in 'y^2 = x^3 + ax + b' */
-	uint8_t buf[32] = {0xff,0xff,0xff,0xff,0x00,0x00,0x00,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xfc};
-	a = crypto_bignum_init_set(buf, 32);
+	a = crypto_ec_get_a_allocated(ec);
```
kapilkedawat, line 597:
```suggestion
	a = crypto_ec_get_b(ec);
```

let's keep the function name similar to upstream https://w1.fi/cgit/hostap/tree/src/crypto/crypto.h#n787

components/wpa_supplicant/src/crypto/crypto.h
```diff
@@ -796,6 +796,14 @@ const struct crypto_bignum * crypto_ec_get_order(struct crypto_ec *e);
  * Internal data structure for EC implementation to represent a point. The
  * contents is specific to the used crypto library.
  */
+struct crypto_ec_point;
```
kapilkedawat, line 799: no need to move this.

kapilkedawat:
Hi @ivq suggested some cleanup for the changes and could you also squash the commits into one.

As always thanks for your contribution for making this project better.

components/wpa_supplicant/esp_supplicant/src/crypto/crypto_mbedtls-ec.c
ivq: Nice point. Mbed TLS already provides API for checking curve type.

components/wpa_supplicant/src/common/sae.c
ivq: OK. Care should be taken when using the API: don't forget to free it.

components/wpa_supplicant/src/crypto/crypto.h
ivq: Please note that the comment for crypto_ec_point is not surrounding it.

components/wpa_supplicant/src/common/sae.c
kapilkedawat: actually that code should be added in `crypto_ec_deinit()`

ivq: > actually that code should be added in `crypto_ec_deinit()`

You mean embed the A in struct crypto_ec?

kapilkedawat: On second thought, we can do this as well
```
const struct crypto_bignum * crypto_ec_get_a(struct crypto_ec *e)
{
    mbedtls_ecp_group *grp = (mbedtls_ecp_group *)e; 
#ifdef MBEDTLS_ECP_SHORT_WEIERSTRASS_ENABLED
    if (mbedtls_ecp_get_type(grp) != MBEDTLS_ECP_TYPE_SHORT_WEIERSTRASS) {
        return NULL;
    }    
    /* Handle Mbed TLS quirk.
     *
     * Mbed TLS default ECP implementation is using grp->A = NULL to represent A = -3 for
     * Short Weierstrass curves(e.g. P-256) thus accessing A needs some tweaking.
     *
     * See mbedtls/ecp.h for details. */
     if (mbedtls_ecp_group_a_is_minus_3(grp)) {
         mbedtls_mpi_sub_int(&grp->A, &grp->P, 3);
     }    
#endif
    return (const struct crypto_bignum *) (&grp->A);
}
```
and keep `sswu()` implementation same as upstream.

ivq: This may not be a good solution. Doing so would interfere with the internal logic of MbedTLS, and the MbedTLS maintainers also do not approve of this approach. See https://github.com/Mbed-TLS/mbedtls/pull/6999#discussion_r1286697700.
Maybe we could embed parameter A in struct crypto_ec like this:
```c
struct crypto_ec {
    /* group must be the first member */
    mbedtls_ecp_group group;
    /* valid only if group is of Short Weierstrass type */
    mbedtls_mpi group_parameter_a;
};
```
And load A in crypto_ec_init(), while free it in crypto_ec_deinit(). In this way, sswu() can be kept same as upstream
and crypto_ec_get_a() can behave the same as upstream. This seems more robust.  However, this will introduce some unnecessary runtime memory overhead when not doing H2E operation.

kapilkedawat: Sorry, I missed this. Currently changes looks okay IMO, will raise internal MR. I don't want to embed A in ec struct since that will unnecessary increase size.

ivq: OK

PR #15759, opened 2025-04-10
Enable ECDSA support on ESP32-P4 (IDFGH-15077)

renzenicolai:
## Description

Enable ECDSA support for ESP32-P4 target. The ESP32-P4 has the hardware required for this feature but the feature is currently disabled in the configuration, probably because of problems with earlier chip revisions.

The ECDSA peripheral appears to work on a version 1.0 ESP32-P4 chip, it is able to generate valid signatures from a key burned into efuse. However I'm not a cryptographic expert by any means so if this feature was intentionally disabled because of hardware problems still present in version 1.0 of the ESP32-P4 then I would like to be informed on what these issues are and what limitations are imposed by them.

## Testing

---

## Checklist

### DIFF
--- a/components/soc/esp32p4/include/soc/Kconfig.soc_caps.in
+++ b/components/soc/esp32p4/include/soc/Kconfig.soc_caps.in
@@ -201,11 +201,11 @@ config SOC_ECC_EXTENDED_MODES_SUPPORTED
 
 config SOC_ECDSA_SUPPORTED
     bool
-    default n
+    default y
 
 config SOC_KEY_MANAGER_SUPPORTED
     bool
-    default n
+    default y
 
 config SOC_FLASH_ENC_SUPPORTED
     bool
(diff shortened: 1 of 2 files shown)


AdityaHPatwardhan:
Hi @renzenicolai,
The ECDSA peripheral on ESP32-P4 is also affected by the advisory mentioned in [ECDSA advisory for ESP32-H2](https://www.espressif.com/sites/default/files/advisory_downloads/ar2024-007_Security%20Advisory%20Concerning%20Timing%20Attacks%20on%20ECDSA%20Peripheral%20in%20ESP32-H2%20EN.pdf) hence we have chosen to keep the hardware support disabled from the start, in order to protect users from being vulnerable to the attack.

We will enable it in ESP32-P4 when a new version of the chip is available which contains the hardware fix in the ECDSA peripheral.

renzenicolai:
Okay, makes sense. For my application the lower security level provided by the current version of the chip is sufficient so I will just add a .patch file for re-enabling the functionality to my project as a workaround. I'm happy to hear that base functionality (generating the signatures) is working normally.

PR #15837, opened 2025-04-23
fix(ulp): Add missing wake up enable and disable functions for LP core

### LINKED ISSUE #15732: RISC-V ULP wake-up timer and gpio (IDFGH-15044)
### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like.

Instead of being able to select only ```ULP_RISCV_WAKEUP_SOURCE_GPIO``` or only ```ULP_RISCV_WAKEUP_SOURCE_TIMER```, I'd like to use both of them at the same time to wakeup periodically and when an interrupt occurs.

```c
ulp_riscv_cfg_t ulp_cfg = {
    .wakeup_source = ULP_RISCV_WAKEUP_SOURCE_GPIO | ULP_RISCV_WAKEUP_SOURCE_TIMER,
};
```

### Describe alternatives you've considered.

_No response_

### Additional context.

This is useful for battery monitoring purposes. On interrupt, we decide if we want to wake-up or not and enter shipping mode (very low power ~10uA) if the battery is nearly empty. We also want to do so periodically, like every hour.

X-Ryl669:
## Description

As the title says, this add the missing wake enable and disable function so it can be called from ULP's LP core code directly without resorting to ugly `rtc_ll` class of functions.

## Related

Truely fixes #15732 (not what the OP did)

## Testing

Tested on ESP32C6 (see related issue for example testing code)
---

## Checklist


sudeep-mohanty:
Hi @X-Ryl669,
Thanks for your contribution. The changes generally look good. However, I would like to understand your use case a bit more. May I know why would you need to configure the GPIO wakeup enable/disable feature from the LP core. Wouldn't configuring it from the HP CPU also suffice? Like shown in this [example](https://github.com/espressif/esp-idf/blob/465b159cd8771ffab6be70c7675ecf6705b62649/examples/system/ulp/lp_core/gpio_wakeup/main/lp_core_gpio_wake_up_example_main.c#L37).

X-Ryl669:
My usage case is simple: 
- My system is running from a supercapacitor and a solar charging circuit so I don't have any margin in terms of power
- The HP Core is doing the WIFI/Zigbee communication stuff, so it's off deep sleeping most of the time, else I don't have more than 2h of run time with the capacitor charged.
- Some open/close cycles are scheduled, so the HP core either sleep until the open time happened. There's only one *possible* wake up: the timer
- The LP core wakes up, reprogram the timer for a timeout time and trigger the closing of the valve, it also program the "valve closed" IO to wakeup too, it then halts (deep sleep)
- Either the IO wake up or the timer fires. In the former case, the LP core disable the wake up for the IO and reprogram the timer to the next communication event. In the latter case, it's an error, the valve should have closed in time so it wakes up the HP core to report it.
- Finally the LP core is woken up again from deep sleep and it wakes up the HP core for communication.

Here, you seen most of the time is spent with the lowest possible consumption, we don't want to wake up the HP core to re-configure the next wake source as it's a huge power hog compared to the LP core (and it's slow, it takes hundred of ms to perform). Doing everything in the LP core and having the LP core halted gives the minimum consumption we are looking for. We've tried with LP core in wait for interrupt mode (it works), but it consume more than 10x more current than while deep sleeping.
We don't care to preserve the program counter/state of the LP core, it can restart from scratch for each wakeup since the state is preserved in the RTC RAM variables anyway.

Currently, we are using the low level functions, but there's no actual reason why it would be required.

X-Ryl669:
Notice that the OP in the initial issue has another common use case: the LP core monitors the battery level and the "power on" button. It's likely desired to disable the IO wakeup when the battery level is below a given threshold to prevent the device from powering on in that case.

components/ulp/lp_core/lp_core/include/ulp_lp_core_gpio.h
```diff
@@ -200,6 +200,27 @@ static inline void ulp_lp_core_gpio_clear_intr_status(void)
...
+ */
+static inline void ulp_lp_core_gpio_wakeup_enable(lp_io_num_t lp_io_num, gpio_int_type_t intr_type)
+{
+    rtcio_ll_wakeup_enable(lp_io_num, intr_type);
+}
+
+/**
+ * @brief Disable wake up for lp io pin
+ *
+ * @param lp_io_num The lp io pin to enable interrupt for
```
sudeep-mohanty, line 217:
```suggestion
 * @param lp_io_num The lp io pin to disable the wake up for
```

sudeep-mohanty:
Thanks for explaining the use-case @X-Ryl669! The changes LGTM. I left a couple of minor nitpicks. If you have the time to fix them then I'll wait for it before pulling in the changes and running it through our tests. Thanks!

sudeep-mohanty (approved):

sudeep-mohanty:
sha=548d253bcb9a4d701ef14e3c2190e4b93cdc612f

X-Ryl669:
Thanks for merging.

PR #15917, opened 2025-05-07
Small clarifications to the ULP I2C peripheral documentation about what is possible (and what's not). (IDFGH-15254)

eriksl:
This clarifies that the limited set of possible transactions using the ULP I2C peripheral is a hardware limitation and not a limitation of the ESP-IDF and exactly what and how it's limited.

### DIFF
--- a/docs/en/api-reference/system/ulp-risc-v.rst
+++ b/docs/en/api-reference/system/ulp-risc-v.rst
@@ -242,7 +242,7 @@ Once the RTC I2C controller is initialized, the I2C slave device address must be
 
 .. note::
 
-    The RTC I2C peripheral always expects a slave sub-register address to be programmed via the :cpp:func:`ulp_riscv_i2c_master_set_slave_reg_addr` API. If it is not, the I2C peripheral uses the ``SENS_SAR_I2C_CTRL_REG[18:11]`` as the sub-register address for the subsequent read or write operations. This could make the RTC I2C peripheral incompatible with certain I2C devices or sensors which do not need any sub-register to be programmed.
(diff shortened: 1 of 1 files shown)


docs/en/api-reference/system/ulp-risc-v.rst
```diff
@@ -242,7 +242,7 @@ Once the RTC I2C controller is initialized, the I2C slave device address must be
...
+    The RTC I2C peripheral can issue two kinds of I2C transactions: 1) READ: [start]->write one byte ("register")->[repeated start]->read N bytes->[stop] and 2) WRITE: [start]->write one byte ("register")->write N bytes->[stop]. Sending of the "register" byte is required and unavoidable. As such the peripheral always expects a slave sub-register address to be programmed via the :cpp:func:`ulp_riscv_i2c_master_set_slave_reg_addr` API. If it is not, the I2C peripheral uses the ``SENS_SAR_I2C_CTRL_REG[18:11]`` as the sub-register address for the subsequent read or write operations. This makes the RTC I2C peripheral incompatible with certain I2C devices or sensors which require a READ transaction without a WRITE transaction or which need a register pointer of 16 bits or wider.
```
KonstantinKondrashov, line 245: I suggest to reformat this note even a bit more. 

```suggestion
    The RTC I2C peripheral issues two kinds of I2C transactions: 

    1) **READ**: [start] → write one byte ("register") → [repeated start] → read N bytes → [stop]  
    2) **WRITE**: [start] → write one byte ("register") → write N bytes → [stop]  
    
    In both cases, sending the "register" byte is required and cannot be disabled. Therefore, the peripheral always expects a slave sub-register address to be set using the :cpp:func:`ulp_riscv_i2c_master_set_slave_reg_addr` API.  
    If it is not set explicitly, the peripheral uses the value in ``SENS_SAR_I2C_CTRL_REG[18:11]`` as the sub-register address for subsequent transactions.
    
    This behavior makes the RTC I2C peripheral incompatible with:
    - Devices that expect a read transaction without a preceding "register" address write.
    - Devices requiring 16-bit or wider register addresses, since only 8-bit addressing is supported.
```

KonstantinKondrashov:
LGTM! I will apply the suggestion for you, if you do not mind. Thanks for the contribution.

KonstantinKondrashov:
sha=3841ae11abdb418d24a2c106a2253960696cf9fc

docs/en/api-reference/system/ulp-risc-v.rst
eriksl: I am totally fine as long as the clarification remains (i.e. what's possible and what's not).

KonstantinKondrashov: Good, I already took this PR in our internal repo. It is in review now.

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/1fa908a7a56af8c8c1d7123e5f6dc9cd6159faf0

KaeLL:
@KonstantinKondrashov 
At the risk of being too late, would you mind adding something similar to vanilla ESP32 ULP docs?

PR #15939, opened 2025-05-12
feat(ota-mqtt): simple example to make an ota over mqtt

juan-lopez-pragmafw:
## Description

This example demonstrates how to perform a simple OTA update over MQTT. The ESP32 connects to Wi-Fi and then to an MQTT broker. A script sends a start signal (e.g., "start") to a topic like ota/update. The ESP32 then listens for firmware data chunks on another topic (e.g., ota/firmware). Each chunk it receives is written to the OTA partition. Once all data has been received, a done signal (e.g., "done") is sent to another topic (e.g., ota/done). The ESP32 then finalizes the update, verifies the firmware, and reboots if the update was successful.

The example includes a script to send the .bin over mqtt to the device.

## Related

This contribution is not related to any current issue.

## Testing

Tested on ESP32 and ESP32-C3 with a local broker, private broker, and public broker.
Note: When using a public broker, message drops may occur under heavy load.

---

## Checklist

### DIFF
--- a/examples/system/ota/simple_mqtt_ota/main/simple_mqtt_ota_example.c
+++ b/examples/system/ota/simple_mqtt_ota/main/simple_mqtt_ota_example.c
@@ -1,0 +1,325 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Unlicense OR CC0-1.0
+ */
+#include "esp_system.h"
+#include "esp_log.h"
+#include "esp_ota_ops.h"
+#include "esp_wifi.h"
+#include "esp_timer.h"
+#include "nvs_flash.h"
+#include "mqtt_client.h"
+#include <string.h>
+#include "protocol_examples_common.h"
+
+#define MQTT_BROKER_URI CONFIG_MQTT_BROKER_URI
+#define TOPIC_FIRMWARE CONFIG_MQTT_TOPIC_FIRMWARE
+#define TOPIC_START_UPDATE CONFIG_MQTT_TOPIC_START_UPDATE
+#define TOPIC_DONE CONFIG_MQTT_TOPIC_DONE
+
+#define HASH_LEN 32
+
+static const char *TAG = "simple_mqtt_ota_example";
+size_t total_bytes_written = 0;
+
+// Global OTA handle
+esp_ota_handle_t ota_handle = 0;
+bool ota_in_progress = false;
+bool ota_data_written = false;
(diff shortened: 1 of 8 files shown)


examples/system/ota/simple_mqtt_ota/main/simple_mqtt_ota_example.c
```diff
@@ -0,0 +1,325 @@
...
+        else if (event->topic_len == strlen(TOPIC_FIRMWARE) &&
+                 strncmp(event->topic, TOPIC_FIRMWARE, event->topic_len) == 0)
+        {
+            if (!ota_in_progress)
+            {
+                ESP_LOGE(TAG, "OTA not started yet. Ignoring firmware chunk.");
+                return;
+            }
+
+            esp_err_t err = esp_ota_write(ota_handle, event->data, event->data_len);
```
Kriechi, line 196: This assumes that each MQTT message arrives in order, which is not guaranteed by the broker.
I suggest to add a payload header of some sort to indicate which location or sequence number this message belongs to, so its payload data bytes can be written to the correct location, see `esp_ota_write_with_offset ` for this use case: https://docs.espressif.com/projects/esp-idf/en/v5.5.1/esp32c6/api-reference/system/ota.html#_CPPv425esp_ota_write_with_offset16esp_ota_handle_tPKv6size_t8uint32_t

euripedesrocha:
Hi @juan-lopez-pragmafw thank you for your contribution. 
Unfortunately, we'll not be taking the newly added example to esp-idf. There are some reasons for that, but the main one is that we already have some examples that cover the details of OTA.  We appreciate the suggestion and effort on putting this together.

KaeLL:
@euripedesrocha 
> We appreciate the suggestion and effort on putting this together.

How about showing that by elaborating the following further?
> There are some reasons for that

It could help others in the future to decide whether to bother or not.

euripedesrocha:
@KaeLL sure I'll try to elaborate. 
OTA is a feature that requires extra test and validation to provide an example. So there will be a lot of maintenance effort from our side on that. 
We already have some examples showing how to use ota related features in IDF. 
esp-mqtt, and related examples, moved from idf to esp-mqtt repository, it is now a managed component. 
Using mqtt as protocol for OTA is something that likely have many system-wide decisions, I must confess here that I would need to spend some time to consider a generalized way to do it, that might impact the example. The comment regarding message ordering is one aspect of that. 

At the moment of evaluation I considered that, unfortunately, adding this example would bring us more trouble than it will be providing guidance for users. If I'm wrong in that assumption, which is of course possible, I'm open to reconsider and either discuss how to work to make this in shape to be accepted or to provide a different path to take the contribution.

Issue #15964, opened 2025-05-16
`idf.py create-project` would fail on linux if idf directory is readonly (IDFGH-15305)


Sped0n:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.4.1-674-gfe75355314

### Operating System used.

Linux

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### What is the expected behavior?

the `test_common.py::test_create_project_with_idf_readonly` should pass

related issue/pr:

- https://github.com/mirrexagon/nixpkgs-esp-dev/issues/75
- https://github.com/espressif/esp-idf/pull/11836

### What is the actual behavior?

https://github.com/mirrexagon/nixpkgs-esp-dev/issues/75

### Steps to reproduce.

1. `docker run --rm -w /opt/esp/idf -u root -it espressif/idf:release-v5.4`, enter the container
2. `chown -R ubuntu:ubuntu .`
3. `su - ubuntu`
4. `cd /opt/esp/idf && ./install.sh --enable-pytest && . ./export.sh`
5. `cd /opt/esp/idf/tools/test_build_system && pytest -k test_create_project`

### Build or installation Logs.

```plain
pytest log:

> using docker image `espressif/idf:release-v5.4`, running test with user `ubuntu`


ubuntu@6f3b46503c73:/opt/esp/idf/tools/test_build_system$ pytest -k test_create_project
================================================================================ test session starts =================================================================================
platform linux -- Python 3.12.3, pytest-8.3.5, pluggy-1.6.0
rootdir: /opt/esp/idf/tools/test_build_system
configfile: pytest.ini
plugins: embedded-1.16.1, ignore-test-results-0.3.0, timeout-2.4.0, rerunfailures-15.1
collected 102 items / 100 deselected / 2 selected

test_common.py::test_create_project
----------------------------------------------------------------------------------- live log call ------------------------------------------------------------------------------------
2025-05-16 06:47:41 INFO Check that command for creating new project will fail if the target folder is not empty.
2025-05-16 06:47:42 INFO Check that command for creating new project will fail if the target path is file.
PASSED                                                                                                                                                                         [ 50%]
[...]

PR #16021, opened 2025-05-24
fix: switch to custom copytree for readonly source handling

Sped0n:
## Description

Replace standard `copytree` with a custom implementation that handles readonly source.

## Related

- #15964
- https://github.com/mirrexagon/nixpkgs-esp-dev/issues/75

## Testing

Same as step to reproduce section in #15964, but with the [patch](https://gist.github.com/Sped0n/a7c3bc67c071ecdf8064cf565bdf1a39) applied. (map the patch into docker container then `git apply`).

---

## Checklist

### DIFF
--- a/tools/idf_py_actions/create_ext.py
+++ b/tools/idf_py_actions/create_ext.py
@@ -2,13 +2,57 @@
 # SPDX-License-Identifier: Apache-2.0
 import os
 import re
+import shutil
 import sys
-from shutil import copyfile
-from shutil import copytree
 from typing import Dict
 
 import click
 from idf_py_actions.tools import PropertyDict
+
+# See https://github.com/espressif/esp-idf/issues/15964
+def custom_copytree(src: str, dst: str, dirs_exist_ok: bool = False) -> str:
+    if not os.path.exists(src):
+        raise FileNotFoundError(f"Source directory '{src}' does not exist.")
+    if not os.path.isdir(src):
+        raise NotADirectoryError(f"Source path '{src}' is not a directory.")
+    if (not dirs_exist_ok) and os.path.exists(dst) and os.listdir(dst):
+        raise FileExistsError(
+            f"Destination directory '{dst}' already exists and is not empty and dirs_exist_ok is False."
+        )
+
+    try:
+        os.makedirs(dst, exist_ok=True)
+    except FileExistsError:
+        raise NotADirectoryError(f"Destination path '{dst}' exists and is not a directory.")
+    except OSError as e:
+        raise shutil.Error(f"Could not create destination directory '{dst}': {e}")
+
+    names = os.listdir(src)
+    errors = []
+    for name in names:
+        srcname = os.path.join(src, name)
+        dstname = os.path.join(dst, name)
+        try:
+            if os.path.islink(srcname):
+                if os.path.isdir(srcname):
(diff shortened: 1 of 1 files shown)


mfialaf:
Hi @Sped0n,

Thanks again for your Pull Request. 
After reviewing and discussing internally, we’ve decided to go with an alternative solution: keeping shutil.copytree with copyfile, and updating the destination directory permissions afterward. This approach minimizes the risk of introducing unexpected behavior from reimplementing core functionality.

I tested this approach using your Docker-based reproduction steps, and the fix works as expected.

The solution would be something like this:
``` diff
--- a/tools/idf_py_actions/create_ext.py
+++ b/tools/idf_py_actions/create_ext.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import os
 import re
+import stat
 import sys
 from shutil import copyfile
 from shutil import copytree
@@ -47,14 +48,37 @@ def is_empty_and_create(path: str, action: str) -> None:
         sys.exit(3)
 
 
+def make_directory_permissions_writable(root_path: str) -> None:
+    """
+    Ensures all directories under `path` have write permission for the owner.
+    Skips files and doesn't override existing permissions unnecessarily.
+    Only applies to POSIX systems (Linux/macOS).
+    """
+    if sys.platform == 'win32':
+        return
+
+    for current_root, dirs, _ in os.walk(root_path):
+        for dirname in dirs:
+            dir_path = os.path.join(current_root, dirname)
+            try:
+                current_perm = stat.S_IMODE(os.stat(dir_path).st_mode)
+                new_perm = current_perm | 0o200  # mask permission for owner (write)
+                if new_perm != current_perm:
+                    os.chmod(dir_path, new_perm)
+            except PermissionError:
+                continue
+
+
 def create_project(target_path: str, name: str) -> None:
     copytree(
         os.path.join(os.environ['IDF_PATH'], 'tools', 'templates', 'sample_project'),
         target_path,
         copy_function=copyfile,
         dirs_exist_ok=True,
     )
+    # since 'copyfile' does preserve directory metadata, we need to make sure the directories are writable
+    make_directory_permissions_writable(target_path)
     main_folder = os.path.join(target_path, 'main')
     os.rename(os.path.join(main_folder, 'main.c'), os.path.join(main_folder, '.'.join((name, 'c'))))
     replace_in_file(os.path.join(main_folder, 'CMakeLists.txt'), 'main', name)
@@ -65,10 +89,12 @@ def create_component(target_path: str, name: str) -> None:
     copytree(
         os.path.join(os.environ['IDF_PATH'], 'tools', 'templates', 'sample_component'),
         target_path,
         copy_function=copyfile,
         dirs_exist_ok=True,
     )
+    # since 'copyfile' does preserve directory metadata, we need to make sure the directories are writable
+    make_directory_permissions_writable(target_path)
     os.rename(os.path.join(target_path, 'main.c'), os.path.join(target_path, '.'.join((name, 'c'))))
     os.rename(
         os.path.join(target_path, 'include', 'main.h'), os.path.join(target_path, 'include', '.'.join((name, 'h')))
```
I hope this solution works for you as well. Let me know if there's anything else I can help with.

Sped0n:
Hi @mfialaf, im good with it, as long as the problem is solved 👍

PR #16047, opened 2025-05-29
fix: detect ksh variants as ksh

tekktonic:
## Description
`ksh` variants are a perfectly usable shell (most notably OpenBSD `ksh` but I also suspect `mksh` works fine,) but they often aren't named `ksh` explicitly to avoid possible conflicts with `ksh93`. This is definitely some special casing but I'm also not aware of another shell whose name isn't usually its name.

## Testing
Fresh build/install on ksh (`oksh` on my system.) 

---

## Checklist

- [x] Documentation is updated as needed. (no updates necessary)


tekktonic:
Ticket was already responded to while I was asleep, a different fix is being implemented.

PR #16064, opened 2025-06-02
feat(freertos): Include pxStack in xTASK_SNAPSHOT

nebkat:
## Description

Adds `pxStack` to task snapshots to allow calculating the size of the stack (`pxEndOfStack - pxStack`).

---

## Checklist

### DIFF
--- a/components/freertos/esp_additions/freertos_tasks_c_additions.h
+++ b/components/freertos/esp_additions/freertos_tasks_c_additions.h
@@ -1051,6 +1051,7 @@ BaseType_t vTaskGetSnapshot( TaskHandle_t pxTask,
 
     TCB_t * pxTCB = ( TCB_t * ) pxTask;
     pxTaskSnapshot->pxTCB = pxTCB;
+    pxTaskSnapshot->pxStack      = ( StackType_t * ) pxTCB->pxStack;
     pxTaskSnapshot->pxTopOfStack = ( StackType_t * ) pxTCB->pxTopOfStack;
     pxTaskSnapshot->pxEndOfStack = ( StackType_t * ) pxTCB->pxEndOfStack;
     return pdTRUE;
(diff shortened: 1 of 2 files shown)


sudeep-mohanty:
Hello @nebkat,
Thank you for your contribution. Would you mind sharing a bit more about your use case for programmatically determining the stack size of a task? Since the stack size is typically specified at the time of task creation, it’s generally known to the application.

nebkat:
@sudeep-mohanty While the task size can be determined from the source code I find it is often not a trivial task (excuse the pun).

By the time one is running into heap issues and having to investigate task stack usage there may be 10+ tasks defined, some of which are outside of user code (timers, lwip, bluetooth). It is then necessary to grep for `xTaskCreate`, follow various macros or functions back to their respective `CONFIG_..._STACK_DEPTH` entries. 

This addition makes it very simple to see at a glance which tasks are worth focusing on to reduce heap usage. In our application we have even created a tool that visualizes this for us:

image

As an alternative this could be placed in a dedicated `uxTaskGetStackDepth(TaskHandle_t)` function to keep `TaskSnapshot_t` unchanged.

sudeep-mohanty:
Thanks for sharing the use case. The changes are overall good. We shall discuss internally on this change and let you know if we can pull this in. As an alternative, you can still get the stack info programmatically from the `TaskSnapshot_t` struct, albeit in a less-straightforward way like this -

```c
    TaskSnapshot_t xTaskSnapshot;
    vTaskGetSnapshot(xTaskGetCurrentTaskHandle(), &xTaskSnapshot);
    printf("Task pxStack: %p\n", ((StaticTask_t *)xTaskSnapshot.pxTCB)->pxDummy6);  // pxDummy6 corresponds to pxStack in the StaticTask_t struct
```

PR #16119, opened 2025-06-12
feat(ws_transport): add header callback hook (IDFGH-15479)

bryghtlabs-richard:
## Description

Add a callback for each line of header responses in the Websocket HTTP->WS transport upgrade. This is necessary to support AWS ALB and F5 cookie-based load-balancer sticky-sessions.

## Related

Please see https://github.com/espressif/esp-protocols/issues/715 . There is another part of this patchset https://github.com/espressif/esp-protocols/pull/827, as well as a different implementation https://github.com/espressif/esp-protocols/pull/794

## Testing

- Existing "Sec-WebSocket-Accept" header-parsing was tested by connecting to a websocket. Failure was tested by intentionally setting `server_key = NULL` and confirming we still reject the connection.
- New headers were tested by logging each header-line with `ESP_LOGI()` at the next layer up, in the esp_websocket_client.

---

## Checklist

### DIFF
--- a/components/tcp_transport/include/esp_transport_ws.h
+++ b/components/tcp_transport/include/esp_transport_ws.h
@@ -29,6 +29,8 @@ typedef enum ws_transport_opcodes {
                                           * from the API esp_transport_ws_get_read_opcode() */
 } ws_transport_opcodes_t;
 
+typedef void (*ws_header_hook)(void * userp, const char * line, int line_len);
+
 /**
  * WS transport configuration structure
  */
@@ -37,6 +39,8 @@ typedef struct {
     const char *sub_protocol;               /*!< WS subprotocol */
     const char *user_agent;                 /*!< WS user agent */
     const char *headers;                    /*!< WS additional headers */
+    ws_header_hook header_hook;             /*!< WS received header */
+    void *header_userp;                     /*!< WS received header user-pointer */
     const char *auth;                       /*!< HTTP authorization header */
     char *response_headers;                 /*!< The buffer to copy the http response header */
     size_t response_headers_len;            /*!< The length of the http response header */
@@ -98,6 +102,31 @@ esp_err_t esp_transport_ws_set_user_agent(esp_transport_handle_t t, const char *
  *      - One of the error codes
  */
 esp_err_t esp_transport_ws_set_headers(esp_transport_handle_t t, const char *headers);
+
+/**
+ * @brief               Set websocket header callback
+ *
+ * @param t             websocket transport handle
+ * @param hook          call function on header received. NULL to disable.
+ *
+ * @return
+ *      - ESP_OK on success
+ *      - One of the error codes
+ */
+esp_err_t esp_transport_ws_set_header_hook(esp_transport_handle_t t, ws_header_hook hook);
+
+
+/**
+ * @brief               Set websocket header callback user-pointer
+ *
+ * @param t             websocket transport handle
+ * @param userp         caller-controlled argument to ws_header_hook
+ *
+ * @return
+ *      - ESP_OK on success
+ *      - One of the error codes
+ */
+esp_err_t esp_transport_ws_set_header_userp(esp_transport_handle_t t, void * userp);
 
 /**
  * @brief               Set websocket authorization headers
--- a/components/tcp_transport/transport_ws.c
+++ b/components/tcp_transport/transport_ws.c
@@ -63,6 +63,8 @@ typedef struct {
     char *sub_protocol;
     char *user_agent;
     char *headers;
+    ws_header_hook header_hook;
+    void * header_userp;
     char *auth;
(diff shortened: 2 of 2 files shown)


components/tcp_transport/transport_ws.c
```diff
@@ -144,29 +146,31 @@ static int esp_transport_read_internal(transport_ws_t *ws, char *buffer, int len
     return to_read;
 }
 
-static char *trimwhitespace(char *str)
```
glmfe, line 147: these functions are being used by another recently added feature (http redirect). so we might need to keep these (trimwhitespace, get_http_header)

components/tcp_transport/transport_ws.c
```diff
@@ -336,7 +325,45 @@ static int ws_connect(esp_transport_handle_t t, const char *host, int port, int
...
+            ESP_LOGD(TAG, "HTTP Header walk found end");
+            break;
+        }
+        int line_len = end_of_line - header_cursor;
+        ESP_LOGD(TAG, "HTTP Header walk line:%.*s", line_len, header_cursor);
+
+        // Find the Sec-WebSocket-Accept header
+        const char * header_sec_websocket_accept = "Sec-WebSocket-Accept: ";
+        size_t header_sec_websocket_accept_len = strlen(header_sec_websocket_accept);
+        if (line_len >= header_sec_websocket_accept_len && !strncmp(header_cursor, header_sec_websocket_accept, header_sec_websocket_accept_len)) {
```
glmfe: As defined in RFC7230, the header fields are "case-insensitive".

image

so we might change this 
```suggestion
        if (line_len >= header_sec_websocket_accept_len && !strncasecmp(header_cursor, header_sec_websocket_accept, header_sec_websocket_accept_len)) {
```

Bryght-Labs-Richard: Good catch - I was unaware.

components/tcp_transport/transport_ws.c
Bryght-Labs-Richard: Good catch, and...

* seems like `get_http_header(ws->buffer, "Location:")` will incorrectly return results for "SomeOtherLocation:" header. Seems risky.

* seems like `get_http_header()` must only ever be called once, because it will null-terminate each found line, so that additional calls to `get_http_header()` become dependent on the response header order. Not a problem today but seems risky.

I propose we unify all features using response headers to the new header parser and remove `trimwhitespace` and `get_http_header`.

glmfe:
sha=0245c5c863fd1bf08ed9fbd913f8cccb297948cc

euripedesrocha:
sha=0245c5c863fd1bf08ed9fbd913f8cccb297948cc

glmfe:
Hi @bryghtlabs-richard,

I've got some comments from colleagues internally that I'll post them here as well

components/tcp_transport/include/esp_transport_ws.h
```diff
@@ -99,6 +103,31 @@ esp_err_t esp_transport_ws_set_user_agent(esp_transport_handle_t t, const char *
...
+ * @brief               Set websocket header callback user-pointer
+ *
+ * @param t             websocket transport handle
+ * @param userp         caller-controlled argument to ws_header_hook
+ *
+ * @return
+ *      - ESP_OK on success
+ *      - One of the error codes
+ */
+esp_err_t esp_transport_ws_set_header_userp(esp_transport_handle_t t, void * userp);
```
glmfe, line 129: Why do we need to set this pointer in a separated function? Could it be a context in the set header hook?

components/tcp_transport/transport_ws.c
```diff
@@ -927,6 +963,14 @@ esp_err_t esp_transport_ws_set_config(esp_transport_handle_t t, const esp_transp
         err = esp_transport_ws_set_headers(t, config->headers);
         ESP_TRANSPORT_ERR_OK_CHECK(TAG, err, return err;)
     }
+    if (config->header_hook) {
+        err = esp_transport_ws_set_header_hook(t, config->header_hook);
+        ESP_TRANSPORT_ERR_OK_CHECK(TAG, err, return err;)
+    }
+    if (config->header_userp) {
```
glmfe, line 970: What if user set this, but the hook isn't configured, should we warn? Is the idea that the context for the callback could change in between calls?

components/tcp_transport/include/esp_transport_ws.h
Bryght-Labs-Richard: Indeed it could. I'm fine with either way.

components/tcp_transport/transport_ws.c
Bryght-Labs-Richard: I'm not sure I would worry about it, but if it's preferred we could. I don't expect anyone would change the header_userp context in-between headers, given that their ordering isn't guaranteed.

glmfe:
Accepted as [11f8534](https://github.com/espressif/esp-idf/commit/11f8534266fb1c3814bf51ddc98ebb5ae09446f6)

Thanks for your contribution!

PR #16138, opened 2025-06-15
feat(esp_eth): Enable EEE (802.3az) on DM9051 (IDFGH-15497)

bmork:
Significant power savings on a 100baseT link.

## Description

This unconditionally enables EEE on DM9051. The resulting power savings are around 200 mW on a 100baseT link.  

## Testing

Briefly run tested on a PoE powered ESP32C3 using a DM9051, using the PoE switch to measure total power consumption for the module.

---

## Checklist

### DIFF
--- a/components/esp_eth/include/esp_eth_mac_spi.h
+++ b/components/esp_eth/include/esp_eth_mac_spi.h
@@ -112,6 +112,14 @@ typedef struct
         .read = NULL,    \
         .write = NULL    \
     }
+
+/**
+ * @brief List of SPI EMAC specific commands for ioctl API
+ *
+ */
+typedef enum {
+    ETH_MAC_SPI_CMD_S_EEE = ETH_CMD_CUSTOM_MAC_CMDS_OFFSET, /*!< Enable or disable Energy Efficient Ethernet */
+} eth_mac_spi_io_cmd_t;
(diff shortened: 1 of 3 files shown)


kostaond:
Hi @bmork thank you for your contribution! However, I'm afraid, the PR cannot be approved as is... I'm still not sure if it's safe approach to always enable EEE. EEE may introduce a latency and I also read bad experience by server maintainers with EEE in general. Maybe, you could consider making it as optional parameter possible to be enabled via `esp_eth_ioctl`.

bmork:
Thanks for reviewing.  I was unsure how such features should be implemented and couldn't find any examples of other drivers implementing EEE.  I'll look at the `esp_eth_ioctl` method.  Thanks for the pointer

bmork:
I remebered that I absolutely hate creating APIs :-) I never seem to get them right.  And then we're stuck with it forever. 

But here's an attempt anyway.  It's absolutely minimalistic, just implementing the actual use case I have:  Enabling EEE with the default timers on the DM9051.  This way I know that I don't define an unusable API.  This is tested and works for me.

Please let me know if you have other preferences.

bmork:
Obsolete with the removal of the SPI ethernet drivers...

PR #16149, opened 2025-06-17
fix(freertos): Avoid core switch deadlock on start (IDFGH-15509)

gonzzor:
## Description
With CONFIG_ESP_MAIN_TASK_AFFINITY_NO_AFFINITY=y the main task could switch core between registering the idle hook and the while loop.

This would cause a deadlock were the current task was waiting for the idle hook to run on the same core it's busy waiting.

Solve this by registering an idle hook on both cores. This ensure that the scheduler is started on the other core.

## Testing
This problem was discovered on an ESP32-S3 running 5.4.1 with QIO and CONFIG_ESP_MAIN_TASK_AFFINITY_NO_AFFINITY=y. It would randomly happen during boot.

With this change the problem hasn't been possible to reproduce.

---

## Checklist

### DIFF
--- a/components/freertos/app_startup.c
+++ b/components/freertos/app_startup.c
@@ -147,7 +147,7 @@ void esp_startup_start_app_other_cores(void)
 static const char* MAIN_TAG = "main_task";
 
 #if !CONFIG_FREERTOS_UNICORE
-static volatile bool s_other_cpu_startup_done[2];
+static volatile bool s_other_cpu_startup_done[2] = {false, false};
 static bool other_cpu_startup_idle_hook_cb(void)
 {
     s_other_cpu_startup_done[xPortGetCoreID()] = true;
@@ -162,7 +162,7 @@ static void main_task(void* args)
     // Wait for FreeRTOS initialization to finish on other core, before replacing its startup stack
     esp_register_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, 0);
     esp_register_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, 1);
-    while (!s_other_cpu_startup_done) {
+    while (!s_other_cpu_startup_done[!xPortGetCoreID()]) {
         ;
     }
     esp_deregister_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, 0);
--- a/components/freertos/app_startup.c
+++ b/components/freertos/app_startup.c
@@ -147,10 +147,10 @@ void esp_startup_start_app_other_cores(void)
 static const char* MAIN_TAG = "main_task";
 
 #if !CONFIG_FREERTOS_UNICORE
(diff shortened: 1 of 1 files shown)


components/freertos/app_startup.c
```diff
@@ -160,11 +160,13 @@ static void main_task(void* args)
     ESP_LOGI(MAIN_TAG, "Started on CPU%d", (int)xPortGetCoreID());
 #if !CONFIG_FREERTOS_UNICORE
     // Wait for FreeRTOS initialization to finish on other core, before replacing its startup stack
-    esp_register_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, !xPortGetCoreID());
+    esp_register_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, 0);
+    esp_register_freertos_idle_hook_for_cpu(other_cpu_startup_idle_hook_cb, 1);
     while (!s_other_cpu_startup_done) {
```
KonstantinKondrashov: `s_other_cpu_startup_done` is now an array.

components/freertos/app_startup.c
```diff
@@ -147,10 +147,10 @@ void esp_startup_start_app_other_cores(void)
 static const char* MAIN_TAG = "main_task";
 
 #if !CONFIG_FREERTOS_UNICORE
-static volatile bool s_other_cpu_startup_done = false;
+static volatile bool s_other_cpu_startup_done[2];
```
KonstantinKondrashov:
```suggestion
static volatile bool s_other_cpu_startup_done[2] = {0};
```

KonstantinKondrashov (changes requested):
I will test this case. I wonder if this is the proper way to fix the issue.

components/freertos/app_startup.c
gonzzor: Yes, my bad. It got lost in rebase process from my internal work. Fixed it.

components/freertos/app_startup.c
gonzzor: Don't think it's needed for static variables, right? Anyway, fixed this as well.

components/freertos/app_startup.c
gonzzor: Also, looking at the code again. The variable name could probably be changed to something better.

gonzzor: Also, looking at the code again. The variable name could probably be changed to something better.

gonzzor:
> I will test this case. I wonder if this is the proper way to fix the issue.

Thanks, let me know if you need more input from me.

KonstantinKondrashov:
I was able to reproduce this behavior in a test; it can occur. I will add this PR to our internal repository. The final solution may differ slightly from your original commit. Thanks.

KonstantinKondrashov (approved):

KonstantinKondrashov:
sha=b97d674836a2c593e118907ca4f5e7b17fb30d3d

PR #16187, opened 2025-06-24
openthread: implement ot_rcp over USB connection (IDFGH-15554)

nagisa:
## Description

This provides a hassle-free, least dependency mechanism to use with `ot-br-posix` and boards that connect their USB port directly to the JTAG USB pins. esp32c6 supermini is such an example.

To be fair using `/dev/secondary` seems like a crude hack, but at the same time I stared at the jtag serial vfs code for half an hour and couldn't figure a way to manually initialize that vfs code in a way that would be reserved to this use-case.

(I also find it quite weird how there's such relatively deep integration of openthread into what feels like a standard library… I'm much more partial to esp-rs/openthread approach, but I wanted to get something working quickly.)

## Testing

with a esp32c6 supermini

```
idf.py menuconfig # select the new RCP mode
idf.py build flash
otbr-agent -d7 --verbose spinel+hdlc+uart:///dev/ttyACM0
```

---

## Checklist


nagisa:
FWIW you're welcome to just steal the idea and integrate it without waiting for me to get around to fixing this PR (or walking me through the correct way(s) to implement this logic)

gytxxsy:
Thank you for your contribution. We are currently working on supporting `RCP over USB` and already have a working patch. Please refer to [Openthread RCP using USB to control it](https://github.com/orgs/openthread/discussions/11627#discussioncomment-13558288).
We've also made related fixes to `usb_serial_jtag` recently to support this functionality and plan to officially support `RCP over USB` soon.
By the way, you may need to open `/dev/usbserjtag` to retrieve the descriptor instead of `/dev/secondary` (of course this is just a guess because I haven’t tested `/dev/secondary`).

nagisa:
I had to use `/dev/secondary` as `/dev/usbserjtag` appears to also be `stdin`/`stdout` and then any operations that relate to `stdout` (such as logging) will interfere with the RCP/Spinel/HDLC communications.

(Using `/dev/secondary` of course requires to makeconfig JTAG serial as a secondary, not primary, but I believe I managed to encode that into makeconfig's dependencies.)

nagisa:
That said can't wait to see the official implementation here. But at the same feel free to take inspiration from what I've done here. This code works okay for me with `ot-br-posix` on the current proper release (v5.4.2) too.

gytxxsy:
> I had to use `/dev/secondary` as `/dev/usbserjtag` appears to also be `stdin`/`stdout` and then any operations that relate to `stdout` (such as logging) will interfere with the RCP/Spinel/HDLC communications.
> 
> (Using `/dev/secondary` of course requires to makeconfig JTAG serial as a secondary, not primary, but I believe I managed to encode that into makeconfig's dependencies.)

Regarding the conflict you mentioned between `RCP/Spinel/HDLC` and `stdin`/`stdout`, this is not an issue caused by USB Serial JTAG. In fact, the same problem exists with RCP over UART if port 0 is used. That said, you can try using my patch first to see if it temporarily meets your needs.

chshu:
It's merged in [51bcec7](https://github.com/espressif/esp-idf/commit/51bcec7ac67efff9e915239fa07fb1909b002d80). @nagisa Thanks for the contribution!

PR #16245, opened 2025-07-03
fix(partition_table): parttool.py handle empty input & erase after all is valid on write_partition (IDFGH-15616)

alwint3r:
## Description

This fixes the following problems:

- No empty input checking. `--input` must be accompanied by a valid filename.
- Partition is erased when either the input is empty or the file is too big for the partition.

## Testing

The change is tested manually because it requires a real ESP32 hardware.

After the change is applied:
```sh
parttool.py write_partition --partition-name storage
```
Output:
```
Connecting....
Detecting chip type... Unsupported detection protocol, switching and trying again...
Connecting....
Detecting chip type... ESP32
Chip is ESP32-PICO-D4 (revision v1.0)
Features: WiFi, BT, Dual Core, 240MHz, Embedded Flash, VRef calibration in efuse, Coding Scheme None
Crystal is 40MHz
MAC: e8:9f:6d:0c:59:cc
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
3072 (100 %)
3072 (100 %)
Read 3072 bytes at 0x00008000 in 0.3 seconds (87.4 kbit/s)...
Hard resetting via RTS pin...
Input file is required
```

No partition is erased in this case.

---

## Checklist

---

> [!NOTE]
> Improves safety and correctness of partition writes and erases.
> 
> - `write_partition` now requires `--input`, validates size against partition, and erases only the remaining space after the input length via `erase_partition(partition_id, content_len)`
> - `erase_partition(partition_id, offset=0)` adds offset support with non-negative check, 4KB alignment, boundary checks, and erases from computed offset to partition end
> - Updates esptool subcommands to underscore variants: `erase_region` and `write_flash`
> - Prevents unintended full-partition erase when input is missing or oversized
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit abce9c7ef8ae8258e1ec323294af2805ec7ffbf4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/partition_table/parttool.py
+++ b/components/partition_table/parttool.py
@@ -150,9 +150,25 @@ class ParttoolTarget:
 
         return partition
 
-    def erase_partition(self, partition_id):
+    def erase_partition(self, partition_id, offset=0):
         partition = self.get_partition_info(partition_id)
-        self._call_esptool(['erase-region', str(partition.offset), str(partition.size)] + self.esptool_erase_args)
+        if offset < 0:
+            raise SystemExit('Offset must not be negative')
+
+        offset = partition.offset + offset
+        if offset % 4096 != 0:
+            offset = ((offset + 4095) // 4096) * 4096
+
+        partition_end = partition.offset + partition.size
+
+        if offset > partition_end:
+            raise SystemExit(f'Offset is outside partition "{partition.name}" boundary')
+
+        if offset == partition_end:
+            return
+
+        size = partition_end - offset
+        self._call_esptool(['erase_region', str(offset), str(size)] + self.esptool_erase_args)
 
     def read_partition(self, partition_id, output):
         partition = self.get_partition_info(partition_id)
@@ -164,7 +180,8 @@ class ParttoolTarget:
         if partition.readonly and not ignore_readonly:
             raise SystemExit(f'"{partition.name}" partition is read-only, (use the --ignore-readonly flag to skip it)')
 
-        self.erase_partition(partition_id)
+        if input is None:
+            raise SystemExit('Input file is required')
 
         with open(input, 'rb') as f:
             content_len = len(f.read())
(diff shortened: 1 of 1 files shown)


components/partition_table/parttool.py
```diff
@@ -165,14 +165,17 @@ def write_partition(self, partition_id, input, ignore_readonly=False):  # noqa:
...
+        if input is None:
+            raise SystemExit('Input file is required')
 
         with open(input, 'rb') as f:
             content_len = len(f.read())
 
             if content_len > partition.size:
                 raise Exception('Input file size exceeds partition size')
 
+        self.erase_partition(partition_id)
```
igrr: It seems unnecessary to erase the part of the partition which will be written in `_call_esptool(...` call below, since writing also involves an erase operation. Should we only erase the part between `content_len` and the end of the partition, instead?

alwint3r: I was thinking the same thing. I'll update the PR. Thanks!

Edit: I removed the call completely, but it seems like I missed your point. Let me figure something out.

igrr: Right, what I meant is that we should probably erase the range between `ROUND_DOWN(content_len, SPI_FLASH_SECTOR_SIZE)` and partition size.

However, feel free to simply add the argument check before erasing the partition — we can remove this redundant erase later.

alwint3r:
Hi @igrr, I have pushed new changes that will erase only the sectors that are not written by new data.

components/partition_table/parttool.py
```diff
@@ -151,9 +151,11 @@ def get_partition_info(self, partition_id):
 
         return partition
 
-    def erase_partition(self, partition_id):
+    def erase_partition(self, partition_id, offset=None, size=None):
         partition = self.get_partition_info(partition_id)
-        self._call_esptool(['erase_region', str(partition.offset), str(partition.size)] + self.esptool_erase_args)
+        offset = offset if offset is not None else partition.offset
```
igrr: Let's change the `offset` argument meaning to be "offset within the partition". Otherwise the function does not necessarily erase part of the partition `partition_id`, it can erase some other partition.

I would also suggest moving the bounds checks and rounding logic inside this function.

igrr:
Almost LGTM, just one nitpick.

components/partition_table/parttool.py
alwint3r: You're right. I think we can discard the additional size parameter since we only want to erase flash sectors not covered by the new data.

igrr (approved):

igrr:
sha=c278159112a7979e5a1dbce31621f2fe0aa4578e

PR #16247, opened 2025-07-03
feat(bootloader): Request factory reset

nebkat:
## Description

Provides a way for the app to request that a factory reset be performed by the bootloader on next boot.

This ensures consistency between factory reset by button and a programmatic reset, and is safer as there is no risk of a runtime crash due to e.g. NVS partition being loaded at runtime detecting corruption during the process.

Also raises log level when performing factory reset to `WARN` to make it easier to see.

## Testing

New test added to verify functionality.

Tested in own project and working as expected.

---

## Checklist


components/bootloader/subproject/main/bootloader_start.c
```diff
@@ -86,20 +86,32 @@ static int selected_boot_partition(const bootloader_state_t *bs)
         return boot_index; // Unrecoverable failure (not due to corrupt ota data or bad partition contents)
     }
     if (esp_rom_get_reset_reason(0) != RESET_REASON_CORE_DEEP_SLEEP) {
-        // Factory firmware.
 #ifdef CONFIG_BOOTLOADER_FACTORY_RESET
+        bool factory_reset = false;
+
+#ifdef CONFIG_BOOTLOADER_RESERVE_RTC_MEM
```
sudeep-mohanty: It would be better to add an explicit config option rather than depending on the RTC reserved memory. You could add another reset option like `CONFIG_BOOTLOADER_FACTORY_RESET_REQUEST_FROM_APP` for this new way of issuing a factory reset.

components/bootloader/subproject/main/bootloader_start.c
```diff
@@ -86,20 +86,32 @@ static int selected_boot_partition(const bootloader_state_t *bs)
...
+        // Check if the factory reset is requested programmatically.
```
sudeep-mohanty: Nitpick: 
```suggestion
        // Check if the factory reset is requested from the application.
```

components/bootloader/subproject/main/bootloader_start.c
```diff
@@ -86,20 +86,32 @@ static int selected_boot_partition(const bootloader_state_t *bs)
...
 #endif
-        if (bootloader_common_check_long_hold_gpio_level(CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET, CONFIG_BOOTLOADER_HOLD_TIME_GPIO, reset_level) == GPIO_LONG_HOLD) {
-            ESP_LOGI(TAG, "Detect a condition of the factory reset");
+
+        // Check if the factory reset is requested by GPIO long hold.
+        factory_reset = factory_reset || bootloader_common_check_long_hold_gpio_level(CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET, CONFIG_BOOTLOADER_HOLD_TIME_GPIO, reset_level) == GPIO_LONG_HOLD;
+#endif // CONFIG_BOOTLOADER_FACTORY_RESET_GPIO
+
+        if (factory_reset) {
+            ESP_LOGW(TAG, "Detect a condition of the factory reset");
```
sudeep-mohanty, line 114: Would be nice to indicate if one or both the sources have requested a factory reset here.

components/bootloader_support/include/bootloader_common.h
```diff
@@ -278,6 +278,21 @@ bool bootloader_common_get_rtc_retain_mem_factory_reset_state(void);
  */
 void bootloader_common_set_rtc_retain_mem_factory_reset_state(void);
 
+/**
+ * @brief Returns True if Factory reset was requested
+ *
+ * Reset the request status after reading it.
```
sudeep-mohanty: The comment here is a bit ambiguous. It isn't clear if the function clears the request automatically or the caller should do it. In this case, it is the function that does it so maybe we could rename the function as `bootloader_common_get_and_clear_rtc_retain_mem_factory_reset_request()`

components/bootloader_support/include/esp_image_format.h
```diff
@@ -52,7 +52,8 @@ typedef struct {
     union {
         struct {
             uint8_t factory_reset_state         : 1;  /* True when Factory reset has occurred */
-            uint8_t reserve                     : 7;  /* Reserve */
+            uint8_t factory_reset_request       : 1;  /* True when Factory reset is requested */
```
sudeep-mohanty, line 55: This field could be wrapped in the new config option.

docs/en/api-guides/bootloader.rst
```diff
@@ -84,14 +84,18 @@ Either or both of these configuration options can be enabled independently.
 
 In addition, the following configuration options control the reset condition:
 
-- :ref:`CONFIG_BOOTLOADER_NUM_PIN_FACTORY_RESET`- The input GPIO number used to trigger a factory reset. This GPIO must be pulled low or high (configurable) on reset to trigger this.
+- :ref:`CONFIG_BOOTLOADER_FACTORY_RESET_GPIO` - enables the factory reset by GPIO functionality in the bootloader. If this is not enabled, the bootloader can only perform a factory reset programmatically.
```
sudeep-mohanty: Let's add the fact that this is the default option.

sudeep-mohanty (commented):
Hi @nebkat,
Thank you for the contribution!

This is a neat feature addition. The overall design LGTM. Left a few minor comments to work upon. Thanks.

components/bootloader/subproject/main/bootloader_start.c
nebkat: Would this be for when a developer explicitly wants to prevent triggering factory reset from app? I figured they would simply not call the request function from app - but I guess it would be better from a security perspective to prevent exploits?

components/bootloader_support/include/bootloader_common.h
nebkat: Agreed, will change.

Side note: the logic here was based on the `bootloader_common_get_rtc_retain_mem_factory_reset_status()` which suffers from the same problem, except slightly worse because it is the responsibility of the app. I feel like the bootloader should be responsible for unsetting the `factory_reset_status` on subsequent boots rather than expecting the app to do it. But I think to fix would be a breaking change, so that is a discussion for another time.

components/bootloader_support/include/esp_image_format.h
nebkat: I think it would be risky to mix different layouts here - e.g. what if there is a new field added in future that does something else, then you OTA a bootloader which newly supports `factory_reset_request` - unintentional factory reset occurs. Worth noting also that `factory_reset_state` is always present regardless of the factory reset config.

sudeep-mohanty: That's a valid point. I agree that it could be risky. Let's keep it as is.

components/bootloader/subproject/main/bootloader_start.c
sudeep-mohanty: Yes, it would be better to have the feature turned on explicitly by conscious users from the security point of view.

components/bootloader_support/include/bootloader_common.h
nebkat: Also it's quite an ugly function name in general - could we alias it to `esp_bootloader_request_factory_reset()` and `esp_bootloader_check_factory_reset_status()`?

sudeep-mohanty: While I must admit that the function names are becoming less readable and less intuitive, unfortunately we must stick with the format matching the other functions in the API, viz., `bootloader_common_xxx`. Having said that, we have agreed internally to do a review of the whole API and if necessary, have changes done in one-go for all the functions. Created a tracking task #16255

nebkat:
@sudeep-mohanty

- Renamed (`sdkconfig.rename`) `CONFIG_BOOTLOADER_FACTORY_RESET` to `CONFIG_BOOTLOADER_FACTORY_RESET_GPIO` to preserve existing behavior.
- Added `CONFIG_BOOTLOADER_FACTORY_RESET_REQUEST_FROM_APP`.
- Added invisible `CONFIG_BOOTLOADER_FACTORY_RESET_ENABLE` that is selected by the above.
- Did _not_ `#ifdef CONFIG_BOOTLOADER_FACTORY_RESET_REQUEST_FROM_APP` around `bootloader_common_set_rtc_retain_mem_factory_reset_request()` as (in theory) the app could be running with different bootloader versions that are configured differently:
    - Example I can think of is a security upgrade was performed that removed the ability to perform factory reset, but old devices still need to be factory reset to perform the upgrade i.e. `CONFIG_BOOTLOADER_FACTORY_RESET_REQUEST_FROM_APP` is not _currently_ set, but is still used by the app.
    - Matches `bootloader_common_get_rtc_retain_mem_factory_reset_state()` lack of `#ifdef`.
- Updated documentation to include separate subsubsections for app triggered and gpio triggered resets.
- Added new log message for when app or GPIO triggers reset.
    - GPIO status is checked even if app request was detected, which in theory may add some unnecessary delay, but helps in testing?
    - Original message is preserved to not affect any tests.  

image

components/bootloader/Kconfig.projbuild
```diff
@@ -145,70 +145,96 @@ menu "Bootloader config"
...
+        config BOOTLOADER_FACTORY_RESET_REQUEST_FROM_APP
+            bool "App request triggers factory reset"
+            default n
+            depends on SOC_RTC_FAST_MEM_SUPPORTED
+            select BOOTLOADER_FACTORY_RESET_ENABLE
+            select BOOTLOADER_RESERVE_RTC_MEM if SOC_RTC_FAST_MEM_SUPPORTED
+            help
+                This option enables performing a factory reset by a request from the app, which may include:
+                - clear one or more data partitions;
+                - reset to boot from "factory" partition.
```
sudeep-mohanty, line 167: This is redundant information. We present it at the "Factory reset" menu level so it can be removed from here.

components/bootloader/Kconfig.projbuild
```diff
@@ -145,70 +145,96 @@ menu "Bootloader config"
...
 
-        config BOOTLOADER_FACTORY_RESET_PIN_LOW
-            bool "Reset on GPIO low"
+                The app can request a factory reset by calling the function
+                `bootloader_common_set_rtc_retain_mem_factory_reset_request`.
+                This stores the request in RTC memory, allowing the bootloader
+                to perform the factory reset on next boot, provided power is preserved.
```
sudeep-mohanty, line 172: Nitpick: 

```suggestion
                to perform the factory reset on next boot, provided power to the RTC memory is preserved.
```

components/bootloader/Kconfig.projbuild
```diff
@@ -145,70 +145,96 @@ menu "Bootloader config"
...
-            The device will boot from "factory" partition (or OTA slot 0 if no factory partition is present) after a
-            factory reset.
+        config BOOTLOADER_FACTORY_RESET_GPIO
+            bool "GPIO triggers factory reset"
+            default y
+            select BOOTLOADER_FACTORY_RESET_ENABLE
+            help
+                This option enables performing a factory reset by long holding a GPIO input, which may include:
+                - clear one or more data partitions;
+                - reset to boot from "factory" partition.
```
sudeep-mohanty, line 182: Same here. Can be removed now.

components/bootloader/Kconfig.projbuild
nebkat: The above `config BOOTLOADER_FACTORY_RESET_ENABLE` is a hidden item. I can make it be an explicit option, but it is useless without one of the two triggers being enabled, so this arrangement made more sense to me.

sudeep-mohanty:
Hi @nebkat,
Thank you for the updates. We had further discussions around this feature request and are now of the opinion that a new method to perform a factory reset from the app may not be necessary. The same could be achieved using [custom bootloader hooks](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/bootloader.html#custom-bootloader). An example demonstrating how to use it to extend the bootloader functionality is [here](https://github.com/espressif/esp-idf/blob/master/examples/custom_bootloader/bootloader_hooks/README.md).

Also, alternatively or in addition to the bootloader hooks, you could in fact manipulate the `BOOTLOADER_NUM_PIN_FACTORY_RESET` directly from the application and issue a reset, triggering a factory reset. Something like this is done in the test_app [here](https://github.com/espressif/esp-idf/blob/bfe5caf58f742fd35c023335f475114a5b88761e/components/app_update/test_apps/test_app_update/main/test_switch_ota.c#L183-L184).

nebkat:
@sudeep-mohanty
My hope with this change was to have something readily available for all developers that is fully tested/maintained considering:
- It is a critical feature that could result in bricked or unintentionally reset devices if wrongly implemented 
- Most of the factory reset logic is already there
- RTC mem is already set up, checksummed, etc

The GPIO trick would mean exposing a GPIO as a factory reset pin, which may not be available or desirable.

The bootloader hooks would be a maintenance overhead for developers, though I'm aware this is just shifting that responsibility to the framework.

PR #17047, opened 2025-07-29
change(esp32): Calling gpio_reset_pin on an input only pin should not produce an error message.

remenyo:
## Description

The `gpio_reset_pin` function attempts to enable internal pullup on input only pins as well, which do not have configurable internal pullups.

For example, in the case of ESP32, 
```c
gpio_reset_pin(GPIO_NUM_39);
```
emits:
```bash
E (4535) gpio: gpio_pullup_en(78): GPIO number error (input-only pad has no internal PU)
```

This change adds the same guard inside `gpio_reset_pin` before calling `gpio_pullup_en` that makes `gpio_pullup_en` print the error in case of calling it with an input only pin.

I'm not familiar with the inside structure of IDF, maybe this is done by purpose, or if not, maybe other parts of the driver should be updated as well. Or maybe the error in gpio_pullup_en should be just a warning.

This change only makes the error log not to show up - the functional effect of calling `gpio_reset_pin` with any parameter remains identical.
---

## Checklist

### DIFF
--- a/components/esp_driver_gpio/src/gpio.c
+++ b/components/esp_driver_gpio/src/gpio.c
@@ -457,8 +457,10 @@ esp_err_t gpio_reset_pin(gpio_num_t gpio_num)
 {
     GPIO_CHECK(GPIO_IS_VALID_GPIO(gpio_num), "GPIO number error", ESP_ERR_INVALID_ARG);
     gpio_intr_disable(gpio_num);
-    // for powersave reasons, the GPIO should not be floating, select pullup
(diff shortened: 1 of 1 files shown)


songruo:
sha=19e5a33b4bd62ebe4c693504a3a0199890be39ef

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/df5bff7a07f12e5c9f484ecd490dd7cd92d011fd

Issue #16891, opened 2025-07-17
Problem with ulp_riscv_delay_cycles (IDFGH-15960)


gestalth:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

5.5.0

### Espressif SoC revision.

S3

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

None

### Development Kit.

Custom

### Power Supply used.

USB

### What is the expected behavior?

ULP runs without freezing.

### What is the actual behavior?

It freezes after several minutes of working correctly using I2C and I have to restart the ULP.

### Steps to reproduce.

1. Setup a RISCV ULP program that continuously reads from I2C.
2. Setup a heartbeat signal to monitor on main core if the ULP is still running.
3. Let run for quite a while. Maybe up to 30 or so minutes. It doesn't happen at the same time each time.
4. Probably the ULP will have frozen at some point in that time.
...

### Debug Logs.

```plain

```

### More Information.

I used trace variables to find that it was freezing due to ulp_riscv_delay_cycles(1) function called from ulp_riscv_i2c_wait_for_interrupt(). 

Currently it looks like this:
void static inline ulp_riscv_delay_cycles(uint32_t cycles)
{
    uint32_t start = ULP_RISCV_GET_CCOUNT();
    /* Off with an estimate of cycles in this function to improve accuracy */
    uint32_t end = start + cycles - 20;

    while (ULP_RISCV_GET_CCOUNT()  < end) {
        /* Wait */
    }
}

I think that the line "while (ULP_RISCV_GET_CCOUNT()  < end)" is creating problems near the roll over. I haven't seen the freeze since changing it to:
void static inline ulp_riscv_delay_cycles(uint32_t cycles)
{    
    uint32_t start = ULP_RISCV_GET_CCOUNT();

    /* Off with an estimate of cycles in this function to improve accuracy */
    if(cycles < 21)
        return;

    cycles = cycles-20;

    while ((ULP_RISCV_GET_CCOUNT() - start)  < cycles) {
        /* Wait */
    }
}

[...]

PR #17050, opened 2025-07-30
esp32s2/3 riscv ulp: fix delay_cycles for timer overflow

pavlohamov:
## Description

ulp_riscv_delay_cycles calculates end cycles as start + delay as math is 32bit: end may be lower that start due to overflow leading to zero dealy

## Related

https://github.com/espressif/esp-idf/issues/16891

## Testing

Verified on esp32s3 with ULP running in superloop mode


SoucheSouche:
Hi @pavlohamov,

The issue reported in https://github.com/espressif/esp-idf/issues/16891 was fixed already.

I will therefore close the PR.

PR #17062, opened 2025-07-31
Pass context to esp_console_run

JVKran:
## Description

One is now able to override the statically configured context with ```esp_console_cmd_register``` by calling the new function ```esp_console_run_with_ctx```. I've taken a look at the original pull-request for the addition of ```esp_console_cmd_func_with_context_t``` and have reused the same context and function.

Specifically, this is very useful when commands are received through (for example) MQTT, HTTP and ESP-NOW. It's great to be able to pass these backends so the command handlers can redirect output to those backends.

## Testing

Changes have been tested in our existing code base. Changes are very basic and simple.

---

## Checklist


o-marshmallow:
Hello @JVKran ,

If I understand correctly, you want to be able to invoke a command with different parameters (context) than the one provided when registering the function, basically, overriding the default parameter. Is it correct?

components/console/commands.c
```diff
@@ -231,7 +231,12 @@ static const cmd_item_t *find_command_by_name(const char *name)
     return cmd;
 }
 
-esp_err_t esp_console_run(const char *cmdline, int *cmd_ret)
+esp_err_t esp_console_run(const char* cmdline, int* cmd_ret) 
+{
+    return esp_console_run_with_ctx(NULL, cmdline, cmd_ret);
+}
+
+esp_err_t esp_console_run_with_ctx(void* invocation_ctx, const char *cmdline, int *cmd_ret)
```
o-marshmallow:
```suggestion
esp_err_t esp_console_run_with_context(void* invocation_ctx, const char *cmdline, int *cmd_ret)
```
Since the type is named `esp_console_cmd_func_with_context_t`, I would suggest to keep `context` instead of `ctx`, even

components/console/commands.c
```diff
@@ -257,7 +262,7 @@ esp_err_t esp_console_run(const char *cmdline, int *cmd_ret)
         *cmd_ret = (*cmd->func)(argc, argv);
     }
     if (cmd->func_w_context) {
-        *cmd_ret = (*cmd->func_w_context)(cmd->context, argc, argv);
+        *cmd_ret = (*cmd->func_w_context)(invocation_ctx, argc, argv);
```
o-marshmallow:
```suggestion
        *cmd_ret = (*cmd->func_w_context)(invocation_ctx ?: cmd->context, argc, argv);
```
I think this is what you meant, else the commands will never ne executed with their default context

components/console/esp_console.h
```diff
@@ -259,6 +259,21 @@ esp_err_t esp_console_cmd_deregister(const char *cmd_name);
  */
 esp_err_t esp_console_run(const char *cmdline, int *cmd_ret);
 
+/**
+ * @brief Run command line with context
+ * @param cmdline command line (command name followed by a number of arguments)
+ * @param[out] cmd_ret return code from the command (set if command was run)
+ * @param invocation_ctx pointer to context, can be used to pass additional data
```
o-marshmallow: I think it's more a context override rather than additional data

components/console/esp_console.h
```diff
@@ -259,6 +259,21 @@ esp_err_t esp_console_cmd_deregister(const char *cmd_name);
...
+ * @param invocation_ctx pointer to context, can be used to pass additional data 
+ * to the command handler. Overrides context given during command registration.
+ * @return
+ *      - ESP_OK, if command was run
+ *      - ESP_ERR_INVALID_ARG, if the command line is empty, or only contained
+ *        whitespace
+ *      - ESP_ERR_NOT_FOUND, if command with given name wasn't registered
+ *      - ESP_ERR_INVALID_STATE, if esp_console_init wasn't called
+ */
+esp_err_t esp_console_run_with_context(void* invocation_ctx, const char* cmdline, int* cmd_ret);
```
o-marshmallow: Since the context is an optional parameter in this function, I think it'd make more sense to put it at the end

JVKran:
Hi @o-marshmallow, I've just incorporated your feedback. Thanks.

o-marshmallow:
@JVKran During an internal review, I was asked about the objective behind using a dynamic context. Because in theory, when a command is registered we already know what context it will be receiving when invoking the command via the console. The only dynamic behavior the command have would be through the `argv` parameters.

And if let's say you have a request coming from HTTP and another other coming from `MQTT` that is very similar with with different way of replying to (output backends), then why using commands at all rather than regular C functions?

JVKran:
Hi @o-marshmallow,

The objective is the following;
- Let's establish that a command has a name, arguments, values, a result (actions occuring, values being changed, etc.) and a return value.
- Currently, the handling of output isn't part of the console component, which I find weird. The only output is the return value of the handler.
- Let's say a command named ```get_value``` can be called through MQTT, HTTP and ESP-NOW. Currently, there's absolutely no way for the command handler to know where to write the requested value to, other than to print it to UART or USB (while I want the output to be sent to MQTT, HTTP or ESP-NOW).
- I want it to be able to write the output (the requested value) to the source backend.

Practically, in my main task, where MQTT payloads, HTTP requests and ESP-NOW packets are received, I can simply call esp_console_run to execute the command. What I want to establish is that I can provide three functions (with the same signature) as result outputs; ```esp_console_run_with_context(&esp_now_send, etc.)```, ```esp_console_run_with_context(&http_reply, etc.)``` In the command handler I can check for the function being passed and then simply call ```context->write(CUSTOM_PAYLOAD```, regardless of source backend.

This desired functionality isn't possible when the context is set during registration. The context set during registration is static, while I want it to be dynamic; they serve different purposes.

Best regards,

Jochem

o-marshmallow:
@JVKran I see why you would need a dynamic context in your case, but the question is why do you need to use a command for this? 
The main purpose of the commands is to be used from the ESP console, and so invoked after parsing a command line from string (usually typed by a user).
In your case, it seems to me that using a command is not necessary, why not having custom C functions for that?

JVKran:
Because I _literally_ need to expose a command line over these interfaces. I have around 30 commands that I need to expose over MQTT, HTTP, ESP-NOW and BLE-SPP. These interfaces must be available to literally do exactly what the otherwise useless ```esp_console_run``` provides, minus a dynamic context.

I want these interfaces to be exactly the same to the users as the REPL's on UART and USB are, minus the logging itself. If I wouldn't be able to use this solution, I would need to not have 30 functions (for the command handlers), but 30 times the amount of interfaces. That's ridiculous.

[Take Zephyr's shell](https://docs.nordicsemi.com/bundle/ncs-1.9.2-dev1/page/zephyr/reference/shell/index.html#dictionary_commands), for example. They provide a pointer to the shell so one can use shell_printf to print to the source command shell. I have no doubt that many others will be very happy if you were to provide an easier way to register logging backends (not just the printf that can be overriden) and would pass the source shell (the shell through which the command is issued) to the handler. This has always been a shortcoming in my opinion. If not, I feel like this dynamic context is a very reasonable and pragmatic solution. I dislike Zephyr in a lot of ways, but console/shell-wise, it's excellent.

o-marshmallow:
Thanks for the explanation, I understand what you mean. Can you please apply the changes I mentioned above so that I can open an internal merge request? (If you prefer I can also directly apply the changes myself on top of your PR when merging it internally)

JVKran:
Hi @o-marshmallow,

I'm happy you understand. I've incorporated all changes you mentioned above and also used the pre-commit hooks. Is there anything else I can/need to assist you with?

Best regards,

Jochem

o-marshmallow:
sha=5fd9fb9e09b672429362c5be6b40df71424f59d7

o-marshmallow:
@JVKran, I got some internal feedbacks, the console component is going to be deprecated in IDF since it will be completely refactored, so it wouldn't make a lot of sense to add a new feature to maintain. However, there is a simpler modification that would help you achieve what you need: make `find_command_by_name` public.

Thanks to it, you will be able to get the registered `cmd_item_t` command, and thus be able to call its callback with any context you want, that would even allow you to customize argc and argv.

If you think this is acceptable, I will open an internal MR to make `find_command_by_name` public.

o-marshmallow:
Hello @JVKran ,

After discussing internally again about this issue, a simpler solution would be to implement a new function similar to `find_command_by_name` that would let you get a pointer to your command function and invoke it with any parameter you want.

If you think this is acceptable, tell me and we will open an internal MR for this feature.

o-marshmallow:
Closing this issue, feel free to reopen if the solution proposed above is acceptable

PR #17435, opened 2025-08-14
feat(tls): Support runtime ATECC608A I2C address configuration in TLS

TheSciBoy:
## Description

Includes changes in MQTT and MBED-TLS to support setting the I2C address of a secure element during runtime. This is required for systems where it cannot be determined at compile-time which secure element (ATEC608A) is installed on the hardware, something which may be, for example, configured during production instead.

This way, a new option is added in the configuration to instead of selecting which secure element is used in hardware, a runtime selection item is chosen instead. Checks are added that require the I2C address to be actively set (default is 0, which is not a valid address).

## Related

None

## Testing

Testing has only been performed on our own proprietary hardware, with the libraries for our own purposes activated. We build locally in a clean docker environment.

## Checklist

- [ x] 🚨 This PR does not introduce breaking changes.
- [ x] All CI checks (GH Actions) pass.
- [ x] Code is well-commented, especially in complex areas.
- [ x] Git history is clean — commits are squashed to the minimum necessary.

### DIFF
--- a/components/esp-tls/esp_tls.h
+++ b/components/esp-tls/esp_tls.h
@@ -213,6 +213,9 @@ typedef struct esp_tls_cfg {
     const int *ciphersuites_list;           /*!< Pointer to a zero-terminated array of IANA identifiers of TLS ciphersuites.
                                                 Please check the list validity by esp_tls_get_ciphersuites_list() API */
     esp_tls_proto_ver_t tls_version;        /*!< TLS protocol version of the connection, e.g., TLS 1.2, TLS 1.3 (default - no preference) */
+#ifdef CONFIG_ATECC608A_RUNTIME_SELECTION
+    uint8_t atecc608a_i2c_addr;            /*!< I2C address of the atecc608a chip */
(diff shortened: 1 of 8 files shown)


AdityaHPatwardhan:
Hi @TheSciBoy, 
Thanks for the PR, Yes, this is indeed helpful. I think one more addition would be to allow dynamically configurable key slot id as well.
We are discussing currently as to how to progress ahead. ( Either to refactor the config to allow more options, or go with the way presented in this PR).
We will take this up soon and update you accordingly.

TheSciBoy:
> Hi @TheSciBoy, Thanks for the PR, Yes, this is indeed helpful. I think one more addition would be to allow dynamically configurable key slot id as well. We are discussing currently as to how to progress ahead. ( Either to refactor the config to allow more options, or go with the way presented in this PR). We will take this up soon and update you accordingly.

Sounds good! I have been busy with other things since I posted the PR and have not had time to dig into the details of contributing yet (such as the CLA), which is why the PR remains in draft. But if you maybe want to refactor this completely, then I will not continue with the PR until you have a decision.

PR #17462, opened 2025-08-19
fix/feat(uart): uart_read_bytes() timeout fixed, added uart_read_bytes_partial()

BitsForPeople:
## Description

**fix:** `uart_read_bytes()` did not correctly handle the given max_wait timeout; when trying to read N bytes from the UART, it could take up to (N+1) * max_wait ticks to return. The new version returns as soon as max_wait ticks have elapsed.
This change may in theory break applications which implicitly relied on the the given time limit not being honored.

**feat:** New function `uart_read_bytes_partial()` - blocks until _any_ data is read from the RX buffer, not until the given buffer is full. Useful when you don't know in advance exactly how many bytes you expect to receive and you still want to react quickly to received data.

## Testing

Wrote a test application using UART loop-back to verify specifically the timing behavior of both uart_read_bytes() and uart_read_bytes_partial().

---

## Checklist

### DIFF
--- a/components/esp_driver_uart/include/driver/uart.h
+++ b/components/esp_driver_uart/include/driver/uart.h
@@ -555,15 +555,17 @@ int uart_write_bytes(uart_port_t uart_num, const void* src, size_t size);
 int uart_write_bytes_with_break(uart_port_t uart_num, const void* src, size_t size, int brk_len);
 
 /**
- * @brief UART read bytes from UART buffer
+ * @brief Read bytes from the UART RX buffer.
+ * This function blocks until either \p length bytes of data have been read or the given timeout
+ * expires. Only when the timeout expires will this return a value between \c 0 and <tt>length - 1</tt>.
  *
- * @param uart_num UART port number, the max port number is (UART_NUM_MAX -1).
- * @param buf     pointer to the buffer.
- * @param length  data length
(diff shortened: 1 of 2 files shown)


songruo:
> uart_read_bytes() did not correctly handle the given ticks_to_wait timeout
I agree.

> New function uart_read_bytes_partial() - blocks until any data is read from the RX buffer, not until the given buffer is full.
You can utilize UART_DATA event to achieve the same purpose. This is some logic you can do in the application layer, using UART_DATA + ticks counting.

BitsForPeople:
You're right in that you _can_ achieve the same overall behavior as uart_read_bytes_partial() by using UART events. That's a whole different programming paradigm though from simple blocking reads.
You can, and people do, also emulate this behavior in user code by using uart_read_bytes() with an initial buffer length of **1** to wait for _any_ data, but that's a bit clumsy and inefficient.

I'd argue that `uart_read_bytes_partial()` actually does what people usually expect `uart_read_bytes()` to do, before they realize that it doesn't. At least that's what a) I'd expect it to do, and b) people somewhat regularily ask for in the forums.

BitsForPeople:
Any updates on the review?

balbet:
It does some long time we are waiting to solve the issue in our devices, is there any update on this? Thanks

Issue #17474, opened 2025-08-21
I2C synchronization Issue (IDFGH-16347)


yazeed-Alshorman:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.2.2

### Espressif SoC revision.

ESP32

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

PowerShell

### Development Kit.

ESP32_WROOM_32E

### Power Supply used.

External 3.3V

### What is the expected behavior?

When the I²C bus is initialized correctly with valid SDA/SCL pins, pull-ups, and a powered device present at a valid 7-bit address, calls such as i2c_master_transmit_receive() should complete the transaction (START → slave address → data → repeated START → read → STOP) and return ESP_OK. In this expected behavior, the slave acknowledges both the address and data phases, no “I2C hardware NACK detected” messages appear in the log, and the function reliably returns the requested data.

### What is the actual behavior?

Even with the I²C bus initialized correctly and the slave device powered and connected, calls to i2c_master_transmit_receive() frequently fail. The driver reports errors such as “I2C hardware NACK detected” and “I2C transaction unexpected NACK detected”. The transaction aborts, the function returns an error instead of ESP_OK, and no valid data is read from the device.
I got this messages:
E (15:35:47.356) i2c.master: I2C hardware NACK detected
E (15:35:47.480) i2c.master: I2C transaction unexpected nack detected
E (15:35:47.564) i2c.master: s_i2c_synchronous_transaction(881): I2C transaction failed
E (15:35:47.675) i2c.master: i2c_master_transmit_receive(1123): I2C transaction failed

### Steps to reproduce.

Initialize the I²C bus with the new ESP-IDF I²C master driver:

```
i2c_master_bus_config_t i2c_mst_config = {
    .clk_source = I2C_CLK_SRC_DEFAULT,
    .i2c_port = I2C_NUM_0,
    .scl_io_num = SCL_GPIO,
    .sda_io_num = SDA_GPIO,
    .glitch_ignore_cnt = 7,
    .flags.enable_internal_pullup = true,
};
[...]

PR #17507, opened 2025-08-29
fix(i2c_master): use atomic operations for event (IDFGH-16368)

raoulh:
## Description

After searching for i2c issues on my board I got a lot of NACK related issues on the console.

```
E (2509388) i2c.master: I2C hardware NACK detected
E (2509388) i2c.master: I2C transaction unexpected nack detected
```

I found out that the i2c driver had a potential bug in the event handling because `i2c_master->event` was not atomic, unlike the status. This creates a race condition where:

- The ISR sets the atomic status to I2C_STATUS_ACK_ERROR.
- But `i2c_master->event` can be overwritten by another thread before being sent to the queue.
- The main thread receives an incorrect or corrupted event.

This PR implements `event` as atomic and prevents race condition and queue corruption.

## Related

Potential fix for #17474

## Testing

Tested on my board, NACK are better handled and I have no more lots of error in my console. 

---

## Checklist

### DIFF
--- a/components/esp_driver_i2c/i2c_master.c
+++ b/components/esp_driver_i2c/i2c_master.c
@@ -535,7 +535,7 @@ static void s_i2c_send_commands(i2c_master_bus_handle_t i2c_master, TickType_t t
 
     // Initialise event queue.
     xQueueReset(i2c_master->event_queue);
-    i2c_master->event = I2C_EVENT_ALIVE;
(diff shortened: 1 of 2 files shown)


BitsForPeople:
The simple use of atomic store and load cannot prevent any kind of lost-update problem.

raoulh:
You are totally right. I'm going to rewrite it differently in a new PR

PR #17568, opened 2025-09-10
Add uart_read_some_bytes() to allow reading into a buffer without waiting until that buffer is full

0xFEEDC0DE64:
The espressif uart api is quite weird to only read some data into my buffer but not wait until its full (I need this for modbus).
My new api fixes this by allowing to do partial buffer reads from uart. Timeout will only expire, if 0 bytes are received

### DIFF
--- a/components/esp_driver_uart/include/driver/uart.h
+++ b/components/esp_driver_uart/include/driver/uart.h
@@ -556,6 +556,7 @@ int uart_write_bytes_with_break(uart_port_t uart_num, const void* src, size_t si
 
 /**
  * @brief UART read bytes from UART buffer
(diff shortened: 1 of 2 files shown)


BitsForPeople:
Already under way in #17462 :)

0xFEEDC0DE64:
@BitsForPeople your diff looks way more complicated and also touches other functions besides the newly added, did I make a mistake in my implementation?

BitsForPeople:
@0xFEEDC0DE64  No, looks ok too and should in practice be pretty much functionally equivalent (except for the timeout thing I also fixed in that PR).

PR #17586, opened 2025-09-14
Fix warning message in install.sh. (IDFGH-16454)

vagran:
Running `install.sh` with bad target name causes this message:
```
WARNING: Targets: "esp_debug_backend" are not supported. Only allowed options are: {", ".join(targets_from_tools_json)}.
```


dobairoland:
sha=e7dcb9bb84addcc9b1a0cac16e3a03910bcf5f47

dobairoland:
@vagran Thank you for the contribution!

PR #17636, opened 2025-09-22
Show stacktrace of task which failed timeout in task watchdog (IDFGH-16499)

0xFEEDC0DE64:
This change adds a stacktrace of tasks that fail the task watchdog timeout.

Too often I had timeouts in the main task but then the stacktrace only shows methods of task watchdog but not of main. This is completely useless.

What exactly in main halted execution so that the watchdog timed out?

This PR adds exactly this

image

### DIFF
--- a/components/esp_system/include/esp_debug_helpers.h
+++ b/components/esp_system/include/esp_debug_helpers.h
@@ -18,6 +18,9 @@ extern "C" {
 #include "esp_err.h"
 #include "esp_cpu.h"
 
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+
 /*
  * @brief   Structure used for backtracing
  *
@@ -128,6 +131,8 @@ esp_err_t esp_backtrace_print(int depth);
  *      - ESP_FAIL  One or more backtraces are corrupt
  */
 esp_err_t esp_backtrace_print_all_tasks(int depth);
+
+esp_err_t esp_backtrace_print_task(TaskHandle_t pxTask, int depth, bool panic);
 
 /**
  * @brief Set a watchpoint to break/panic when a certain memory range is accessed.
--- a/components/esp_system/port/arch/xtensa/debug_helpers.c
+++ b/components/esp_system/port/arch/xtensa/debug_helpers.c
@@ -250,3 +250,70 @@ ipc_err:
 malloc_err:
     return ret;
 }
+
+esp_err_t esp_backtrace_print_task(TaskHandle_t pxTask, int depth, bool panic)
+{
+    esp_err_t ret = ESP_OK;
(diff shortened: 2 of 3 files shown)


0xFEEDC0DE64:
Is this some new AI bullshit?
Ok im done, I will never pull-request any of my improvements to esp-idf again if you start wasting my time so much.
Having CI to check on commit messages but not even on the code itself is already quite stupid on its own.

If any of the comments created by AI are incorrect this will be my last improvement shared for sure.

ESP-Marius:
@0xFEEDC0DE64 sorry that the BOT annoyed you. It is a trial to see if it can indeed provide anything useful or not. If we find that it is mostly just spamming useless comments we will deactivate it.

Just trying to understand your situation a bit better: am I correct in that you are using `esp_task_wdt_add` to watch your main task? Because in those cases the backtrace will indeed not print much useful stuff.

0xFEEDC0DE64:
yes correct. my main task kept hitting the watchdog and I need its stacktrace to eliminate blocking operations.

and for the bot: I think there is general consent that AI is only a big waste of time and many other places forbid AI generated content completely. I think you should stop this bot from demotivating other developers asap

meetpatelvgec:
Hi @0xFEEDC0DE64 

I have carefully reviewed your suggestion and code changes in your PR, and also discussed it with our team. We really appreciate your efforts for submitting the changes to make ESP-IDF better for everyone.

Looking at the original issue that you reported and some previous comments, I can see that you are getting the task watchdog timeout in the main task, but the CPU0 backtrace as well as current running task shows info of IDLE task because probably the main task is waiting for some resource or completing a delay, and IDLE task is running in the meantime.

In such scenarios, there are some inbuilt debug features of ESP-IDF that you can use to find out the reason for WDT timeout as explained below:

1. **esp_task_wdt API**
esp_task_wdt APIs are explained in [watchdog documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/wdts.html) as well as in example program (examples/system/task_watchdog). With these APIs, you can monitor not just individual tasks that are responsible for wdt timeout, but also register multiple functions inside these tasks to find out which function takes too long to execute and then optimise it.

2. **Core dump**
Let's say if you still need the backtrace information of the task which caused watchdog timeout, you can use [core dump](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/core_dump.html) to print backtrace and detailed information of all the tasks that were running when a problem happened. You can refer one of our test app to understand how to use it: tools/test_apps/system/panic

We believe the above mentioned debug tools should be more than enough to debug task watchdog timeout issues that the users face. Hence making additional changes as given in your PR will be redundant. So unfortunately, we won't be able to take in your changes, and hope that you understand the justification.

PR #17641, opened 2025-09-23
Allow the https server to request client certs only with OPTIONAL (IDFGH-16506)

0xFEEDC0DE64:
I want to serve my webinterface via https to smartphones and the manufacturer should have special permissions and we authenticate developers or production using client key / cert. The end customer will not have such a cert. The connection should not fail, but our internal APIs will then refuse to work.

### DIFF
--- a/components/esp-tls/esp_tls_mbedtls.c
+++ b/components/esp-tls/esp_tls_mbedtls.c
@@ -694,6 +694,8 @@ static esp_err_t set_server_config(esp_tls_cfg_server_t *cfg, esp_tls_t *tls)
         if (esp_ret != ESP_OK) {
             return esp_ret;
         }
+        if (cfg->cacert_authmode_optional)
+            mbedtls_ssl_conf_authmode(&tls->conf, MBEDTLS_SSL_VERIFY_OPTIONAL);
     } else {
 #ifdef CONFIG_ESP_TLS_SERVER_MIN_AUTH_MODE_OPTIONAL
         mbedtls_ssl_conf_authmode(&tls->conf, MBEDTLS_SSL_VERIFY_OPTIONAL);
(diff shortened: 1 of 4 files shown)


mahavirj:
@0xFEEDC0DE64 

We have a configuration knob for this here:
https://github.com/espressif/esp-idf/blob/5c5eb99eabc7d3e7f704656ffe9ebf8f5a77afbd/components/esp-tls/Kconfig#L63

Please see if that can help!

Edit: Probably we should be using this config in the `if` block where `cacert_buf` is not NULL CC @Ashish285

0xFEEDC0DE64:
Hi thanks for sharing, but it does not help, since this completely breaks the security of my mqtt broker that verifies clients. I only want it optional for the https server

0xFEEDC0DE64:
Do you think just offering this option all the time is a risk? If applications don't set it, behaviour should stay the same. That's why its set to true to reduce the cert checking.

Ashish285:
Hi @0xFEEDC0DE64 , My suggestion to have compile time switch is to enforce the configuration (as is the case currently). Along with the runtime check that you have enabled, we can cater to all different use cases. Having a compile time switch also helps to enforce a secure profile.

Let me know your thoughts, we will be happy to include this PR.

components/esp-tls/esp_tls_mbedtls.c
```diff
@@ -694,6 +694,8 @@ static esp_err_t set_server_config(esp_tls_cfg_server_t *cfg, esp_tls_t *tls)
         if (esp_ret != ESP_OK) {
             return esp_ret;
         }
+        if (cfg->cacert_authmode_optional)
+            mbedtls_ssl_conf_authmode(&tls->conf, MBEDTLS_SSL_VERIFY_OPTIONAL);
```
hurricanefrog, line 698: "External opinion": I would suggest adding curly braces here, to stay consistent with the rest of the code. (Which also adds curly braces even for single statements, at least as far as I can see in the context).

Ashish285:
Hi @0xFEEDC0DE64 , thanks for the PR. I will accept it and make any changes regarding the compile time switches internally if required.

Ashish285:
sha=6213a906c049d08f5bc1a372aad1577dee26fdf1

PR #17657, opened 2025-09-25
ESP_GATT_MAX_ATTR_LEN = 512

nebkat:
## Description

The maximum attribute length as per the BLE GATT spec is 512 - see https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/attribute-protocol--att-.html#UUID-d24e0156-618e-728d-f19b-01349ced952b:~:text=The%20maximum%20length%20of%20an%20attribute%20value%20shall%20be%20512%20octets.

---

## Checklist

### DIFF
--- a/components/bt/host/bluedroid/api/include/api/esp_gatt_defs.h
+++ b/components/bt/host/bluedroid/api/include/api/esp_gatt_defs.h
@@ -478,7 +478,7 @@ typedef uint8_t esp_gatt_char_prop_t;
  *
  * This definition specifies the maximum number of bytes that a GATT attribute can hold.
  */
-#define ESP_GATT_MAX_ATTR_LEN   517 /*!< As same as GATT_MAX_ATTR_LEN. */
+#define ESP_GATT_MAX_ATTR_LEN   512 /*!< As same as GATT_MAX_ATTR_LEN. */
 
 /**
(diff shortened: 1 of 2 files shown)


esp-zhp:
We cannot change the max attribute length to 512, as it would cause other issues.

esp-zhp:
we encountered a data loss issue, which we found was related to setting the max ATT length to 512. May I ask what impact setting it to 517 would have on you?

nebkat:
@esp-zhp unless I have completely misunderstood the BLE spec I believe it is invalid for any attribute to contain a _value length_ greater than 512, even if the MTU can in theory be up to 517.

If there is data loss occurring with a 512 max it is presumably because something is trying to send an attribute with length >512? 

We must strictly adhere to the standard as otherwise another BLE stack may receive our packets and experience data loss on their end as they expect a max length of 512 bytes. Alternatively if this data loss is occurring when receiving data from another device, it is the fault of that device for sending a packet that is too large.

esp-zhp:
For maximum compatibility on ESP, we retain data even when its length exceeds 512 bytes.
If ESP attempts to send data larger than 512 bytes, a warning will be provided to help ensure that the ESP32 sends data in accordance with the BLE specification. At the same time, the system is capable of handling cases where the data exceeds 512 bytes.

PR #17742, opened 2025-10-17
fix(i2s_hal): fix typo in PDM2PCM check (IDFGH-16642)

agocsdaniel:
I run into a bug while trying to use I2S HAL. It seems that there is a typo in one of the macros being checked. I checked that no other places mention the typoed macro.


suda-morris:
sha=a047b6af247c5bdad5a6741f2a3ee57d8ad62a1b

suda-morris:
merged in https://github.com/espressif/esp-idf/commit/041130244cf7139d7a0aa258b82a59cf136cc975

PR #17752, opened 2025-10-20
fix: indexing in DNS server IP address retrieval #17751 (IDFGH-16654)

### LINKED ISSUE #17751: Bug in DNS server: Incorrect array indexing when checking IP address (IDFGH-16653)
### Answers checklist.

### General issue report

https://github.com/espressif/esp-idf/blob/ab149384e12ace1b8e8f7858d52ae47ab8af4563/examples/protocols/http_server/captive_portal/components/dns_server/dns_server.c#L161

The DNS server does not use the correct entry index when checking the IP address.
```c
h->entry->ip.addr 
```
instead of 
```c
h->entry[i].ip.addr
```
which causes it to always check entry 0 regardless of which entry matched in the loop.

This bug prevents proper DNS resolution when multiple entries are configured and the matching entry is not at index 0.

jojo2massol:
## Description
Fix #17751 :

Bug in DNS server: Incorrect array indexing when checking IP address (IDFGH-16653)

### Answers checklist.

https://github.com/espressif/esp-idf/blob/ab149384e12ace1b8e8f7858d52ae47ab8af4563/examples/protocols/http_server/captive_portal/components/dns_server/dns_server.c#L161

The DNS server does not use the correct entry index when checking the IP address.
```c
h->entry->ip.addr 
```
instead of 
```c
h->entry[i].ip.addr
```
which causes it to always check entry 0 regardless of which entry matched in the loop.

This bug prevents proper DNS resolution when multiple entries are configured and the matching entry is not at index 0.

## Checklist

### DIFF
--- a/examples/protocols/http_server/captive_portal/components/dns_server/dns_server.c
+++ b/examples/protocols/http_server/captive_portal/components/dns_server/dns_server.c
@@ -158,7 +158,7 @@ static int parse_dns_request(char *req, size_t req_len, char *dns_reply, size_t 
                         esp_netif_get_ip_info(esp_netif_get_handle_from_ifkey(h->entry[i].if_key), &ip_info);
                         ip.addr = ip_info.ip.addr;
                         break;
-                    } else if (h->entry->ip.addr != IPADDR_ANY) {
+                    } else if (h->entry[i].ip.addr != IPADDR_ANY) {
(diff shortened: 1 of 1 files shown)


mahavirj:
sha=b801fa0f7086d212e2443f67af05a64efa9855bf

mahavirj (approved):

PR #17760, opened 2025-10-22
fix(esp-tls): clean up is_tls flag (IDFGH-16662)

erkia:
## Description

Commit https://github.com/espressif/esp-idf/commit/2dd280f1264bb8c77a05930c81e8890395867ad1 added a neat feature to support STARTTLS by doing something like this:
```c
esp_tls_t *ftp;
esp_tls_cfg_t cfg;

memset(&cfg, 0, sizeof(cfg));

ftp = esp_tls_init();

cfg.is_plain_tcp = true;
esp_tls_conn_new_sync("example.com", strlen("example.com"), 21, &cfg, ftp);

// ... AUTH TLS or STARTTLS or what not ...

cfg.is_plain_tcp = false;
esp_tls_set_conn_state(ftp, ESP_TLS_CONNECTING);
esp_tls_conn_new_sync("example.com", strlen("example.com"), 21, &cfg, ftp);
```
However, when doing this, `tls->is_tls` flag is never set to `true` internally ([as it is only set in ESP_TLS_INIT state](https://github.com/espressif/esp-idf/blob/ab149384e12ace1b8e8f7858d52ae47ab8af4563/components/esp-tls/esp_tls.c#L473)), which prevents calling `mbedtls_net_free()` [here](https://github.com/espressif/esp-idf/blob/ab149384e12ace1b8e8f7858d52ae47ab8af4563/components/esp-tls/esp_tls_mbedtls.c#L461C17-L461C33).

Luckily, despite its name, `mbedtls_net_free()` does not free anything and only does a graceful shutdown of the socket, so it is not a big issue (socket is closed by esp-tls anyway). But as this is the only usecase for `tls->is_tls` flag, then it makes sense to clean it up, fixing the small inconsistency in the process.

### DIFF
--- a/components/esp-tls/esp_tls.c
+++ b/components/esp-tls/esp_tls.c
@@ -468,15 +468,12 @@ static int esp_tls_low_level_conn(const char *hostname, int hostlen, int port, c
     switch (tls->conn_state) {
     case ESP_TLS_INIT:
         tls->sockfd = -1;
-        if (cfg != NULL && cfg->is_plain_tcp == false) {
-            _esp_tls_net_init(tls);
-            tls->is_tls = true;
-        }
+        _esp_tls_net_init(tls);
         if ((esp_ret = tcp_connect(hostname, hostlen, port, cfg, tls->error_handle, &tls->sockfd)) != ESP_OK) {
             ESP_INT_EVENT_TRACKER_CAPTURE(tls->error_handle, ESP_TLS_ERR_TYPE_ESP, esp_ret);
             return -1;
         }
-        if (tls->is_tls == false) {
+        if (cfg && cfg->is_plain_tcp) {
             tls->read = tcp_read;
             tls->write = tcp_write;
             ESP_LOGD(TAG, "non-tls connection established");
--- a/components/esp-tls/esp_tls_mbedtls.c
+++ b/components/esp-tls/esp_tls_mbedtls.c
@@ -456,12 +456,10 @@ void esp_mbedtls_conn_delete(esp_tls_t *tls)
 {
     if (tls != NULL) {
         esp_mbedtls_cleanup(tls);
-        if (tls->is_tls) {
-            if (tls->server_fd.fd != -1) {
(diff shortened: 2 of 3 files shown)


components/esp-tls/esp_tls.c
```diff
@@ -468,15 +468,12 @@ static int esp_tls_low_level_conn(const char *hostname, int hostlen, int port, c
     switch (tls->conn_state) {
     case ESP_TLS_INIT:
         tls->sockfd = -1;
-        if (cfg != NULL && cfg->is_plain_tcp == false) {
-            _esp_tls_net_init(tls);
-            tls->is_tls = true;
-        }
+        _esp_tls_net_init(tls);
```
erkia, line 471: `_esp_tls_net_init()` is called by `esp_tls_init()` anyway, no matter if plain or TLS. 

It's only job is to set `tls->server_fd.fd = -1;` for mbedtls. For wolfssl, it is no-op. So it is safe to call it here always, especially, as `tls->sockfd = -1;` is forced also.

Ashish285:
Hi @erkia , thanks for reporting and creating this issue.


Will setting the `tls->is_tls` flag inside `ESP_TLS_CONNECTING` state a better fix for this issue? The `mbedtls_net_free()` might not be doing much right now but it may in the future so we would probably want to keep that API for tls connections.

erkia:
> Hi @erkia , thanks for reporting and creating this issue.
> 
> Will setting the `tls->is_tls` flag inside `ESP_TLS_CONNECTING` state a better fix for this issue? The `mbedtls_net_free()` might not be doing much right now but it may in the future so we would probably want to keep that API for tls connections.

`mbedtls_net_init()` is called unconditionally by `esp_tls_init()` anyway, so it makes no sense to call `mbedtls_net_free()` conditionally.

Ashish285:
> mbedtls_net_init() is called unconditionally by esp_tls_init() anyway


Sure, I believe that's also not the correct way as it should be initialised only when tls will be used.

PR #17797, opened 2025-10-30
feat(esp_event): Allow an event carry more data without malloc (IDFGH-16709)

jimmyw:
This is both a feature and an optimization.

Feature:
Adjustable size of the internal storage in esp_event queue, currently used by ISR posting, as they wont be able to make a malloc.

Optimization:
When non-isr is posting an event, use the inernal storage in the struct instead of always allocating a new heap for the data. Most events in esp-idf only contains a few bytes event information, and we have that allocation payed for anyway.

This solved in a big part our memory fragmentation issue, as events happens freqvently and used to create small memory allocations for just 4 bytes, and then in the event handler we usually allocated a bigger chunk of heap for our feature. When returning from the event handler, the 4 byte allocation was freed, leaveing a hole in the heap.

### DIFF
--- a/components/esp_event/Kconfig
+++ b/components/esp_event/Kconfig
@@ -14,6 +14,14 @@ menu "Event Loop Library"
         help
             Enable posting events from interrupt handlers.
 
+    config ESP_EVENT_POST_FROM_ISR_SIZE
+        int "Max size of data from events"
+        default 4
+        depends on ESP_EVENT_POST_FROM_ISR
+        help
+            Enable posting events from interrupt handlers.
+
+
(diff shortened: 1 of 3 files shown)


KonstantinKondrashov (approved):
LGTM!

KonstantinKondrashov:
sha=6f552182c6622bf8f2a7aab096ed01cd7e658d1a

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/e2a6653680e3a8e2fa0d259ca428352c6f7b3662.

Issue #17773, opened 2025-10-26
Stop pushing compiler flags on users (IDFGH-16679)


higaski:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

I've spent too much time lately adopting projects to Espressif's misuse of CMake. Instead of setting per target compiler options you force your own set of flags on everyone using ESP-IDF. This is not how this is supposed to work. ESP-IDF is a framework and not an application, you can't just set all your options globally and force them on whatever comes down the dependency tree...

A great example of this is what I have just encountered
```sh
error: initializer-string for array of 'char' truncates NUL terminator but destination lacks 'nonstring' attribute (17 chars into 16 available) [-Werror=unterminated-string-initialization]
   63 | static const char sigma[16] = "expand 32-byte k";
```

Yes this definition is missing the `nonstring` attribute, but it's C code written probably before Espressif has been founded and it's not going to change.

higaski:
It's even worse than I thought. ESP-IDF also breaks `PROJECT_IS_TOP_LEVEL` which has been my go-to solution for conditionally setting compiler flags on CMake projects or not. This also means that the `ESP_PLATFORM` variable leaks... because it must be checked in addition to `PROJECT_IS_TOP_LEVEL`...

This is bad... really bad.

/edit
Maybe this issue should be renamed in "ESP-IDF is fundamentally incompatible with CMake".

igrr:
@higaski IDF sets compiler flags for all the "IDF components". If you have a CMake library that isn't linked to an IDF component (as in, "idf_as_lib" type project), it shouldn't get compiler flags from IDF, except the ones set in the toolchain file.

> Maybe this issue should be renamed in "ESP-IDF is fundamentally incompatible with CMake".

I think it could be, but that issue has already been largely discussed in https://github.com/espressif/esp-idf/issues/10579. The existing build system won't change, it would break too many projects. However the new IDF build system which is coming to IDF v6 will allow you to develop your project as a "standard CMake" project, without suffering all the downsides of "idf_as_lib" (i.e. lack of monitor, menuconfig, etc.)

[...]

PR #17830, opened 2025-11-05
fix(tools): set *_IS_TOP_LEVEL cmake values at project() calls (IDFGH-16745)

benedekkupper:
## Description

As the comment at line 607 explains what's done here:
> Set the variables that project() normally sets

However, with [cmake 3.21](https://cmake.org/cmake/help/v3.21/command/project.html), two more variables are added to this list:
[PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/v3.21/variable/PROJECT_IS_TOP_LEVEL.html#variable:PROJECT_IS_TOP_LEVEL),
[\<PROJECT-NAME\>_IS_TOP_LEVEL](https://cmake.org/cmake/help/v3.21/variable/PROJECT-NAME_IS_TOP_LEVEL.html#variable:%3CPROJECT-NAME%3E_IS_TOP_LEVEL)

When these aren't set correctly, builds fail in various ways, with rather vague error messages:
https://github.com/IntergatedCircuits/esp-idf-examples/actions/runs/19098336995/job/54563853753

With this change added, these variables are set as expected.

The condition used to set the value for these variables might be subject to change, I'm open for adjusting it as needed.

## Checklist

---

> [!NOTE]
> Sets `PROJECT_IS_TOP_LEVEL` and `<PROJECT-NAME>_IS_TOP_LEVEL` in `tools/cmake/project.cmake` for CMake >= 3.21 based on whether the project is top-level.
> 
> - **Build/CMake**:
>   - Update `tools/cmake/project.cmake` to set `PROJECT_IS_TOP_LEVEL` and `<PROJECT-NAME>_IS_TOP_LEVEL` (CMake >= `3.21`) after the `project()` call, based on `CMAKE_CURRENT_SOURCE_DIR == CMAKE_SOURCE_DIR`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9c61f0403cef79988e9218d0b0601237b52e194. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tools/cmake/project.cmake
```diff
@@ -632,6 +632,15 @@ macro(project project_name)
         set(${PROJECT_NAME}_VERSION_TWEAK "${${PROJECT_NAME}_VERSION_TWEAK}" PARENT_SCOPE)
         set(${PROJECT_NAME}_DESCRIPTION "${${PROJECT_NAME}_DESCRIPTION}" PARENT_SCOPE)
         set(${PROJECT_NAME}_HOMEPAGE_URL "${${PROJECT_NAME}_HOMEPAGE_URL}" PARENT_SCOPE)
+        if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.21")
+            if((CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME) AND (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR))
```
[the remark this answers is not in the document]
benedekkupper: I have updated the change as requested.

tools/cmake/project.cmake
```diff
@@ -632,6 +632,15 @@ macro(project project_name)
         set(${PROJECT_NAME}_VERSION_TWEAK "${${PROJECT_NAME}_VERSION_TWEAK}" PARENT_SCOPE)
         set(${PROJECT_NAME}_DESCRIPTION "${${PROJECT_NAME}_DESCRIPTION}" PARENT_SCOPE)
         set(${PROJECT_NAME}_HOMEPAGE_URL "${${PROJECT_NAME}_HOMEPAGE_URL}" PARENT_SCOPE)
+        if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.21")
+            if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+                set(PROJECT_IS_TOP_LEVEL ON PARENT_SCOPE)
```
fhrbata, line 637: I think this does not mimic how cmake sets `PROJECT_IS_TOP_LEVEL`.  Specifically 

> The variable value will be true in:
   a directory added by add_subdirectory() that does not also contain a project() call`

as described in [docs](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html)

benedekkupper: There's also a change in the documentation between 3.23 and 3.24, so if we want to handle these variables completely up to spec, this should be reflected too. I'm a bit out of my depth here, do you have a proposal for the condition here?

fhrbata: > There's also a change in the documentation between 3.23 and 3.24, so if we want to handle these variables completely up to spec, this should be reflected too.

Ah, nice catch, I haven't noticed the change between 3.23 and 3.24. 

> I'm a bit out of my depth here, do you have a proposal for the condition here?

TBH I'm not sure neither. I guess it would be best to fully understand how the [IsRootMakefile()](https://github.com/Kitware/CMake/blob/master/Source/cmProjectCommand.cxx#L145) function, used to set the `PROJECT_IS_TOP_LEVEL` variable, is implemented. I tried to take a look, but I would definitely need more time to understand the logic.

My worry is that with this change, which does not fully mimic cmake's behavior, we may fix problems for one set of people, but at the same time, we may change the expected behavior for others and break their components. At this point, I'm not sure this issue can be fully resolved, but I could be completely off here.

We are currently working on a [next version of the build system](https://github.com/espressif/esp-idf/issues/17833), which is currently available as technical preview and it does not have this issue.

Thank you very much for looking into this problem!

benedekkupper: Does that mean that there won't be further releases from 5.5? What is the timeline for releasing this new build system?

dobairoland: > Does that mean that there won't be further releases from 5.5?

Please see https://github.com/espressif/esp-idf?tab=readme-ov-file#esp-idf-release-support-schedule. We will release bugfix releases for v5.5 up until 2028. All PRs should be targeted to the master branch and we can backport them to v5.5 if possible (patch is without breaking changes).

The build system what you know in v5.5 will be available in v6.0 as well. This is referred as cmakev1 in https://github.com/espressif/esp-idf/issues/17833. cmakev2 is available from v6.0 in preview state. cmakev2 will became the default build some later version of ESP-IDF. I'm unable to tell when will this happen.

benedekkupper: I still think it is better to fix this in 5.5 one way or another, I don't believe that anybody is depending on the current wrong behavior and gets a worse experience with this fix.

fhrbata: Hello @benedekkupper ,

> I still think it is better to fix this in 5.5 one way or another, I don't believe that anybody is depending on the current wrong behavior and gets a worse experience with this fix.

This build system quirk is present for a long time and people may have worked around it already. For example there is another issue #17773 and the [ESP_PLATFORM](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/build-system.html#preset-component-variables) variable was used as a workaround(`if(PROJECT_IS_TOP_LEVEL AND NOT ESP_PLATFORM)`). Maybe this is something that could be used for your use case too. Note I don't think that your change would break this particular workaround, but we cannot be sure how the `PROJECT_IS_TOP_LEVEL`  may be used in other components.

IIUC with the change you proposed we would actually set the `PROJECT_IS_TOP_LEVEL`  variable always to `OFF` for components, which is now always `ON`. As stated above, I think this may break components which already worked around this unfortunate behavior.

I agree that it would be best to fix this, but I'm not sure if replacing one quirky behavior with another is the ideal solution. I of course can be wrong and maybe others have a different opinion.

Thank you very much

PR #17880, opened 2025-11-16
feat(tools): allow to specify python binary by ESP_PYTHON_CUSTOM environment… (IDFGH-16803)

garywill:
## Description

If environment variable `ESP_PYTHON_CUSTOM` already set, `detect_python.sh` uses that python binary, otherwise, looks for python version same or greater than minimal required version on path.

That allow user to choose a particular python version or python binary

Notice: 
1. Assuming no Python 4 (as the code has always been)
2. I don't know how we should deal with when user specifies a large version of python which not exist in our list, for example python 3.19. Assuming allow. Maintainers should decide.

## Testing

I did some simple test with bash

---

## Checklist

---

> [!NOTE]
> Adds `ESP_PYTHON_CUSTOM` support to select a specific Python and makes install/export scripts fail fast on Python detection errors; docs updated (EN/ZN).
> 
> - **Tools/Scripts**:
>   - **Python detection**: Refactor `tools/detect_python.sh` and `tools/detect_python.fish` to honor `ESP_PYTHON_CUSTOM`, select the first supported interpreter, export `ESP_PYTHON`, and return non-zero with stderr on failure.
>   - **Install/Export scripts**: Update `install.sh`, `install.fish`, `export.sh`, `export.fish` to source detection with `|| exit/return 1` and proceed using `"$ESP_PYTHON"`.
> - **Docs**:
>   - Mention `ESP_PYTHON_CUSTOM` usage in `docs/en/*` and `docs/zh_CN/*` install/setup guides with example (`export ESP_PYTHON_CUSTOM=/opt/bin/python3.13`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d3eee7d7c0ae612e17f7a67f9244bcf701b4ad0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/detect_python.fish
+++ b/tools/detect_python.fish
@@ -31,7 +31,7 @@ test -n "$ESP_PYTHON"; or return 1
 
 $ESP_PYTHON --version
 echo "$ESP_PYTHON has been detected"
-set -x ESP_PYTHON "$ESP_PYTHON"
+set -gx ESP_PYTHON "$ESP_PYTHON"
 
 end
 detect_python # Make sure at last line call function, because outside caller need its return value


fhrbata:
Hello @garywill ,

thank you for your contribution. I generally have no objections, but there is one issue that might cause problems. With these changes, the `tools/detect_python.sh` and consequently the `export.sh` scripts become non-POSIX compliant due to the use of `[[` and arrays. I believe we want to keep `export.sh` POSIX compliant so it can be used with as many shells as possible.

```bash
$ shellcheck --shell=sh export.sh tools/detect_python.sh 

In tools/detect_python.sh line 14:
if [[ -z "$ESP_PYTHON" ]]; then
   ^--------------------^ SC3010 (warning): In POSIX sh, [[ ]] is undefined.


In tools/detect_python.sh line 15:
    PYTHON_CANDIDATES=(python3 python python3.10 python3.11 python3.12 python3.13)
                      ^-- SC3030 (warning): In POSIX sh, arrays are undefined.


In tools/detect_python.sh line 18:
    PYTHON_CANDIDATES=("$ESP_PYTHON")
                      ^-------------^ SC3030 (warning): In POSIX sh, arrays are undefined.


In tools/detect_python.sh line 22:
for p_cmd in "${PYTHON_CANDIDATES[@]}"; do
              ^---------------------^ SC3054 (warning): In POSIX sh, array references are undefined.


In tools/detect_python.sh line 34:
[[ -n "$ESP_PYTHON" ]] &&  "$ESP_PYTHON" --version 2>/dev/null || {
^--------------------^ SC3010 (warning): In POSIX sh, [[ ]] is undefined.
                       ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.

For more information:
  https://www.shellcheck.net/wiki/SC3010 -- In POSIX sh, [[ ]] is undefined.
  https://www.shellcheck.net/wiki/SC3030 -- In POSIX sh, arrays are undefined.
  https://www.shellcheck.net/wiki/SC3054 -- In POSIX sh, array references are...

```

Perhaps we can use `read` and a here document to replace the array, something like this.

```sh
OLDEST_PYTHON_SUPPORTED_MAJOR=3
OLDEST_PYTHON_SUPPORTED_MINOR=10

PYTHON_CANDIDATES="python3
python
python3.10
python3.11
python3.12
python3.13"

if [ -n "$ESP_PYTHON" ]; then
    echo "Reading ESP_PYTHON from environment: \"$ESP_PYTHON\""
    PYTHON_CANDIDATES="$ESP_PYTHON"
    unset ESP_PYTHON
fi

while IFS= read -r p_cmd; do
    echo "Checking \"$p_cmd\" ..."

    "$p_cmd" --version >/dev/null 2>&1 || continue
    "$p_cmd" -c "import sys; exit(1) if sys.version_info.major < int(\"$OLDEST_PYTHON_SUPPORTED_MAJOR\") else exit(0);" || continue
    "$p_cmd" -c "import sys; exit(1) if sys.version_info.minor < int(\"$OLDEST_PYTHON_SUPPORTED_MINOR\") else exit(0);" || continue

    ESP_PYTHON="$p_cmd"
    break
done << EOF
$PYTHON_CANDIDATES
EOF

if [ -z "$ESP_PYTHON" ]; then
    echo "Python ${OLDEST_PYTHON_SUPPORTED_MAJOR}.${OLDEST_PYTHON_SUPPORTED_MINOR}+ is not installed! Please see the documentation for how to install it." >&2
    exit 1
fi

echo "\"$ESP_PYTHON\" has been detected"
```

Or maybe a different approach to ensure the script remains POSIX compliant.

Thank you!

garywill:
I make it POSIX compatitable, and did some changes:
1. Use `ESP_PYTHON_CUSTOM` as user custom variable name, instead of internal `ESP_PYTHON.      Because if we 'unset ESP_PYTHON' during python binary check,  it may break user's custom environment.    So we should separate those two variables
2. Found a bug that has been in your original code logic: `exit 1` kills the process running `detect_python.sh`. If user is running `. export.sh` from terminal, user's terminal shell will be killed. Fix this bug by wrapping code into function (haven't added indent yet).

fhrbata:
> I make it POSIX compatitable, and did some changes:

Thank you, it's much appreciated!

> 1. Use `ESP_PYTHON_CUSTOM` as user custom variable name, instead of internal `ESP_PYTHON.      Because if we 'unset ESP_PYTHON' during python binary check,  it may break user's custom environment.    So we should separate those two variables

I agree, but not because "it may break the user's custom environment". This is an internal variable, and it has always been explicitly set to `python` with `ESP_PYTHON=python`. So, in my opinion, it's fine to unset it at the beginning of the script, as the purpose of this script is to set this variable correctly. In other words, yes, we should not mix the script output (`ESP_PYTHON`) with the user's preference (`ESP_PYTHON_CUSTOM`).

Without disregarding the previously set `ESP_PYTHON`, the entire check can be currently bypassed, and the `ESP_PYTHON_CUSTOM` is used more as a recommendation rather than an explicit request.

```bash
$ ESP_PYTHON_CUSTOM=python3.11 . ./tools/detect_python.sh 
Reading ESP_PYTHON_CUSTOM from environment: "python3.11"
Checking "python3.11" ...
"python3.11" has been detected
$ echo $ESP_PYTHON
python3.11
$ ESP_PYTHON_CUSTOM=NOPYTHON . ./tools/detect_python.sh 
Reading ESP_PYTHON_CUSTOM from environment: "NOPYTHON"
"python3.11" has been detected
$ echo $ESP_PYTHON # previously python is used, instead of error(ENOENT)
python3.11

```

I also think this might not work as expected from within already activated(after `. export.sh`)  esp-idf environment. 

1. `. ./install.sh`
2. `. ./export.sh`
3. `python --version` # Python 3.12.8
4. ESP_PYTHON_CUSTOM=python3.11 ./install.sh # this will still install in `idf6.1_py3.12_env`, not in `idf6.1_py3.11_env` as one would expect

same with `export.sh`

1. `. ./export.sh`
2. `python --version` # Python 3.12.8
3. ESP_PYTHON_CUSTOM=python3.11 . ./export.sh
4. `python --version` # Python 3.12.8 # not 3.11 as one would expect

The problem is with [IDF_PYTHON_ENV_PATH](https://github.com/espressif/esp-idf/blob/master/tools/idf_tools.py#L1860) which is set with `export.sh` and used to set the python venv. 

I guess we can unset it if `ESP_PYTHON_CUSTOM` is specified.

> 2. Found a bug that has been in your original code logic: `exit 1` kills the process running `detect_python.sh`. If user is running `. export.sh` from terminal, user's terminal shell will be killed. Fix this bug by wrapping code into function (haven't added indent yet).

Yes, this problem is there probably for a long time. I guess that it's a very rare condition(no python) that actually triggers this code path, so nobody noticed or complained. Agreed this should fixed.

I've been playing with this slightly modified sh version
```sh
# Ignore any previously set ESP_PYTHON value. It is the responsibility of this script to set it correctly.
unset ESP_PYTHON

OLDEST_PYTHON_SUPPORTED_MAJOR=3
OLDEST_PYTHON_SUPPORTED_MINOR=10

PYTHON_CANDIDATES="python3
python
python3.10
python3.11
python3.12
python3.13"

if [ -n "${ESP_PYTHON_CUSTOM:-}" ]; then
    echo "Reading ESP_PYTHON_CUSTOM from environment: \"$ESP_PYTHON_CUSTOM\""
    PYTHON_CANDIDATES="$ESP_PYTHON_CUSTOM"
    # If a custom Python version is requested, ignore any previously set Python
    # environment path.
    unset IDF_PYTHON_ENV_PATH
fi

while IFS= read -r p_cmd; do
    "$p_cmd" --version >/dev/null 2>&1 || continue
    echo "Checking \"$p_cmd\" ..."

    "$p_cmd" -c "import sys; exit(1) if sys.version_info.major < int(\"$OLDEST_PYTHON_SUPPORTED_MAJOR\") else exit(0);" || continue
    "$p_cmd" -c "import sys; exit(1) if sys.version_info.minor < int(\"$OLDEST_PYTHON_SUPPORTED_MINOR\") else exit(0);" || continue

    ESP_PYTHON="$p_cmd"
    break
done << EOF
$PYTHON_CANDIDATES
EOF

unset PYTHON_CANDIDATES

if [ -z "${ESP_PYTHON:-}" ]; then
    echo "Python ${OLDEST_PYTHON_SUPPORTED_MAJOR}.${OLDEST_PYTHON_SUPPORTED_MINOR}+ is not installed! Please see the documentation for how to install it." >&2
    return 1
fi
echo "\"$ESP_PYTHON\" has been detected"

```

which seems to be working, but there might be other issues. This still might require some additional work.

Note: I dropped the function, because IMHO it's not needed based on the [dot](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot) and [return](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#return) description. I also dropped the ` export ESP_PYTHON`. But these are just nitpicks.

In my experience, almost every time we touch these scripts, something breaks for someone :). 

Thank you very much!

PR #17887, opened 2025-11-16
Refactor TWAI driver by removing redundant checks (IDFGH-16810)

diplfranzhoepfinger:
Removed second chance check for hardware availability in TWAI driver.

🧐 Potential Issues in esp_twai_onchip.c
1. 🚨 Thread Safety and Concurrency (Atomic Operations)
The code uses _Atomic variables and atomic_load/atomic_store for fields like state, hw_busy, and rx_isr. However, there's a potential weak point in the _node_queue_tx function:

Issue in _node_queue_tx (Second Chance Check): The logic attempts a "second chance check" to start transmission if the hardware became free while a task was waiting on the queue.

```
// Second chance check for hardware availability
false_var = false;
if (atomic_compare_exchange_strong(&twai_ctx->hw_busy, &false_var, true)) {
    // ... dequeue logic ...
}
```

If this block executes in a Task Context (not ISR), the following scenario can lead to a race condition:

Task A queues a message and releases the CPU (e.g., waiting for ticks_to_wait).

The ISR runs, finishes the previous transmission, checks the queue, finds Task A's message, sets hw_busy to true, and starts transmission.

Task A wakes up. The atomic_compare_exchange_strong check will fail because hw_busy is now true.

Task A's dequeuing logic runs, which is unnecessary and potentially removes a valid frame from the queue that the ISR should have processed later. The assert(false && "should always get frame at this moment") suggests the developer was aware of a potential logic issue here.

Recommendation: The logic for dequeuing and starting transmission should primarily reside in the ISR (TWAI_HAL_EVENT_TX_BUFF_FREE event) to ensure serialization. The non-ISR path of _node_queue_tx should only focus on queueing the message and the initial attempt to claim the hardware if idle. The "second chance" logic for task context should be removed or made more robust by ensuring the task does not dequeue a frame that the ISR might process.

---

> [!NOTE]
> Simplifies `_node_queue_tx` by removing the post-enqueue "second chance" hardware-busy check/dequeue, relying on ISR to start transmissions.
> 
> - **Driver (TWAI)**
>   - In `components/esp_driver_twai/esp_twai_onchip.c`:
>     - Simplifies transmit path in `_node_queue_tx` by removing the post-enqueue "second chance" block that attempted to claim `hw_busy` and dequeue/start TX immediately.
>     - Transmit queuing remains; ISR (`TWAI_HAL_EVENT_TX_BUFF_FREE`) now solely advances transmissions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 66c74b069ddaaa4910505e271c8181c10b51afd1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_driver_twai/esp_twai_onchip.c
+++ b/components/esp_driver_twai/esp_twai_onchip.c
@@ -615,23 +615,6 @@ static esp_err_t _node_queue_tx(twai_node_handle_t node, const twai_frame_t *fra
             ESP_RETURN_ON_FALSE(xQueueSend(twai_ctx->tx_mount_queue, &frame, ticks_to_wait), ESP_ERR_TIMEOUT, TAG, "tx queue full");
         }
 
-        // Second chance check for hardware availability
-        false_var = false;
-        if (atomic_compare_exchange_strong(&twai_ctx->hw_busy, &false_var, true)) {
-            BaseType_t dequeue_result;
-            if (is_isr_context) {
-                dequeue_result = xQueueReceiveFromISR(twai_ctx->tx_mount_queue, &twai_ctx->p_curr_tx, &yield_required);
-            } else {
-                dequeue_result = xQueueReceive(twai_ctx->tx_mount_queue, &twai_ctx->p_curr_tx, 0);
-            }
-
-            if (dequeue_result == pdTRUE) {
-                _node_start_trans(twai_ctx);
-            } else {
-                assert(false && "should always get frame at this moment");
-            }
-        }
-
         // Handle ISR yield if required
         if (is_isr_context && yield_required) {
             portYIELD_FROM_ISR();


diplfranzhoepfinger:
@esp Team. 
Pls review and check carefully. 
I am not sure if what gemini said is true. 

True is: this new Can driver makes some troubles; and we have to improve the reliability. 


Tnx

wanckl:
@diplfranzhoepfinger Hi, Thanks for your review and suggestion

In your step:
> Task A wakes up. The atomic_compare_exchange_strong check will fail because hw_busy is now true.

If you are saying the 2nd check, the `dequeue` is in `true` branch not `false`. So `hw_busy:true -> 2nd check fail` is expect and do nothing.

And the 2nd check is required because we don't prevent ISR from executing during `send` API, we need check if ISR is exited between 1st check and enqueue, As the cursor says above

franz-ms-muc:
so, close !

PR #18229, opened 2026-02-10
fix(bootloader): prevent unsigned overflow in partition table validation (IDFGH-17230)

Eun0us:
## Summary

The bounds check in `esp_partition_table_verify()` uses:
```c
if (pos->offset > chip_size || pos->offset + pos->size > chip_size)
```

When both `pos->offset` and `pos->size` are large `uint32_t` values, their addition can silently wrap around to a small value, bypassing the validation and allowing an invalid partition entry to pass.

**Fix:** Replace with `pos->size > chip_size - pos->offset`, which is safe because the first condition already guarantees `pos->offset <= chip_size`.

## How it was found

Found via [esp-fuzzer](https://github.com/Eun0us/esp-fuzzer) with `-fsanitize=integer`.

## Test plan

- [x] Verified with crafted partition table where `offset + size` wraps `UINT32_MAX`
- [x] Normal partition tables still validate correctly (subtraction produces same result when no overflow)

### DIFF
--- a/components/bootloader_support/src/flash_partitions.c
+++ b/components/bootloader_support/src/flash_partitions.c
@@ -23,7 +23,7 @@ esp_err_t esp_partition_table_verify(const esp_partition_info_t *partition_table
 
         if (part->magic == ESP_PARTITION_MAGIC) {
             const esp_partition_pos_t *pos = &part->pos;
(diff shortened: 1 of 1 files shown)


mahavirj (approved):

mahavirj:
sha=fc29190dbaf9ca2671f9b5d00ac84af7080bd6e1

PR #18241, opened 2026-02-14
fix(esp_https_server): SSL context leak when HTTPS server fails to start (IDFGH-17248)

### LINKED ISSUE #18230: esp_https_server: Leak of SSL context when HTTPS server fails to start (IDFGH-17232)
### Answers checklist.

### General issue report

In the `httpd_ssl_start` function of the `components/esp_https_server/src/https_server.c` file, a leak is possible.

This can happen because the variable `ssl_ctx` is not cleaned up properly. 

After the following call, `ssl_ctx` may point to dynamic memory:

```c
ret = create_secure_context(config, &ssl_ctx);
```

If the later call to `httpd_start` fails, the error path frees `ssl_ctx` but not the memory that it was pointing to. This will cause a memory leak.

This error path may be triggered as a result of the HTTPS server failing to bind, for example.

A possible fix for this is to reuse the `free_secure_context` function to free any dynamic memory that may have been allocated as a result of `create_secure_context`. `ssl_ctx` may be `NULL` (if the `transport_mode` is not secure), so check for it to avoid failing an assert in `free_secure_context`.

```diff
    if (ret != ESP_OK) {
-       free(ssl_ctx);
-       ssl_ctx = NULL;
+       if (ssl_ctx) {
+           free_secure_context(ssl_ctx);
+           ssl_ctx = NULL;
+       }
```

david-slay:
## Description

Fixes SSL context memory leak when an HTTPS server fails to start. Frees dynamic memory that may have been allocated when starting the HTTPS server.

## Related

Fixes #18230 

---

> [!NOTE]
> **Low Risk**
> Small, localized cleanup change on an error path; minimal behavior impact aside from preventing a memory leak.
> 
> **Overview**
> Fixes a resource leak when `httpd_ssl_start()` fails: on `httpd_start()` error, the code now tears down the HTTPS global SSL context via `free_secure_context()` (including TLS config/cert/key buffers) instead of only freeing the top-level pointer.
> 
> Also updates the SPDX copyright year to 2026.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d5f58a14470e101eda0a19d5f773c07846974026. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_https_server/src/https_server.c
+++ b/components/esp_https_server/src/https_server.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2018-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2018-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -448,8 +448,9 @@ esp_err_t httpd_ssl_start(httpd_handle_t *pHandle, struct httpd_ssl_config *conf
 
     ret = httpd_start(&handle, &config->httpd);
     if (ret != ESP_OK) {
-        free(ssl_ctx);
-        ssl_ctx = NULL;
+        if (ssl_ctx) {
+            free_secure_context(ssl_ctx);
+        }
         return ret;
     }
 


components/esp_https_server/src/https_server.c
```diff
@@ -448,8 +448,10 @@ esp_err_t httpd_ssl_start(httpd_handle_t *pHandle, struct httpd_ssl_config *conf
 
     ret = httpd_start(&handle, &config->httpd);
     if (ret != ESP_OK) {
-        free(ssl_ctx);
-        ssl_ctx = NULL;
+        if (ssl_ctx) {
+            free_secure_context(ssl_ctx);
```
atanisoft, line 452: does `free_secure_context` also free the pointer internally? Based on other code in this method I believe you will still need to retain the call to `free`.

david-slay: Yes, it will free `ssl_ctx` and any dynamic memory it may have been pointing to. The call to `free` should not be necessary.

components/esp_https_server/src/https_server.c
```diff
@@ -448,8 +448,10 @@ esp_err_t httpd_ssl_start(httpd_handle_t *pHandle, struct httpd_ssl_config *conf
...
+            ssl_ctx = NULL;
```
mahavirj: This can be removed, the function shall return after this.

mahavirj:
@david-slay One comment - otherwise looks good!

components/esp_https_server/src/https_server.c
david-slay: Removed the set to `NULL`.

mahavirj:
sha=d5f58a14470e101eda0a19d5f773c07846974026

PR #18249, opened 2026-02-17
feat(ethernet/ptp): added support for gPTP (IDFGH-17258)

scrambletools:
## Description

- Added support in ptpd.c for peer delay requests/responses/followups. Added necessary changes for BMCA, peer delay and other calculations to support gPTP profile
- Added minor changes to ptpv2.h to support message data needed for gPTP.
- Added gPTP related settings to menuconfig
- Motivation is to allow for ESP32P4 to have a fully functional PTP stack with support for gPTP to enable AVB functionality on a low-cost solution

## Related

This is a re-submission of a previous PR that is defunct (#15001 reviewed by @kostaond) due to an unrecoverable error in rebasing

## Testing

Tested on esp32p4 rev0.1 using 2 Waveshare ESP32 P4 Nanos connected directly to each other.
Enable/disabled gPTP via menuconfig and tested with different menuconfig settings.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core PTP synchronization logic (BMCA selection, packet formats, delay calculation, and clock adjustment), which can affect time accuracy and interoperability across networks, though it is gated behind a new gPTP config option.
> 
> **Overview**
> Adds an optional **gPTP profile mode** (`CONFIG_NETUTILS_PTPD_GPTP_PROFILE`) to `ptpd`, switching it to 802.1AS behavior (LLDP multicast destination, gPTP flags/SDO ID, mandatory two-step sync, and updated BMCA comparisons).
> 
> Implements **peer-delay (PDelay)** support end-to-end: new message types/structs and TLVs in `ptpv2.h`, new state tracking (`peer_delay_ns`/correction field), handling of `PDELAY_REQ`/`PDELAY_RESP`/`PDELAY_RESP_FOLLOW_UP`, and updated clock-correction math to use peer delay + correction field when in gPTP.
> 
> Updates Kconfig to expose gPTP and timing knobs (renames `*_INTERVAL_MSEC` to `*_INTERVAL_MS`, replaces server delay interval config with `NETUTILS_PTPD_DELAYREQ_INTERVAL_MS`, adds peer-delay stability tuning), and extends status reporting (`ptpd.h`) to include `peer_delay_ns`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 94f59e5de8073f9705d43a0b506bd3346c7aaa3b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


scrambletools:
@kostaond I just resubmitted the PR, please let me know when you can look at it.

kostaond:
@scrambletools thank you for the update. Our `ptp` daemon is based on Nuttx implementation. I've just noticed that there was an [update recently](https://github.com/apache/nuttx-apps/commits/master/netutils/ptpd/ptpd.c). I would like to keep track with the Nuttx upstream version. It unfortunately means another delay for your PR :(  

I understand that this waiting must be very frustrating for you. Let me propose a compromise.

I can extract `ptp` into a separate component and publish it via the [Espressif Component Registry](https://components.espressif.com/) — this is planned anyway. You could then fork it and distribute your solution under your own namespace in the registry. Once things settle, we can merge your fork back upstream.

Would this approach work for you? If so, I should be able to prepare the standalone `ptp` component within a week or two.

scrambletools:
@kostaond that sounds ok. I have a few questions if you don't mind:
How far behind is it from Nuttx master?
What is the cadence for sync'ing with Nuttx?
Some features are gated behind ESP-specific configs. How will those features be enabled for Nuttx?
Once ptp is moved to the component registry, it will be maintained there (not in the example tree) correct?

scrambletools:
@kostaond I plan to support gPTP over wifi in a future update (using FTM to calc peer delay) so you may want to maintain ptp somewhere that is not specific to ethernet.

kostaond:
> I have a few questions if you don't mind:
* TBH, I ported it in rush at the time so I did't note specifically but I think ESP-IDF version is at 86f3671 of Nuttx
* As you can see in the history, there were no functional updates in Nuttx for almost 2 years
* I'm not sure if I fully understand to the question. However, I'm planning to migrate ESP-IDF version to https://github.com/espressif/esp-protocols/tree/master/components repo as it is. Hence the daemon should stay more or less the same.
* The pdp daemon component will be maintained at https://github.com/espressif/esp-protocols/tree/master/components repo and distributed over Component registry. Opened question is if to keep the ptp example in the IDF which would be dependent on ptp component or migrate the example to Protocols repo, or combination of both. [ptpd](https://github.com/espressif/esp-idf/tree/master/examples/ethernet/ptp/components/ptpd) will definitely be separate component, the question is what to do with [esp_eth_time](https://github.com/espressif/esp-idf/tree/master/examples/ethernet/ptp/components/esp_eth_time)(that was probably your question, right?). Ideally, it would be better to be part of IDF. I'll start like that and I will see if it's possible.

kostaond:
> I plan to support gPTP over wifi in a future update 

That would be super neat! We discussed it too, but we currently don't have resources. Anyway, having ptp as separate component should make your life easier with distribution of your solution to the public.

scrambletools:
@kostaond regarding where you plan to maintain ptpd, that makes sense. My other question was regarding how to keep the features of ptpd working in both nuttx and esp-idf. I am not testing for nuttx, only esp-idf.

kostaond:
I see, we are not planning to merge to Nuttx upstream. We will maintain just our version and time to time sync with Nuttx.

examples/ethernet/ptp/components/ptpd/ptpd.c
```diff
@@ -1713,26 +2055,40 @@ static int ptp_process_rx_packet(FAR struct ptp_state_s *state,
     case PTP_MSGTYPE_SYNC:
       ptpinfo("Got sync packet, seq %ld\n",
               (long)ptp_get_sequence(&state->rxbuf.header));
+      if (!state->selected_source_valid) {
+        return OK;
+      } // ignore if operating as a server in gPTP profile
       return ptp_process_sync(state, &state->rxbuf.sync);
```
safocl, line 2086: C draft "ISO/IEC 9899:TC2  Committee Draft — May 6, 2005  WG14/N1124":

> 
> **6.5.2.3 Structure and union members**
> [...]
> 5 One special guarantee is made in order to simplify the use of unions: if a union contains several structures that share a common initial sequence (see below), and if the union object currently contains one of these structures, it is permitted to inspect the common initial part of any of them anywhere that a declaration of the complete type of the union is visible. Two structures share a common initial sequence if corresponding members have compatible types (and, for bit-fields, the same widths) for a sequence of one or more initial members.

> 8 The following is not a valid fragment (_because the union type is not visible within function f_):
> ```c
> struct t1 { int m; };
> struct t2 { int m; };
> int f(struct t1 *p1, struct t2 *p2)
> {
>   if (p1->m < 0)
>   p2->m = -p2->m;
>   return p1->m;
> }
> int g()
> {
> union {
>   struct t1 s1;
>   struct t2 s2;
> } u;
> /* ... */
>   return f(&u.s1, &u.s2);
> }
> ```

> **Annex J**
> (informative)
> **Portability issues**
> 
> J.1 Unspecified behavior
> The following are **unspecified**:
> [...]
> — The value of a union member other than the last one stored into (6.2.6.1).

> 
> **6.2.6 Representations of types
> 6.2.6.1 General**
> [...]
> 7 When a value is stored in a member of an object of union type, the bytes of the object
> representation that do not correspond to that member but do correspond to other members
> take unspecified values.

(in C++ this is undefined behavior).

You can simply wrap it in a separate structure `struct ptp_header_s header;` (but accordingly change access everywhere through an additional union member):
```c
typedef union
{
  struct { struct ptp_header_s header; } general;
  struct ptp_announce_s announce;
  struct ptp_sync_s sync;
  struct ptp_follow_up_s follow_up;
  struct ptp_delay_req_s delay_req;
  struct ptp_delay_resp_s delay_resp;
  struct ptp_delay_resp_follow_up_s delay_resp_follow_up;
  uint8_t raw[128];
} ptp_msgbuf;

ptp_msgbuf u = {.sync = {}};
if (u.general.header.messagetype == MAGIC_NUMBER_FOR_SYNC){
  struct ptp_sync_s s1 = u.sync;
} else if (u.general.header.messagetype == MAGIC_NUMBER_FOR_FOLLOW){
  struct ptp_follow_up_s s1 = u.follow_up;
}
```

safocl: > **6.7.2.1 Structure and union specifiers**
> 
> 13 Within a structure object, the non-bit-field members and the units in which bit-fields reside have addresses that increase in the order in which they are declared. A pointer to a structure object, suitably converted, points to its initial member (or if that member is a bit-field, then to the unit in which it resides), and vice versa. There may be unnamed padding within a structure object, but not at its beginning.
> 14 The size of a union is sufficient to contain the largest of its members. The value of at most one of the members can be stored in a union object at any time. A pointer to a union object, suitably converted, points to each of its members (or if a member is a bit-field, then to the unit in which it resides), and vice versa.

I can't find any provisions here about the possibility of converting a pointer to one type **A** into a pointer to another type **B** when there is a third type **C** that can be _pointer-interconvertible (suitably converted)_ to **B** when type **B** is nested in type **A**, and then **C** can be _pointer-interconvertible (suitably converted)_ to **A**.

as in C++ standard:
https://eel.is/c++draft/basic.compound#5

> 5 Two objects a and b are [pointer-interconvertible](https://eel.is/c++draft/basic.compound#def:pointer-interconvertible) if
> [...]
> [(5.4)](https://eel.is/c++draft/basic.compound#5.4) there exists an object c such that a and c are pointer-interconvertible, and c and b are pointer-interconvertible[.](https://eel.is/c++draft/basic.compound#5.sentence-1)

then you could simply convert the pointer to the union into a header type:
```c
ptp_msgbuf u = {.sync = {}};
if (((ptp_header_s*)u)->messagetype == MAGIC_NUMBER_FOR_SYNC){
  struct ptp_sync_s s1 = u.sync;
} else if (u.general.header == MAGIC_NUMBER_FOR_FOLLOW){
  struct ptp_follow_up_s s1 = u.follow_up;
}
```

scrambletools:
@kostaond do you have an estimate when ptpd will move to https://github.com/espressif/esp-protocols/tree/master/components?

examples/ethernet/ptp/components/ptpd/ptpd.c
scrambletools: All structs in the union contain only uint8_t, it is used for convenience to reference offsets in the message buffer. This is the way the union was used before this PR.

safocl: > All structs in the union contain only uint8_t, it is used for convenience to reference offsets in the message buffer. This is the way the union was used before this PR.

you example contains:
```C
struct ptp_header_s
{
  uint8_t messagetype;
};

struct ptp_announce_s
{
  struct ptp_header_s header;
};

typedef union
{
  struct ptp_header_s header;
  struct ptp_announce_s announce;
  uint8_t raw[128];
} ptp_msgbuf;

void foo(){
    ptp_msgbuf u.announce = {0}; // ptp_msgbuf::announce is an active member
    int i = u.header.messagetype; // this is a _unspecified behavior_ in C and a _undefined behavior_ in C++ -- access to inactive member of union.
}
```
but the following is well-defined:
```C
struct ptp_header_s
{
  uint8_t messagetype;
};

struct ptp_announce_s
{
  struct ptp_header_s header;
};

typedef union
{
  struct {struct ptp_header_s header} general;
  struct ptp_announce_s announce;
  uint8_t raw[128];
} ptp_msgbuf;

void foo(){
    ptp_msgbuf u.announce = {0}; // ptp_msgbuf::announce is an active member
    int i = u.general.header.messagetype; // expected behavior via the common initial sequence
}
```

safocl: ~This example shows how code can be generated in the case of such a member access -- anything can be generated, including, in this particular case, the check is removed (_UB_ with `#if 1`):~
~https://godbolt.org/z/6oe7ovc7n~
upd. oh no... they simply calculated that it was the same thing in two branches...
But this doesn't negate the above. It seems the compilers in this example are simply aggressively relaxing the rules and performing type-punning.

kostaond:
@scrambletools as you correctly pointed in our previous discussion, there's a need to address EMAC clock integration into ESP-IDF time functions (`get/set_time`, etc.). I've already proposed solution and it's in internal review for some time as it's architectural change, it requires broader discussion...

examples/ethernet/ptp/components/ptpd/ptpd.c
kostaond: At this point, there is no need to address these comments as our plan is to maintain the ptp at separate repo in near feature.

scrambletools:
I see, thanks for letting me know.

scrambletools:
@kostaond we published our esp_avb component to the registry, and to make things simple we also published a fork of ptpd (we call it esp_ptp) to the component registry as it was a dependency of the first module.

scrambletools:
@kostaond ideally we would hope to merge our changes into ptpd mainline and drop esp_ptp, but for now we need something stable that supports gPTP.

kostaond:
@scrambletools nice job! I'm currently in process of syncing ESP-IDF ptpd version to the upstream Nuttx. Once done, I'm planning to migrate to our external `protocols` GH repository and publish over the component registry. The whole process is bit slower then I expected...

PR #18288, opened 2026-02-27
Fixed memory leaks in ESP HTTP CLIENT and TLS (IDFGH-17303)

b00lean:
## Description

I have identified several memory leaks while I was building my application.
Changes are trivial yet they have big positive impact when you have application that does regular HTTPS calls.

No screenshots are needed.

## Testing

No need to test.

---

## Checklist

- [X ] 🚨 This PR does not introduce breaking changes.
- [ X] All CI checks (GH Actions) pass.
- [ X] Documentation is updated as needed.
- [ X] Tests are updated or added as necessary.
- [ X] Code is well-commented, especially in complex areas.
- [X ] Git history is clean — commits are squashed to the minimum necessary.

---

> [!NOTE]
> **Medium Risk**
> Touches core networking cleanup and async TLS connection state, so regressions could show up as unexpected reconnect/reset behavior or latent use-after-free if callers assume previous pointer lifetimes. Changes are mostly defensive frees/nulling and should be low-impact if cleanup paths are correct.
> 
> **Overview**
> Fixes multiple memory-leak scenarios across networking components by making teardown paths idempotent and ensuring allocations are always released.
> 
> `esp_http_client` now frees connection/auth fields conditionally, nulls pointers after free (including password/scheme/url), treats missing `auth_data` as a no-op, and hardens `esp_http_client_cleanup()` to avoid double-frees and to fully release request/response buffers and headers.
> 
> `lwIP`’s `sys_mbox_free()` keeps the assert but always frees the mailbox allocation even in release builds, and `transport_ssl` now destroys and resets the TLS context on fatal async connect errors to avoid leaking the `esp_tls` object.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b5a6b57828290bedfbeaacb79304aacb7c8e37d1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


components/esp_http_client/esp_http_client.c
```diff
@@ -648,33 +648,72 @@ static esp_err_t _set_config(esp_http_client_handle_t client, const esp_http_cli
...
 static esp_err_t _clear_connection_info(esp_http_client_handle_t client)
 {
-    free(client->connection_info.path);
-    free(client->connection_info.host);
-    free(client->connection_info.query);
-    free(client->connection_info.username);
+    if (client->connection_info.path) {
+        free(client->connection_info.path);
+        client->connection_info.path = NULL;
+    }
```
igrr, line 654: This seems equivalent to the original code, isn't it?

`free(ptr);` is the same thing as `if (ptr) {free(ptr);}`, and the `connection_info` structure is zeroed out at the end of the function, anyway?

components/esp_http_client/esp_http_client.c
```diff
@@ -1097,38 +1136,77 @@ esp_err_t esp_http_client_cleanup(esp_http_client_handle_t client)
...
+        free(client->current_header_value);
+        client->current_header_value = NULL;
+    }
+    if (client->location) {
+        free(client->location);
+        client->location = NULL;
+    }
+    if (client->auth_header) {
+        free(client->auth_header);
+        client->auth_header = NULL;
```
igrr, line 1208: And clearing the fields here doesn't matter because of `free(client)` below, right?

components/esp_http_client/esp_http_client.c
b00lean: Why do we free the individual fields before zeroing the structure?
 
  The reason is that simply zeroing (memset) the structure only sets the bytes to zero.
  However, many fields in connection_info_t are pointers to dynamically allocated memory.
  If we do not free these pointers before zeroing the struct, we will introduce memory leaks.
  For example, strdup() is used to assign strings, so each of those fields must be explicitly freed.
  In addition, zeroing a pointer field does NOT free its underlying memory.
  
  For passwords, there is an extra reason: we explicitly overwrite the memory buffer
  (memset to zero) before freeing, to erase sensitive information from RAM.
  
  So, this pattern ensures all memory is deallocated, and sensitive strings are erased appropriately.

components/esp_http_client/esp_http_client.c
b00lean: again freeing client doesn't free the underlying pointers to auth_header etc.

components/esp_http_client/esp_http_client.c
igrr: The original code also frees the fields of connection_info, e.g. `free(client->connection_info.path);`

> For passwords, there is an extra reason: we explicitly overwrite the memory buffer
(memset to zero) before freeing

That would make sense, however you can't rely on memset for security-sensitive scenarios. C compiler can optimize memset out, you have to use one of the secure memset functions for that purpose, e.g. mbedtls_zeroize.

components/esp_http_client/esp_http_client.c
igrr: The underlying fields are free'ed in the original code as well. My point is that `client->auth_header = NULL;` has no effect because the whole `client` structure is free'ed a few lines below.

components/tcp_transport/transport_ssl.c
```diff
@@ -84,6 +84,13 @@ static int esp_tls_connect_async(esp_transport_handle_t t, const char *host, int
                 esp_tls_conn_destroy(ssl->tls);
                 return -1;
             }
+        } else if (progress != ESP_TLS_ERR_SSL_WANT_READ && progress != ESP_TLS_ERR_SSL_WANT_WRITE) {
```
bryghtlabs-richard, line 87: Some of this patch is spelling fixes, some of this patch seems to be stylist (free,null) rather than (free all, memset0 all). Outside of those, @igrr , what do you think about this part though?

Issue #18227, opened 2026-02-09
esp_http_server: PONG frames not dispatched to WebSocket handler despite comment saying they should be (IDFGH-17227)


pback34:
## Answers checklist

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

## IDF version

v5.5.2 (also confirmed unfixed on current `master`)

## Espressif SoC revision

ESP32-S3

## Operating System used

Linux

## How did you build your project?

Command line with idf.py

## If you are using Windows, please specify command line type.

None

## Development Kit

Seeed Studio XIAO ESP32S3 Sense

## Power Supply used

USB

## What is the expected behavior?

When the server sends WebSocket PING frames and the client responds with PONG, the PONG frame should be dispatched to the user's WebSocket handler so that application code (e.g., heartbeat tracking) can process it.

The code in `httpd_parse.c` line 802-804 even has a comment acknowledging this:

```c
} else if (ra->ws_type == HTTPD_WS_TYPE_PONG) {
    /* Pass the PONG frames to the handler as well, as user app might send PINGs */
    ESP_LOGD(TAG, LOG_FMT("Received PONG frame"));
}
```

## What is the actual behavior?

PONG frames are **never dispatched** to the user handler. The dispatch condition at line 808-809 excludes them:

```c
if (ret == ESP_OK &&
    (ra->ws_type < HTTPD_WS_TYPE_CLOSE || sd->ws_control_frames)) {
    ret = sd->ws_handler(r);
}
```

Since `HTTPD_WS_TYPE_PONG = 0xA` and `HTTPD_WS_TYPE_CLOSE = 0x8`, the condition `PONG < CLOSE` is `false`, so the handler is never called for PONG frames.

**This has a critical secondary effect:** `httpd_ws_get_frame_type()` only reads byte 0 (the opcode) of the PONG frame. Since the handler is never called, `httpd_ws_recv_frame()` is never invoked, and bytes 1-5 of the PONG frame (second_byte + 4-byte mask_key for a zero-length PONG) remain unconsumed in the TCP receive buffer.

[...]

PR #18309, opened 2026-03-05
fix(esp_http_server): dispatch PONG frames to WebSocket handler (IDFGH-17330)

pback34:
## Description

PONG frames (opcode `0xA`) are never dispatched to the user's WebSocket handler despite an existing comment at `httpd_parse.c:808` stating they should be ("Pass the PONG frames to the handler as well, as user app might send PINGs").

The dispatch condition `ra->ws_type < HTTPD_WS_TYPE_CLOSE` excludes PONG (`0xA`) since `HTTPD_WS_TYPE_CLOSE` is `0x8`.

This causes a **critical secondary bug**: when the server sends PING frames and the client responds with PONG, `httpd_ws_recv_frame()` is never called for the PONG, leaving bytes 1–5 (second_byte + 4-byte mask_key) unconsumed in the TCP buffer. On the next WebSocket read, these orphaned bytes are misinterpreted as a new frame header, causing either:
- "WS frame is not properly masked" errors
- EAGAIN timeouts with garbage length values

This effectively destroys the connection every time a PONG arrives.

## Fix

Add `ra->ws_type == HTTPD_WS_TYPE_PONG` to the dispatch condition in `httpd_parse.c` so PONG frames reach the user handler, which calls `httpd_ws_recv_frame()` to properly consume the frame bytes from the socket.

```c
if (ret == ESP_OK &&
    (ra->ws_type < HTTPD_WS_TYPE_CLOSE ||
     ra->ws_type == HTTPD_WS_TYPE_PONG ||
     sd->ws_control_frames)) {
    ret = sd->ws_handler(r);
}
```

Closes https://github.com/espressif/esp-idf/issues/18227

## Testing

Tested on ESP32-S3 (Seeed XIAO ESP32S3 Sense) with a WebSocket server sending PING frames every 10 seconds to browser clients. Before the fix, connections would fail with masking errors after the first PONG. After the fix, PONG frames are correctly dispatched and connections remain stable indefinitely.

---

> [!NOTE]
> **Medium Risk**
> Changes WebSocket frame dispatch in the core HTTP server loop, which can affect application handler behavior by invoking callbacks for PONG frames that were previously skipped. Scope is small and targeted, but sits on a hot path for all WebSocket connections.
> 
> **Overview**
> Ensures incoming WebSocket `PONG` frames are dispatched to the user’s WS handler in `httpd_parse.c`, even when control frames aren’t generally forwarded.
> 
> Updates the dispatch condition (and comment) so `PONG` frames get consumed via the handler’s `httpd_ws_recv_frame()` path, preventing leftover bytes from desynchronizing subsequent WS reads and improving heartbeat/keepalive handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 049a5dc957a8a55abf23414c714cc56fefa8134f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_http_server/src/httpd_parse.c
+++ b/components/esp_http_server/src/httpd_parse.c
@@ -809,9 +809,16 @@ esp_err_t httpd_req_new(struct httpd_data *hd, struct sock_db *sd)
             ESP_LOGD(TAG, LOG_FMT("Received PONG frame"));
         }
 
-        /* Call handler if it's a non-control frame (or if handler requests control frames, as well) */
+        /* Call handler if it's a non-control frame, a PONG frame,
+         * or if handler requests control frames as well.
+         * PONG must be dispatched so that:
+         *  1. User code that sends PINGs can track responses (heartbeat)
+         *  2. The PONG frame bytes are consumed from the socket via
+         *     httpd_ws_recv_frame(), preventing TCP stream misalignment */
         if (ret == ESP_OK &&
-            (ra->ws_type < HTTPD_WS_TYPE_CLOSE || sd->ws_control_frames)) {
+            (ra->ws_type < HTTPD_WS_TYPE_CLOSE ||
+             ra->ws_type == HTTPD_WS_TYPE_PONG ||
+             sd->ws_control_frames)) {
             ret = sd->ws_handler(r);
         }
 


pback34:
Closing in favor of a new PR with lowercase branch name and fixed commit message (DangerJS warnings).

PR #18347, opened 2026-03-15
Fix/jpeg p4 cache alignment for input buffers on P4 (IDFGH-17377)

pskasten:
### **Problem**
On the **ESP32-P4**, the JPEG driver currently fails when using PSRAM-backed input buffers because it does not enforce cache-line alignment for the `JPEG_DEC_ALLOC_INPUT_BUFFER` (decoder) or `JPEG_ENC_ALLOC_INPUT_BUFFER` (encoder) directions.  Specifically, the allocation sizes are correctly aligned, but the start address of the allocations are incorrect.

Because the P4 utilizes a **write-back cache**, `esp_cache_msync` requires that both the **start address** and the **buffer size** be aligned to the cache line size (64 bytes). The existing allocation logic uses standard `heap_caps_calloc` for input buffers, which often results in a 4-byte offset, leading to a hard failure: `ESP_ERR_INVALID_ARG (0x102)` during cache synchronization.

### **Solution**
I have updated `jpeg_alloc_decoder_mem` and `jpeg_alloc_encoder_mem` to check for **`SOC_CACHE_WRITEBACK_SUPPORTED`**. 

* When this macro is true, the `buffer_direction` is ignored for alignment purposes.
* Both input and output buffers are now force-aligned using `JPEG_ALIGN_UP` for size and `heap_caps_aligned_calloc` for the address.
* This ensures full compatibility with the P4’s cache requirements while preserving the more optimized/relaxed allocation for write-through chips (like the original ESP32).

### **Testing performed**
* **Hardware:** ESP32-P4 (Waveshare 10.1-inch HMI All-in-One)
* **IDF Version:** v5.5
* **Result:** Confirmed that `esp_cache_msync` now returns `ESP_OK` (0x0). The previous alignment `LOGE` errors are resolved and JPEG decoding/encoding functions as expected.

### **Other Changes**
* Fixed a minor typo in the `jpeg_alloc_decoder_mem` and `jpeg_alloc_encoder_mem` header comments (corrected "FOr" to "For").
---

## Checklist

- [?] Tests are updated or added as necessary.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes allocation behavior for JPEG encode/decode buffers on write-back-cache targets and could impact memory usage/alignment assumptions, but it is gated by `SOC_CACHE_WRITEBACK_SUPPORTED`.
> 
> **Overview**
> Ensures JPEG encode/decode PSRAM buffers are always cache-line aligned on write-back-cache targets (e.g., ESP32-P4) by ignoring `buffer_direction` and using `heap_caps_aligned_calloc` with size rounded up via `JPEG_ALIGN_UP`.
> 
> Keeps the prior behavior for non-write-back targets (only aligning decoder/encoder *output* buffers), and updates the allocation comments/typo to document the stricter alignment requirement.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4a0da9a8b9ee5dbf793f03a05555e7ce3432588b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_driver_jpeg/jpeg_decode.c
+++ b/components/esp_driver_jpeg/jpeg_decode.c
@@ -363,10 +363,20 @@ void *jpeg_alloc_decoder_mem(size_t size, const jpeg_decode_memory_alloc_cfg_t *
     /*
        Principle of buffer align.
        For output buffer(for decoder is 2DDMA write to PSRAM), both address and size should be aligned according to cache invalidate.
-       FOr input buffer(for decoder is PSRAM write to 2DDMA), no restriction for any align (both cache writeback and requirement from 2DDMA).
+       For input buffer(for decoder is PSRAM write to 2DDMA), no restriction for any align (both cache writeback and requirement from 2DDMA).
+       NOTE: If the chip uses write-back cache (like the ESP32-P4), the input buffer address and size must both be aligned; viz. the buffer_direction is insignificant.  (See also: jpeg_alloc_encoder_mem()).
     */
     size_t cache_align = 0;
     esp_cache_get_alignment(MALLOC_CAP_SPIRAM, &cache_align);
+
+#if SOC_CACHE_WRITEBACK_SUPPORTED
+    /*
+      As noted above, mem_cfg->buffer_direction is ignored here; address and size must both be aligned regardless.
+    */ 
+    size = JPEG_ALIGN_UP(size, cache_align);
+    *allocated_size = size;
+    return heap_caps_aligned_calloc(cache_align, 1, size, MALLOC_CAP_SPIRAM);
+#else
     if (mem_cfg->buffer_direction == JPEG_DEC_ALLOC_OUTPUT_BUFFER) {
         size = JPEG_ALIGN_UP(size, cache_align);
         *allocated_size = size;
@@ -375,6 +385,7 @@ void *jpeg_alloc_decoder_mem(size_t size, const jpeg_decode_memory_alloc_cfg_t *
         *allocated_size = size;
         return heap_caps_calloc(1, size, MALLOC_CAP_SPIRAM);
     }
+#endif
 }
 
 /****************************************************************
--- a/components/esp_driver_jpeg/jpeg_encode.c
+++ b/components/esp_driver_jpeg/jpeg_encode.c
@@ -392,9 +392,19 @@ void *jpeg_alloc_encoder_mem(size_t size, const jpeg_encode_memory_alloc_cfg_t *
        Principle of buffer align.
        For output buffer(for decoder is 2DDMA write to PSRAM), both address and size should be aligned according to cache invalidate.
        For input buffer(for decoder is PSRAM write to 2DDMA), no restriction for any align (both cache writeback and requirement from 2DDMA).
+       NOTE: If the chip uses write-back cache (like the ESP32-P4), the input buffer address and size must both be aligned; viz. the buffer_direction is insignificant.  (See also: jpeg_alloc_decoder_mem()).
     */
     size_t cache_align = 0;
     esp_cache_get_alignment(MALLOC_CAP_SPIRAM, &cache_align);
+
+#if SOC_CACHE_WRITEBACK_SUPPORTED
+    /*
+      As noted above, mem_cfg->buffer_direction is ignored here; address and size must both be aligned regardless.
+    */ 
+    size = JPEG_ALIGN_UP(size, cache_align);
+    *allocated_size = size;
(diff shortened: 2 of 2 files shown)


mythbuster5:
Hi, besides return value, do you see any other error logs?

pskasten:
Oh!  Good catch.  I meant to include this info.

Although operationally everything seemed ok, I noticed this error log for every allocation; it was exactly this log message that started me looking into the issue:

E (15517) cache: esp_cache_msync(112): start address: 0x487ebec4, or the size: 0x1a280 is(are) not aligned with cache line size (0x40)

Although the size was always aligned properly, the start address (of the allocated memory) was always off by 4.  I originally implemented a workaround in my code:  precomputing the extra amount needed so that I could "slide" the memory window in the allocated region to start at an aligned address and still have an aligned size.  At that point, the error log message and error codes stopped. 

Thanks,
Phil

pskasten:
Hi @mythbuster5.  I just wanted to follow up, to make sure I answered your question sufficiently.  I have been using my proposed fix extensively (testing "under fire" so to speak) and it has been 100% clean.  Prior to the change, the failure occurred 100% of the time ("failure" in the sense of the error code and the log message; operationally, the images were processed ok).

My change was intended to be done in the smallest, safest scope possible.  An alternative approach I considered was to augment the
`jpeg_decode_memory_alloc_cfg_t` and `jpeg_encode_memory_alloc_cfg_t` definitions with an explicit alignment member.  But, this would shift the responsibility of knowing what's required, in terms of alignment, to the client code -- an idea I wasn't in favor of.

pskasten:
Checking back in on this. I've just rebased the PR against the latest master to ensure everything is current and passing CI.

Since it has been a couple of months and all previous questions from @mythbuster5 have been addressed, is there another maintainer on the esp-idf-core or multimedia team who can assist with the final review? This fix is an important correction for JPEG memory alignment that prevents hardware decoder failures on the P4.

mythbuster5:
Sorry for the late followup, I'm working on another thing recently. I have a question. What about adding MALLOC_CAP_DMA in heap_caps_aligned_calloc(cache_align, 1, size, MALLOC_CAP_SPIRAM);? I think this can solve your problem as well. @pskasten

pskasten:
Thank you @mythbuster5 for the follow-up question. I was trying to make my changes to jpeg_alloc_decoded_mem() and jpeg_alloc_encoder_mem() in the spirit of the original code -- using JPEG_ALIGN_UP().  In the end (MALLOC_CAP_DMA or JPEG_ALIGN_UP), the math will be the same to achieve the address and size alignment goals.

Also, it seems that the alignment requirements are more lenient for the input buffer (per the original top-of-function commentary and the  original implementation).  My proposed change makes the code path and content determination at compile time.

Are you ok with the changes I am proposing?

Thanks!
Phil

mythbuster5:
I understand your changes, but after reviewing them carefully, I still have some questions.

Before your modification, the behavior was that the output buffer was aligned, while the input buffer was not aligned.

So based on the error log you shared above, I’m guessing the issue was caused by the input buffer, correct?

If that’s the case, then for the input buffer we should only need to do a write-back, and on P4, write-back (cache → PSRAM direction) does not require alignment.

Also, in both the encoder and decoder, ESP_CACHE_MSYNC_FLAG_UNALIGNED has already been added for the input buffer write-back path.

And in the sync function, there is also this logic: 

```
uint32_t cache_line_size = cache_hal_get_cache_line_size(cache_level, cache_type);
    if ((flags & ESP_CACHE_MSYNC_FLAG_UNALIGNED) == 0) {
        bool aligned_addr = (((uint32_t)addr % cache_line_size) == 0) && ((size % cache_line_size) == 0);
        ESP_RETURN_ON_FALSE_ISR(aligned_addr, ESP_ERR_INVALID_ARG, TAG, "start address: 0x%" PRIx32 ", or the size: 0x%" PRIx32 " is(are) not aligned with cache line size (0x%" PRIx32 ")B", (uint32_t)addr, (uint32_t)size, cache_line_size);
    }
```

So when ESP_CACHE_MSYNC_FLAG_UNALIGNED is set, this alignment check is skipped, and it should not print the corresponding log either.

That’s why I’d like to take a look at your initialization code, or trace the exact call path where the error is triggered, because I want to make sure this fix is as precise and accurate as possible.

mythbuster5:
No respond for a pretty long time, and we have already added very strict check on jpeg side. So I will close this issue. Please feel free to reopen if there is anything more to discuss.

PR #18359, opened 2026-03-18
fix(esp_http_client): reset response buffer and raw length before new… (IDFGH-17389)

idolpx:
… HTTP requests

When making multiple requests using the same client connection, if the response isn't read or flushed before the next request it can cause the following crash.

```
assert failed: http_on_body esp_http_client.c:316 (res_buffer->orig_raw_data == res_buffer->raw_data)
```

## Testing

Applying this patch resolved the issue with manual testing in my project.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core HTTP client request lifecycle; while small, it changes how response buffering state is reset between requests and could affect applications relying on cached unread data across calls.
> 
> **Overview**
> Prevents crashes when reusing an `esp_http_client` handle for multiple requests by **clearing any cached response body buffer and resetting `raw_len`** at the start of `esp_http_client_prepare()`.
> 
> This ensures `res_buffer->raw_data` and `orig_raw_data` are consistent before parsing a new response, avoiding assertion failures if a previous response wasn’t fully read or flushed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f36127a5f89716ac46e26255db2c9f545e28056. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -188,6 +188,7 @@ static const char *HTTP_METHOD_MAPPING[] = {
     "REPORT"
 };
 
+static void esp_http_client_cached_buf_cleanup(esp_http_buffer_t *res_buffer);
 esp_err_t esp_http_client_request_send(esp_http_client_handle_t client, int write_len);
 static esp_err_t esp_http_client_connect(esp_http_client_handle_t client);
 static esp_err_t esp_http_client_send_post_data(esp_http_client_handle_t client);
@@ -742,6 +743,12 @@ esp_err_t esp_http_client_prepare(esp_http_client_handle_t client)
     client->process_again = 0;
     client->response->data_process = 0;
     client->first_line_prepared = false;
+
+    // Reset the response buffer before each new request to ensure raw_data == orig_raw_data.
+    esp_http_client_cached_buf_cleanup(client->response->buffer);
+    // Also unconditionally reset raw_len.
+    client->response->buffer->raw_len = 0;
+
     /**
      * Clear location field before making a new HTTP request. Location
      * field should not be cleared in http_on_header* callbacks because


BitsForPeople:
> if the response isn't read or flushed before the next request it can cause the following crash.

I don't think the HTTP client supports pipelining, so sending a new request over the same (persistent) connection without first consuming all of the server's response can easily mess up the data received (server still sending remaining data for request 1 while the client has already moved on and expects the response to request 2).

idolpx:
This fixes the issue without the need to read/flush the unneeded/unexpected response.

idolpx:
Did I submit this wrong?  Please advise.

Ashish285:
Hi @idolpx , thanks for raising this PR. As mentioned earlier by @BitsForPeople , the HTTP client doesn't support pipelining. Although your changes fixes the issue in your use case, I have concerns about silently dropping unconsumed response data as it may mask application bugs.

May I suggest using `esp_http_client_flush_response` on the application level if you do not care about previously cached data instead of silently dropping it for every use case?

Let me know if you have any other thoughts, thanks!

idolpx:
Oh... I'm not trying to implement pipelining.  I believe I tried flushing the response and it wasn't clearing the buffer.
Since this is opening a new request and resetting other client properties it made sense to free the buffer here.
If it's not freed it causes a crash.

In my use case the client is reading data and the user breaks in the middle then makes a new request using the same connection.
This is normal for my use case.  I traced the crash to this cause.

The only other way to free the buffer was to close the connection and reopen it which of course slows things down noticably.

Ashish285:
Hey @idolpx , will an API that allows to reset the buffers be enough? This can then be handled at the application level explicitly.

idolpx:
Yeah... that would work.  Sorry so slow.  Just catching up on some things.

Ashish285:
Hey @idolpx , can you please check if [esp_http_client_clear_response_buffer()](https://github.com/espressif/esp-idf/blob/055ba9d3f9c6fd9a0efacd4993a2a942972dd65d/components/esp_http_client/include/esp_http_client.h#L831) is the API for your use case? Thanks!

idolpx:
Finally getting back to this.  I updated to v5.5.3 with that function in it and it seems good now.
Thanks.  :)

Ashish285:
Hey @idolpx , thanks for confirming. In that case I will close this PR, please feel free to raise a new issue if you encounter any other issue, thanks!

PR #18385, opened 2026-03-24
feat(mcp): Add <idf.py monitor> support for MCP

jinw06k:
## Description

Add monitor support to the ESP-IDF MCP server for interactive serial workflows.

This PR introduces PTY-backed monitor tools that allow MCP clients to start an `idf.py monitor` session, send input to the device, read buffered serial output, stop the session, and capture boot logs in a one-shot flow.

### Changes included
- add `monitor_start`
- add `monitor_send`
- add `monitor_read`
- add `monitor_stop`
- add `device_reset_and_capture`
- add PTY/session management for `idf.py monitor`
- guard monitor-only functionality on platforms without PTY support
- factor repeated `idf.py` command construction into `_idf_cmd()`

### Motivation
`idf.py monitor` requires a TTY-like environment, so it cannot be integrated the same way as the existing build/flash/clean tools. This change adds a PTY-backed monitor session model so MCP clients can interact with device serial output in a more complete way while keeping the existing non-monitor tools unchanged.

## Related
N/A

## Testing
Tested manually on a local MacOS ESP-IDF setup.

### Tested flows
- `monitor_start` resets the board and returns initial boot output inline
- `monitor_read` polls buffered output with remaining-line indicator
- `monitor_send` writes commands to the device; response retrieved via `monitor_read`
- `monitor_stop` kills the full process tree and drains remaining output
- `monitor_boot` captures boot log in a one-shot flow with clean cleanup
- Repeated start/stop cycles work back-to-back — no zombie processes or port locks
- Wrong port: reports "Monitor failed to start" with available ports; all other tools return clear "no session" messages
- Unexpected process exit: monitor_read returns buffered data with exit note; monitor_stop detects and cleans up the dead session

### Platform notes
- monitor tools require PTY support and are guarded on platforms where PTY is unavailable
- existing build/flash/clean/set-target behavior remains available even when PTY support is unavailable

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces long-lived subprocess/PTY session management and process-group termination logic, which can affect serial port usage and cleanup behavior across platforms. Non-monitor tools are mostly unchanged but rely on new `_idf_cmd()` IDF_PATH validation.
> 
> **Overview**
> Adds **interactive serial monitoring** to the MCP server by introducing PTY-backed tools: `monitor_start`, `monitor_send`, `monitor_read`, `monitor_stop`, plus a one-shot `monitor_boot` that resets and captures boot logs for a bounded duration.
> 
> Refactors `idf.py` invocation into `_idf_cmd()` (with explicit `IDF_PATH` validation) and adds a `MonitorSession` with background output buffering, ANSI stripping, and robust process-group cleanup; monitor features are *guarded* on platforms without PTY support and sessions are cleaned up on server exit.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 78b3581dc785ec6041223453f1a0f314f2de7022. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/idf_py_actions/mcp_ext.py
+++ b/tools/idf_py_actions/mcp_ext.py
@@ -1,10 +1,13 @@
 # SPDX-FileCopyrightText: 2024-2026 Espressif Systems (Shanghai) CO LTD
 # SPDX-License-Identifier: Apache-2.0
-
 import json
 import os
+import re
+import signal
 import subprocess
 import sys
+import threading
+import time
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


tools/idf_py_actions/mcp_ext.py
```diff
@@ -195,7 +420,271 @@ def clean_project() -> str:
...
+            max_lines = max(1, min(max_lines, 1000))
+            timeout = max(0, min(timeout, 30))
+
+            # Accumulate output, optionally waiting for new data or a pattern
+            collected: list[str] = []
+            remaining = 0
+            deadline = time.monotonic() + timeout if timeout > 0 else 0
+
+            while True:
+                lines, remaining = monitor_session.read(max_lines=max_lines)
```
dobairoland, line 633: **`monitor_read` can return more than `max_lines` lines**

The polling loop passes the original `max_lines` to every `read()` call without decrementing it by the lines already collected. When `timeout > 0`, each iteration can fetch up to `max_lines` additional lines, so the total may significantly exceed the caller's requested limit.

Suggested fix:
```python
budget = max_lines
while True:
    lines, remaining = monitor_session.read(max_lines=budget)
    if lines:
        collected.extend(lines)
        budget -= len(lines)
        if wait_for and wait_for in ''.join(collected):
            break
        if budget <= 0:
            break
    if deadline == 0:
        break
    if time.monotonic() >= deadline:
        break
    if not monitor_session.is_running:
        break
    time.sleep(0.3)
```

tools/idf_py_actions/mcp_ext.py
```diff
@@ -23,14 +26,258 @@
...
+    while time.monotonic() < deadline:
+        remaining = max(0.1, deadline - time.monotonic())
+        ready, _, _ = select.select([master_fd], [], [], min(remaining, 0.5))
+        if ready:
+            try:
+                chunk = os.read(master_fd, 4096)
+                if not chunk:
+                    break
+                parts.append(chunk.decode('utf-8', errors='replace'))
+                if wait_for and wait_for in ''.join(parts):
```
dobairoland, line 129: **O(n²) join for `wait_for` check in `_pty_read_for`**

On each received chunk, `''.join(parts)` re-joins the entire accumulated output to scan for `wait_for`. For long captures this is quadratic in total output size.

A simple fix is to check only the tail of the buffer — the match can only span the boundary between the previous accumulated text and the new chunk:
```python
parts.append(chunk.decode('utf-8', errors='replace'))
if wait_for:
    # Only the last (len(wait_for)-1 + new chunk) bytes can contain a new match
    tail = ''.join(parts)[-(len(wait_for) + len(parts[-1])):]  
    if wait_for in tail:
        break
```
Or maintain a running concatenated string instead of a list of parts.

tools/idf_py_actions/mcp_ext.py
```diff
@@ -23,14 +26,258 @@
...
+# existing build/flash/clean/set-target tools remain usable on Windows.
+try:
+    import pty
+    import select
+    import termios
+    _PTY_SUPPORTED = True
+except ImportError:
+    _PTY_SUPPORTED = False
+
+_ANSI_RE = re.compile(r'\x1b\[[0-9;]*[A-Za-z]')
```
dobairoland, line 45: **Incomplete ANSI escape sequence stripping**

The current regex `\x1b\[[0-9;]*[A-Za-z]` only handles CSI sequences. Several common sequences emitted by `idf_monitor.py` will leak through:
- CSI sequences with `?` modifier, e.g. `\x1b[?25h` (cursor show/hide)
- OSC sequences: `\x1b]0;title\x07` (terminal title)
- Single-char escapes: `\x1b(B` (character set)

A more comprehensive pattern:
```python
_ANSI_RE = re.compile(
    r'\x1b(?:'
    r'\[[0-9;?]*[A-Za-z]'    # CSI sequences (including ?)
    r'|\][^\x07]*(?:\x07|\x1b\\)'  # OSC sequences
    r'|[()][A-Z]'              # Character set designations
    r'|[A-Z]'                  # Single-char Fe escapes
    r')'
)

File: tools/idf_py_actions/mcp_ext.py [hunk 4]
```diff
@@ -195,7 +420,271 @@ def clean_project() -> str:
                 print(f'ERROR: Error cleaning: {str(e)}', file=sys.stderr)
                 return f'Error cleaning: {str(e)}'
 
+        # === Monitor Tools ===
+
+        def _require_pty(tool_name: str) -> str | None:
```
dobairoland, line 425: **`_require_pty` captures nothing from the enclosing scope — move to module level**

This helper is defined inside `start_mcp_server` but doesn't reference any local variables from that scope. Placing it at module level alongside the other `_pty_*` helper functions would make the organisation cleaner and easier to test in isolation.

dobairoland:
Thank you for the contribution. The main issue I see that this is not compatible with Windows which is a must for all of our tools. Also the documentation section about MCP support needs to be updated.

dobairoland:
Also please note that the contribution guide states that pre-commit hooks must pass for contributions. See the failed check for more information.

jinw06k:
@dobairoland Thanks for the feedback. The current implementation wraps `idf.py monitor` in a PTY to bypass the `isatty()` check. However, Python pty is Unix-only, so it doesn't work on Windows.

I've been exploring two approaches to fix this and wanted to get your input on which direction to take:

**Option A: pyserial + address decoding**
- Replace the monitor path with pyserial, which removes the PTY dependency and keeps the implementation cross-platform with **no additional dependency**. However, this would no longer run `idf.py monitor` directly, so it wouldn't be a 100% IDF Monitor equivalent. For the MCP use cases, panic/backtrace decoding could be reproduced by invoking the toolchain’s `addr2line` against the project ELF when a `Backtrace:` pattern is detected.

**Option B: PTY on Unix + PyWinpty/ConPTY on Windows**
- Keep the current `idf.py monitor` design on Unix, and use a Windows pseudoterminal backend such as PyWinpty/ConPTY on Windows. That would preserve all of the original `idf.py monitor`’s behavior, but it would **add an external dependency**.

My preference is **Option A** unless preserving full idf.py monitor parity is a requirement.

dobairoland:
@jinw06k I think option B is the best and then we will have to test this if we haven't missed anything Window-related.

The issue with option A is that we would start to develop a new monitor implementation which would have to be maintained in addition to esp-idf-monitor. Here, we could prepare a pyserial object and pass it to esp-idf-monitor's internals but the port, baud and similar would also have to be re-implemented.

We could also remove the writing capability (and hence the necessity of PTY) but then interactive applications (eg. https://github.com/espressif/esp-idf/tree/44c77cbf46844cd056c923277ece745173cb270d/examples/system/console/advanced ) could not be used. And I think it would be useful if the MCP server would able to control the ESP application that way.

So all in all, we need to evaluate option B on Windows with a console-based interactive application.

jinw06k:
@dobairoland Thanks for the feedback on Option B. Before I go down the PyWinpty path, I wanted to share a third option (prototyped and tested).

**Option C: Add --non-interactive mode to [`esp-idf-monitor`](https://github.com/espressif/esp-idf-monitor)**
I traced the two things that block `idf.py monitor` from running in a subprocess pipe:

1. The `isatty()` check in main() — rejects non-TTY stdin
2. `miniterm.Console()` in Monitor.__init__() — calls `termios.tcgetattr()` which fails on a pipe

Just around 15 lines of code changes to `esp-idf-monitor` (gated on a new `ESP_IDF_MONITOR_NON_INTERACTIVE` env var) fixes both, in the same pattern as [`ESP_IDF_MONITOR_TEST`](https://github.com/search?q=repo%3Aespressif%2Fesp-idf-monitor%20ESP_IDF_MONITOR_TEST&type=code) which already bypasses `isatty()` and sets `socket_test_mode` to skip console input:

1. Skip the [`isatty()`](https://github.com/espressif/esp-idf-monitor/blob/master/esp_idf_monitor/idf_monitor.py#L426) check (same as ESP_IDF_MONITOR_TEST)
3. Set [`socket_test_mode=True`](https://github.com/espressif/esp-idf-monitor/blob/master/esp_idf_monitor/idf_monitor.py#L159) so ConsoleReader doesn't read interactive input (same as ESP_IDF_MONITOR_TEST)
4. Replace `miniterm.Console()` with a dummy that writes to stdout
5. In `console_reader.py`, catch the termios.error and keep the thread alive without reading

I tested this locally, and all IDF monitor features work normally. The MCP server can just set the env var and read stdout. Option C does not require PTY, platform-specific code, or new dependencies. It does, however, require another PR to the esp-idf-monitor repo.

Happy to go with Option B if you prefer keeping changes to one repo, but Option C would also benefit in the long run.

dobairoland:
Yes, I didn't explain it in too much details but essentially this was option C:

> We could also remove the writing capability (and hence the necessity of PTY) but then interactive applications (eg. https://github.com/espressif/esp-idf/tree/44c77cbf46844cd056c923277ece745173cb270d/examples/system/console/advanced ) could not be used. And I think it would be useful if the MCP server would able to control the ESP application that way.

The issue with this that we'd do a lot of hacking and then comes someone who really needs this to be interactive and then we will have to implement option B.

jinw06k:
@dobairoland Interactive writing works in my Option C prototype.

The `ConsoleReader` patch reads from `sys.stdin.buffer` when no TTY is present and forwards to the device via `TAG_KEY` events. I tested `monitor_send()` and was able to get responses back.

My "non-interactive mode" wording may have been misleading — it's more of a headless mode. Full read and write work, just without the `miniterm.Console` TTY setup (which required PTY).

dobairoland:
Thank you for the explanation. Option B still sounds the best to me.

jinw06k:
Understood. I will update with option B!

PR #18476, opened 2026-04-17
fix(lwip): prevent integer overflow in sys_arch_sem_wait timeout (IDFGH-17541)

jjsch-dev:
Description / Motivation:
This PR fixes a critical integer overflow vulnerability in sys_arch_sem_wait that causes an infinite spin-loop (starvation) when a maximum or near-maximum timeout value is passed, particularly when the RTOS tick rate is configured to 1000 Hz.

Root Cause Analysis:
When a third-party LwIP application (e.g., Barracuda App Server) attempts to wait infinitely by casting -1 to u32_t, the timeout argument becomes 0xFFFFFFFF.
The current macro used to round up ticks is:
TickType_t timeout_ticks = ((timeout + portTICK_PERIOD_MS - 1) / portTICK_PERIOD_MS) + 1;

If configTICK_RATE_HZ == 1000, portTICK_PERIOD_MS is 1.
The operation 0xFFFFFFFF + 1 overflows to 0. The math evaluates to exactly 0.
This passes 0 to xSemaphoreTake, resulting in an instant errQUEUE_EMPTY instead of blocking. The caller receives SYS_ARCH_TIMEOUT immediately, causing an infinite retry loop that hijacks the CPU (Starvation) and freezes the system.

(Note: At 100 Hz, the overflow results in a 1-tick delay, causing a silent performance drop rather than a hard freeze, which is likely why this went unnoticed).

The Fix:
Added a boundary check before the math operations:
if (!timeout || timeout >= (UINT32_MAX - portTICK_PERIOD_MS))
If the requested timeout is zero (standard LwIP infinite) or large enough to trigger an overflow when adding portTICK_PERIOD_MS, it is safely forced to portMAX_DELAY. This makes the function mathematically robust across all configTICK_RATE_HZ configurations.

---

> [!NOTE]
> **Medium Risk**
> Touches FreeRTOS/LwIP semaphore wait timing; while small, it changes blocking behavior for extreme timeout values and could affect edge-case timing in networking/concurrency paths.
> 
> **Overview**
> `sys_arch_sem_wait` now guards against 32-bit overflow when converting millisecond timeouts to FreeRTOS ticks by treating `timeout == 0` and near-`UINT32_MAX` values as an infinite wait (`portMAX_DELAY`). This prevents large/"-1 cast" timeouts from wrapping to a zero-tick wait and causing immediate `SYS_ARCH_TIMEOUT` returns (and potential busy-loop starvation).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33feb7d6dd8fda16d2e5d51035e5040101dfb87d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/lwip/port/freertos/sys_arch.c
+++ b/components/lwip/port/freertos/sys_arch.c
@@ -164,8 +164,12 @@ u32_t
 sys_arch_sem_wait(sys_sem_t *sem, u32_t timeout)
 {
   BaseType_t ret;
-
-  if (!timeout) {
+  /* * Protect against Integer Overflow: 
+   * If timeout is 0 (LwIP infinite wait), or if it's so large that adding 
+   * portTICK_PERIOD_MS would overflow the 32-bit unsigned integer,
+   * we force it to wait infinitely.
+   */
+  if (!timeout || timeout >= (UINT32_MAX - portTICK_PERIOD_MS)) {
     /* wait infinite */
     ret = xSemaphoreTake((QueueHandle_t)sem, portMAX_DELAY);
     LWIP_ASSERT("taking semaphore failed", ret == pdTRUE);


jjsch-dev:
Hi team! Just following up on this PR.

We are currently finalizing the porting of the Barracuda App Server (Xedge32) to the new ESP32-P4 and ESP32-S3 on IDF v6.1. During our stress tests, we identified severe freezing/lockup issues during cold and hard resets on these targets.

We traced the root cause down to this specific lwIP behavior. Applying this patch completely resolves the lockups and stabilizes the network stack for the P4 and S3.

Since this is a critical stability blocker for our upcoming release (and likely affects other users on these new SoCs), I wanted to gently ping this thread. You can check out our project and the porting progress here: [Xedge32](https://github.com/RealTimeLogic/Xedge32/)

Please let me know if you need any additional logs, memory dumps, or testing from our P4/S3 hardware to help validate and merge this fix. Thanks for your time!

david-cermak (approved):
Thanks for the fix and the detailed root cause analysis and the overflow math:
* with portTICK_PERIOD_MS = 1, passing 0xFFFFFFFF causes timeout_ticks to wrap to 0, which makes xSemaphoreTake return immediately and puts the caller into an infinite busy loop.

**Worth noting**: 
LwIP convention for "wait forever" is `timeout = 0`, so the ideal fix would be in the middleware layer (the Barracuda App Server?) to use 0 instead of (u32_t)-1. 

But the guard you've added is a sensible defensive measure, the affected values were already producing completely wrong behavior due to overflow, and converting them to infinite waits is strictly better than a system freeze. No correct caller would be passing a ~49.7-day semaphore timeout on an embedded system, so there's no realistic risk of breaking existing behavior.

jjsch-dev:
Hi @david-cermak,

Thank you for the review and the approval!

You are completely right about the timeout = 0 convention in LwIP. Your observation was spot on—I shared this insight with the Real Time Logic team, and Wilfred has already updated the Barracuda App Server middleware to catch negative timeouts and convert them to 0 before they even reach the LwIP port.

I'm glad the defensive guard in the port layer still makes sense as a safety net to prevent system freezes.

Thanks again for your time, the detailed feedback, and the quick merge!

Best regards,
Juan

jjsch-dev:
I noticed my changes were merged internally by @david-cermak in commit d0dcfda. Closing this PR to keep the tracker clean. Thanks!

jjsch-dev:
I noticed my changes were merged internally by @david-cermak in commit d0dcfda. Closing this PR to keep the tracker clean. Thanks!

PR #18494, opened 2026-04-20
fix(sd): Disable warning about input line delay not supported (IDFGH-17561)

nebkat:
## Description

Remove confusing log message introduced in ESP-IDF v6.0.

Alternatively it should only be printed if value is anything other than 0, in which case an `SDMMC_DELAY_LINE_DISABLED = 0` or similar should be introduced.

## Checklist

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to SDMMC configuration defaults/logging and introduce a clearer sentinel value for a disabled delay line, with minimal behavioral impact beyond when warnings are emitted and when delay-line config is applied.
> 
> **Overview**
> Reduces noisy SDMMC init logging by only warning about unsupported input delay lines when the caller explicitly requests a delay line.
> 
> Introduces `SDMMC_DELAY_LINE_DISABLED = 0` and updates slot configuration to treat this as the “do not configure delay line” sentinel (replacing `!= 0` checks), keeping delay-line state untouched unless enabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b621c3496f33bd0e2162d65dafb21c601fcf46d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_driver_sdmmc/src/sd_host_sdmmc.c
+++ b/components/esp_driver_sdmmc/src/sd_host_sdmmc.c
@@ -201,7 +201,9 @@ static esp_err_t sd_host_slot_sdmmc_configure(sd_host_slot_handle_t slot, const 
 #if SOC_SDMMC_UHS_I_SUPPORTED
     ESP_RETURN_ON_FALSE(config->delayline < (SOC_SDMMC_DELAY_PHASE_NUM + 1), ESP_ERR_INVALID_ARG, TAG, "invalid delay line");
 #else
-    ESP_LOGW(TAG, "input line delay not supported, fallback to 0 delay");
+    if (config->delayline != SDMMC_DELAY_LINE_DISABLED) {
+        ESP_LOGW(TAG, "input line delay not supported, fallback to 0 delay");
+    }
 #endif
 
 #if CONFIG_IDF_TARGET_ESP32P4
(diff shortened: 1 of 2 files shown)


L-KAYA:
Hi @nebkat , thanks for your PR! You mean the warning log prints without any condition, which makes users think that their configuration is not correct when just migrated the IDF version, right?

Totally agree that this is misleading, but deleting the warning log directly is not a good idea for those who thought there should be `input line delay` feature. So could you update your PR, add a condition for this warning instead? Like:

```c
#if SOC_SDMMC_UHS_I_SUPPORTED
    ESP_RETURN_ON_FALSE(config->delayline < (SOC_SDMMC_DELAY_PHASE_NUM + 1), ESP_ERR_INVALID_ARG, TAG, "invalid delay line");
#else
    if (config->delayline > 0) {
        ESP_LOGW(TAG, "input line delay not supported, fallback to 0 delay");
    }
#endif
```

nebkat:
@L-KAYA Sure, or would it be even better to #ifdef the field away? Similar to:

https://github.com/espressif/esp-idf/blob/338f3411101a8820e12ed824c5a9815ac787ae47/components/esp_adc/include/esp_adc/adc_cali_scheme.h#L75-L83

Or alternatively shouldn't we return an error?

L-KAYA:
sha=9b621c3496f33bd0e2162d65dafb21c601fcf46d

PR #18627, opened 2026-05-16
fix(uart): clear IOMUX function on RX pin when routed via GPIO matrix (IDFGH-17675)

### LINKED ISSUE #18363: uart_set_pin(): RX GPIO matrix path does not clear IOMUX function select, causing TX-to-RX loopback
### Answers checklist.

### IDF version.

v5.5.3

### Espressif SoC revision.

ESP32-C3 (confirmed), likely affects all ESP32 variants

### Operating System used.

macOS

Command line with CMake

_No response_

### Development Kit.

ESP32-C3-DevKitM-1

### Power Supply used.

USB

When calling `uart_set_pin(UART_NUM_0, tx_io_num=20, rx_io_num=21, ...)` on ESP32-C3 (where the UART0 defaults are `U0RXD_GPIO_NUM=20`, `U0TXD_GPIO_NUM=21`), the UART should transmit and receive independently with no internal loopback.

After boot, GPIO21 retains its IOMUX function as UART0 TX (configured by the ROM bootloader for console output). When `uart_set_pin()` reassigns GPIO21 as RX via GPIO matrix, the old IOMUX TX function is **not** cleared. GPIO21 simultaneously acts as UART0 TX (via IOMUX) and UART0 RX (via GPIO matrix), causing all transmitted data to loop back into the RX FIFO.

The device receives its own TX data as an echo. Connected peripherals (e.g. Midea AC units, LD2410 presence sensors) do not respond because the echoed frame consumes the RX processing, and in some protocols the wrong protocol byte is learned from the echo.

### Steps to reproduce.

1. Use an ESP32-C3 board with default UART0 pins: GPIO20 (default RX) and GPIO21 (default TX).
2. Configure UART with the pins swapped from default: `tx_pin: GPIO20`, `rx_pin: GPIO21`.
3. Connect a UART peripheral (e.g. Midea AC, LD2410 sensor) to these pins.
4. Boot the device. The ROM bootloader configures GPIO21 as UART0 TX via IOMUX for console output.
5. Application calls `uart_set_pin(UART_NUM_0, 20, 21, -1, -1)`.
6. Since GPIO20 is not the default TX pin and GPIO21 is not the default RX pin, both go through the GPIO matrix path (not IOMUX).
7. Transmit data on UART0.
8. **Observe**: TX data appears in the RX buffer (internal loopback). The connected peripheral either doesn't respond or its response is lost behind the echo.

**Workaround**: Calling `gpio_func_sel(21, PIN_FUNC_GPIO)` before `uart_set_pin()` clears the residual IOMUX function and eliminates the loopback.

### Debug Logs.

```
Working (with gpio_func_sel workaround applied before uart_set_pin):
TX: AA 21 AC 8D 00 00 00 00 02 03 41 81 ...
RX: AA 23 AC 00 00 00 00 00 02 03 C0 00 4A 66 7F 7F ...
(AC responds normally)

Broken (without workaround):
TX: AA 21 AC 8D 00 00 00 00 00 03 41 81 ...
RX: AA 21 AC 8D 00 00 00 00 00 03 41 81 ...
(RX is identical to TX — internal loopback, then response timeout)
```

_No response_

## Root cause analysis

The asymmetry is in `uart_set_pin()` between the TX and RX GPIO matrix paths:

**TX path** ([uart.c#L1159](https://github.com/espressif/esp-idf/blob/v5.5.3/components/esp_driver_uart/src/uart.c#L1159)): calls `gpio_func_sel(tx_io_num, PIN_FUNC_GPIO)` which clears the IOMUX function select before connecting the TX signal via GPIO matrix. Any residual IOMUX function from boot is removed. ✓

**RX path** ([uart.c#L1174](https://github.com/espressif/esp-idf/blob/v5.5.3/components/esp_driver_uart/src/uart.c#L1174)): calls only `gpio_input_enable(rx_io_num)` + `esp_rom_gpio_connect_in_signal(...)`. There is no call to `gpio_func_sel()` or equivalent to clear the IOMUX function select. ✗

[`gpio_input_enable()`](https://github.com/espressif/esp-idf/blob/v5.5.3/components/esp_driver_gpio/src/gpio.c#L189-L194) → `gpio_hal_input_enable()` → `gpio_ll_input_enable()` only sets `FUN_IE` (bit 9) in the IO MUX register. It does not touch the function select field (`MCU_SEL`, bits 12-14).

## Suggested fix

Add `gpio_func_sel(rx_io_num, PIN_FUNC_GPIO)` before `gpio_input_enable(rx_io_num)` in the RX GPIO matrix path of `uart_set_pin()`, matching what the TX path already does.

## Impact

This bug has affected hundreds of users of the [ESPHome](https://esphome.io/) project (1.2M+ devices) who use ESP32-C3 boards with UART0 default pins. It has been a persistent source of regressions over multiple release cycles. The full history of ESPHome PRs and issues related to this can be found at:

https://github.com/esphome/esphome/issues?q=sort%3Aupdated-desc%20label%3Aesp-idf-uart-regressions

Key issues and PRs:

- https://github.com/esphome/esphome/issues/10132 — Original report: UART0 default pins broken on ESP32-C3
- https://github.com/esphome/esphome/issues/11823 — `gpio_reset_pin()` workaround broke external components
- https://github.com/esphome/esphome/issues/14612 — LD2410 sensor broken on ESP32-C3 after reverting workaround
- https://github.com/esphome/esphome/issues/14973 — Midea AC broken on ESP32-C3 after reverting workaround
- https://github.com/esphome/esphome/pull/14130 — ESPHome added `gpio_reset_pin()` workaround for UART0 pins
- https://github.com/esphome/esphome/pull/14363 — ESPHome reverted workaround, trusting ESP-IDF 5.5.2 fix
- https://github.com/esphome/esphome/pull/14978 — ESPHome targeted fix using `gpio_func_sel()` instead

anoshyn:
## Summary

Fix `uart_set_pin()` so the RX pin's IOMUX function is reset to `PIN_FUNC_GPIO` when the pin is routed via the GPIO matrix. Without this, residual IOMUX function from boot keeps driving the pad in parallel with the matrix-routed RX signal, producing on-pin loopback when the default UART0 TX pin is reassigned as RX.

Closes #18363.

## Root cause

The matrix RX and matrix TX paths inside `_uart_set_pin6()` are asymmetric:

- **TX path** → `gpio_matrix_output()` → `gpio_hal_matrix_out()` first calls `gpio_ll_func_sel(..., PIN_FUNC_GPIO)` before connecting the output signal (`components/esp_hal_gpio/gpio_hal.c:60-69`).
- **RX path** → `gpio_matrix_input()` → `gpio_hal_matrix_in()` only enables the input; it never clears the IOMUX function (`components/esp_hal_gpio/gpio_hal.c:48-58`).

If the ROM bootloader (or any prior init code) had selected an IOMUX peripheral function on that pin — e.g. UART0 TX on GPIO21 (ESP32-C3) or GPIO16 (ESP32-C6) — that function keeps driving the pad after `uart_set_pin()` routes UART RX through the matrix. The pin is now simultaneously driven by IOMUX (as TX) and read by the matrix (as RX), so whatever UART TX writes is echoed back into the same UART's RX FIFO. This is the regression ESPHome has been carrying workarounds for across multiple ESP-IDF releases (see issue #18363 for the full ESPHome PR history).

## Fix

Call `gpio_func_sel(rx_io_num, PIN_FUNC_GPIO)` on the RX matrix path before `gpio_matrix_input()`. Matches what TX does implicitly and what `i2c_common.c`, `twai.c`, `i3c_master.c`, `i2s_common.c`, `esp_eth_mac_esp_gpio.c` already do before routing inputs via the GPIO matrix.

```diff
 if (uart_num < SOC_UART_HP_NUM) {
+    gpio_func_sel(rx_io_num, PIN_FUNC_GPIO);
     gpio_matrix_input(rx_io_num, UART_PERIPH_SIGNAL(uart_num, SOC_UART_PERIPH_SIGNAL_RX), false);
 }
```

No build-system changes: `uart.c` already includes `esp_private/gpio.h` and `esp_driver_gpio` is already in `priv_requires`.

## Verification

Hardware: ESP32-C6 DevKit.

A small test program forces `IO_MUX[GPIO11].MCU_SEL` to a non-GPIO value, then calls `uart_set_pin(UART_NUM_1, tx=10, rx=11, -1, -1)`, then reads `MCU_SEL` back:

| Build | Pre | Post | Verdict |
|---|---|---|---|
| master (unpatched) | 0 | **0** (unchanged) | bug reproduced |
| master + this fix | 0 | **1** (`PIN_FUNC_GPIO`) | fixed |

Also: `components/esp_driver_uart/test_apps/uart` builds cleanly for esp32c6 with the patch (no new warnings).

## Alternative considered

The deeper asymmetry sits at the HAL: `gpio_hal_matrix_in()` could itself call `gpio_ll_func_sel(..., PIN_FUNC_GPIO)` so every peripheral driver routing input via the GPIO matrix gets the same hygiene as the output path. That would be a wider behavioural change touching every peripheral and is left as a follow-up if maintainers prefer it — happy to send a separate PR.

## Test plan

- [x] Builds cleanly for esp32c6 (`components/esp_driver_uart/test_apps/uart`, no new warnings)
- [x] Bug reproduced on ESP32-C6 hardware on unpatched master
- [x] Fix verified on ESP32-C6 hardware (MCU_SEL → PIN_FUNC_GPIO)
- [ ] Regression check: default-pin UART0 path (IOMUX, not matrix) still works on ESP32-C3 — recommended for reviewer / CI coverage

### DIFF
--- a/components/esp_driver_uart/src/uart.c
+++ b/components/esp_driver_uart/src/uart.c
@@ -947,6 +947,9 @@ esp_err_t _uart_set_pin6(uart_port_t uart_num, int tx_io_num, int rx_io_num, int
         if (tx_rx_same_io || !uart_try_set_iomux_pin(uart_num, rx_io_num, SOC_UART_PERIPH_SIGNAL_RX)) {
             io_reserve_mask &= ~BIT64(rx_io_num); // input IO via GPIO matrix does not need to be reserved
             if (uart_num < SOC_UART_HP_NUM) {
+                // Clear any residual IOMUX function so the pin isn't simultaneously driven by IOMUX
+                // while we route it via the GPIO matrix. Symmetric with gpio_matrix_output().
+                gpio_func_sel(rx_io_num, PIN_FUNC_GPIO);
                 gpio_matrix_input(rx_io_num, UART_PERIPH_SIGNAL(uart_num, SOC_UART_PERIPH_SIGNAL_RX), false);
             }
 #if SOC_LP_GPIO_MATRIX_SUPPORTED && (SOC_UART_LP_NUM >= 1)


songruo:
Hi @anoshyn, I have stated the reason pretty clear in https://github.com/espressif/esp-idf/issues/18363#issuecomment-4396070139. Also, if you are sticking with the "symmetry", the hardware design of the IOMUX/GPIO Matrix is not exactly symmetric for input and output side.

image

Please check this screenshot from the TRM, for a signal to go through GPIO Matrix to a pad, the func sel has to be selected to specific value, but it is not true for an input signal from the pad to go through the GPIO Matrix, the func sel is skipped.

I hope you can understand. And I have provided solutions in the thread I linked above.

anoshyn:
Thanks @songruo for the detailed explanation and the TRM reference — that clears it up. You're right: the `func_sel` requirement is asymmetric at the silicon level (output side needs it, input side doesn't), and unconditionally clearing IOMUX on the matrix-input path would silently break the IOMUX-output + GPIO-matrix-input shared-pin use case. The "symmetric with `gpio_matrix_output()`" framing in my PR description was wrong — the asymmetry is by design, not a bug.

For the original report (#18363), `CONFIG_ESP_CONSOLE_NONE` / `CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG` releasing the default UART0 IOs is the correct path — addresses the root cause (the ROM-bootloader-driven default TX pin still being driven) rather than papering over it inside `uart_set_pin()`.

I'll close this PR. If it would be useful, I'd be happy to send a small docs-only follow-up — e.g. a note on `uart_set_pin()` (or the UART peripheral page) that calls out the `CONFIG_ESP_CONSOLE_*` requirement when reassigning default-UART0 pins, so the next person who hits this finds the answer without needing to dig through the matrix/IOMUX semantics. Let me know if you'd want that, and against which file you'd prefer it.

Thanks again for taking the time to walk through the hardware reasoning.

songruo:
@anoshyn This is not specific to UART only. The usage of signal mapping is standardized in IDF peripheral drivers, and documented in https://github.com/espressif/esp-idf/blob/master/components/esp_driver_gpio/README.md, so in most of the IDF peripheral drivers (there are still a few not refactored yet), you should not see `gpio_func_sel` getting called for GPIO Matrix input.

We added a LOGI in the startup stage to notify the users about the default UART TX/RX pins, but seems like having limited effect. We will consider adding a migration note to the version where the change was introduced.

anoshyn:
Thanks for the pointer to `components/esp_driver_gpio/README.md` and for the plan to add a migration note — that addresses the discoverability gap better than anything I could land here. Closing.

PR #18656, opened 2026-05-26
esp_netif: retain RX pbuf free callback (IDFGH-17719)

dhimasardinata:
## Summary
- Store the driver RX-buffer free callback and driver handle in each custom pbuf when it is allocated.
- Keep delayed `pbuf_free()` independent from later `esp_netif` driver config teardown.
- Add an `esp_netif` regression test that clears the driver config before freeing the custom pbuf.

## Root cause
`esp_pbuf_allocate()` kept an `esp_netif_t *` inside the custom pbuf. If the netif driver config was cleared before a queued pbuf was drained, `esp_pbuf_free()` could later call through the cleared/stale `esp_netif->driver_free_rx_buffer` path.

## Verification
- `git diff --check`
- `idf.py -C examples/wifi/getting_started/station -B /tmp/espnetif_pbuf_station_build -D SDKCONFIG=/tmp/espnetif_pbuf_station_sdkconfig set-target esp32 build`
- `idf.py -C components/esp_netif/test_apps/test_app_esp_netif -B /tmp/espnetif_test_app_build -D SDKCONFIG=/tmp/espnetif_test_app_sdkconfig set-target esp32 build`

Refs espressif/arduino-esp32#12517

### DIFF
--- a/components/esp_netif/lwip/netif/esp_pbuf_ref.c
+++ b/components/esp_netif/lwip/netif/esp_pbuf_ref.c
@@ -6,12 +6,12 @@
 /**
  * @file esp_pbuf reference
  * This file handles lwip custom pbufs interfacing with esp_netif
- * and the L2 free function esp_netif_free_rx_buffer()
+ * and L2 driver RX buffer free callbacks
  */
 
 #include "lwip/mem.h"
 #include "lwip/esp_pbuf_ref.h"
-#include "esp_netif_net_stack.h"
+#include "esp_netif_lwip_internal.h"
 
 /**
  * @brief Specific pbuf structure for pbufs allocated by ESP netif
@@ -20,7 +20,8 @@
 typedef struct esp_custom_pbuf
 {
     struct pbuf_custom p;
-    esp_netif_t *esp_netif;
+    void* driver_handle;
+    void (*driver_free_rx_buffer)(void *h, void* buffer);
     void* l2_buf;
 } esp_custom_pbuf_t;
 
@@ -33,7 +34,7 @@ typedef struct esp_custom_pbuf
 static void esp_pbuf_free(struct pbuf *pbuf)
 {
     esp_custom_pbuf_t* esp_pbuf = (esp_custom_pbuf_t*)pbuf;
-    esp_netif_free_rx_buffer(esp_pbuf->esp_netif, esp_pbuf->l2_buf);
+    esp_pbuf->driver_free_rx_buffer(esp_pbuf->driver_handle, esp_pbuf->l2_buf);
     mem_free(pbuf);
 }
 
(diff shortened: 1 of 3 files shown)


components/esp_netif/lwip/netif/esp_pbuf_ref.c
```diff
@@ -33,7 +34,7 @@ typedef struct esp_custom_pbuf
 static void esp_pbuf_free(struct pbuf *pbuf)
 {
     esp_custom_pbuf_t* esp_pbuf = (esp_custom_pbuf_t*)pbuf;
```
TD-er, line 36: Maybe also check for nullptr?

components/esp_netif/lwip/netif/esp_pbuf_ref.c
```diff
@@ -54,7 +55,9 @@ struct pbuf* esp_pbuf_allocate(esp_netif_t *esp_netif, void *buffer, size_t len,
         return NULL;
     }
     esp_pbuf->p.custom_free_function = esp_pbuf_free;
-    esp_pbuf->esp_netif = esp_netif;
+    /* The pbuf can outlive esp_netif driver config teardown. */
+    esp_pbuf->driver_handle = esp_netif->driver_handle;
+    esp_pbuf->driver_free_rx_buffer = esp_netif->driver_free_rx_buffer;
     esp_pbuf->l2_buf = l2_buff;
     p = pbuf_alloced_custom(PBUF_RAW, len, PBUF_REF, &esp_pbuf->p, buffer, len);
     if (p == NULL) {
```
TD-er, line 65: Just curious, has `esp_pbuf` already been destructed when the alloc fails?

components/esp_netif/lwip/netif/esp_pbuf_ref.c
dhimasardinata: Added the null guard in eb14700. The previous helper path also assumed the callback existed, but this keeps the stored-callback custom pbuf free path from dereferencing NULL while still freeing the custom pbuf itself.

components/esp_netif/lwip/netif/esp_pbuf_ref.c
dhimasardinata: No, the caller still owns `esp_pbuf` on that path. `pbuf_alloced_custom()` uses caller-provided custom-pbuf storage; if it returns NULL it did not attach/own that storage, so the explicit `mem_free(esp_pbuf)` is the needed cleanup.

components/esp_netif/lwip/netif/esp_pbuf_ref.c
```diff
@@ -20,7 +20,8 @@
 typedef struct esp_custom_pbuf
 {
     struct pbuf_custom p;
-    esp_netif_t *esp_netif;
+    void* driver_handle;
```
david-cermak, line 23: Now, do we need to keep the driver active after `esp_netif` gets destroyed? It's not an issue with current WiFi and Ethernet glues, but something to keep in mind, or perhaps document

david-cermak (approved):
Thanks for the fix!

It makes sense to fix it this way, if there still might be some pbufs active after deleting netif. I'll think about how we could documented. but I think we can merge the fix. Thank you!

components/esp_netif/lwip/netif/esp_pbuf_ref.c
dhimasardinata: I agree this lifetime expectation is worth documenting

dhimasardinata: Resolving this here since the remaining lifetime documentation note can be handled separately if needed.

dhimasardinata: Added the lifetime note to `esp_pbuf_allocate()` documentation in e5b822a.

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/a396ab922f7eb604e3a2d14e2b9bd4dd83371f22

Issue #13116, opened 2024-02-05
Inconsistent behaviour of uart_set_rx_timeout


ScumCoder:
https://github.com/espressif/esp-idf/blob/5454d37d496a8c58542eb450467471404c606501/components/esp_driver_uart/src/uart.c#L1775

Out of the three functions that set various thresholds - `uart_set_rx_full_threshold`, `uart_set_tx_empty_threshold` and `uart_set_rx_timeout` - the first two only do that after checking that the corresponding interrupt (`UART_INTR_RXFIFO_FULL` and `UART_INTR_TXFIFO_EMPTY` respectively) is set.
The latter, however, does not check `UART_INTR_RXFIFO_TOUT` and calls `uart_hal_set_rx_timeout` straight away.

nopnop2002:
Did you specify the correct value for tout_thresh?
```
tout_thresh -- This parameter defines timeout threshold in uart symbol periods. The maximum value of threshold is 126. tout_thresh = 1, defines TOUT interrupt timeout equal to transmission time of one symbol (~11 bit) on current baudrate. If the time is expired the UART_RXFIFO_TOUT_INT interrupt is triggered. If tout_thresh == 0, the TOUT feature is disabled.
```

ScumCoder:
> Did you specify the correct value for tout_thresh?

Not sure I understand your question... I did not specify anything, this post is just an observation I had while reading the IDF source code (see the linked line).

nopnop2002:
If you use uart_set_rx_timeout(), it will not work properly unless you specify the correct value for tout_thresh.

john30:
this is indeed a weird way of implementation, at least for uart_set_tx_empty_threshold() and uart_set_rx_full_threshold() as in contrast to the documentation, the threshold is only set on the UART if the interrupt is already enabled. there is no error result if it wasn't applied, which I would have expected at least.

so basically it depends on some timing as well as on data already having arrived or sent on the UART whether the threshold will be set or not (as the driver enables the corresponding interrupts if needed)

PR #18677, opened 2026-05-31
uart: fix threshold configuration loss when interrupts are disabled (IDFGH-17749)

AbanoubSalah:
## Description
Description: Fixes an asymmetric flaw in the UART driver where uart_set_rx_full_threshold and uart_set_tx_empty_threshold silently drop developer configuration updates and return ESP_OK if executed before the hardware interrupts are enabled.

Problem: Checking uart_hal_get_intr_ena_status inside setters prevents developer from setting-up initialization config. If a developer configures hardware thresholds before calling uart_driver_install or unmasking interrupts, the settings never reach HAL while the API reports ESP_OK. Meanwhile, uart_set_rx_timeout applies configurations unconditionally.

Solution: Refactored both functions to write directly to their respective HAL threshold registers unconditionally inside the critical section, aligning their architectural behavior with uart_set_rx_timeout.

## Related
Fixes issue #13116

## Testing
It compiles perfectly with v6.1-dev

---

## Checklist

### DIFF
--- a/components/esp_driver_uart/src/uart.c
+++ b/components/esp_driver_uart/src/uart.c
@@ -2219,9 +2219,7 @@ esp_err_t uart_set_rx_full_threshold(uart_port_t uart_num, int threshold)
         return ESP_ERR_INVALID_STATE;
     }
     UART_ENTER_CRITICAL(&(uart_context[uart_num].spinlock));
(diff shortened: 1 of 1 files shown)


songruo:
sha=1ef578caac16dbbb99d7736e40802dab453b0185

PR #18684, opened 2026-06-02
feat(httpd): avoid useless string copy by introducing httpd_req_get_url_query_str_ptr() (IDFGH-17758)

0xFEEDC0DE64:
esp webserver always forced us to create just another copy to access the query parameter and this patch changes that.

Now we can direct access to the underlying memory structure containing the query parameter.

This also finally allows working with non-null-terminated strings for C++ (std::string_view).

The memory management is clear, that the pointers should not be kept or accessed after the uri handler has finished


BitsForPeople:
Would be great to have this also for the header fields.

Ashish285:
sha=3c64eea401b137dc230ffe75628ed3a298e26d4b

Ashish285 (approved):

Issue #11627, opened 2023-06-09
System crashes when we try to enter deep sleep after terminating the BLE( Properly Terminating the BLE as per docs ) (IDFGH-10369)


AbIoTSolutionsPrivateLimited:
### Answers checklist.

- [X] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [X] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

5.2.0

### Operating System used.

Windows

### How did you build your project?

Eclipse IDE

### If you are using Windows, please specify command line type.

None

### Development Kit.

ESP32S3

### Power Supply used.

USB

### What is the expected behavior?

The system must enter the deep sleep without any crash.

1. We first terminate the ble using the following commands as mentioned in the docs:
2. `esp_bluedroid_disable();
        esp_bluedroid_deinit();
        esp_bt_controller_disable();
        esp_bt_controller_deinit();`
3. Then after a small delay of few seconds enter in deep sleep mode.
4. Now the system must disable ble and enter the deep sleep.

### What is the actual behavior?

The system crashes at an random place:
0x40375aca: panic_abort at C:/Users/suhas/esp/esp-idf/components/esp_system/panic.c:452

0x40381a75: esp_system_abort at C:/Users/suhas/esp/esp-idf/components/esp_system/port/esp_system_chip.c:90

0x403876ea: abort at C:/Users/suhas/esp/esp-idf/components/newlib/abort.c:38

0x4037705b: lock_acquire_generic at C:/Users/suhas/esp/esp-idf/components/newlib/locks.c:130

0x40377171: _lock_acquire_recursive at C:/Users/suhas/esp/esp-idf/components/newlib/locks.c:158

0x40377240: __retarget_lock_acquire_recursive at C:/Users/suhas/esp/esp-idf/components/newlib/locks.c:314 (discriminator 3)

0x4208dde1: __env_lock at /builds/idf/crosstool-NG/.build/HOST-x86_64-w64-mingw32/xtensa-esp32s3-elf/src/newlib/newlib/libc/stdlib/envlock.c:43

[...]

PR #18706, opened 2026-06-10
fix(usb_serial_jtag): flush TX FIFO at end of usb_serial_jtag_write (IDFGH-17790)

### LINKED ISSUE #17432: ESP32-C6/H2 USB_SERIAL_JTAG not initialised correctly when logging is disabled (IDFGH-16276)
### Answers checklist.

### IDF version.

v6.0-dev-1489-g4e036983a7

### Espressif SoC revision.

ESP32-C6/H2

### Operating System used.

Linux

Command line with idf.py

None

### Development Kit.

ESP32-C6-DevKitC1

### Power Supply used.

USB

When using USB_SERIAL_JTAG as the primary console, all application output/input should be directed over that USB connection. If the device is reset using the RST button, by Hangup over UART or by Hangup over USB_SERIAL_JTAG CDC, the device should initialise the USB_SERIAL_JTAG CDC and output any data over that. If the device is plugged in to that port, it should work as soon as it has been enumerated by the host and is connected using a serial terminal application.

When starting `idf.py monitor` the device is reset as expected and output over the USB_SERIAL_JTAG works. When the device is reset in hardware, the USB_SERIAL_JTAG phy is correctly put into reset but appears to not be properly initialised IF LOGGING IS DISABLED. if logging is enabled, it works just fine, however it should work  

### Steps to reproduce.

1. Build the `usb_serial_echo` example, disable logging in menuconfig
2. flash device
3. open `idf.py monitor` and observe correct interaction over serial terminal.
4. reset device using  HARDWARE RESET (not Ctrl+T,Ctrl+R) or by unplugging and replugging the USB (not UART) port, ensuring device is fully powered down.
5. observe that nothing happens over the USB_SERIAL_JTAG interface and no output/input happens. (incorrect behaviour)
6. reset device using Ctrl+T,Ctrl+R
7. observe that serial monitor now behaves correctly again.
...

### Debug Logs.

```plain
N/A, only reproducable without logs as far as i can tell.
```

### More Information.

It looks like the console is trying to fallback to the secondary output because the USB_SERIAL_JTAG is not enumerated by the host in time for the application?

MaikVermeulen:
## Summary

`usb_serial_jtag_write()` (the USB-Serial/JTAG VFS write path) emits every byte but never flushes the TX FIFO. The final bytes of a write — and the zero-length packet that terminates a 64-byte-multiple USB transfer — can therefore linger in the FIFO and never be delivered across a USB re-enumeration after a chip reset. On ESP32-C6/H2 this intermittently wedges the host CDC pipe:

- **Console (no-driver path):** missing `printf`/log output on cold boot, especially with logging disabled.
- **OpenThread RCP (driver-backed path):** OpenThread sends every spinel frame via `otPlatUartSend → write() → usb_serial_jtag_write`, and `otPlatUartFlush()` is a no-op, so the spinel link goes dead and the host's TLS/handshake never gets a response. Recovery requires a physical unplug.

The fix mirrors the existing `'\n'` handling in `usb_serial_jtag_tx_char_no_driver()` and the `fsync` path, which already flush to "make sure the line doesn't linger in fifo" — we simply flush once at the end of every write.

```c
        s_ctx.tx_func(fd, c);
    }
    usb_serial_jtag_ll_txfifo_flush();   // <-- added
    _lock_release_recursive(&s_ctx.write_lock);
    return size;
```

## Validation

Reproduced and fixed on physical ESP32-C6 hardware via an A/B cold-plug test (physical VBUS power-cycle per iteration, one handshake-over-USB attempt per boot, logging disabled — the condition that exposes the bug). The only difference between the two images was this flush:

| Build | Flush | Result |
|---|---|---|
| Baseline | absent | ~33% timeouts (host never gets a response) |
| Fixed | present | 0 timeouts over 20 consecutive cold-plugs |

## References

- Closes #17432 (ESP32-C6/H2 USB_SERIAL_JTAG not initialised correctly when logging is disabled)
- Related: openthread/openthread discussion #11627 (USB OT-RCP intermittent replug failures), where the same "flush after sending" fix was identified.

The fix was first posted by me in the discussion on #17432.

### DIFF
--- a/components/esp_driver_usb_serial_jtag/src/usb_serial_jtag_vfs.c
+++ b/components/esp_driver_usb_serial_jtag/src/usb_serial_jtag_vfs.c
@@ -202,6 +202,11 @@ static ssize_t usb_serial_jtag_write(__attribute__((unused)) void *ctx, int fd, 
         }
         s_ctx.tx_func(fd, c);
     }
+    // Flush the TX FIFO so the final bytes of the write reach the host. Otherwise
+    // the trailing bytes (and the zero-length packet that terminates a
+    // 64-byte-multiple transfer) can linger and never be delivered across a USB
+    // re-enumeration, wedging the host CDC pipe after a chip reset. See #17432.
+    usb_serial_jtag_ll_txfifo_flush();
     _lock_release_recursive(&s_ctx.write_lock);
     return size;
 }


mythbuster5:
Sorry for the late reply. I remember I've seen  a similar issue before. And we have provided a fix here https://github.com/espressif/esp-idf/commit/132db34cfebaf4f03cbce55af427e54c4e9f7f2f . Which version are you using, is it contain this fix?

MaikVermeulen:
> Sorry for the late reply. I remember I've seen a similar issue before. And we have provided a fix here [132db34](https://github.com/espressif/esp-idf/commit/132db34cfebaf4f03cbce55af427e54c4e9f7f2f) . Which version are you using, is it contain this fix?

Yes, like shown in the references in the PR description, this fix is actually based on the one you reference. However, that fix is only for Openthread, while this issue can actually happen on all ESP USB traffic. So this PR moves it to the appropriate place.

We've used it extensively in the field for 2 months now, and all USB related issues have been resolved, while previously we had ~12% of installations reporting issues. So I'm very confident that this PR is the proper fix.

igrr:
Please note that POSIX `write` on its own is not guaranteed to immediately flush the output to the device. If you are calling write and expecting to immediately get the data written, the usual pattern is to also call fsync to tell the driver to flush the write.

If you are calling write via a libc function, you need to call `fsync(fileno(fp))`.

The proposed change would flush the output for every single write, which would be a regression for applications which rely on the hardware FIFO for coalescing short writes.

(I think the linked commit should have used fsync instead of pulling in an LL header)

PR #18729, opened 2026-06-17
fix(tools): add UTF-8 encoding and Windows short path support for Unicode paths (IDFGH-17822)

omeryildiz58:
## Description

On Windows, when the user's home directory contains non-ASCII characters (e.g. Turkish `Ö`, Chinese, Japanese, Korean characters), several ESP-IDF build tools fail because they read UTF-8 encoded build files using the system's default ANSI code page (e.g. `cp1254` for Turkish locale).

### Root Cause

CMake generates build configuration files (`config.env`, `ldgen_libraries`) encoded in UTF-8. However, the Python scripts that consume these files use `argparse.FileType('r')` **without** specifying an encoding, causing Python to open them with the system's default ANSI code page.

For example, on a Turkish Windows system (code page `cp1254`), the UTF-8 bytes for `Ö` (`0xC3 0x96`) are incorrectly decoded as `Ã–` (U+00C3 U+2013). This corrupts all file paths containing the user's home directory, leading to `FileNotFoundError` during the build.

Additionally, the GCC cross-compiler toolchain's `objdump` does not support Unicode characters in file paths on Windows, even when the paths are correctly decoded by Python.

### Affected Tools

- `tools/kconfig_new/prepare_kconfig_files.py` — fails to write `kconfigs.in` because the output path is corrupted
- `tools/ldgen/ldgen.py` — fails when calling `objdump` on library files because paths contain non-ASCII characters

### Fix

1. **Explicit UTF-8 encoding**: Changed `argparse.FileType('r')` to `argparse.FileType('r', encoding='utf-8')` for `--env-file` and `--libraries-file` arguments, ensuring build configuration files are read with the correct encoding regardless of the system locale.

2. **Windows 8.3 short path resolution**: Added a `get_short_path_name()` helper in `ldgen.py` that converts long Unicode paths to DOS 8.3 short paths (e.g. `C:\Users\MERYIL~1\...`) before passing them to `objdump`. This works around the GCC toolchain's lack of Unicode path support on Windows.

3. **Robust output decoding**: Changed `.decode()` to `.decode('utf-8', errors='replace')` for `objdump` output to prevent decoding crashes.

### Testing

Tested on Windows 11 with:
- User path: `C:\Users\Ömer YILDIZ\workspace\scan`
- System locale: Turkish (cp1254)
- ESP-IDF v6.0.1
- Both clean and incremental builds succeed after the fix

### Related

This fix also requires a corresponding change in the `esp-idf-kconfig` package (`kconfgen/core.py`) which uses the same `argparse.FileType('r')` pattern for `--env-file`. A separate PR will be submitted to that repository.

### DIFF
--- a/tools/kconfig_new/prepare_kconfig_files.py
+++ b/tools/kconfig_new/prepare_kconfig_files.py
@@ -87,7 +87,7 @@ def main() -> None:
 
     parser.add_argument(
         '--env-file',
-        type=argparse.FileType('r'),
+        type=argparse.FileType('r', encoding='utf-8'),
         help='Optional file to load environment variables from. Contents '
         'should be a JSON object where each key/value pair is a variable.',
     )
--- a/tools/ldgen/ldgen.py
+++ b/tools/ldgen/ldgen.py
@@ -3,6 +3,7 @@
 # SPDX-FileCopyrightText: 2021-2026 Espressif Systems (Shanghai) CO LTD
 # SPDX-License-Identifier: Apache-2.0
 #
+import ctypes
 import errno
 import hashlib
 import json
@@ -17,7 +18,6 @@ from io import StringIO
 import rich_click as click
 from esp_pylib.cli_options import MutuallyExclusiveOption
 from esp_pylib.cli_options import OptionEatAll
-import ctypes
 
 def get_short_path_name(long_name):
     if os.name != 'nt':
--- a/tools/ldgen/ldgen.py
+++ b/tools/ldgen/ldgen.py
@@ -17,6 +17,31 @@ from io import StringIO
 import rich_click as click
 from esp_pylib.cli_options import MutuallyExclusiveOption
 from esp_pylib.cli_options import OptionEatAll
+import ctypes
+
+def get_short_path_name(long_name):
+    if os.name != 'nt':
+        return long_name
+    try:
+        dirname, filename = os.path.split(long_name)
+        normalized_dir = os.path.normpath(dirname)
+        buf_size = ctypes.windll.kernel32.GetShortPathNameW(normalized_dir, None, 0)
+        if buf_size == 0:
+            normalized_name = os.path.normpath(long_name)
+            buf_size = ctypes.windll.kernel32.GetShortPathNameW(normalized_name, None, 0)
+            if buf_size == 0:
+                return long_name
+            output_buf = ctypes.create_unicode_buffer(buf_size)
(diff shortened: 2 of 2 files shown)


jakub-kocka:
Hello, @omeryildiz58,

Thank you for your Pull Request. However, based on your description and changes, I believe that is not the course we would like to take, and this PR won't be addressed as is. We are improving these encoding/locale issues, and a couple of fixes are still planned.

Here is why this won't be addressed:

- You mentioned IDF v6.0.1 — the encoding/locale improvements are already on master ([ea7a7db](https://github.com/espressif/esp-idf/commit/ea7a7db4b146aeb009aac0428ed4011c60f05dbd), [6e4ed2c](https://github.com/espressif/esp-idf/commit/6e4ed2c204696db1e2fc1518e8e105cb500863a5)) were merged after the v6.0.1 release (2026-04-22) they are expected in an upcoming release. Your PR goes against the direction of that work: configure a UTF-8-capable environment rather than adding per-tool legacy code-page workarounds and 8.3 short-path hacks.

- I understand, it can be frustrating that you are required to configure Unicode encoding, but your mentioned cp1254 is not deprecated, but is stated to be legacy in favour of UTF-8.

- Not all the files that would need this change are addressed - incomplete coverage of scripts

- ❗ Your Windows 8.3 short-path workaround (GetShortPathNameW) is a legacy compatibility hack. It converts Unicode paths to ASCII aliases for tools that do not handle non-ASCII paths. Relying on 8.3 names is fragile (they can be disabled on NTFS and are not guaranteed) and not a foundation we want to build on.

-  The `.decode('utf-8', errors='replace')` change is unrelated to the reported failure. The actual problem is objdump being invoked with a non-ASCII file path on Windows. Changing how we decode objdump stdout does not address that. In practice, ldgen only parses ELF section names (.text, .data, …), which are ASCII regardless; this decode change adds little and can mask real encoding mismatches on legacy code-page systems

---

To fix your issue, please go with the recommended ways introduced by the mentioned commits:

> Recommended fix (persistent):
     - Enable "Beta: Use Unicode UTF-8 for worldwide language support" in Windows language/region settings.   
     If you still see issues or this warning persists, you can also try one of these workarounds:
                 - Run "chcp 65001" in the current console before running idf.py.
                 - Set PYTHONUTF8=1 before running idf.py (or set it globally).
                Note: "chcp 65001" must be run in each new console session unless your console/profile is configured to apply it automatically.


Thank you for your understanding, and the points you raised that will be considered in the task that is targeted at this problem

PR #18731, opened 2026-06-17
Add preferred SBC sink codec capabilities for sample frequencies (IDFGH-17826)

cociweb:
Add preferred SBC sink codec capabilities for sample frequencies

## Description

Adding preferred SBC codec possibility at compile time.

## Related

Fixes issue when the source is always negotiated. this PR make it possible to limit (and force) the advertised sample rates.

## Testing
Docker

---

## Checklist

### DIFF
--- a/components/bt/host/bluedroid/btc/profile/std/a2dp/bta_av_co.c
+++ b/components/bt/host/bluedroid/btc/profile/std/a2dp/bta_av_co.c
@@ -88,8 +88,16 @@ const tA2D_SBC_CIE bta_av_co_sbc_caps = {
 };
 
 /* SBC SINK codec capabilities */
+#if defined(BTC_AV_SBC_SINK_SAMP_FREQ_48_ONLY)
+const tA2D_SBC_CIE bta_av_co_sbc_sink_caps = {
+    A2D_SBC_IE_SAMP_FREQ_48, /* samp_freq - 48kHz only */
+#elif defined(BTC_AV_SBC_SINK_SAMP_FREQ_44_ONLY)
+const tA2D_SBC_CIE bta_av_co_sbc_sink_caps = {
+    A2D_SBC_IE_SAMP_FREQ_44, /* samp_freq - 44.1kHz only */
+#else
 const tA2D_SBC_CIE bta_av_co_sbc_sink_caps = {
     (A2D_SBC_IE_SAMP_FREQ_48 | A2D_SBC_IE_SAMP_FREQ_44), /* samp_freq */
+#endif
     (A2D_SBC_IE_CH_MD_MONO | A2D_SBC_IE_CH_MD_STEREO | A2D_SBC_IE_CH_MD_JOINT | A2D_SBC_IE_CH_MD_DUAL), /* ch_mode */
     (A2D_SBC_IE_BLOCKS_16 | A2D_SBC_IE_BLOCKS_12 | A2D_SBC_IE_BLOCKS_8 | A2D_SBC_IE_BLOCKS_4), /* block_len */
     (A2D_SBC_IE_SUBBAND_4 | A2D_SBC_IE_SUBBAND_8), /* num_subbands */


Finney-esp:
Hi @cociweb , the code you provided seems incomplete. Could you please clarify what function needs to be implemented? Thanks.
If you want to set the preferred configuration of SBC, we have provided an API (esp_a2d_source_set_pref_mcc) to modify the configuration.

cociweb:
Hello @Finney-esp, as of now, both 44.1 and 48 kHz capabilities are advertised by the sink towards the source, and source (any 3rd party) selects from it. With this adjustment, we can force one codec (or the other) at compile time. It is very useful to be forced if the output is narrowed to one (or the other) sample rate, without cpu conversion. So `BTC_AV_SBC_SINK_SAMP_FREQ_44_ONLY` and `BTC_AV_SBC_SINK_SAMP_FREQ_48_ONLY` would be a build flag. Otherwise, both codecs would be compiled.

the esp_a2d_source_set_pref_mcc is an api for the source that is not relevant in our case. This mod is useful for the sink.

Please guide me if I missed anything and point me to where I should make further adjustments. I would be very happy if 44.1/48(/both by default) kHz would be selectable during compile time for the sink.

Finney-esp:
> Hello @Finney-esp, as of now, both 44.1 and 48 kHz capabilities are advertised by the sink towards the source, and source (any 3rd party) selects from it. With this adjustment, we can force one codec (or the other) at compile time. It is very useful to be forced if the output is narrowed to one (or the other) sample rate, without cpu conversion. So `BTC_AV_SBC_SINK_SAMP_FREQ_44_ONLY` and `BTC_AV_SBC_SINK_SAMP_FREQ_48_ONLY` would be a build flag. Otherwise, both codecs would be compiled.
> 
> the esp_a2d_source_set_pref_mcc is an api for the source that is not relevant in our case. This mod is useful for the sink.
> 
> Please guide me if I missed anything and point me to where I should make further adjustments. I would be very happy if 44.1/48(/both by default) kHz would be selectable during compile time for the sink.

Hello @cociweb , thanks for your reply.
According to the A2DP profile, when using the SBC codec, the sink must support 44.1 and 48 kHz; otherwise, the sink's capabilities do not meet the protocol requirements and may result in the failure of configuration negotiation. Therefore, we cannot separately set 44.1 or 48 kHz for the sink's capabilities.

cociweb:
Ok, but in this case the source select (an unoptomal) sample rate. Usually the (extetnal) source has compute capability, but if an esp is the sink, than we have a limited compute resource to transcode it agan, if the source selects an unoptomal from the speaker pov. - that is why it is required to force the sample rate by the sink.

Finney-esp:
> Ok, but in this case the source select (an unoptomal) sample rate. Usually the (extetnal) source has compute capability, but if an esp is the sink, than we have a limited compute resource to transcode it agan, if the source selects an unoptomal from the speaker pov. - that is why it is required to force the sample rate by the sink.

Does your sink device support initiating a connection to the source? If so, there might be a way to set the preferred configuration. If your sink device can only passively accept connections, then the sink will not be able to actively set the preferred configuration, and as a sink device, it must support both 48 and 44.1 kHz sampling rates.

cociweb:
No, the source init the connection. It requests the sink capabilities, then the source negotiate the connection by selecting the best commonly supported codec.

Finney-esp:
> No, the source init the connection. It requests the sink capabilities, then the source negotiate the connection by selecting the best commonly supported codec.

In this case, it is up to the source to decide which configuration to adopt. The sink only needs to response its own capabilities to the source. Therefore, the sink cannot decide whether to choose 44.1 or 48 kHz.

cociweb:
> source to decide which configuration to adopt. 

Exactly! That's why it is required to force it by the sink. If we limit the sink, then the negotiation will be forced to the source.

Finney-esp:
> > source to decide which configuration to adopt.
> 
> Exactly! That's why it is required to force it by the sink. If we limit the sink, then the negotiation will be forced to the source.

Given the official profile requirements and the compatibility with the source, we cannot modify the IDF to make the sink support only a certain sampling rate. If you really need to modify the IDF to adapt to your usage conditions, you can do it yourself by modifying the IDF, such as removing a certain sampling rate from the sink capability. However, there is a risk of failing to negotiate with the source.

cociweb:
I'm sad to read this. Actually, it is not a modification, it is an addition by compile switch which goves more freedom (keeping the original wow).

Anyways, I accept your decline. Thank you for your time&effort.

PR #16202, opened 2025-06-27
fix(esp_http_server): Fix regression in httpd_cookie_key_value introduced by #15288

iranl:
## Description

PR #15288 introduced a regression in `httpd_cookie_key_value`. 

The API reference states:

`val_size -- [inout] Pointer to size of the user buffer "val". This variable will contain cookie length if ESP_OK is returned and required buffer length in case ESP_ERR_HTTPD_RESULT_TRUNC is returned.`

Since PR #15288 `val_size` will not be modified in case ESP_ERR_HTTPD_RESULT_TRUNC is returned and thus will not contain the required buffer length as stated in the API reference.

## Related

- #15288 
- https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_server.html#_CPPv424httpd_req_get_cookie_valP11httpd_req_tPKcPcP6size_t

## Testing

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [x] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [x] Documentation is updated as needed.
- [x] Tests are updated or added as necessary.
- [x] Code is well-commented, especially in complex areas.
- [x] Git history is clean — commits are squashed to the minimum necessary.

PR #18734, opened 2026-06-18
Fix header value truncation detection in httpd_req_get_hdr_value_str (IDFGH-17829)

brabema:
## Description

This PR fixes truncation reporting in `httpd_req_get_hdr_value_str()`.

The function detected truncation with `val_size < full_size`, where `full_size` is the return value of `strlcpy()`. However, `strlcpy()` returns `strlen(src)`, excluding the terminating null byte. Therefore truncation occurs when `strlen(src) >= val_size`.

At the boundary case where `strlen(src) == val_size`, the value is copied as `val_size - 1` characters followed by NUL, so the returned header value is truncated. However, the function previously returned `ESP_OK`, so the caller was not informed that truncation occurred.

This PR changes the check to `val_size <= full_size` and corrects the misleading comment about `strlcpy()`'s return value.

No memory-safety impact is intended: for `val_size > 0`, `strlcpy()` still writes a NUL-terminated destination buffer. The issue is that truncation was not reported to the caller at the exact-size boundary.

## Related

Related to #16202, which fixed the same truncation-reporting class for `httpd_cookie_key_value()`. `httpd_req_get_hdr_value_str()` still had the old boundary check.

## Testing

Tested locally with a header value whose length exactly matches `val_size`.

Verified that:

- before the fix, `httpd_req_get_hdr_value_str()` returned `ESP_OK` even though the copied value was truncated to `val_size - 1` characters plus NUL;
- after the fix, the same boundary case returns `ESP_ERR_HTTPD_RESULT_TRUNC`;
- non-truncated header values still return `ESP_OK`;
- the destination buffer remains NUL-terminated when `val_size > 0`.

---

## Checklist

### DIFF
--- a/components/esp_http_server/src/httpd_parse.c
+++ b/components/esp_http_server/src/httpd_parse.c
@@ -1121,12 +1121,13 @@ esp_err_t httpd_req_get_hdr_value_str(httpd_req_t *r, const char *field, char *v
         }
 
         /* Get the NULL terminated value and copy it to the caller's buffer.
-         * Note `strlcpy()` will always return the size of the source string
-         * including terminimating null.*/
+         * strlcpy() returns strlen() of the source string (the terminating
+         * null is NOT counted), so only val_size - 1 characters fit. */
         size_t full_size = strlcpy(val, val_ptr, val_size);
 
-        /* If buffer length is smaller than needed, return truncation error */
-        if (val_size < full_size) {
+        /* If the value did not fit in the buffer it was truncated, i.e. its
+         * length reached or exceeded val_size. Return truncation error. */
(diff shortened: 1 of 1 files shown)


Ashish285 (approved):

Ashish285:
Hey @brabema , thanks for the find and the fix. LGTM!

Ashish285:
sha=c565ed9e478c3704ed0287e39c2180023a3aeab6

Alvin1Zhang:
Thanks for contribution, changes have been merged with https://github.com/espressif/esp-idf/commit/131e66a071d6a49adf6494ccf4d83b4e2aac74f8.

PR #18748, opened 2026-06-21
fix(bootloader): sync IDF_TARGET env in subproject (IDFGH-17847)

### LINKED ISSUE #1883: Build bug: bootloader subproject can use stale IDF_TARGET in VS Code multi-root workspaces (IDFGH-17846) (VSC-1968)
### Answers checklist.

### IDF version.

v6.0.1

### Operating System used.

Windows

VS Code IDE

CMD

In a VS Code multi-root workspace that contains several independent ESP-IDF project folders, each folder should build with its own selected target.

For example:

```text
workspace
├── p4/  -> esp32p4
├── c5/  -> esp32c5
├── c2/  -> esp32c2
└── s3/  -> esp32s3
```
After switching the active workspace folder from p4 to c5, building the c5 folder should configure both the main project and the bootloader subproject as esp32c5.
If the parent build invokes the bootloader subproject with:
```text
-DIDF_TARGET=esp32c5
```
the bootloader subproject should consistently use esp32c5, regardless of a stale IDF_TARGET value inherited from the parent process environment.

After switching between different ESP-IDF project folders in the same VS Code multi-root workspace, the main project is configured with the correct target, but the bootloader subproject can still see a stale target from the previously active folder.

For example, after configuring/building the `p4` folder, switching to the `c5` folder can leave the inherited environment with:

```text
IDF_TARGET=esp32p4
```
The c5 project then invokes the bootloader subproject with the correct command-line target:
```text
-DIDF_TARGET=esp32c5
```
but bootloader configuration fails with:
```text
IDF_TARGET 'esp32c5' in CMake cache does not match currently selected IDF_TARGET 'esp32p4'. To change the target, clear the build directory and sdkconfig file, and build the project again.
```

### Steps to reproduce.

1. Create or open a VS Code multi-root workspace on Windows.
2. Add multiple independent ESP-IDF project folders to the workspace.
3. Configure the folders with different targets, for example:
```text
   p4 -> esp32p4
   c5 -> esp32c5
   c2 -> esp32c2
   s3 -> esp32s3
```
4. Select the p4 workspace folder in the ESP-IDF extension and configure/build it as esp32p4.
5. Without closing VS Code, switch the ESP-IDF extension context to the c5 workspace folder.
6. Configure/build the c5 folder as esp32c5. The generated CMake invocation contains the correct target:
```text
-DIDF_TARGET=esp32c5
```
7. Observe that the main project starts configuring/building as esp32c5.
8. Observe that the bootloader subproject configure step fails with an esp32c5 vs esp32p4 target mismatch.
The same failure can be reduced to a stale environment target by setting a previous folder's target before configuring a different folder:
```powershell
$env:IDF_TARGET = "esp32p4"
```
then configuring/building an `esp32c5` project whose generated bootloader configure command passes `-DIDF_TARGET=esp32c5`.

### Build or installation Logs.

```plain
Performing configure step for 'bootloader'
FAILED: bootloader-prefix/src/bootloader-stamp/bootloader-configure

cmake.exe -DSDKCONFIG=.../sdkconfig
          -DIDF_PATH=...
          -DIDF_TARGET=esp32c5
          -DPYTHON_DEPS_CHECKED=1
          -DPYTHON=...
          -DEXTRA_COMPONENT_DIRS=.../components/bootloader
          -DPROJECT_SOURCE_DIR=...
          -DIGNORE_EXTRA_COMPONENT=
          -GNinja
          -S .../components/bootloader/subproject
          -B .../build/bootloader

CMake Error at tools/cmake/targets.cmake:98 (message):
  IDF_TARGET 'esp32c5' in CMake cache does not match currently selected
  IDF_TARGET 'esp32p4'. To change the target, clear the build directory and
  sdkconfig file, and build the project again.
```

_No response_

### More Information.

This is not a normal "change target without cleaning build directory" case. The failing workspace folder has its own build directory and sdkconfig, and the generated bootloader configure command contains the correct `-DIDF_TARGET`.

The mismatch appears because the bootloader subproject receives the correct CMake variable/cache value while still inheriting a stale `IDF_TARGET` environment value from the VS Code multi-root task/process environment.

The bootloader subproject already requires `IDF_TARGET` to be passed by the parent build:

```cmake
if(NOT IDF_TARGET)
    message(FATAL_ERROR "Bootloader subproject expects the IDF_TARGET variable to be passed "
        "in by the parent build process.")
endif()
```
However, `tools/cmake/targets.cmake` can still observe `$ENV{IDF_TARGET}`. If the parent process environment contains the target from a previously active workspace folder, the bootloader subproject can report a mismatch even though the subproject CMake command received the correct `-DIDF_TARGET`.
Synchronizing the environment in the bootloader subproject before including the ESP-IDF project machinery fixes the mismatch:
```cmake
set(ENV{IDF_TARGET} "${IDF_TARGET}")
```
After this change, the same stale-environment test configures the bootloader successfully and generates an ESP32-C5 bootloader image.
I have a proposed fix here:
https://github.com/espressif/esp-idf/compare/master...kxt116:esp-idf:fix/bootloader-idf-target-env

kxt116:
## Description

The bootloader subproject receives `IDF_TARGET` from the parent build via `-DIDF_TARGET=...`, and it already treats a missing `IDF_TARGET` CMake variable as fatal.

However, the ESP-IDF target initialization code can still observe `$ENV{IDF_TARGET}`. In environments where the parent process has a stale target, the bootloader subproject can fail with a target mismatch even though the generated bootloader CMake command passes the correct `-DIDF_TARGET`.

One practical case is a Windows VS Code ESP-IDF multi-root workspace containing several independent ESP-IDF project folders, where each folder targets a different chip. For example:

```text
workspace
├── p4/  -> esp32p4
├── c5/  -> esp32c5
├── c2/  -> esp32c2
└── s3/  -> esp32s3
```

After configuring/building one folder, switching the ESP-IDF extension context to another folder can leave the task/process environment with the previous folder's `IDF_TARGET`. The new folder's main project is configured with the requested target, but the bootloader subproject can inherit the stale environment value and fail during target initialization.

For example, when switching from an `esp32p4` folder to an `esp32c5` folder, the generated bootloader configure command can correctly contain:

```text
-DIDF_TARGET=esp32c5
```

but bootloader configuration still fails if the inherited environment contains:

```text
IDF_TARGET=esp32p4
```

This change synchronizes the bootloader subproject environment with the explicitly passed `IDF_TARGET` before including the ESP-IDF project machinery:

```cmake
set(ENV{IDF_TARGET} "${IDF_TARGET}")
```

That keeps the CMake variable/cache value and the environment value consistent inside the bootloader subproject.

## Related issue

Fixes espressif/vscode-esp-idf-extension#1883

## Testing

Verified locally on Windows with ESP-IDF v6.0.1.

Reproduced the failure in a VS Code multi-root workspace with multiple independent ESP-IDF firmware folders configured for different targets (`esp32p4`, `esp32c5`, `esp32c2`, and `esp32s3`). After switching from the `esp32p4` folder to the `esp32c5` folder, the `esp32c5` bootloader configure step failed because the inherited environment still selected `esp32p4`.

Also reduced the issue to a direct stale-environment reproduction by setting:

```powershell
$env:IDF_TARGET = "esp32p4"
```

Then configured/built an `esp32c5` project whose generated bootloader configure command passed:

```text
-DIDF_TARGET=esp32c5
```

Before this change, the bootloader configure step failed with:

```text
IDF_TARGET 'esp32c5' in CMake cache does not match currently selected IDF_TARGET 'esp32p4'
```

After this change, the bootloader configured successfully and generated an ESP32-C5 bootloader image.

## Notes

This is intentionally limited to the bootloader subproject. The parent build still owns target selection; the subproject only mirrors the explicitly passed target into its own process environment so ESP-IDF target initialization sees a consistent value.

---

> [!NOTE]
> **Low Risk**
> Single-line CMake env sync in the bootloader subproject only; no auth, data, or application runtime behavior changes.
> 
> **Overview**
> Fixes bootloader configure failures when the process still has a **stale `IDF_TARGET` in the environment** (e.g. switching chip targets in a VS Code multi-root ESP-IDF workspace) even though the parent passes the correct `-DIDF_TARGET=...`.
> 
> After the existing fatal check that `IDF_TARGET` is defined, the bootloader subproject now runs **`set(ENV{IDF_TARGET} "${IDF_TARGET}")`** so ESP-IDF target initialization sees the same target in CMake and in `$ENV{IDF_TARGET}`, avoiding cache-vs-environment mismatch errors. Scope is limited to the bootloader subproject; the parent build still selects the target.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbd12709ab9214d630446cd66e4a8af273f057e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/bootloader/subproject/CMakeLists.txt
+++ b/components/bootloader/subproject/CMakeLists.txt
@@ -14,6 +14,7 @@ if(NOT IDF_TARGET)
     message(FATAL_ERROR "Bootloader subproject expects the IDF_TARGET variable to be passed "
         "in by the parent build process.")
 endif()
+set(ENV{IDF_TARGET} "${IDF_TARGET}")
 
 if(IDF_BUILD_V2)
     include(CMakeLists_v2.txt)


meetpatelvgec:
Hi @kxt116 
Thanks for raising the MR to fix IDF_TARGET sync issue. We are reviewing it internally and will get back to you soon.

meetpatelvgec (approved):

sudeep-mohanty:
sha=dbd12709ab9214d630446cd66e4a8af273f057e2

meetpatelvgec:
Hi @kxt116 
For my local verification, can you please share the exact steps you used to reproduce this problem on your side?
Please also share which branch/commit, esp-idf version and VS code esp-idf extension version are you using?

fhrbata:
Hi @kxt116,

First of all, thank you for the detailed report and fix. 

Your diagnosis matches what I see. The bootloader is configured as a separate
CMake invocation during the build (via `ExternalProject_Add`), and ESP-IDF target
initialization reads `$ENV{IDF_TARGET}` in addition to the `-DIDF_TARGET` passed
on the command line. When the two disagree, target initialization reports the
mismatch.

I am not sure the build system is the right place to fix this, and I
would like to explain my reasoning:

- The mismatch is reported intentionally. When the value passed on the command
  line (stored in the CMake cache) and the value in the environment disagree,
  there is no defined precedence between them, so the build system deliberately
  reports an error instead of silently choosing one. This is also why `idf.py`
  runs `_check_idf_target` early (in `ensure_build_directory`), before invoking
  any CMake target: an inconsistent environment is caught up front rather than
  surfacing later, mid-build, when the subproject is configured.

- The change makes the passed `-DIDF_TARGET` win for the subproject by overwriting
  `$ENV{IDF_TARGET}`. In your multi-root workspace case that is exactly the
  outcome you want, but in the general case the build system cannot tell whether
  the environment value was set intentionally by the user or is simply stale, so
  overriding it silently could also hide a genuine misconfiguration.

- It introduces an inconsistency in behavior. With this change, a target mismatch
  would pass in the bootloader subproject, while the same mismatch in the
  application build (and anything else going through `__target_init`) is still
  reported as an error. The same situation would then behave differently depending
  on where it occurs.

- It is not really bootloader-specific either. `esp_tee` uses the same mechanism
  (`ExternalProject_Add` + `-DIDF_TARGET` + a `subproject/CMakeLists.txt` that
  includes `project.cmake`), so it has the same latent behavior and would not be
  covered by a one-line change in the bootloader.

If we did want the command-line/cache value to take precedence, I think that
decision is better made once, at the build-system level, inside the actual
consistency check, and applied uniformly, rather than as a one-off environment
override in a single subproject.

Given all of the above, in my opinion this is better addressed in the VS Code
extension, which is what actually triggers the situation and introduces the `IDF_TARGET` environment inconsistency.

This is just my personal view, so of course others may weigh in with a different preference.

Thank you!

meetpatelvgec:
I agree with @fhrbata's suggestion to not address this in esp-idf, but rather in VS Code extension. I will mark this MR as closed now, and request you to create an internal ticket for addressing it in VS Code extension.
Thanks again @kxt116

PR #18768, opened 2026-06-27
fix(ppa): bypass SRM macro block order for P4 rotation (IDFGH-17870)

Triwalt:
## Description

This extends the existing ESP32-P4 SRM macro block order workaround so 90/270-degree rotation also enables macro block order bypass before starting the SRM engine.

`esp_lvgl_adapter` currently documents a downstream patch for ESP32-P4 display freezes when `ESP_LV_ADAPTER_TEAR_AVOID_MODE_TRIPLE_PARTIAL` is used with screen rotation and PPA acceleration. ESP-IDF already has a `DIG-734` workaround in this area using `ppa_ll_srm_bypass_mb_order()`, so this change keeps the existing LL path instead of adding a direct register write.

## Testing

- Compared this branch against `espressif:master`: 1 file changed, +2/-1.
- Checked the touched file for trailing whitespace.

Not run:

- ESP-IDF build.
- ESP-IDF pre-commit hooks.
- ESP32-P4 hardware reproduction test.

The local environment used for preparing this PR could not clone from `github.com` with `git` and does not have `idf.py` available, so this is opened as a draft for maintainer review.

---

> [!NOTE]
> **Medium Risk**
> Touches ESP32-P4-only display DMA/PPA workaround logic; wrong dimension mapping could mis-set bypass and affect stability under rotation, but scope is a single guarded code path aligned with existing rotation math in the same file.
> 
> **Overview**
> **ESP32-P4 DIG-734 workaround:** the logic that decides when to call `ppa_ll_srm_bypass_mb_order()` now uses **post-rotation** block size and scale factors, matching how output geometry is already computed elsewhere in SRM.
> 
> For **90° and 270°** rotation, effective width/height and X/Y scale integers/fragments are swapped before computing `w_out`, `h_left`, and the multi–macro-block split condition. **0°/180°** paths keep the previous direct use of `in.block_w/h` and `scale_x/y_*`.
> 
> This should let the existing hardware workaround trigger correctly under rotation (e.g. LVGL partial tear-avoid + PPA), instead of underestimating output size and skipping bypass when it was needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e8513a3d1528bcea1365d0f39012d870c0edd1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_driver_ppa/src/ppa_srm.c
+++ b/components/esp_driver_ppa/src/ppa_srm.c
@@ -156,18 +156,35 @@ bool ppa_srm_transaction_on_picked(uint32_t num_chans, const dma2d_trans_channel
 
 #if CONFIG_IDF_TARGET_ESP32P4
     // Hardware bug workaround (DIG-734)
-    uint32_t w_out = srm_trans_desc->in.block_w * srm_trans_desc->scale_x_int + srm_trans_desc->in.block_w * srm_trans_desc->scale_x_frag / PPA_LL_SRM_SCALING_FRAG_MAX;
+    uint32_t in_block_w, in_block_h;
+    uint32_t scale_x_int, scale_x_frag, scale_y_int, scale_y_frag;
+    if (srm_trans_desc->rotation_angle == PPA_SRM_ROTATION_ANGLE_90 || srm_trans_desc->rotation_angle == PPA_SRM_ROTATION_ANGLE_270) {
+        in_block_w = srm_trans_desc->in.block_h;
+        in_block_h = srm_trans_desc->in.block_w;
+        scale_x_int = srm_trans_desc->scale_y_int;
+        scale_x_frag = srm_trans_desc->scale_y_frag;
+        scale_y_int = srm_trans_desc->scale_x_int;
+        scale_y_frag = srm_trans_desc->scale_x_frag;
+    } else {
+        in_block_w = srm_trans_desc->in.block_w;
+        in_block_h = srm_trans_desc->in.block_h;
+        scale_x_int = srm_trans_desc->scale_x_int;
+        scale_x_frag = srm_trans_desc->scale_x_frag;
+        scale_y_int = srm_trans_desc->scale_y_int;
+        scale_y_frag = srm_trans_desc->scale_y_frag;
+    }
+    uint32_t w_out = in_block_w * scale_x_int + in_block_w * scale_x_frag / PPA_LL_SRM_SCALING_FRAG_MAX;
     uint32_t w_divisor = (ppa_out_color_mode == PPA_SRM_COLOR_MODE_ARGB8888 || ppa_out_color_mode == PPA_SRM_COLOR_MODE_RGB888) ? 32 : 64;
     uint32_t w_left = w_out % w_divisor;
     w_left = (w_left == 0) ? w_divisor : w_left;
     uint32_t h_mb = (ppa_ll_srm_get_mb_size(platform->hal.dev) == PPA_LL_SRM_MB_SIZE_16_16) ? 16 : 32;
-    uint32_t h_in_left = srm_trans_desc->in.block_h % h_mb;
+    uint32_t h_in_left = in_block_h % h_mb;
     h_in_left = (h_in_left == 0) ? h_mb : h_in_left;
-    uint32_t h_left = h_in_left * srm_trans_desc->scale_y_int + h_in_left * srm_trans_desc->scale_y_frag / PPA_LL_SRM_SCALING_FRAG_MAX;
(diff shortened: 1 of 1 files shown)


songruo:
Hi @Triwalt, we are actually aware of this issue, but right now we lose the environment to reproduce the issue, so the fix is on hold... I think the fix you raised is not complete, the block width and height need to be flipped additionally, and maybe also for scale w/h. If you are able to encounter the problem in your application, can you help test the patch below?

[0001-fix-ppa-fix-for-SRM-operation-potential-block-if-do-.patch](https://github.com/user-attachments/files/29455541/0001-fix-ppa-fix-for-SRM-operation-potential-block-if-do-.patch)

Triwalt:
Thanks for the patch and explanation. I updated this PR to use your suggested approach: for 90/270-degree rotation, the workaround now flips the input block width/height and the scale x/y values before computing the DIG-734 macro block order bypass condition.

What I tested locally:

- Applied the equivalent change to an ESP-IDF v5.5.3 test worktree.
- Built the current application far enough to compile `esp_driver_ppa`; `esp-idf/esp_driver_ppa/libesp_driver_ppa.a` was generated and `ppa_srm.c.obj` was built successfully.
- `git diff --check -- components/esp_driver_ppa/src/ppa_srm.c` passed for the touched file.

Relevant local app configuration I found:

- Main display uses `ESP_LV_ADAPTER_ROTATE_90` + `ESP_LV_ADAPTER_TEAR_AVOID_MODE_TRIPLE_PARTIAL`.
- `CONFIG_SOC_PPA_SUPPORTED=y`.
- The BSP display profile currently has `.enable_ppa_accel = false`, and `CONFIG_LV_USE_PPA` is not set.

Because of that last point, I cannot honestly claim that I reproduced or verified the display-freeze path in the adapter PPA-accelerated rotation mode yet. So far this is a compile-level verification of the proposed driver change in my app environment, not a hardware reproduction result.

Triwalt:
Hi @songruo, I tested the SRM path again on an ESP32-P4 board with a patch equivalent to the DIG-734 change you suggested:

- for 90/270 degree rotation, use the rotated block width/height in the DIG-734 workaround calculation
- for 90/270 degree rotation, swap scale_x/scale_y in that workaround calculation

Build validation:

```text
Target: esp32p4
App: components/esp_driver_ppa/test_apps
Command:
python D:\Espressif\frameworks\esp-idf-v5.5.3-codex-ppa-test\tools\idf.py -B build_codex_ppa_srm_current -DIDF_TARGET=esp32p4 build

Result:
[660/690] Building C object esp-idf/esp_driver_ppa/CMakeFiles/__idf_esp_driver_ppa.dir/src/ppa_srm.c.obj
[671/690] Linking C static library esp-idf\esp_driver_ppa\libesp_driver_ppa.a
Generated .../build_codex_ppa_srm_current/ppa_test.bin
Project build complete.
```

Board validation was done through an application-level SRM diagnostic, because the official test app build uses a 2 MB single factory-app flash layout and this board currently uses a 32 MB OTA layout.

Board status after controlled restart into the tested image:

```text
SVDBG:PART running=ota_0@0x510000 configured=ota_0@0x510000 next=ota_1@0xa10000
SVDBG:STATUS ... uptime_ms=79189 reset=software(3) stage=MAIN: boot finished ...
```

Results:

```text
SVDBG:PPA_SRM_SEARCH DONE result=ESP_OK cases_run=34 cases_total=34
SVDBG:PPA_SRM_STRESS DONE result=ESP_OK iterations_run=10000 iterations_total=10000 cases_run=40000 failure_case=0
```

One representative 90-degree case where the old and new workaround decisions differ:

```text
cm=RGB565 rotation=1 block=17x35 scale=1.500/1.875
bypass_old16=0 new16=1 old_actual=0 new_actual=1
reg_before=0 reg_after=1 op=ESP_OK completed=1 result=ESP_OK
```

I did not reproduce the original DIG-734 block directly on this board, so this is verification of the corrected calculation and smoke/stress stability rather than direct reproduction of the old internal failure.

Separately, I found a YUV420 validation case that can block SRM before entering the patched validation path:

```text
cm=YUV420 rotation=90 block=18x34 scale=1.000/1.500 out_block=51x18 out=52x18
before validation fix: op=ESP_OK completed=0 unreg=ESP_ERR_INVALID_STATE result=ESP_ERR_TIMEOUT
after validation fix:  op=ESP_ERR_INVALID_ARG result=ESP_ERR_INVALID_ARG
subsequent RGB565 SRM: op=ESP_OK completed=1 result=ESP_OK
```

I would keep that YUV420 validation issue separate from the DIG-734 workaround PR unless you prefer a combined patch. I also cannot verify the YUV422 side on this current board/config, because YUV422 SRM is disabled for the configured ESP32-P4 rev < 3 target.

PR #18825, opened 2026-07-11
fix(log): fix out-of-bounds reads in binary log mode's output_pointer() (IDFGH-17956)

94xhn:
## Bug

`output_pointer()` (`components/log/src/log_format_binary.c`), used when
`CONFIG_LOG_MODE_BINARY_EN` is enabled, formats a pointer argument either as
a raw address (if it points into the ELF, e.g. a flash-resident format
string) or, otherwise, as an embedded length-prefixed byte string:

```c
static unsigned output_pointer(const char *ptr, pkg_info_t *pkg_info)
{
    unsigned pkg_len = 0;
    uintptr_t addr = (uintptr_t)ptr;
    if (PRESENT_IN_ELF(addr) || addr == 0) {
        pkg_len = output(&addr, sizeof(addr), pkg_info);
    } else {
        int len = (pkg_info->buffer_len) ? pkg_info->buffer_len : strlen(ptr);
        int16_t pkg_str_len = 1 - len;
        pkg_len = output(&pkg_str_len, sizeof(pkg_str_len), pkg_info);
        for (unsigned i = 0; i < MAX(len, 2); i++) {
            pkg_len += output(&ptr[i], sizeof(uint8_t), pkg_info);
        }
    }
    return pkg_len;
}
```

This branch has two independent bugs, both reachable through the fully
public, documented `ESP_LOG_BUFFER_HEX()` / `ESP_LOG_BUFFER_CHAR()` /
`ESP_LOG_BUFFER_HEXDUMP()` macros with a caller-supplied `buff_len` of `0` or
`1` (see `components/log/src/buffer/log_buffers.c`, which always calls
`esp_log()` with args `(buff_len, (const char*)buffer, (void*)buffer)`, i.e.
`buff_len` becomes `pkg_info->buffer_len` before `output_pointer()` is called
on `buffer`):

1. **`MAX(len, 2)` always reads at least 2 bytes**, even when the caller
   declared `buff_len == 1`, e.g. `ESP_LOG_BUFFER_HEX(tag, buf, 1, level)`.
   This unconditionally reads one byte past the end of a 1-byte buffer.

2. **`buffer_len == 0` is indistinguishable from "not set"**. `buffer_len`
   is a plain `int` doubling as its own "is this a buffer-log call at all"
   flag: `0` means both "not applicable, fall back to `strlen()`" (for a
   real NUL-terminated format/tag string) *and* "the caller's buffer is 0
   bytes long" — a legitimate case, e.g. logging an empty receive buffer.
   When `buff_len == 0`, the code silently takes the `strlen()` branch
   instead, scanning the (non-NUL-terminated, arbitrary binary) buffer for
   the first zero byte — an unbounded read with no relation to the buffer's
   actual (zero) size.

## Fix

- Add a dedicated `BUFFER_LEN_NOT_SET` (`-1`) sentinel, initialized once in
  `esp_log_format_binary()`, so an explicit length of `0` is no longer
  confusable with "not applicable". `output_pointer()` now falls back to
  `strlen()` only when no explicit length was ever provided.
- Read exactly `len` bytes — guarded by the `for (int i = 0; i < len; i++)`
  loop bound itself, so `len <= 0` correctly reads zero bytes — instead of
  forcing a minimum of 2.

Diff is intentionally minimal and confined to `output_pointer()` and the one
`pkg_info_t` initializer; `output_arguments()` (which sets `buffer_len` from
the caller's actual `buff_len` argument) needed no changes.

## Verification

Verified with a standalone host-side C reproduction using a Windows
`VirtualAlloc` guard page (a committed page immediately followed by an
uncommitted/`PAGE_NOACCESS` page, with the 0/1-byte "buffer" placed at the
very end of the committed page), mirroring the exact `(pkg_info->buffer_len)
? ... : strlen(ptr)` / `MAX(len, 2)` logic vs. the patched logic:

```
########## BUGGY, len=1 (expect CRASH) ##########
  [buggy] declared_buffer_len=1 -> computed len=1, will read 2 bytes (MAX(len,2))
    read ptr[0] = 0x41 (still inside the committed page)
Segmentation fault   <-- crashes reading ptr[1], past the declared 1-byte buffer

########## BUGGY, len=0 (expect CRASH) ##########
Segmentation fault   <-- crashes inside strlen(ptr), ptr[0] is already past the buffer

########## FIXED, len=1 (expect clean exit) ##########
  [fixed] declared_buffer_len=1 -> computed len=1, will read exactly 1 bytes
    read ptr[0] = 0x41 (still inside the committed page)
  Completed without touching the guard page. bytes_read_ok=1
exit code: 0

########## FIXED, len=0 (expect clean exit) ##########
  [fixed] declared_buffer_len=0 -> computed len=0, will read exactly 0 bytes
  Completed without touching the guard page. bytes_read_ok=0
exit code: 0
```

i.e. the unpatched logic crashes with an access violation for both
`buff_len == 1` and `buff_len == 0`, in exactly the byte position
corresponding to the read past the declared buffer; the patched logic
completes cleanly, touching only the bytes within the declared length, for
both cases.

I checked `components/log/host_test/log_test` and `components/log/test_apps`
— neither has any CI configuration enabling `CONFIG_LOG_MODE_BINARY_EN`
together with the buffer-logging macros, so this specific combination had no
prior regression coverage.

## Disclosure

This fix was prepared with AI assistance (Claude) and reviewed by me before
submission.

---

> [!NOTE]
> **Medium Risk**
> Touches memory-boundary logic on the binary log encode path; change is small and localized but fixes real OOB/`strlen` misuse on arbitrary buffer data.
> 
> **Overview**
> Fixes **out-of-bounds reads** in binary log formatting when `output_pointer()` embeds non-ELF pointers as length-prefixed strings (notably `ESP_LOG_BUFFER_*` with small or zero `buff_len`).
> 
> Introduces **`BUFFER_LEN_NOT_SET` (-1)** so `pkg_info->buffer_len == 0` means “caller buffer is empty” instead of being treated like “use `strlen()` on a C string.” **`strlen()`** is only used when no explicit buffer length was set (format/tag strings). The emit loop now copies **exactly `len` bytes** instead of **`MAX(len, 2)`**, which previously always read at least two bytes. **`calc_pkg_len()`** and **`esp_log_format_binary()`** initialize/reset `buffer_len` to the sentinel so length calculation matches the actual encode path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f3465aa84fabb5e6120389f6b2cc4d015d93457. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/log/src/log_format_binary.c
+++ b/components/log/src/log_format_binary.c
@@ -252,6 +252,7 @@ static unsigned calc_pkg_len(esp_log_msg_t *message, pkg_info_t *pkg_info)
     pkg_len += output_arguments(message, args, pkg_info);
     va_end(args);
     pkg_len += sizeof(uint8_t); // crc8
+    pkg_info->buffer_len = BUFFER_LEN_NOT_SET;
     pkg_info->len_calculation_stage = false;
     return pkg_len;
 }
--- a/components/log/src/log_format_binary.c
+++ b/components/log/src/log_format_binary.c
@@ -136,13 +136,16 @@ static unsigned output_pointer(const char *ptr, pkg_info_t *pkg_info)
         // *actual* length of the caller-supplied buffer (see output_arguments() below), which
         // may legitimately be 0 (e.g. logging an empty buffer). Only fall back to strlen() when
         // no explicit length has been provided, i.e. ptr is a real NUL-terminated C string.
-        int len = (pkg_info->buffer_len != BUFFER_LEN_NOT_SET) ? pkg_info->buffer_len : (int)strlen(ptr);
-        int16_t pkg_str_len = 1 - len;
+        int data_len = (pkg_info->buffer_len != BUFFER_LEN_NOT_SET) ? pkg_info->buffer_len : (int)strlen(ptr);
+        int encoded_len = MAX(data_len, 2);
+        int16_t pkg_str_len = 1 - encoded_len;
         pkg_len = output(&pkg_str_len, sizeof(pkg_str_len), pkg_info);
-        // Emit exactly `len` bytes: previously this was MAX(len, 2), which always read at
-        // least 2 bytes even when len was 0 or 1, reading past the end of the caller's buffer.
-        for (int i = 0; i < len; i++) {
+        for (int i = 0; i < data_len; i++) {
             pkg_len += output(&ptr[i], sizeof(uint8_t), pkg_info);
+        }
+        const uint8_t padding = 0;
+        for (int i = data_len; i < encoded_len; i++) {
+            pkg_len += output(&padding, sizeof(padding), pkg_info);
         }
     }
     return pkg_len;
--- a/components/log/src/log_format_binary.c
+++ b/components/log/src/log_format_binary.c
@@ -59,6 +59,12 @@ typedef struct {
 
 _Static_assert(sizeof(control_t) == 3, "control_t must be exactly 24 bits (3 bytes)");
 
+// Sentinel for pkg_info_t.buffer_len meaning "no explicit buffer length was supplied for the
+// pointer currently being formatted" (i.e. treat it as a real NUL-terminated C string). This is
+// distinct from a real, explicit length of 0 (e.g. ESP_LOG_BUFFER_HEX() called with buff_len==0),
+// which must be honored as-is instead of falling back to strlen() on non-string buffer data.
+#define BUFFER_LEN_NOT_SET (-1)
+
 typedef struct {
     uint8_t crc;
     bool buffer_hex_log;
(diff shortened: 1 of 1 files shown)


components/log/src/log_format_binary.c
```diff
@@ -269,7 +282,7 @@ void esp_log_format_binary(esp_log_msg_t *message)
         .buffer_hex_log = message->format == __ESP_BUFFER_HEX_FORMAT__,
         .buffer_char_log = message->format == __ESP_BUFFER_CHAR_FORMAT__,
         .buffer_hexdump_log = message->format == __ESP_BUFFER_HEXDUMP_FORMAT__,
-        .buffer_len = 0,
+        .buffer_len = BUFFER_LEN_NOT_SET,
```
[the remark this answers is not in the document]
94xhn: Confirmed. calc_pkg_len() left the explicit buffer length in pkg_info, so the emission pass encoded the format and tag with stale state. Commit 0f3465aa84 resets that transient state after sizing. An actual-source host harness reproduced declared/emitted lengths of 68/18 for length 0 and 69/21 for length 1 before the fix; both are now 68/68 and 69/69.

components/log/src/log_format_binary.c
```diff
@@ -125,10 +131,17 @@ static unsigned output_pointer(const char *ptr, pkg_info_t *pkg_info)
...
+        // no explicit length has been provided, i.e. ptr is a real NUL-terminated C string.
+        int len = (pkg_info->buffer_len != BUFFER_LEN_NOT_SET) ? pkg_info->buffer_len : (int)strlen(ptr);
         int16_t pkg_str_len = 1 - len;
         pkg_len = output(&pkg_str_len, sizeof(pkg_str_len), pkg_info);
-        for (unsigned i = 0; i < MAX(len, 2); i++) {
+        // Emit exactly `len` bytes: previously this was MAX(len, 2), which always read at
+        // least 2 bytes even when len was 0 or 1, reading past the end of the caller's buffer.
+        for (int i = 0; i < len; i++) {
             pkg_len += output(&ptr[i], sizeof(uint8_t), pkg_info);
         }
```
KonstantinKondrashov: MAX(len, 2) is essential; we can not simply drop it. We need to send padding. 

```c
int data_len = pkg_info->buffer_len != BUFFER_LEN_NOT_SET ? pkg_info->buffer_len : strlen(ptr);
int encoded_len = MAX(data_len, 2);
int16_t pkg_str_len = 1 - encoded_len;

pkg_len = output(&pkg_str_len, sizeof(pkg_str_len), pkg_info);

for (int i = 0; i < data_len; i++) {
    pkg_len += output(&ptr[i], sizeof(uint8_t), pkg_info);
}

const uint8_t padding = 0;
for (int i = data_len; i < encoded_len; i++) {
    pkg_len += output(&padding, sizeof(padding), pkg_info);
}
```

ref: https://github.com/espressif/esp-idf-monitor/blob/master/esp_idf_monitor/base/binlog.py#L139-L142

KonstantinKondrashov (commented):
Thank you for the fix! One comment regarding output padding.

components/log/src/log_format_binary.c
94xhn: Updated in 437676b8af. The encoded length remains at least two bytes for the monitor framing, while only data_len bytes are read from the caller buffer; any remaining encoded bytes are now emitted as explicit zero padding. The 0-, 1-, and 2-byte cases pass the MinGW host regression and WSL ASan/UBSan run.

KonstantinKondrashov (approved):

KonstantinKondrashov:
sha=437676b8af355e02e0698f26d418fb3cf74a8dfb

Alvin1Zhang:
Thanks for contribution again, changes have been merged with https://github.com/espressif/esp-idf/commit/62e507ab8e43446764928553a8a88311dcc3792e

PR #18890, opened 2026-07-25
fix(esp_libc): provide picolibc timespec_get (IDFGH-18035)

### LINKED ISSUE #18889: ESP32-S3 picolibc declares timespec_get but does not define it (IDFGH-18034)
### Answers checklist

- [x] I have read the ESP-IDF Programming Guide and did not find this behavior documented.
- [x] I checked the current `release/v6.0` branch (`bed9759dbb42e74ebe16301acff21475d925a9a7`); it still selects the affected `esp-15.2.0_20251204` Xtensa toolchain.
- [x] I searched the ESP-IDF, Espressif crosstool-NG, and picolibc issue trackers and did not find an existing `timespec_get` report.

### IDF version

`v6.0` (`662a3be354759d9487bf4b1a629fadb766cb1800`).

The current `release/v6.0` branch at `bed9759dbb42e74ebe16301acff21475d925a9a7` and the `v6.0.2` tag at `7101770dc6db2667b3c477cc31365dd1acd6db4e` still select the same `esp-15.2.0_20251204` Xtensa toolchain release.

### Operating System used

Linux (Ubuntu 24.04 on GitHub Actions). The same minimal reproducer also fails on macOS with the corresponding Apple-hosted archive from the same toolchain release.

Command line with `idf.py`.

Code that includes `<time.h>` and calls the declared C11 function `timespec_get()` should link successfully for ESP32-S3, or the header should not declare the function when its implementation is unavailable.

The picolibc `<time.h>` shipped by `esp-15.2.0_20251204` declares:

```c
int timespec_get(struct timespec *_ts, int _base);
```

However, the ESP32-S3 picolibc archives do not define `timespec_get`. Linking any caller fails with an undefined reference.

The missing implementation is not specific to a larger application: a three-line `app_main()` is sufficient to reproduce it. Inspecting all `.a` files in the installed Xtensa toolchain with `xtensa-esp32s3-elf-nm` also finds no defined `timespec_get` symbol.

### Steps to reproduce

Create a normal ESP-IDF project with the following files.

`CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.16)

include($ENV{IDF_PATH}/tools/cmake/project.cmake)
project(timespec_get_repro)
```

`main/CMakeLists.txt`:

```cmake
idf_component_register(SRCS "main.c")
```

`main/main.c`:

```c
#include <time.h>

void app_main(void) {
    struct timespec now;
    (void)timespec_get(&now, TIME_UTC);
}
```

Then run:

```console
./install.sh esp32s3
. ./export.sh
idf.py set-target esp32s3
idf.py build
```

### Build or installation logs

```text
xtensa-esp32s3-elf-ld: esp-idf/main/libmain.a(main.c.obj):
(.literal.app_main+0x0): undefined reference to `timespec_get'
xtensa-esp32s3-elf-ld: esp-idf/main/libmain.a(main.c.obj):
in function `app_main':
main/main.c:5:(.text.app_main+0x7): undefined reference to `timespec_get'
collect2: error: ld returned 1 exit status
```

The linked archive is:

```text
~/.espressif/tools/xtensa-esp-elf/esp-15.2.0_20251204/
  xtensa-esp-elf/picolibc/xtensa-esp-elf/lib/esp32s3/no-rtti/libc.a
```

### More Information

`clock_gettime(CLOCK_MONOTONIC, ...)` and `clock_gettime(CLOCK_REALTIME, ...)` are available through ESP-IDF, but they are not substitutes for a C library function that its public header declares.

This was first encountered when a portable dependency used `timespec_get()` as a fallback after `clock_gettime()`. A local compatibility symbol can work around the link failure, but fixing the header/library mismatch in the ESP-IDF toolchain would avoid each consumer having to provide its own libc shim.

idy:
## Description

The Picolibc `<time.h>` shipped with the ESP-IDF v6.0 toolchain declares the
C11 `timespec_get()` API, but the Picolibc archives do not provide the symbol.
Any ESP-IDF application that calls the declared API therefore fails to link.

Add the missing implementation to the Picolibc-specific `esp_libc` sources.
`TIME_UTC` is mapped to the existing ESP-IDF `CLOCK_REALTIME` implementation,
while unsupported time bases and clock failures return zero as required by
C11.

The implementation lives in a separate archive object so a future toolchain
that provides `timespec_get()` can satisfy the reference from libc without
pulling in this compatibility object.

Fixes #18889

## Testing

- Confirmed the issue's minimal ESP32-S3 reproducer fails to link with
  `esp-15.2.0_20251204` before the fix:
  `undefined reference to 'timespec_get'`.
- Built the same ESP32-S3 project successfully with the compatibility source
  included; the resulting ELF defines both `clock_gettime` and `timespec_get`.
- Added an `esp_libc` Unity regression test for `TIME_UTC`, the returned
  nanosecond range, and an unsupported time base.
- Ran the applicable pre-commit hooks for all three changed files.

---

> [!NOTE]
> **Low Risk**
> Small, isolated libc compatibility shim with no auth or data-path changes; behavior is delegated to existing clock_gettime.
> 
> **Overview**
> Picolibc builds now ship a **C11 `timespec_get()`** in `esp_libc` so apps that call the API declared in `<time.h>` link successfully instead of failing with an undefined reference.
> 
> For **`TIME_UTC`**, the new `timespec_get` fills the `timespec` via existing **`clock_gettime(CLOCK_REALTIME)`** and returns `TIME_UTC`; unsupported bases or clock failure return **0** per C11. The source is wired into the Picolibc-only `esp_libc` object list (not Newlib).
> 
> A Unity test in **`test_time.c`** checks successful `TIME_UTC` behavior, sane nanosecond range, and rejection of an invalid base.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7de8574ce534ced7fbbedf0e5b32eb0087d5ecba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_libc/CMakeLists.txt
+++ b/components/esp_libc/CMakeLists.txt
@@ -76,6 +76,7 @@ else()
     list(APPEND srcs
          "src/picolibc/picolibc_init.c"
          "src/picolibc/rand.c"
+         "src/picolibc/timespec_get.c"
          "src/picolibc/open_memstream.c"
          "src/picolibc/errno.c"
          "src/picolibc/bufio_setvbuf.c") # TODO IDF-15494: remove this and the following lines
--- a/components/esp_libc/src/picolibc/timespec_get.c
+++ b/components/esp_libc/src/picolibc/timespec_get.c
@@ -1,0 +1,15 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <time.h>
+
+int timespec_get(struct timespec *ts, int base)
+{
+    if (base != TIME_UTC || clock_gettime(CLOCK_REALTIME, ts) < 0) {
+        return 0;
+    }
+    return base;
+}
--- a/components/esp_libc/test_apps/newlib/main/test_time.c
+++ b/components/esp_libc/test_apps/newlib/main/test_time.c
@@ -465,6 +465,17 @@ TEST_CASE("test posix_timers clock_... functions", "[newlib]")
     test_posix_timers_clock();
 }
 
+TEST_CASE("timespec_get returns UTC time", "[newlib]")
+{
+    struct timespec ts = {0};
+
+    TEST_ASSERT_EQUAL_INT(TIME_UTC, timespec_get(&ts, TIME_UTC));
+    TEST_ASSERT_GREATER_OR_EQUAL_INT64(0, ts.tv_sec);
+    TEST_ASSERT_GREATER_OR_EQUAL_INT32(0, ts.tv_nsec);
+    TEST_ASSERT_LESS_THAN_INT32(1000000000L, ts.tv_nsec);
+    TEST_ASSERT_EQUAL_INT(0, timespec_get(&ts, -1));
+}
+
 #ifndef _USE_LONG_TIME_T
 
 static struct timeval get_time(const char *desc, char *buffer)
--- a/components/esp_libc/test_apps/newlib/main/test_time.c
+++ b/components/esp_libc/test_apps/newlib/main/test_time.c
@@ -465,7 +465,8 @@ TEST_CASE("test posix_timers clock_... functions", "[newlib]")
     test_posix_timers_clock();
 }
 
-TEST_CASE("timespec_get returns UTC time", "[newlib]")
+#if CONFIG_LIBC_PICOLIBC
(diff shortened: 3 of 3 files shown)


SoucheSouche:
@idy, thanks for the fix. Everything looks good to me. I will submit it internally.

I will close https://github.com/espressif/esp-idf/issues/18889, so we don't have to track 2 tickets for the same issue.

PR #18944, opened 2026-08-06
freertos: yield immediately after unicore self deletion (IDFGH-18104)

### LINKED ISSUE #18460: vTaskDelete results in IllegalInstruction Exception with clang -O2 (IDFGH-17524)
### Answers checklist.

### IDF version.

v6.0

### Operating System used.

Linux

Command line with idf.py

None

Compilation completes successfully

`vTaskDelete(nullptr)` results in an IllegalInstruction Exception on newly-created tasks.

Curiously, though maybe this is an expected quirk, `vTaskDelete(nullptr)` of the `app_main` task doesn't cause an issue for me. Only a newly-created task does.

```
Guru Meditation Error: Core  0 panic'ed (IllegalInstruction). Exception was unhandled.
Memory dump at 0x42009d90: db82810a 1d0008e0 000000f0
--- 0x42009d90: (anonymous namespace)::task(void*) at [test].cpp:29
Core  0 register dump:
PC      : 0x42009d97  PS      : 0x00060030  A0      : 0x00000000  A1      : 0x3fca3430  
--- 0x42009d97: (anonymous namespace)::task(void*) at [test].cpp:31
A2      : 0x00000000  A3      : 0x00000000  A4      : 0x00000000  A5      : 0x00000000  
A6      : 0x00000000  A7      : 0x00000000  A8      : 0x82009d97  A9      : 0x3fca3410  
A10     : 0x3fca359c  A11     : 0x00000000  A12     : 0x40386f48  A13     : 0x3fca1e2c  
--- 0x40386f48: uxListRemove at [HOME]/esp/esp-idf/components/freertos/FreeRTOS-Kernel/list.c:207
A14     : 0x3fca35a0  A15     : 0xffffffff  SAR     : 0x00000008  EXCCAUSE: 0x00000000  
EXCVADDR: 0x00000000  LBEG    : 0x00000000  LEND    : 0x00000000  LCOUNT  : 0x00000000  


Backtrace: 0x42009d94:0x3fca3430
--- 0x42009d94: (anonymous namespace)::task(void*) at [test].cpp:30
```

### Steps to reproduce.

[sdkconfig.txt](https://github.com/user-attachments/files/26694567/sdkconfig.txt)

The below example is enough to cause it for me. (If the `xTaskCreateStatic` is removed, no issue is observed. The exception is seen inside `task()`, as shown by the dump.
```cpp
namespace {

StaticTask_t taskInfo;
std::array<StackType_t, 4096> stack;

auto task(void *) {
    vTaskDelete(nullptr);
};

} // namespace

extern "C" void app_main() {
    xTaskCreateStatic(
        task,
        "Test",
        stack.size(),
        nullptr,
        24,
        stack.data(),
        &taskInfo
    );
    vTaskDelete(nullptr);
}
```

### Build or installation Logs.

```plain

```

_No response_

### More Information.

This doesn't occur with `-Og` on clang, or with GCC at either `-Og` or `-O2`

Adding `[[clang::optnone]]` to the definition of `vTaskDelete()` in `tasks.c` avoids the optimization which causes the issue.

RobertEmbeddedProject:
### Description
Fix a failure where a task deleting itself can continue executing long enough to return from its task entry function in a unicore Xtensa build.

In the current self-delete path, vTaskDelete() calls portYIELD_WITHIN_API(). On the Xtensa port, this requests a yield through the cross-core interrupt mechanism and then it returns. In a unicore Clang -O2 build, the deleted task can return from vTaskDelete() and subsequently return from its entry function before the pending yield interrupt is serviced, resulting in an IllegalInstruction panic.

For unicore builds, portYIELD() should be used instead. This performs the intended context switch immediately and does not return through the deleted task’s execution path.

The existing portYIELD_WITHIN_API() behavior is preserved for multicore builds, where the code executes inside the SMP-only kernel critical section and the yield must remain deferred.

This fixes #18460.

### Testing
Tested on ESP32-S3 with ESP-IDF v6.0 and an AITRIP ESP32-S3-DevKitC-1-N8R2.

### Test results

**Clang 20.1.1, -O2, unicore, original code**
Result: IllegalInstruction

**Clang 20.1.1, -O2, unicore, patched**
Result: Pass

**Clang 20.1.1, -Og, unicore, patched**
Result: Pass

**GCC, -O2, unicore, patched**
Result: Pass

**Clang 20.1.1, -O2, dual-core, patched**
Result: Build pass

**Official FreeRTOS "FreeRTOS Delete Tasks" test, GCC -O2, unicore, patched**
Result: Pass
        

The official FreeRTOS deletion test completed with:

FreeRTOS Delete Tasks:PASS
1 Tests 0 Failures 0 Ignored

---

> [!NOTE]
> **Medium Risk**
> Changes scheduler yield behavior on the task self-delete path in the kernel, but only for unicore builds and multicore behavior is unchanged.
> 
> **Overview**
> Fixes a unicore bug where a task that deletes itself could keep running after `vTaskDelete()` and return from its entry function before a deferred yield runs (e.g. Xtensa Clang `-O2`), causing an **IllegalInstruction** panic.
> 
> In the self-delete reschedule path in `tasks.c`, **`portYIELD()`** is used when `configNUMBER_OF_CORES == 1` so the context switch happens immediately instead of via **`portYIELD_WITHIN_API()`**. Multicore builds still use **`portYIELD_WITHIN_API()`** so yield stays deferred while inside the SMP-only kernel critical section.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c7298f57dbd10ba7ce816f7977c2e148f0daddc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/freertos/FreeRTOS-Kernel/tasks.c
+++ b/components/freertos/FreeRTOS-Kernel/tasks.c
@@ -1472,7 +1472,12 @@ static void prvAddNewTaskToReadyList( TCB_t * pxNewTCB )
                 if( xSelfDelete == pdTRUE )
                 {
                     configASSERT( taskIS_SCHEDULER_SUSPENDED() == pdFALSE );
-                    portYIELD_WITHIN_API();
+
+                    #if ( configNUMBER_OF_CORES == 1 )
+                        portYIELD();
+                    #else
+                        portYIELD_WITHIN_API();
+                    #endif
                 }
                 else
                 {


esp-rctb:
sha=1c7298f

esp-rctb:
Hi @RobertEmbeddedProject , thanks for the submitting this PR. I was able to reproduce the issue, should be able to merge once it passes internal tests and reviews.

esp-rctb:
sha=1c7298f57dbd10ba7ce816f7977c2e148f0daddc

PR #18983, opened 2026-08-16
fix(http_client): default Content-Type for post field (IDFGH-18153)

### LINKED ISSUE #18980: esp_http_client_set_post_field returns ESP_ERR_NOT_FOUND when no Content-Type is set (IDFGH-18150)
### Answers checklist.

### IDF version.

IDF 6.0.1

### Espressif SoC revision.

ESPS3

### Operating System used.

Linux

Other (please specify in More Information)

None

### Development Kit.

Any

### Power Supply used.

USB

When no Content-Type is not set in the header  or set to NULL ( esp_http_client_set_header(client, "Content-Type", nullptr) ) the esp_http_client_set_post_field would accept post data and set "Content-Type" to  "application/x-www-form-urlencoded". When content type is set, it should keep the already set content type and use the post data. 

On esp_http_client_set_header(client, "Content-Type", nullptr) esp_http_client_set_post_field returns ESP_ERR_NOT_FOUND.
When  "Content-Type" is not set in header is esp_http_client_set_post_field returns ESP_ERR_NOT_FOUND. 

Basically the only working solution is to manually set the Content-Type before calling the esp_http_client_set_post_field. This is not mentioned in the IDF 6.x documentation. Also this is changed after IDF5.x ( ret=esp_http_client_set_header(client, "Content-Type", "application/x-www-form-urlencoded");). Noticed it after upgrading to IDF6.x

In the code below has dead code, see line below. The  esp_http_client_get_header(client, "Content-Type", &value))  above it returns an error when value is NULL or does not exist. (  esp_http_client_set_header(client, "Content-Type", nullptr) )

```
if (value == NULL) {
            err = esp_http_client_set_header(client, "Content-Type", "application/x-www-form-urlencoded");
        }
```

Code in the framework:
```
esp_err_t http_header_get(http_header_handle_t header, const char *key, char **value)
{
    http_header_item_handle_t item;

    item = http_header_get_item(header, key);
    if (item) {
        *value = item->value;
    } else {
        *value = NULL;
        return ESP_ERR_NOT_FOUND;
    }

    return ESP_OK;
}



esp_err_t esp_http_client_get_header(esp_http_client_handle_t client, const char *key, char **value)
{
    if (client == NULL || client->request == NULL || client->request->headers == NULL || key == NULL) {
        return ESP_ERR_INVALID_ARG;
    }

    return http_header_get(client->request->headers, key, value);
}

esp_err_t esp_http_client_set_post_field(esp_http_client_handle_t client, const char *data, int len)
{
    esp_err_t err = ESP_OK;
    client->post_data = (char *)data;
    client->post_len = len;
    ESP_LOGD(TAG, "set post file length = %d", len);
    if (client->post_data) {
        char *value = NULL;
        if ((err = esp_http_client_get_header(client, "Content-Type", &value)) != ESP_OK) {
            return err;
        }
        if (value == NULL) {
            err = esp_http_client_set_header(client, "Content-Type", "application/x-www-form-urlencoded");
        }
    } else {
        client->post_len = 0;
        err = esp_http_client_set_header(client, "Content-Type", NULL);
    }
    return err;
}
``

> [esp_err_t](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-reference/system/esp_err.html#_CPPv49esp_err_t) esp_http_client_set_post_field([esp_http_client_handle_t](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-reference/protocols/esp_http_client.html#_CPPv424esp_http_client_handle_t) client, const char *data, int len)[](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c2/api-reference/protocols/esp_http_client.html#_CPPv430esp_http_client_set_post_field24esp_http_client_handle_tPKci)
> 
>     Set post data, this function must be called before esp_http_client_perform. Note: Post data persists between requests and must be cleared manually. To avoid post data carryover, set the post data to NULL before starting any subsequent requests.
> 
>     Parameters:
> 
>             client -- [in] The esp_http_client handle
> 
>             data -- [in] post data pointer
> 
>             len -- [in] post length
> 
>     Returns:
> 
>             ESP_OK
> 
>             ESP_FAIL`

### Steps to reproduce.

esp_http_client_init
esp_http_client_set_post_field -> returns ESP_ERR_NOT_FOUND


### Debug Logs.

```plain

```

_No response_

### More Information.

_No response_

Zhanfg:
## Description

Fix `esp_http_client_set_post_field()` so an absent `Content-Type` header is treated as the defaultable case instead of returning `ESP_ERR_NOT_FOUND` before the documented default can be applied.

The function now continues only for `ESP_ERR_NOT_FOUND`; other errors from `esp_http_client_get_header()` are still propagated. If no Content-Type is present, the existing `application/x-www-form-urlencoded` default path runs. An explicitly configured Content-Type remains unchanged.

## Related

Fixes #18980.

## Testing

- Added a regression test for an absent Content-Type and the default `application/x-www-form-urlencoded` value.
- Added a regression test verifying an explicit `application/json` Content-Type is preserved.
- Built `components/esp_http_client/test_apps` successfully on the fork CI runner.

## Checklist

- [x] This PR does not introduce breaking changes.
- [ ] All upstream CI checks pass (currently awaiting maintainer approval for fork workflows).
- [x] Documentation is not required for this behavior restoration.
- [x] Code change is intentionally minimal.
- [x] Git history is clean and contains one commit.

### DIFF
--- a/components/esp_http_client/esp_http_client.c
+++ b/components/esp_http_client/esp_http_client.c
@@ -2065,7 +2065,7 @@ esp_err_t esp_http_client_set_post_field(esp_http_client_handle_t client, const 
     ESP_LOGD(TAG, "set post file length = %d", len);
     if (client->post_data) {
         char *value = NULL;
-        if ((err = esp_http_client_get_header(client, "Content-Type", &value)) != ESP_OK) {
+        if ((err = esp_http_client_get_header(client, "Content-Type", &value)) != ESP_OK && err != ESP_ERR_NOT_FOUND) {
             return err;
         }
         if (value == NULL) {
--- a/components/esp_http_client/test_apps/main/test_http_client.c
+++ b/components/esp_http_client/test_apps/main/test_http_client.c
@@ -188,6 +188,46 @@ TEST_CASE("esp_http_client_set_header() should not return error if header value 
     esp_http_client_cleanup(client);
 }
 
+TEST_CASE("set_post_field adds default Content-Type when missing", "[esp_http_client]")
+{
+    const esp_http_client_config_t config = {
+        .url = "http://localhost",
+    };
+    esp_http_client_handle_t client = esp_http_client_init(&config);
+    TEST_ASSERT_NOT_NULL(client);
+
+    const char post_data[] = "foo=bar";
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_set_post_field(client, post_data, strlen(post_data)));
+
+    char *content_type = NULL;
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_get_header(client, "Content-Type", &content_type));
+    TEST_ASSERT_NOT_NULL(content_type);
+    TEST_ASSERT_EQUAL_STRING("application/x-www-form-urlencoded", content_type);
+
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_cleanup(client));
+}
+
+TEST_CASE("set_post_field preserves explicit Content-Type", "[esp_http_client]")
+{
+    const esp_http_client_config_t config = {
+        .url = "http://localhost",
+    };
+    esp_http_client_handle_t client = esp_http_client_init(&config);
+    TEST_ASSERT_NOT_NULL(client);
+
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_set_header(client, "Content-Type", "application/json"));
+
+    const char post_data[] = "{}";
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_set_post_field(client, post_data, strlen(post_data)));
+
+    char *content_type = NULL;
+    TEST_ASSERT_EQUAL(ESP_OK, esp_http_client_get_header(client, "Content-Type", &content_type));
+    TEST_ASSERT_NOT_NULL(content_type);
+    TEST_ASSERT_EQUAL_STRING("application/json", content_type);
+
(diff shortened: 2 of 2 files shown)


components/esp_http_client/test_apps/main/test_post_field.c
```diff
@@ -0,0 +1,44 @@
+#include <string.h>
+
+#include "esp_http_client.h"
+#include "unity.h"
+
+TEST_CASE("set_post_field adds default Content-Type when missing", "[esp_http_client]")
```
Ashish285: I think the tests can be moved to the existing test file

Zhanfg: Agreed. I’ll move these regression cases into the existing HTTP client test file instead of keeping a separate test_post_field.c.

Zhanfg: Done. I moved both regression cases into `components/esp_http_client/test_apps/main/test_http_client.c` and removed the separate `test_post_field.c`. The branch is squashed back to a single commit.

Zhanfg: The thread is now outdated because the standalone test file was removed; the two cases are in the existing test file as requested.

Ashish285:
The changes LGTM! Thanks for the PR

Ashish285:
sha=22bb11ef2288679a0a016bafddfc34927ae9972f

PR #18987, opened 2026-08-16
esp_lcd: Enable DMA for PSRAM in SPI transactions (IDFGH-18159)

djpnewton:
We found in our project that when the image buffer was in PSRAM that the lcd driver was allocated 8k blocks of memory in DRAM fragmenting our DRAM enough to cause other issues.

This is a workaround to make the lcd driver use DMA to avoid the DRAM allocations (and therefore the fragmentation).

We would like get this option merged upstream if possible.

I could make it a KConfig option so that users can opt-in to the functionality.

I welcome feedback on this issue.

Cheers!


components/esp_lcd/spi/esp_lcd_panel_io_spi.c
suda-morris, line 378: @djpnewton  have you tried this option?

suda-morris: flags.psram_dma_direct

djpnewton: Thanks I was on esp idf v5.5.4 but I see that option is in newer versions

suda-morris: FYI, the recent v5.5.5 contains this option already. @djpnewton

PR #18960, opened 2026-08-11
Fix default value type of FATFS_PRINT_LLI and FATFS_PRINT_FLOAT (IDFGH-18122)

mutatrum:
## Description

Fixes compiler warning:
```
FATFS_PRINT_LLI: 'default 0' is not a valid bool value (only 'y' and 'n' are 
allowed). Value is treated as 'n'.
```
and
```
FATFS_PRINT_FLOAT: 'default 0' is not a valid bool value (only 'y' and 'n' are 
allowed). Value is treated as 'n'.
```

Functionality unchanged, just the format of the defaults are aligned with the type of the value.

---

> [!NOTE]
> **Low Risk**
> Kconfig-only default syntax fix with no functional or build logic changes beyond silencing warnings.
> 
> **Overview**
> Fixes Kconfig warnings for **`FATFS_PRINT_LLI`** and **`FATFS_PRINT_FLOAT`** in `components/fatfs/Kconfig` by changing their defaults from **`default 0`** to **`default n`**, which is the correct syntax for `bool` options.
> 
> Runtime behavior is unchanged: Kconfig already treated `0` as disabled (`n`), and `ffconf.h` still maps these symbols to `FF_PRINT_LLI` / `FF_PRINT_FLOAT` the same way.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4188ab1026a5e59301ba9260edfd854746a891f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #18990, opened 2026-08-17
Bool Kconfig options must use 'y'/'n' for defaults, not '0'/'1'. (IDFGH-18162)

richardsonnick:
## Description
The kconfiglib parser warns on 'default 0' and falls back to 'n'. Change the five remaining instances to 'default 0' to silence the warnings and follow the [Kconfig parsers expectations](https://github.com/espressif/esp-idf-kconfig/blob/v3.12.0/esp_kconfiglib/core.py#L3650). The parser explicitly states that '0' is invalid in the [INVALID_BOOL_LITERALS set](https://github.com/espressif/esp-idf-kconfig/blob/master/esp_kconfiglib/constants.py#L49).

## Related

There are some existing PRs open to fix this (i.e. https://github.com/espressif/esp-idf/pull/18960) but these fixes are narrow in scope; missing other instances of invalid bool values.

## Testing
Verified by rebuilding a project with the modified Kconfig files and confirming the four NOTE: ... 'default 0' is not a valid bool value warnings no longer appear. The change is behaviorally a no-op since the parser already treated 0 as n so this only silences the warnings and aligns with the Kconfig language spec.

---

> [!NOTE]
> **Low Risk**
> Documentation-style Kconfig default fixes in examples only; no functional change.
> 
> **Overview**
> Replaces invalid bool defaults in example Kconfig files with **`default n`**, matching kconfiglib’s requirement that bool options use **`y`**/**`n`** rather than **`0`**, **`1`**, or **`false`**.
> 
> **`EXAMPLE_ENABLE_TWAI_FD`** in the TWAI utils example and **`EXAMPLE_USE_SCAN_CHANNEL_BITMAP`** in the Wi‑Fi scan example are updated. Runtime behavior is unchanged because the parser already treated those literals as off.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12f90f39ddd2e2f0f2381d46b79ff40aae87a6be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/peripherals/twai/twai_utils/main/Kconfig.projbuild
+++ b/examples/peripherals/twai/twai_utils/main/Kconfig.projbuild
@@ -6,7 +6,7 @@ menu "TWAI Configuration"
     config EXAMPLE_ENABLE_TWAI_FD
         bool "Enable TWAI-FD Support"
         depends on SOC_TWAI_FD_SUPPORTED
-        default false
+        default n
         help
             Enable TWAI-FD (Flexible Data-rate) support.
             Allows up to 64 bytes of data per frame and dual bit rates.
--- a/examples/wifi/scan/main/Kconfig.projbuild
+++ b/examples/wifi/scan/main/Kconfig.projbuild
@@ -9,7 +9,7 @@ menu "Example Configuration"
 
     config EXAMPLE_USE_SCAN_CHANNEL_BITMAP
         bool "Scan only non overlapping channels using Channel bitmap"
-        default 0
+        default n
         help
             Enable this to scan only the non overlapping channels i.e 1,6,11 by mentioning a channel bitmap
             in scan config. If you wish to scan a different set of specific channels, please edit the channel_list


fhrbata:
Hello @richardsonnick,

Thank you very much for the patch, and for taking the wider sweep here.

GitHub is a mirror of our internal repository and the sync lags a little, so `master` here is behind. Against our current internal `master`:

- `BTDM_CTRL_LEGACY_AUTH_VENDOR_EVT_EFF` and `BT_NIMBLE_MESH_PROVISIONER` are already `default n`, so those two hunks would be no-ops.
- The two `fatfs` lines are covered by #18960, which is already queued internally.
- `EXAMPLE_USE_SCAN_CHANNEL_BITMAP` in `examples/wifi/scan` is the one line here that is not covered anywhere else.

There seems to be one more instance of the same problem: `examples/peripherals/twai/twai_utils/main/Kconfig.projbuild:9`, where `EXAMPLE_ENABLE_TWAI_FD` has `default false`. I scanned the whole tree against the full set of invalid bool literals and that appears to be the last one.

Would you be willing to adjust the PR to keep just the `examples/wifi/scan` fix and add the `twai_utils` one? That would close out the rest nicely. If you would rather not, that is absolutely no problem, we can take it as it is and make the adjustment on our side.

richardsonnick:
@fhrbata Of course happy to help. Should be fixed. Lmk if there's anything else I can help with here :)

fhrbata:
Hello @richardsonnick,

Thank you very much for the quick update. The change is exactly right now, both lines are what we needed and nothing else is outstanding on the content side.

I owe you an apology: I should have caught this in the previous round and spared you another trip. Our project follows the Conventional Commits style for commit messages, and the current message does not quite match that format, so it gets flagged once the change reaches our internal repository. That is entirely my oversight for not mentioning it when I reviewed the rest.

Would you mind amending the commit message to something along these lines?

```
fix(examples): use 'n' instead of '0'/'false' for bool Kconfig defaults

'0' and 'false' are not valid values for a bool symbol. The parser emits
a note and falls back to 'n', so the resulting configuration is correct
but the build is noisy. Set the two remaining example symbols to
'default n' to make the value explicit and silence the note.
```

The structure we follow is `<type>(<scope>): <summary>`, with the summary under 72 characters, a blank line after the subject, and body lines wrapped at 100 characters. Allowed types are `change`, `ci`, `docs`, `feat`, `fix`, `refactor`, `remove`, `revert` and `test`. Please feel free to reword the text however you prefer, only the structure matters.

A `git commit --amend` followed by a force push to your branch is all that is needed. Your authorship is of course preserved.

One thing that may be useful for future contributions: this particular check runs as a local pre-commit hook rather than in CI, which makes it very easy to miss, and I am sorry that it was not visible to you anywhere on the pull request. If you would like it to run automatically, it can be installed with:

```
pip install pre-commit
pre-commit install --allow-missing-config -t pre-commit -t commit-msg
```

Details are here: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/install-pre-commit-hook.html

Thank you again for the contribution and for your patience with the extra round.

richardsonnick:
@fhrbata no worries. Looking better?

fhrbata:
Hello @richardsonnick,

Perfect, that is exactly it. Thank you very much for the quick turnaround, and for putting up with the extra round.

Nothing further is needed from your side. I will sync the pull request to our internal repository, where it goes through our CI, and once it is merged there it will be pushed back out to GitHub. This pull request will then close itself as merged, with your commit and authorship preserved. That can take a little while depending on the CI queue, so please do not worry if it sits open for a few days.

Thank you again for the fix, and for the wider sweep that turned these up in the first place.

fhrbata:
sha=acebd798941768491be3e2c7b55ff52e515ea4eb

Issue #18800, opened 2026-07-07
ESP32-P4 EMAC: MTL RX FIFO can latch dead after overflow in threshold mode (IDFGH-17915)


scrambletools:
### Chip / IDF version

ESP32-P4 (chip rev 1.3), IDF v6.1-dev-4859 (code unchanged on current master).

### Problem

After an RX FIFO overflow under burst load, the P4 EMAC can stop forwarding frames to the DMA permanently.

Observed live register state in the stuck condition:
- `dmastatus` receive process state = 011 (Running: waiting for receive packet)
- all RX descriptors DMA-owned/free, remaining frames = 0
- overflow flag (bit 4) latched in `dmastatus`
- no further EMAC interrupts of any kind (the overflow interrupt summary only edge-triggers, so once cleared nothing re-fires)
- the MAC continues receiving (wire carries full-rate traffic; missed/overflow counters climb), but the MTL RX FIFO never hands another byte to the DMA

Not recoverable by: receive poll demand, MAC receiver enable toggle, or `esp_eth_stop()`/`esp_eth_start()` (start performs no software reset). The only recovery found is the full DMA software reset (`dmabusmode` SWR) followed by complete MAC/DMA reconfiguration. Note the software reset requires active PHY RX/TX clocks to complete, so it cannot simply be added to the start path — at boot it races PHY bring-up and times out; it is only safe with the link established, which is the only state the lockup occurs in anyway.

### Trigger

RX descriptor exhaustion during multi-millisecond RX-task starvation (e.g. flash-cache-disable windows during NVS writes) while sustained traffic (~16k pkt/s bidirectional) keeps arriving, overflowing the FIFO. PR #18799 fixes two driver defects (RBU never waking the RX task; descriptor leak on overflow-truncated frames) that make this trigger far more likely; with that PR the lockup becomes rare but the underlying FIFO state remains reachable.

### Why P4 is exposed

The P4 RX FIFO is 256 B, so receive store-and-forward cannot hold a full frame and `emac_hal_init_dma_default` disables it — threshold mode is mandatory. The lockup appears specific to threshold-mode operation after overflow.

### Ask

[...]

PR #19007, opened 2026-08-21
fix(esp_eth): avoid per-frame descriptor ring scan in RX task (IDFGH-18183)

scrambletools:
The RX task's drain loop called emac_esp_dma_get_remain_frames() after every received frame, an O(backlog) walk of the descriptor ring with a cache invalidate per descriptor. Under sustained load the scan cost grows with the backlog, so the deeper the backlog gets, the more time goes into scanning instead of draining, until the task can no longer keep up and the DMA drops frames continuously.

The loop only needs to know whether at least one more complete frame is waiting, which emac_esp_dma_alloc_recv_buf() already answers by checking a single descriptor, so the loop now exits on that instead. The full remain-frames count is kept only under CONFIG_ETH_SOFT_FLOW_CONTROL, which genuinely needs it for the pause-frame watermarks.

Measured on ESP32-P4 (20 RX descriptors, 16k pkt/s bidirectional AVB load): ~3000 missed frames/s before, 0 after, with dmamissedfr confirming. Discussion and measurements in #18800.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path RX task behavior for all Ethernet users; incorrect drain/exit logic could drop frames or leave pause asserted, though the change is narrowly scoped to loop control.
> 
> **Overview**
> **Reworks the EMAC RX drain loop** so it no longer calls `emac_esp_dma_get_remain_frames()` after every received frame to decide whether to keep draining.
> 
> The inner loop now relies on `emac_esp_dma_alloc_recv_buf()` to detect whether another frame is ready. When allocation returns no buffer and `frame_len` is zero, it only then scans remaining frames to distinguish an empty ring from a bad descriptor with more frames behind it; with soft flow control enabled it clears any standing pause before idling when the ring is drained.
> 
> **`emac_esp_dma_get_remain_frames()` is limited to** that empty-or-continue path and to `CONFIG_ETH_SOFT_FLOW_CONTROL`, where descriptor counts are still needed for pause-frame watermarks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4bc48f3ac4155daa9d87654addf02dbc8180247e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_eth/src/mac/esp_eth_mac_esp.c
+++ b/components/esp_eth/src/mac/esp_eth_mac_esp.c
@@ -615,7 +615,7 @@ static void emac_esp32_rx_task(void *arg)
     while (1) {
         // block indefinitely until got notification from underlay event
         ulTaskNotifyTake(pdTRUE, portMAX_DELAY);
-        do {
+        while (1) {
             /* set max expected frame len */
             uint32_t frame_len = ETH_MAX_PACKET_SIZE;
             buffer = emac_esp_dma_alloc_recv_buf(emac->emac_dma_hndl, &frame_len);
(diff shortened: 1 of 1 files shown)


scrambletools:
Good catch on the ambiguous exit: emac_esp_dma_alloc_recv_buf returns (NULL, len=0) both for an empty ring and after emac_esp_dma_get_valid_recv_len flushes an ErrSummary frame, and the original version of this PR would stop draining in the second case with complete frames still waiting, and could skip the final pause-clear under soft flow control.

Fixed in bac0f6b: on that path the loop now runs one emac_esp_dma_get_remain_frames() to disambiguate, exits only when the ring is truly drained (releasing any standing pause frame first when soft flow control is enabled), and otherwise continues draining past the flushed frame. The walk is paid once per drain exit or per erroneous frame rather than per received frame, so the hot path stays O(1) per frame, which is the point of the change.

components/esp_eth/src/mac/esp_eth_mac_esp.c
```diff
@@ -645,17 +645,37 @@ static void emac_esp32_rx_task(void *arg)
                 ESP_LOGE(TAG, "no mem for receive buffer");
                 /* ensures that interface to EMAC does not get stuck with unprocessed frames */
                 emac_esp_dma_flush_recv_frame(emac->emac_dma_hndl);
+            } else {
+                /* alloc_recv_buf saw no valid frame. This is ambiguous: the
```
kostaond: @scrambletools could you please make the comment more brief? Two lines maximum.

scrambletools: Done, reduced to a single line in 4bc48f3.

kostaond:
sha=4bc48f3ac4155daa9d87654addf02dbc8180247e

PR #19024, opened 2026-08-27
feat(driver_twai): support queueless TX with tx_queue_depth 0 (IDFGH-18205)

nebkat:
Adds a `tx_queue_depth == 0` mode to the on-chip TWAI driver. Instead of a TX queue, the node holds a single TX permit, so a submitted frame goes straight to the hardware and is never parked in the driver.

## Why

We want to manage the transmit queue entirely in the application, mainly so that a frame which has not gone out yet is still ours to reorder or drop. The driver's frame queue gives no way to do that — `twai_frame_queue` has no remove and no flush — so the only way to keep a frame retractable is to never hand it over in the first place.

`tx_queue_depth = 1` is not a substitute: one frame in hardware plus one in the queue is effectively a queue of two, and the queued one is already out of reach.

## How

- `tx_queue_depth == 0` (and not listen-only) creates a binary semaphore, given once, in place of `tx_queue`. It means "one frame in flight", independent of `tx_slot_num`.
- `twai_node_transmit()` takes the permit, blocking up to its timeout (ISR-safe take on the ISR path), then starts the transaction directly.
- The TX-done ISR returns the node to idle and hands the permit back **before** invoking `on_tx_done`, so transmitting from inside that callback picks the permit straight back up and starts the next frame with no driver-side hop. Doing the whole idle transition before the handoff also means there is no read-then-act after it — everything is ordered by the permit itself.
- `twai_node_transmit_wait_all_done()` just acquires and releases the permit, which is exact, so `TWAI_IDLE_EVENT_BIT` is unused in this mode. That also keeps the chain-submit path from reaching the non-ISR `xEventGroupClearBits()` call in `_node_queue_tx()`.
- Bus-off recovery hands the permit back if a frame was in flight.
- `tx_queue_remaining` reports the permit count: 1 when a frame can be submitted without blocking, 0 while busy.

The queue path is unchanged. The one line that moved on it is in the TX-done ISR loop, which now hoists `done_tx_frame` and clears `p_curr_tx[tx_idx]` before the callback rather than after, so both modes share one path. That is behaviourally identical for the queue mode: `hw_busy` is still true during the callback, so a transmit from inside it takes the queue path and its second-chance CAS fails either way, and nothing else reads `p_curr_tx` from a callback.

## Status

Draft — this is not heavily tested. It was built and smoke-tested on ESP32-S3 against v6.1 in an application that uses TWAI, then rebased onto master; there are no new unit tests yet.

Opening it mainly to find out whether this is a direction you'd want in the driver at all, and whether the ISR ordering above reads correctly to you. Happy to add test coverage and docs if the approach is welcome.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/esp_driver_twai/esp_twai_onchip.c
+++ b/components/esp_driver_twai/esp_twai_onchip.c
@@ -50,6 +50,7 @@ typedef struct {
     intr_handle_t intr_hdl;
     intr_handle_t timer_intr_hdl;
     twai_frame_queue_t tx_queue;
+    SemaphoreHandle_t idle_sem;
     EventGroupHandle_t event_group;
     twai_clock_source_t curr_clk_src;
     uint32_t src_freq_hz;
@@ -258,7 +259,15 @@ static void _node_isr_main(void *arg)
         }
         // node recover from busoff, restart remain tx transaction
         if ((e_data.old_sta == TWAI_ERROR_BUS_OFF) && (e_data.new_sta == TWAI_ERROR_ACTIVE)) {
-            if (_node_start_tx_batch_from_isr(twai_ctx, &do_yield)) {
+            if (twai_ctx->idle_sem) {
+                // permit mode: the frame in flight (if any) is lost, hand its permit back
+                memset(twai_ctx->p_curr_tx, 0, sizeof(twai_ctx->p_curr_tx));
+                if (atomic_exchange(&twai_ctx->hw_busy, false)) {
+                    BaseType_t sem_yield = pdFALSE;
+                    xSemaphoreGiveFromISR(twai_ctx->idle_sem, &sem_yield);
+                    do_yield |= sem_yield;
+                }
+            } else if (_node_start_tx_batch_from_isr(twai_ctx, &do_yield)) {
                 atomic_store(&twai_ctx->hw_busy, true);
             } else {
                 _node_mark_tx_idle(twai_ctx);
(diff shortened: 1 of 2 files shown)


nebkat:
@wanckl fyi — this builds on your TX queue rework, thought you'd want to see it.

components/esp_driver_twai/esp_twai_onchip.c
```diff
@@ -619,6 +641,26 @@ static esp_err_t _node_queue_tx(twai_node_handle_t node, const twai_frame_t *fra
...
+        BaseType_t taken;
+        if (xPortInIsrContext()) {
+            taken = xSemaphoreTakeFromISR(twai_ctx->idle_sem, &yield_required);
+        } else {
+            taken = xSemaphoreTake(twai_ctx->idle_sem, (timeout == -1) ? portMAX_DELAY : pdMS_TO_TICKS(timeout));
+        }
+        ESP_RETURN_ON_FALSE_ISR(taken == pdTRUE, ESP_ERR_TIMEOUT, TAG, "tx busy");
+        atomic_store(&twai_ctx->hw_busy, true);
+        twai_ctx->p_curr_tx[0] = frame;
+        _node_start_trans(twai_ctx, 0);
```
wanckl, line 657: This just start trans and not block till finish

components/esp_driver_twai/esp_twai_onchip.c
```diff
@@ -648,6 +690,13 @@ static esp_err_t _node_wait_tx_all_done(twai_node_handle_t node, int timeout)
     TickType_t ticks_to_wait = (timeout == -1) ? portMAX_DELAY : pdMS_TO_TICKS(timeout);
     ESP_RETURN_ON_FALSE(atomic_load(&twai_ctx->state) != TWAI_ERROR_BUS_OFF, ESP_ERR_INVALID_STATE, TAG, "node is bus off");
 
+    if (twai_ctx->idle_sem) {
+        // permit mode: holding the permit means nothing is in flight, give it straight back
+        ESP_RETURN_ON_FALSE(xSemaphoreTake(twai_ctx->idle_sem, ticks_to_wait) == pdTRUE, ESP_ERR_TIMEOUT, TAG, "wait tx timeout");
```
wanckl, line 695: So a call to `wait_all_done` still required even queue is not used, it different from you say

`and `twai_node_transmit` blocks until the hardware is free`

wanckl:
But I think you are more want a `blocking` like API right?

Driver use queue only when hw busy,  as you see tx function try get atomic lock first.    

If you wrap all your send as:
```c
esp_error_t twai_node_tx_blocking() {
node_transmmit();
node_wait_all_done();
}
```

It will don't use queue whatever how depth is, WDYT?

Issue #19023, opened 2026-08-27
PPA SRM: ppa_do_scale_rotate_mirror() never returns for certain rotated block geometries under concurrent DSI scanout (ESP32-P4, DIG-734) (IDFGH-18204)


c0ball:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

v5.5.3

### Espressif SoC revision.

ESP32-P4 v1.3

### Operating System used.

macOS

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

None

### Development Kit.

Elecrow CrowPanel Advanced 9" V1.2 (ESP32-P4 + ESP32-C6, EK79007 MIPI-DSI panel, 1024×600, 32 MB PSRAM)

### Power Supply used.

USB

### What is the expected behavior?

`ppa_do_scale_rotate_mirror()` called with `PPA_TRANS_MODE_BLOCKING` returns for
every accepted block geometry — either with `ESP_OK` or with an error.

### What is the actual behavior?

For certain block geometries under 270° rotation the call never returns, if the
destination is a MIPI-DSI framebuffer being scanned out concurrently. The
calling task then blocks permanently in

```c
// components/esp_driver_ppa/src/ppa_core.c:459
xSemaphoreTake(trans_elm->sem, portMAX_DELAY); // Given in the ISR
```

There is no timeout, no error and no log output. The task watchdog does not
catch it either, because the task is blocked rather than spinning — so the
symptom is a silently frozen display while the rest of the firmware keeps
running normally.

### Steps to reproduce.

Reproducer: <https://github.com/c0ball/PPA-SRM-stall-reproducer>

It brings up the panel, obtains the two DSI framebuffers via
`esp_lcd_dpi_panel_get_frame_buffer()`, and then repeats the cycle
*rotate a chunk into the inactive buffer → present at VSYNC → resync the
now-inactive buffer* — the same sequence a partial-rendering LVGL display
driver performs. All submissions are RGB565 → RGB565, scale 1:1, PPA client
with `.oper_type = PPA_OPERATION_SRM` and `.max_pending_trans_num = 1`.

[...]

Issue #18999, opened 2026-08-19
DMA2D dma2d_connect hangs indefinitely on ESP32-P4 rev v1.3 under continuous PPA load (IDFGH-18171)


ccatlett1984:
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.
- [x] I have searched the issue tracker for a similar issue and not found a similar issue.

### IDF version.

6.0.0, 6.0.2, 6.1-RC1

### Espressif SoC revision.

ESP32-P4, silicon revision v1.3

### Operating System used.

Windows

### How did you build your project?

VS Code IDE

### If you are using Windows, please specify command line type.

CMD

### Development Kit.

Waveshare esp32-p4-wifi6-touch-lcd-5

### Power Supply used.

USB

### What is the expected behavior?

I expect dma2d to work without causing a cpu panic.

### What is the actual behavior?

dma2d_connect() enters a busy-loop with no timeout in dma2d_hal_tx_reset_channel() waiting for dma2d_ll_tx_is_reset_avail() to return true. That register bit never gets set, the interrupt WDT fires, and the app panics on CPU 0.

Failure timing under sustained load (14 fps camera pipeline that pushes one PPA SRM transaction per frame): stable for ~1000 frames (~70 seconds) and then hangs deterministically. Same call site, same stack trace, every time.

Reduced-load workarounds (skipping PPA when a menu overlay is on-screen) delay but do not prevent the hang.

Stack trace (representative)
Guru Meditation Error: Core  0 panic'ed (Interrupt wdt timeout on CPU0).
MEPC    : 0x4ff0566e — dma2d_ll_tx_is_reset_avail at hal/dma2d_ll.h:866
                       (inlined by dma2d_hal_tx_reset_channel at dma2d_hal.c:20)
RA      : 0x4ff0409c — dma2d_connect at esp_driver_dma/src/dma2d.c:659
...
  ppa_srm_transaction_on_picked at esp_driver_ppa/src/ppa_srm.c:89
  dma2d_enqueue at esp_driver_dma/src/dma2d.c:1077
  ppa_dma2d_enqueue at esp_driver_ppa/src/ppa_core.c:418
  ppa_do_operation at esp_driver_ppa/src/ppa_core.c:519
[...]

PR #19047, opened 2026-09-04
fix(ppa): decide engine idleness under the engine spinlock (IDFGH-18236)

gingerbeardman:
## Summary

`esp_driver_ppa` can strand a queued transaction when the submitting task runs on a different core from the PPA/2D-DMA interrupt. The engine's "idle" state is a binary semaphore that the completion ISR gives only *after* leaving the engine spinlock, while `ppa_do_operation` inserts its transaction under that spinlock and then tries the semaphore with a zero timeout. A task on the other core can insert between the ISR's "queue empty" decision and its give, find the semaphore still taken, assume the ISR will chain-start the transaction, and be left at the head of an idle engine's queue. It only starts when a later submission's completion picks it up. A lone last submission never completes, and a caller waiting on its done callback blocks forever.

On the ISR's own core the two steps cannot interleave, which is why single-core users never see it.

## Reproduction

ESP32-P4 (v1.3), IDF v5.5.5 and master. PPA/2D-DMA interrupt on core 0 (allocated where the client was registered). A core-1 task submits single SRM transactions (rotation 0, 480×40 RGB565, `PPA_TRANS_MODE_NON_BLOCKING`) spaced at the transaction's own measured duration ±60 µs, so that submissions land inside the completion interrupt, each tagged with its index in `user_data`; the `on_trans_done` callback counts completions that arrive out of index order and their latency.

| submitting task | before | after |
|---|---|---|
| core 1 | ~130 stranded per 20000 (completed out of order, ~3 durations late — only after the next submission ran) | 0 |
| core 0 | 0 | 0 |

In the original application (a two-core render pipeline: raster + PPA rotate on core 1, simulation on core 0) the last strip rotate of a frame was stranded about once a minute and the frame loop deadlocked on its completion.

## Fix

Replace the engine semaphore with a `busy` flag that both sides read and write only under the engine spinlock. The submitter inserts and, if the engine is idle, marks it busy and starts the transaction itself; the ISR clears the flag in the same critical section in which it finds the queue empty. Behaviour is otherwise unchanged: at most one transaction runs per engine, FIFO order, the ISR chains the next queued transaction.

Also in this commit: blocking mode clears a stale give on the recycled per-transaction semaphore. The ISR gives `trans_elm->sem` unconditionally and non-blocking callers never take it, so a later blocking operation on a recycled element returned before its transaction had finished.

## Notes

Possibly related open reports, though they look like different mechanisms: #19023 (blocking SRM call never returns under DSI scan-out for some 270° geometries) and #18999 (`dma2d_connect` reset-bit hang under sustained load).

I have signed the contributor agreement / will sign when prompted.

---

> [!NOTE]
> **Medium Risk**
> Changes core PPA transaction scheduling and synchronization across cores; incorrect locking could still strand ops or affect PM lock lifetime, though the fix is narrowly scoped to engine idle detection.
> 
> **Overview**
> Fixes a **multi-core race** where non-blocking PPA submissions from a different core than the completion ISR could leave a transaction at the head of an **idle** engine queue (never started until a later submission), including possible **permanent deadlock** on the last transaction of a frame.
> 
> **Engine scheduling** no longer uses a per-engine binary semaphore. `ppa_engine_t` now has a **`busy` flag** updated only under the engine spinlock: `ppa_do_operation` enqueues, sets `busy` and kicks DMA when the engine was idle; `ppa_transaction_done_cb` clears `busy` in the same critical section when the queue becomes empty (instead of giving the semaphore after releasing the lock). PM lock behavior when the queue drains is unchanged.
> 
> **Blocking mode**: before waiting on the per-transaction semaphore, the driver **drains a stale give** on recycled transaction elements (ISR always gives; non-blocking callers never take).
> 
> SRM/blend engine registration teardown no longer allocates or frees the removed engine semaphores.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15abeb729a291ebfae26bccb22cb6191ab044e6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/esp_driver_ppa/src/ppa_core.c
+++ b/components/esp_driver_ppa/src/ppa_core.c
@@ -96,17 +96,14 @@ static esp_err_t ppa_engine_acquire(const ppa_engine_config_t *config, ppa_engin
     if (config->engine == PPA_ENGINE_TYPE_SRM) {
         if (!s_platform.srm) {
             ppa_srm_engine_t *srm_engine = heap_caps_calloc(1, sizeof(ppa_srm_engine_t), PPA_MEM_ALLOC_CAPS);
-            SemaphoreHandle_t srm_sem = xSemaphoreCreateBinaryWithCaps(PPA_MEM_ALLOC_CAPS);
             dma2d_descriptor_t *srm_tx_dma_desc = (dma2d_descriptor_t *)heap_caps_aligned_calloc(alignment, 1, s_platform.dma_desc_mem_size, MALLOC_CAP_DMA | MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
             dma2d_descriptor_t *srm_rx_dma_desc = (dma2d_descriptor_t *)heap_caps_aligned_calloc(alignment, 1, s_platform.dma_desc_mem_size, MALLOC_CAP_DMA | MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT);
-            if (srm_engine && srm_sem && srm_tx_dma_desc && srm_rx_dma_desc) {
+            if (srm_engine && srm_tx_dma_desc && srm_rx_dma_desc) {
                 srm_engine->dma_tx_desc = srm_tx_dma_desc;
                 srm_engine->dma_rx_desc = srm_rx_dma_desc;
                 srm_engine->base.platform = &s_platform;
                 srm_engine->base.type = PPA_ENGINE_TYPE_SRM;
                 srm_engine->base.spinlock = (portMUX_TYPE)portMUX_INITIALIZER_UNLOCKED;
-                srm_engine->base.sem = srm_sem;
-                xSemaphoreGive(srm_engine->base.sem);
                 STAILQ_INIT(&srm_engine->base.trans_stailq);
                 s_platform.srm = srm_engine;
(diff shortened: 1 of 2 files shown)


gingerbeardman:
Minimal reproducer (ESP32-P4 with PSRAM, no display): https://github.com/gingerbeardman/esp-idf-ppa-race-repro

The SRM client is registered on core 0 (so the 2D-DMA interrupt lives there); a core-1 task submits single non-blocking copies spaced at the transaction's own duration ±60 µs, tagged with their index, and the done callback counts out-of-order completions. A second phase submits pairs and waits with a timeout, which turns a stranded second transaction into the application hang.

Measured on ESP32-P4 v1.3, IDF v5.5.5:

| submitter | driver | out of order / 20000 | hangs / 2000 pairs |
|---|---|---|---|
| core 1 | stock | 195 (each ~3 durations late) | 18 |
| core 0 | stock | 0 | 0 |
| core 1 | this PR | 0 | 0 |
| core 0 | this PR | 0 | 0 |

songruo:
sha=15abeb729a291ebfae26bccb22cb6191ab044e6c

PR #19061, opened 2026-09-09
fix(tcp_transport): clear stale TLS client session (IDFGH-18254)

### LINKED ISSUE #19033: tcp_transport: esp_transport_ssl_session_ticket_operation leaves cfg.client_session dangling after a failed SAVE, causing a use-after-free in the next handshake (IDFGH-18217)
### Environment

- ESP-IDF version: reproduced against v5.5.2; the code is byte-identical on
  `master`, `release/v5.5`, `release/v5.4` and `release/v5.3` at time of writing
- Target: ESP32 (original), no PSRAM
- Component: `tcp_transport` (`esp_transport_ssl_session_ticket_operation`),
  reached through `esp_http_client_config_t.save_client_session`
- Relevant config: `CONFIG_ESP_TLS_CLIENT_SESSION_TICKETS=y`,
  `CONFIG_MBEDTLS_SSL_PROTO_TLS1_3` **not** set (the ESP32 default)

### Summary

`esp_transport_ssl_session_ticket_operation()` can leave
`ssl->cfg.client_session` pointing at freed memory. A later connect passes that
config to `esp_tls_conn_new_sync()`, and `esp_mbedtls_handshake()` dereferences
the freed object.

The trigger is an allocation failure inside `esp_tls_get_client_session()`, so
this fires under memory pressure — the condition where a device is least able to
survive heap corruption, and where the failure is hardest to attribute.

### Root cause

`components/tcp_transport/transport_ssl.c`:

```c
case ESP_TRANSPORT_SESSION_TICKET_SAVE:
    esp_tls_free_client_session(ssl->session_ticket);
    ssl->session_ticket = esp_tls_get_client_session(ssl->tls);
    break;
case ESP_TRANSPORT_SESSION_TICKET_USE:
    if (ssl->session_ticket == NULL) {
        return ESP_ERR_INVALID_STATE;
    }
    ssl->cfg.client_session = ssl->session_ticket;
    break;
```

Two gaps:

1. `SAVE` frees the object that a previous `USE` published into
   `ssl->cfg.client_session`, and reassigns `ssl->session_ticket` without
   touching `cfg`. If the reassignment yields NULL, `cfg` is left dangling.
2. `USE` returns early when `ssl->session_ticket` is NULL **without clearing
   `ssl->cfg.client_session`**, so the stale pointer survives the one operation
   that would otherwise repair it.

Nothing else clears it: `base_close()` touches neither `cfg` nor
`session_ticket`, so closing the connection does not help.

`esp_mbedtls_get_client_session()` in `components/esp-tls/esp_tls_mbedtls.c`
returns NULL on `calloc` failure, and on `mbedtls_ssl_get_session()` failure —
which returns `MBEDTLS_ERR_SSL_ALLOC_FAILED` from `mbedtls_ssl_session_copy()`,
since that deep-copies the peer certificate.

### Sequence (TLS 1.2)

With one `esp_http_client_handle_t` reused across requests and
`save_client_session = true`:

1. **Connect 1.** `USE` is skipped (no ticket yet). `SAVE` stores `S1`.
   `cfg.client_session` is still NULL.
2. **Connect 2.** `USE` sets `cfg.client_session = S1`. Handshake resumes
   normally. `SAVE` then frees `S1` and calls `esp_tls_get_client_session()`
   again — **this call fails under memory pressure and returns NULL**, so
   `ssl->session_ticket = NULL` while `cfg.client_session` still points at the
   freed `S1`.
3. **Connect 3.** `USE` sees `session_ticket == NULL` and returns
   `ESP_ERR_INVALID_STATE` early, leaving `cfg.client_session` untouched.
   `ssl_connect()` passes `&ssl->cfg` to `esp_tls_conn_new_sync()`, and
   `esp_mbedtls_handshake()` runs:

   ```c
   if (cfg->client_session != NULL) {
       ESP_LOGD(TAG, "Reusing the already saved client session context");
       if ((ret = mbedtls_ssl_set_session(&tls->ssl, &(cfg->client_session->saved_session))) != 0 ) {
   ```

   The NULL guard passes, because the pointer is not NULL — it is freed.
   `mbedtls_ssl_set_session()` then deep-copies from freed memory.

Step 2 is the only step that needs an unusual condition, and it is an ordinary
one on a constrained target: the device I found this on runs Bluetooth, WiFi and
TLS on an ESP32 with no PSRAM, where the heap routinely has plenty of free bytes
but no large contiguous block.

### Why TLS 1.3 does not show it

Under TLS 1.3 the save never succeeds in the first place —
`esp_mbedtls_get_client_session()` returns NULL because the ticket has not
arrived yet, which is the separate defect reported in #18919. So
`cfg.client_session` never becomes non-NULL and the dangling pointer cannot
arm. The hazard is specific to TLS 1.2, which is the default on the ESP32 since
`CONFIG_MBEDTLS_SSL_PROTO_TLS1_3` is off there.

### Suggested fix

Clear `cfg.client_session` wherever the object it points at stops being valid:

```c
case ESP_TRANSPORT_SESSION_TICKET_SAVE:
    esp_tls_free_client_session(ssl->session_ticket);
    ssl->cfg.client_session = NULL;          /* the old ticket is gone */
    ssl->session_ticket = esp_tls_get_client_session(ssl->tls);
    break;
case ESP_TRANSPORT_SESSION_TICKET_USE:
    if (ssl->session_ticket == NULL) {
        ssl->cfg.client_session = NULL;      /* nothing to offer */
        return ESP_ERR_INVALID_STATE;
    }
    ssl->cfg.client_session = ssl->session_ticket;
    break;
case ESP_TRANSPORT_SESSION_TICKET_FREE:
    esp_tls_free_client_session(ssl->session_ticket);
    ssl->cfg.client_session = NULL;
    ssl->session_ticket = NULL;
    break;
```

The `FREE` branch has the same gap, though it is currently unreachable from
`esp_http_client`.

### How this was found

Reviewing a change that enabled `save_client_session` on an ESP32 BLE-to-Signal K
gateway. The setting was dropped rather than shipped, so this is a code-reading
result with the failure sequence traced through the sources rather than a captured
crash — I have not built a deliberate allocation-failure harness to observe the
corruption. The dangling assignment itself is plain in the source, and every step
above is cited.

wyf-777:
## Summary

Clear `cfg.client_session` when a saved TLS session is freed or replaced, and when `USE` finds no saved session.

A previous `USE` can publish a session pointer into the connection configuration. If a subsequent `SAVE` frees that session but fails to obtain a replacement, the configuration retains the freed pointer. The next connection can then dereference it during the TLS handshake. Explicit `FREE` has the same stale-pointer problem.

Add a Catch2 regression test that inspects the configuration passed to the next TLS connection after:
- a failed replacement `SAVE`, both with and without a subsequent `USE`;
- an explicit `FREE`.

Enable CMock callbacks for the ESP-TLS mock and propagate the session-ticket definition from that mock component so the mock, transport, and test use the same configuration layout.

Fixes #19033.

## Validation

- Full ESP32 build of `components/tcp_transport/test_apps` with `CONFIG_ESP_TLS_CLIENT_SESSION_TICKETS=y`: passed (710 build steps). No hardware execution was performed.
- Host-test CMake configuration: passed; verified the session-ticket definition is present for the ESP-TLS mock, transport, and new test.
- Generated the actual ESP-TLS mocks with the repository's CMock generator and compiled the new C++ test translation unit against the generated header: passed (MinGW, with a local BSD-header shim). This is a compile-only check, not execution of the Linux test suite.
- Applicable pre-commit checks: passed except the stock error-code registration hook, which reports Windows backslash paths as unregistered against CMake forward-slash paths. Re-running that check with in-memory path normalization passed; its source was not changed by this PR.
- The Linux Catch2/ASan host-test suite was not run: this environment is Windows without a usable Linux runtime. Linux CI validation is still needed.

---

> [!NOTE]
> **Medium Risk**
> Changes TLS session-ticket handling so connection config no longer retains freed session pointers; limited scope but affects handshake configuration on reconnect.
> 
> **Overview**
> Fixes a **stale `client_session` pointer** in the SSL transport when TLS session tickets are saved, used, or freed. After `USE`, `cfg.client_session` could still reference memory that `SAVE` or `FREE` had already released (including when a replacement `SAVE` fails to obtain a new session), so the next connect could pass a dangling pointer into the TLS handshake.
> 
> **`esp_transport_ssl_session_ticket_operation`** now sets `ssl->cfg.client_session` to `NULL` on `SAVE` (before replacing the ticket), on failed `USE` (no saved ticket), and on `FREE`.
> 
> Adds a **Catch2 host test** (`test_ssl_transport.cpp`) that stubs `esp_tls_conn_new_sync` and asserts the `esp_tls_cfg_t` passed on the next connect matches the expected session pointer for normal use, failed replacement save, explicit free, and save-without-use paths. Host-test **CMake** registers the new test and defines `CONFIG_ESP_TLS_CLIENT_SESSION_TICKETS=1` on the mocked `esp-tls` component; the **ESP-TLS CMock** config enables the **callback** plugin for connect stubs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 931a886e780bdcaebbbf3b392aa75b6bfdd4c19a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/components/tcp_transport/host_test/main/CMakeLists.txt
+++ b/components/tcp_transport/host_test/main/CMakeLists.txt
@@ -1,4 +1,4 @@
-idf_component_register(SRCS "test_socks_transport.cpp" "test_websocket_transport.cpp"
+idf_component_register(SRCS "test_socks_transport.cpp" "test_ssl_transport.cpp" "test_websocket_transport.cpp"
                         REQUIRES tcp_transport mocked_transport
                         INCLUDE_DIRS "$ENV{IDF_PATH}/tools"
                         WHOLE_ARCHIVE)
@@ -7,6 +7,9 @@ idf_component_get_property(lwip_component lwip COMPONENT_LIB)
 idf_component_get_property(esp_timer_component esp_timer COMPONENT_LIB)
 idf_component_get_property(tcp_transport_component tcp_transport COMPONENT_LIB)
 target_link_libraries(${tcp_transport_component} PUBLIC ${lwip_component} ${esp_timer_component})
+idf_component_get_property(esp_tls_component esp-tls COMPONENT_LIB)
+# The mocked esp-tls component does not expose the session ticket Kconfig option.
+target_compile_definitions(${esp_tls_component} PUBLIC CONFIG_ESP_TLS_CLIENT_SESSION_TICKETS=1)
 target_compile_options(${COMPONENT_LIB} PUBLIC -fsanitize=address -fconcepts)
 target_link_options(${COMPONENT_LIB} PUBLIC -fsanitize=address)
 
--- a/components/tcp_transport/host_test/main/test_ssl_transport.cpp
+++ b/components/tcp_transport/host_test/main/test_ssl_transport.cpp
@@ -21,17 +21,26 @@ namespace {
 esp_tls_client_session_t saved_session{};
 esp_tls_client_session_t *session_to_save;
 esp_tls_t *fake_tls = reinterpret_cast<esp_tls_t *>(&saved_session);
+bool connect_called;
+bool connect_config_present;
+esp_tls_client_session_t *connected_session;
+esp_tls_t *connected_tls;
 
 int connect_callback(const char *, int, int, const esp_tls_cfg_t *cfg, esp_tls_t *tls, int)
 {
-    REQUIRE(cfg != nullptr);
-    REQUIRE(cfg->client_session == nullptr);
-    REQUIRE(tls == fake_tls);
+    connect_called = true;
+    connect_config_present = cfg != nullptr;
+    connected_session = cfg ? cfg->client_session : nullptr;
+    connected_tls = tls;
     return -1;
 }
 
-void expect_connect_without_client_session(esp_transport_handle_t transport)
+void expect_connect_with_client_session(esp_transport_handle_t transport, esp_tls_client_session_t *expected_session)
 {
+    connect_called = false;
+    connect_config_present = false;
+    connected_session = nullptr;
+    connected_tls = nullptr;
     esp_tls_init_ExpectAndReturn(fake_tls);
     esp_tls_conn_new_sync_Stub(connect_callback);
     esp_tls_get_error_handle_ExpectAnyArgsAndReturn(ESP_FAIL);
@@ -39,6 +48,10 @@ void expect_connect_without_client_session(esp_transport_handle_t transport)
 
     REQUIRE(esp_transport_connect(transport, "localhost", 443, 100) == -1);
     Mockesp_tls_Verify();
+    REQUIRE(connect_called);
+    REQUIRE(connect_config_present);
+    REQUIRE(connected_session == expected_session);
+    REQUIRE(connected_tls == fake_tls);
 }
 
 } // namespace
@@ -54,7 +67,7 @@ extern "C" void esp_tls_free_client_session(esp_tls_client_session_t *)
 {
 }
 
-TEST_CASE("TLS session ticket operations clear the configured client session", "[tcp_transport][tls]")
+TEST_CASE("TLS session ticket operations update the configured client session", "[tcp_transport][tls]")
 {
     Mockesp_tls_Init();
     session_to_save = &saved_session;
@@ -63,6 +76,11 @@ TEST_CASE("TLS session ticket operations clear the configured client session", "
 
     REQUIRE(esp_transport_ssl_session_ticket_operation(transport.get(), ESP_TRANSPORT_SESSION_TICKET_SAVE) == ESP_OK);
     REQUIRE(esp_transport_ssl_session_ticket_operation(transport.get(), ESP_TRANSPORT_SESSION_TICKET_USE) == ESP_OK);
+    esp_tls_client_session_t *expected_session = nullptr;
+
+    SECTION("when the saved session is used") {
+        expected_session = &saved_session;
+    }
 
     SECTION("when saving the replacement session fails") {
         REQUIRE(esp_transport_ssl_session_ticket_operation(transport.get(), ESP_TRANSPORT_SESSION_TICKET_SAVE) == ESP_OK);
@@ -78,5 +96,5 @@ TEST_CASE("TLS session ticket operations clear the configured client session", "
         REQUIRE(esp_transport_ssl_session_ticket_operation(transport.get(), ESP_TRANSPORT_SESSION_TICKET_FREE) == ESP_OK);
     }
 
-    expect_connect_without_client_session(transport.get());
+    expect_connect_with_client_session(transport.get(), expected_session);
 }
--- a/components/tcp_transport/host_test/main/test_ssl_transport.cpp
+++ b/components/tcp_transport/host_test/main/test_ssl_transport.cpp
@@ -1,0 +1,82 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <memory>
+#include <type_traits>
+
+#include <catch2/catch_test_macros.hpp>
+
+extern "C" {
+#include "Mockesp_tls.h"
+#include "esp_transport.h"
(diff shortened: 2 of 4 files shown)


andrewfree:
**Independent Linux host-test validation of `8ab63afdc`**

The in-tree `test_transport_on_host` CI job runs this suite but is currently `allow_failure: true` (IDFCI-2781), so I built and ran it independently for a hard signal:

| Environment | Arch | Toolchain | Result |
|---|---|---|---|
| Ubuntu 24.04 (container) | aarch64 | gcc 13.3, glibc 2.39 | All tests passed (107 assertions in 5 test cases) |
| Ubuntu 24.04.4 (bare metal, kernel 6.8) | x86_64 | gcc 13.3, glibc 2.39 | All tests passed (107/5) |
| Debian 12 (container) | aarch64 | gcc 12.2, glibc 2.36 | All tests passed (107/5) |
| Fedora 41 (container) | aarch64 | gcc 14.3.1, glibc 2.40 | All tests passed (107/5) |

(Fedora needed `-Wno-tautological-compare -Wno-dangling-reference` for gcc-14 `-Werror` diagnostics in the vendored fmt v9 headers — unrelated to this PR.)

**Mutation check.** Reverting only the three `ssl->cfg.client_session = NULL;` lines and rebuilding makes the new test case fail in all four environments: `REQUIRE(cfg->client_session == nullptr)` observes the stale pointer at the next `esp_transport_connect` — the sequence from #19033. Restoring the fix returns the suite to 107/5, so the regression test is genuinely sensitive to the bug it guards. (Side note: 3 of the 6 failed assertions in the mutated run are Catch2 "exception translation disabled" artifacts from `REQUIRE` throwing inside the CMock callback through C frames — `CHECK`, or capturing the pointer and asserting after `esp_transport_connect()` returns, would make the failure output cleaner.)

**Coverage suggestion.** All config assertions in the new test are negative (`client_session == nullptr`). Deleting the USE-branch assignment entirely — which would silently disable session resumption — still passes all three sections. A section doing a successful SAVE → USE and asserting `cfg->client_session == &saved_session` at connect time would pin both directions.

**Behavior question.** SAVE now clears `cfg.client_session` on success too, so callers must USE before every connect. `esp_http_client` does exactly that (and ignores the USE return value, so the SAVE-side clear is the load-bearing line for it). Is esp-mqtt's usage pattern compatible?

Scope: host/mock only — no on-target run, no real TLS 1.2 handshake. The build is ASan-linked with no reports, but the test's `esp_tls_free_client_session` is a no-op over a static session, so this validates pointer hygiene as a proxy for the UAF rather than reproducing it under a sanitizer. The change itself reads correct: `cfg.client_session` is written only in those four places and read only via `&ssl->cfg` at connect, and `esp_transport_esp_tls_destroy` frees the enclosing struct immediately, so I don't see a remaining dangling path. Thanks for the fix!

wyf-777:
Thanks for the independent validation and the coverage suggestions. The follow-up adds a successful SAVE → USE section that checks the saved session pointer at connection time. The callback now only captures its arguments; all Catch2 assertions run after `esp_transport_connect()` returns, including a check that the callback actually ran. The failed SAVE and explicit FREE cases remain covered.

For the esp-mqtt question, I inspected its current master at [`4d36e707320c7029952b7c21f37cb80d7c84115f`](https://github.com/espressif/esp-mqtt/blob/4d36e707320c7029952b7c21f37cb80d7c84115f/mqtt_client.c#L2030). Its standard connection code initializes the SSL transport and calls `esp_transport_connect()`, but does not call the session-ticket SAVE/USE API. I therefore found no affected session-resumption path in that code. This is a source-level check of that revision, not a claim about custom transports or all versions. The esp_http_client path calls USE before reconnecting with a saved ticket, as noted in your analysis.

Validation of this follow-up: applicable pre-commit checks and compilation of the updated test translation unit against the actual CMock-generated ESP-TLS header passed. The production fix is unchanged. I still cannot execute the Linux Catch2/ASan suite in this Windows environment, so your four-environment runtime results apply to the previous commit, not this updated test. No on-target or real TLS handshake validation is claimed.

PR #19102, opened 2026-09-17
feat(bootloader_support): Read the partition table offset from the bootloader description (IDFGH-18310)

nebkat:
Applications locate the partition table at their own `CONFIG_PARTITION_TABLE_OFFSET`, so an application built with a different offset than the bootloader that booted it finds no partitions and aborts:

```
E (1664) partition: No MD5 found in partition table
E (1664) partition: load_partitions returned 0x105
assert failed: esp_partition_get_running_partition partition_target.c:251 (it != NULL)
```

That makes it impossible to move the partition table on devices already in the field, because the update carrying the new layout has to run under the old bootloader first. The need comes up with Secure Boot: a signed bootloader is padded to a sector boundary plus a signature sector, and on targets whose ROM loads the bootloader from 0x2000 (ESP32-P4, C5, H4, S31) only 0x6000 fits in front of the default 0x8000 table.

The bootloader is the one component that has to know where the table is, so this lets it say:

- `esp_bootloader_desc_t` gains `partition_table_offset`, set from `CONFIG_PARTITION_TABLE_OFFSET` when the bootloader is built. It takes 4 of the `reserved2` bytes, so the struct stays 80 bytes and older bootloaders read 0.
- New `esp_partition_table_get_offset()`, which always takes the offset from a bootloader description. The bootloader uses its own, so the offset it records is the one it uses. An application reads the field once from the bootloader that booted it (on targets with a recovery bootloader, the one at `esp_rom_get_bootloader_offset()`) and caches it; under a bootloader that records 0 or has no description it uses the new `CONFIG_PARTITION_TABLE_OFFSET_LEGACY_BOOTLOADER`, which defaults to `CONFIG_PARTITION_TABLE_OFFSET`, so behaviour is unchanged unless it is set. Any other invalid value is an error in both.
- The bootloader and applications use it wherever they read or bound the table: loading it (`bootloader_utility`, `esp_partition`), flash encryption, the protected flash region, the bootloader bounds in `app_update` and `esp_image_verify`, and `bootloader_common_erase_part_type_data()`. ESP-TEE and pure RAM applications keep the configured offset.

Tools can use the same field. `parttool.py`, `otatool.py` and flashers currently assume the table is at 0x8000 unless told otherwise with `--partition-table-offset`. Reading 80 bytes at the chip's bootloader offset + 0x20 on a connected device, or the same position in a `bootloader.bin` or merged image, gives them the actual offset, and `esptool image-info` could print it alongside the rest of the bootloader description.

## Testing

The `esp_bootloader_format` test app is extended to check `partition_table_offset` and `esp_partition_table_get_offset()`; it passes on ESP32 in QEMU and builds for ESP32-P4.

A hello_world variant that lists the partitions and initialises NVS, on ESP32 in QEMU:

| Bootloader | Table at | App built for | Result |
|---|---|---|---|
| this branch | 0x10000 | 0x10000 | reads 0x10000, partitions and NVS OK |
| this branch | 0x10000 | 0x8000 | follows the bootloader to 0x10000, OK |
| v6.1 (no field) | 0x8000 | 0x10000, `LEGACY_BOOTLOADER=0x8000` | falls back to 0x8000, OK |
| v6.1 (no field) | 0x8000 | 0x10000, default | aborts as above, unchanged |
| built for 0x8000, description overridden to 0x10000 | 0x10000 | 0x8000 | bootloader and application both use 0x10000, OK |
| description overridden to 0x10001 | 0x10000 | 0x8000 | bootloader logs `part_table: Invalid offset 0x10001` and aborts |

The bootloader grows by 96 bytes on ESP32 and 112 bytes on ESP32-P4.

## Related

No related issues.

---

> [!NOTE]
> **High Risk**
> Changes early boot partition discovery and bootloader sizing/verification; a wrong legacy offset or bad cached value would prevent partition load or brick OTA transitions.
> 
> **Overview**
> Apps no longer rely only on their own `CONFIG_PARTITION_TABLE_OFFSET` to find the partition table. **Bootloaders record the table location** in `esp_bootloader_desc_t::partition_table_offset` (filled at build time; older bootloaders effectively expose 0).
> 
> **`esp_partition_table_get_offset()`** is the single runtime source: the bootloader build returns `CONFIG_PARTITION_TABLE_OFFSET`; a normal app reads and caches the value from the bootloader that actually booted it, with validation and fallback to the new **`CONFIG_PARTITION_TABLE_OFFSET_LEGACY_BOOTLOADER`** when the field is missing or invalid. Partition loading, OTA/bootloader bounds checks, image verify, and bootloader erase paths are wired to this helper instead of fixed macros.
> 
> This allows an OTA app built with a different table offset than the installed bootloader to still boot (e.g. moving the table for Secure Boot padding on some chips) while staying compatible with legacy bootloaders when legacy offset is configured.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2d789c2c800ad5a5f91d76db83d532ac593a15c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Alvin1Zhang:
Thanks for contribution.

